PR #719, opened 2020-10-12
Fix issues from fv3config public bucket switch to requester pays

frodre:

Significant internal changes:

- [ ] Tests added


nbren12:
Seems old

PR #1054, opened 2021-03-02
Add training with backpropagation through time

mcgibbon:
This PR adds code necessary to train a model using backpropagation through time, and use it in our offline report and prognostic runs.

Reviewers please note that I would suggest I move all files stored under `workflows/bptt` into the explore repo. I am keeping it in this PR only for the initial review, so there is some opportunity to get review feedback on this code.

Ideally I'd like to define GCMCell and RecurrentModel outside of fv3fit (in explore), but I could not think of a way to do so.

Added public API:
- Added `ExternalModel`, `RecurrentModel`, and `GCMCell` to `fv3fit.keras`

Refactored public API:
- fv3fit.load now supports recurrent-keras models.
- StandardScaler now supports 3D arrays (e.g. [sample, time, height])

- [x] Tests added (some)


mcgibbon:
Superceded by #1103

PR #1074, opened 2021-03-05
Offline report figures accomodate additional "dataset" dim

AnnaKwa:
This makes a few changes to the plotting in the offline report to handle the multiple dataset training data that have an additional "dataset" dimension (e.g. unperturbed SSTs, +4K SSTs, etc). The snapshot transect plots and global time averaged maps are expanded into rows for each dataset. The remaining plots of RMSE/bias vertical profiles and predicted vertical profiles are averaged over all datasets in the test sample.

Example:
https://storage.googleapis.com/vcm-ml-public/annak/2021-04-28-test-offline-diags-multidataset/index.html


spencerkclark (approved):
Nice, looks good to me.

PR #1085, opened 2021-03-12
Plot nudging tendencies in heating/moistening movies

oliverwm1:
Add a "dQ or nQ" derived variable which is useful for the "column integrated heating and moistening" movie, which we want to work for either prognostic run or nudged runs.

Significant internal changes:
- make the colorbar limits half as large for the ML/nudging panel of heating+moistening movies.

For example, see a movie for nudged run [here](https://storage.googleapis.com/vcm-ml-public/argo/2021-04-09-test-nudged-movies/nudged/column_heating_moistening.mp4) and a prognostic run [here](https://storage.googleapis.com/vcm-ml-public/argo/2021-04-09-test-nudged-movies/prognostic/column_heating_moistening.mp4).


constraints.txt
```diff
@@ -99,7 +99,7 @@ imagesize==1.2.0
 importlib-metadata==1.1.3
 iniconfig==1.0.1
 intake-xarray==0.3.1
-intake==0.6.0
+intake==0.6.2
```
oliverwm1, line 102: gets ride of a warning that happens every time we import `intake`:

```
/home/OliWM/miniconda3/envs/fv3net/lib/python3.7/site-packages/intake/__init__.py:39: UserWarning: deprecation: intake.Catalog now references the base class intake.catalog.base.Catalog
 If you want to open a generic URL, you should use intake.open_catalog
  warnings.warn('deprecation: intake.Catalog now references the base class intake.catalog.base.Catalog\n '
```

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/cli.py
```diff
@@ -11,7 +11,9 @@ def dissoc(namespace, key):
 
 def get_parser():
     parser = argparse.ArgumentParser(description="Prognostic run diagnostics")
-    subparsers = parser.add_subparsers(help="Prognostic run diagnostics")
+    subparsers = parser.add_subparsers(
+        help="Prognostic run diagnostics", required=True, dest="command"
```
oliverwm1, line 15: without this, calling `prognostic_run_diags` without any command raises an error instead of printing usage.

constraints.txt
AnnaKwa: Building with this new constraints file works for the fv3net image but not the prognostic run image, can you try put this intake requirement in `docker/prognostic_run/requirements/base.txt` and then regenerate the constraints with `pip-compile`? Maybe it will be more consistent that way.

AnnaKwa (approved):
The feature to plot nudging and/or ML tendencies looks good to me, could you set the new intake requirement in the prognostic run requirements and autogenerate constraints.txt? It may help with the build issue. Also, if someone else recompiles the constraints later on, your update to intake may be overwritten if that's the only place it shows up.

constraints.txt
oliverwm1: I reverted this change. Requirements changes should be isolated to separate PRs.

PR #1088, opened 2021-03-15
Prognostic run: set surface radiative fluxes in prephysics step

brianhenn:
Allows setting surface radiative fluxes from an external dataset in the prephysics step of the runfile, i.e., for setting values to the fine-res reference in nudge-to-fine run. I established a fine-res reference dataset with standard variable names and dimensions at `gs://vcm-ml-intermediate/2021-03-fine-res-surface-radiative-fluxes/fine-res-surface-radiative-fluxes.zarr` using [this](https://github.com/VulcanClimateModeling/explore/blob/91bc47e6f1217067fbc0f5ba291917d4468cf69c/brianh/2021-03-29-create-surface-radiation-reference/2021-03-29-create-surface-radiation-reference.ipynb) notebook.  Also along with tests, [this](https://github.com/VulcanClimateModeling/explore/blob/master/brianh/2021-01-04-N2F-compare-to-fine/2021-03-26-N2F-rad-setting-verify.ipynb) notebook verifies that the prescribed fluxes in an example run match those in the catalog dataset. 

Added public API:
- `PrescriberConfig` for specifying the `prephysics` entry of the `UserConfig`
- `Prescriber` stepper class for setting values from an external dataset within the runfile's prephysics step

Significant internal changes:
- the `prephysics` step of the runfile now can either execute state setting from a prescribed dataset or from an ML model

- [X] Tests added


workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def _get_prescribed_ds(
+    dataset_key: str,
+    variables: Sequence[str],
+    timesteps: Optional[Sequence[cftime.DatetimeJulian]],
+    consolidated: bool = True,
+) -> Tuple[xr.Dataset, xr.DataArray]:
+    logger.info(f"Setting up dataset for state setting: {dataset_key}")
+    ds = _open_ds(dataset_key, consolidated)
```
brianhenn, line 168: Would have liked to call vcm.fv3.metadata.standardize_fv3_diagnostics here but can't because the time-rounding part of it doesn't work in the prognostic run image, so there's some cutting and pasting of its functionality.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        self._config = config
+        self._communicator = communicator
+        self._rename: Mapping[
+            Hashable, Hashable
+        ] = config.rename if config.rename else {}
+        self._timesteps = timesteps
+        self._prescribed_ds: xr.Dataset = self._scatter_prescribed_ds()
+
+    def _scatter_prescribed_ds(self):
```
oliverwm1, line 113:
```suggestion
    def _load_and_scatter_prescribed_ds(self):
```

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+    time_coord = ds.coords["time"]
+    return ds.drop_vars(names="time").load(), time_coord
+
+
+def _open_ds(dataset_key: str, consolidated: bool) -> xr.Dataset:
+    try:
+        zarr_path = CATALOG[dataset_key].urlpath
+    except KeyError:
+        zarr_path = dataset_key
+    return intake.open_zarr(zarr_path, consolidated=consolidated).to_dask()
```
oliverwm1, line 185: Is there a reason to use `intake.open_zarr` instead of `xr.open_zarr` here?

I think the typical pattern for opening a dataset with intake would be to do:

`ds = CATALOG[dataset_key].to_dask()`

which avoids grabbing the `urlpath` or baking in `zarr` format assumption in the code.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+def _round_time_coord(ds: xr.Dataset) -> xr.Dataset:
+    # done this way because vcm.round_time's singledispatch doesn't
+    # work in the prognostic_run container, which uses python 3.6.9
+    time_coord = ds.coords["time"]
+    rounded_time_coord = [round_time(time.item()) for time in time_coord]
+    return ds.assign_coords({"time": rounded_time_coord})
+
+
+def _add_net_shortwave(ds: xr.Dataset) -> xr.Dataset:
+    net_shortwave = ds["DSWRFsfc"] - ds["USWRFsfc"]
```
oliverwm1, line 266: hard-coded non-standard variable names here.

oliverwm1 (commented):
Thanks for the PR! Excited to do some runs with this capability soon.

I'm a little concerned about maintainability of the low-level data cleaning functions in `prescriber.py`. I think the wrapper/runtime code should know as little as possible about the Fortran world of the model (in terms of variable/dim names, etc.).

As an idea: what do you think about doing all this data cleaning as a separate one-off step, and basically requiring that the provided dataset has the same conventions as the wrapper in terms of naming and time coordinate? And as part of that, could do things like "compute net SW" which is a very specific function that doesn't really seem to belong here.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: My reason for not doing this is that the catalog approach doesn't allow for opening with `consolidated=True` (as far as I can tell?)

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+    ds = _open_ds(dataset_key, consolidated)
+    ds = _remove_name_suffix(ds)
+    ds = get_variables(ds, variables)
+    ds = _rename_dims(ds)
+    ds = _set_missing_units_attr(ds)
+    ds = _round_time_coord(ds)
+    if timesteps is not None:
+        ds = ds.sel(time=timesteps)
+    time_coord = ds.coords["time"]
+    return ds.drop_vars(names="time").load(), time_coord
```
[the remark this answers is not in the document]
brianhenn: This is a good strategic question...I think the approach here is probably only for setting 2D variables, and I'm not sure we have yet a use case for setting 3D variables. I'm hesitant to introduce another I/O call every timestep, since this already doesn't work all that great for the nudge-to-fine I/O, though I agree that with appropriate chunking maybe this is feasible.

workflows/prognostic_c48_run/prepare_config.py
```diff
@@ -145,7 +139,7 @@ def user_config_from_dict_and_args(config_dict: dict, args) -> UserConfig:
         )
 
     return UserConfig(
-        prephysics=prephysics,
+        prephysics=config_dict.get("prephysics"),
```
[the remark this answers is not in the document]
brianhenn: I think this does exactly what you're suggesting. The dict is converted to a dataclass since that `Union` is defined in the `UserConfig` attributes.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,274 @@
...
+    timesteps: Optional[Sequence[cftime.DatetimeJulian]],
+    consolidated: bool = True,
+) -> Tuple[xr.Dataset, xr.DataArray]:
+    logger.info(f"Setting up dataset for state setting: {dataset_key}")
+    ds = _open_ds(dataset_key, consolidated)
+    ds = _remove_name_suffix(ds)
+    ds = get_variables(ds, variables)
+    ds = _rename_dims(ds)
+    ds = _set_missing_units_attr(ds)
+    ds = _round_time_coord(ds)
```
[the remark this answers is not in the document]
brianhenn: I'm on board this with, would just need to do a quick PR to add a "clean" dataset to the catalog.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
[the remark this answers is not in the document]
brianhenn: On the other hand if we're running relatively short segmented runs then it's less of a problem to download everything once.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
oliverwm1: Got it. The catalog entries do contain a `consolidated` boolean as metadata, so I'd hope intake would use that, but I'm not sure TBH.

workflows/prognostic_c48_run/prepare_config.py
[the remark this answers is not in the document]
oliverwm1: I think you'd need an explicit `dacite.from_dict` call to convert it to a dataclass. But it doesn't end up mattering since we call `dataclasses.asdict(user_config)` in line 276 anyway (i.e. for the purposes of serializing, we want the UserConfig as a dict anyway).

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: Oh OK, will just let intake do that then.

workflows/prognostic_c48_run/prepare_config.py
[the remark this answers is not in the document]
brianhenn: OK I see the issue you're pointing out: the code as written sent a dict to the `UserConfig` did not check its contents via dacite. But the problem is that dacite can't call `from_dict` with the first argument being a type, it actually has to be a concrete class. That is frustrating. Doing what you suggest requires a hardcoded if statement to decide which type of config in the `Union` is there, which I added. The alternative would be to define a new class which has an attribute that is a `Union` of the two types (dacite can handle this; if you look at their Union example page that's how it's done), but that adds more complexity and confusion I think.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: Moot since code deleted

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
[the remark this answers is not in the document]
brianhenn: As discussed the clean reference dataset is at `gs://vcm-ml-intermediate/2021-03-fine-res-surface-radiative-fluxes/fine-res-surface-radiative-fluxes.zarr` now

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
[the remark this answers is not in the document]
brianhenn: Leaving this as is as discussed

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: Moot since deleted

workflows/prognostic_c48_run/prepare_config.py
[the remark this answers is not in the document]
oliverwm1: I see. Yeah... ideally we would just do a `dacite.from_dict(UserConfig, config_dict)` here—which is what is done at runtime—but the insertion of all the defaults makes this not straightforward. Maybe could change the order, i.e. generate the `UserConfig` dataclass using dacite, and then fill in the custom defaults as necessary.

Anyway, I don't think this should hold up the PR. I'm fine to merge as is... the prepare config script is already a mess of dicts and dataclasses, so this is not making it worse!

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,196 @@
...
+
+    dataset_key: str
+    variables: Sequence[str]
+    rename: Optional[Mapping[Hashable, Hashable]] = None
+    consolidated: bool = True
+
+
+class Prescriber:
+
+    net_moistening = "net_moistening"
```
oliverwm1, line 77: does this get used anywhere?

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -0,0 +1,196 @@
...
+    Attributes:
+        dataset_key (str): location of the dataset that provides prescribe values;
+            the routine first will try `dataset_key` as a `vcm.catalog` key, and if
+            not present it will attempt to use it as a (local or remote) path to
+            a zarr dataset
+        variables (Sequence[str]): sequence of "standardized" ("_coarse" suffix removed)
+            variable names to prescribe
+        rename (Mapping[Hashable, Hashable]): mapping of "standardized" ("_coarse"
+            suffix removed) names in the external dataset to variable names desired
+            for the runfile
```
oliverwm1, line 52: could you remove references to the `_coarse` suffix which no longer seems relevant. Also in principle we don't really need the renaming capability anymore but I'm okay to leave it in. I'm not really sure what `"standardized"` means here.

oliverwm1 (approved):
Thanks for the changes! Just a couple small comments but I don't need to review again. I'd consider just getting rid of the renaming functionality, but totally your call.

workflows/prognostic_c48_run/prepare_config.py
```diff
@@ -100,11 +101,17 @@ def user_config_from_dict_and_args(config_dict: dict, args) -> UserConfig:
         config_dict.get("namelist", {}).get("fv_core_nml", {}).get("nudge", False)
     )
 
-    prephysics: Optional[MachineLearningConfig]
-    if "prephysics" in config_dict:
+    prephysics: Optional[Union[MachineLearningConfig, PrescriberConfig]]
```
oliverwm1, line 104: hopefully we can replace this whole function with `dacite.from_dict(UserConfig, config_dict)` soon which will remove the necessity for all fo this.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: No, but it's a legacy requirement of the `Stepper` base class.

Issue #1048, opened 2021-02-26
Use new *Coarsen.reduce functionality for coarsening with custom reductions


spencerkclark:
Xarray 0.17.0 adds a `.reduce` method to `DataArrayCoarsen` and `DatasetCoarsen`, which natively enables coarsening DataArrays and Datasets with custom reduction functions.  This obviates some functions in vcm, namely [`xarray_block_reduce` and `horizontal_block_reduce`](https://github.com/VulcanClimateModeling/fv3net/blob/fd446563574e23978ac9137e1ac4e8232edc5c80/external/vcm/vcm/cubedsphere/coarsen.py#L463-L554) (as well as some associated private functions).  Some care will be needed to maintain dask compatibility for some operations, but it should hopefully be possible without too much trouble.

The main motivation for doing this is that it should allow us to drop the dependency on scikit-image.

spencerkclark:
@mcgibbon given that @nbren12 vendored the required functionality from scikit-image (removing our dependency on it) I consider this refactor to be low priority (if necessary at all).  I would be fine with closing this if we'd like to clean up the issue tracker.

mcgibbon:
If the refactor isn't necessary I'd say to close it to keep things clean, if you think this is something I should spend time on in let's say the next two months then it can stay open. If not I'd think it isn't likely to ever get done.

spencerkclark:
Yeah, I don't think this is worth spending your time on.  Let's close.

PR #1091, opened 2021-03-17
Vendorize skimage

nbren12:
Scikit-image is a difficult dependency to manage (especially since it makes strong assumptions about pooch). Even an import of scikit-image mutates the users ~/.cache folder, which can lead to hard to debug permissions errors and other issues. We work around this via careful dependency pinning, but I still frequently run into this problem when e.g. updating versions. We only use two small functions from scikit-image, so it's easy to vendorize them in this repo.

We can reimplement this using `xarray.coarsen` at another time (see #1048).

Added public API:
- vcm.testing.checksum_dataarray
- vcm.testing.checksum_dataarray_mapping

Significant internal changes:
- rewired the coarening tests to use regression checksums instead.

Requirement changes:
- removed scikit-image
- I would've liked to remove pooch too, but metpy uses it too, albeit only in one module that we probably don't import.


external/vcm/tests/test_cubedsphere.py
```diff
@@ -260,27 +260,16 @@ def input_dataset(input_dataarray):
...
+):
     if use_dask:
         input_dataarray = input_dataarray.chunk({"x": 2, "y": 2, "z": -1})
-
     block_sizes = {"x": 2, "y": 2}
     result = _xarray_block_reduce_dataarray(
         input_dataarray, block_sizes, reduction_function
     )
-    assert_identical_including_dtype(result, expected)
+    print(vcm.testing.checksum_dataarray(result), file=regtest)
```
spencerkclark, line 272: It looks like `checksum_dataarray` only takes the hash of the values of the DataArray (and ignores coordinate values and attributes).  Propagating attributes in particular was an issue with the coarsening functions in the past (#513), so I think we may want to revert this change (or modify the way `checksum_dataarray` operates).

spencerkclark (commented):
It's great this was so simple to vendor over.  I just have a minor comment about the modification to the test of `_xarray_block_reduce_dataarray`.

external/vcm/tests/test_cubedsphere.py
nbren12: yah i can do that.

spencerkclark (approved):

PR #1100, opened 2021-03-21
specify z dim in interpolating prognostic output

brianhenn:
Fixes two bugs in the prognostic run diagnostics data loading. Both stem from the switch to use the `state_after_timestep.zarr` instead of fortran outputs.
- specify `z` as the vertical dimension in the pressure interpolation (previously assumed fortran dim `pfull`)
- Keep a subset of the 3D output (T, q, dQ's, W). The previous bug was caused by the computation trying to interpolate everything in the output, and choking on the `x_wind, y_wind` interface variables that were saved in the state_after_timestep output. Also, the pressure interpolation is slow, so we should limit the number of variables computed here.

Verified that the fixes worked by generating this report on the run that initially ran into this issue: https://storage.googleapis.com/vcm-ml-public/annak/2021-03-24-debug-prog-diags-3d/output/index.html

### DIFF
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
@@ -124,7 +124,9 @@ def load_3d(
         pressure_vars = [var for var in ds.data_vars if "z" in ds[var].dims]
         for var in pressure_vars:
             ds_interp[var] = vcm.interpolate_to_pressure_levels(
-                field=ds[var], delp=ds["pressure_thickness_of_atmospheric_layer"]
+                field=ds[var],
+                delp=ds["pressure_thickness_of_atmospheric_layer"],
+                dim="z",
             )
 
         # open verification
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
@@ -95,22 +95,7 @@ def _load_prognostic_run_3d_output(url: str):
             zarr_name = os.path.basename(item)
             path = os.path.join(url, zarr_name)
             outputs.append(_load_standardized(path))
-        merged = xr.merge(outputs)
-        # interpolation is slow, so only keep a subset of 3D outputs
-        keep_vars = [
-            "air_temperature",
-            "specific_humidity",
-            "eastward_wind",
-            "northward_wind",
-            "vertical_wind",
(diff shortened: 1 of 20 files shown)


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
```diff
@@ -87,15 +87,29 @@ def _load_prognostic_run_3d_output(url: str):
     prognostic_3d_output = [
         item
         for item in fs.ls(url)
-        if item.endswith("3d.zarr") or item.endswith("state_after_timestep.zarr")
+        if item.endswith("diags_3d.zarr") or item.endswith("state_after_timestep.zarr")
```
AnnaKwa, line 90: `"diags_3d.zarr"` is the name that the default example saves ML tendencies and physics tendencies to.

brianhenn:
Thanks @AnnaKwa if you want to approve "my" PR and merge I'm good with that.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py
```diff
@@ -87,15 +87,30 @@ def _load_prognostic_run_3d_output(url: str):
...
     if len(prognostic_3d_output) > 0:
         outputs = []
         for item in prognostic_3d_output:
             zarr_name = os.path.basename(item)
             path = os.path.join(url, zarr_name)
             outputs.append(_load_standardized(path))
-        return xr.merge(outputs)
+        merged = xr.merge(outputs)
+        # interpolation is slow, so only keep a subset of 3D outputs
+        keep_vars = [
```
oliverwm1, line 100: could you change this subsetting to use the standard `subset_variables` transform that is used for all the other diagnostics? i.e. do it as part of the diagnostic calculation instead of the loading. Shouldn't make too much of a efficiency difference since data is loaded lazily (but let me know if the speed is hugely different... maybe even setting up the task graph for the interpolation is really slow?)

oliverwm1: e.g. like this: https://github.com/VulcanClimateModeling/fv3net/blob/149b545c5dd633cae746f938e84fb82906623bc0/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py#L244

AnnaKwa (approved):
Approving (as Brian since he opened the PR and I approved his contribution)

I will update the prognostic report test data with the correctly formatted version (previous version used fortran output instead of state_after_timestep) and rerun the test before merging.

PR #1101, opened 2021-03-23
Fix SST setting logic

### LINKED ISSUE #1037: nudge-to-fine workflow not updating SSTs
The nudging runs are supposed to update coarse-model SSTs with the reference state but the N2F workflow in the sklearn_runfile is missing the actual state update call after it gets the reference SSTs:

https://github.com/VulcanClimateModeling/fv3net/blob/f646892b3c7fc752efadae9d70df94a45e512489/workflows/prognostic_c48_run/runtime/steppers/nudging.py#L89

spencerkclark:
This PR restores the logic needed to update the sea surface temperature via the Python wrapper.

Resolves #1037

### DIFF
--- a/workflows/prognostic_c48_run/docs/config-usage.rst
+++ b/workflows/prognostic_c48_run/docs/config-usage.rst
@@ -63,6 +63,15 @@ Notice how these configurations correspond with
 :py:attr:`runtime.config.UserConfig.nudging`. Refer to those docs as a
 reference.
 
+In addition to the above settings, to prescribe the sea surface temperatures to
+the values in the reference restart files, the
+``gfs_physics_nml.use_climatological_sst`` namelist parameter must be set to
+``false``::
+
+    namelist:
+        gfs_physics_nml:
+            use_climatological_sst: false
+
 .. _nudge to obs:
 
 Nudge to obs
--- a/workflows/prognostic_c48_run/docs/config-usage.rst
+++ b/workflows/prognostic_c48_run/docs/config-usage.rst
@@ -63,15 +63,6 @@ Notice how these configurations correspond with
 :py:attr:`runtime.config.UserConfig.nudging`. Refer to those docs as a
 reference.
 
-In addition to the above settings, to prescribe the sea surface temperatures to
-the values in the reference restart files, the
-``gfs_physics_nml.use_climatological_sst`` namelist parameter must be set to
-``false``::
-
-    namelist:
-        gfs_physics_nml:
-            use_climatological_sst: false
-
 .. _nudge to obs:
 
 Nudge to obs
--- a/workflows/prognostic_c48_run/runtime/nudging.py
+++ b/workflows/prognostic_c48_run/runtime/nudging.py
@@ -238,8 +238,9 @@ def get_nudging_tendency(
 
 def get_reference_surface_temperatures(state: State, reference: State) -> State:
     """
-    Set the sea surface temperature model state to values in a reference state.
-    Useful for maintaining consistency between a nudged run and reference state.
+    Set the sea surface and surface temperatures in a model state to values in
+    a reference state. Useful for maintaining consistency between a nudged run
+    and reference state.
     """
     state = {
         SST_NAME: _sst_from_reference(
--- a/workflows/prognostic_c48_run/runtime/nudging.py
+++ b/workflows/prognostic_c48_run/runtime/nudging.py
@@ -250,11 +250,12 @@ def get_reference_surface_temperatures(state: State, reference: State) -> State:
             reference[TSFC_NAME], state[TSFC_NAME], state[MASK_NAME]
         ),
     }
+    return state
     # TODO fix this bug in a follow-up non-refactor PR
     # this logic replicates a bug in the previous nudged run
     # the SSTs were never actually applied to the fv3gfs-wrapper state
     # This bug could be scientifically significant.
-    return {}
+    # return {}
 
 
 def _sst_from_reference(
--- a/workflows/prognostic_c48_run/runtime/nudging.py
+++ b/workflows/prognostic_c48_run/runtime/nudging.py
@@ -242,6 +242,9 @@ def get_reference_surface_temperatures(state: State, reference: State) -> State:
     Useful for maintaining consistency between a nudged run and reference state.
     """
     state = {
+        TSFC_NAME: _sst_from_reference(
+            reference[TSFC_NAME], state[TSFC_NAME], state[MASK_NAME]
+        ),
         SST_NAME: _sst_from_reference(
             reference[TSFC_NAME], state[SST_NAME], state[MASK_NAME]
         ),
--- a/workflows/prognostic_c48_run/runtime/nudging.py
+++ b/workflows/prognostic_c48_run/runtime/nudging.py
@@ -242,11 +242,11 @@ def get_reference_surface_temperatures(state: State, reference: State) -> State:
     Useful for maintaining consistency between a nudged run and reference state.
     """
     state = {
-        TSFC_NAME: _sst_from_reference(
-            reference[TSFC_NAME], state[TSFC_NAME], state[MASK_NAME]
-        ),
         SST_NAME: _sst_from_reference(
             reference[TSFC_NAME], state[SST_NAME], state[MASK_NAME]
+        ),
+        TSFC_NAME: _sst_from_reference(
+            reference[TSFC_NAME], state[TSFC_NAME], state[MASK_NAME]
         ),
     }
     return state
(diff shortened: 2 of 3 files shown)


workflows/prognostic_c48_run/runtime/nudging.py
```diff
@@ -238,23 +238,15 @@ def get_nudging_tendency(
...
-    a reference state. Useful for maintaining consistency between a nudged run
-    and reference state.
+    Set the sea surface temperature model state to values in a reference state.
+    Useful for maintaining consistency between a nudged run and reference state.
     """
     state = {
         SST_NAME: _sst_from_reference(
             reference[TSFC_NAME], state[SST_NAME], state[MASK_NAME]
         ),
-        TSFC_NAME: _sst_from_reference(
```
spencerkclark, line 249: Note we do not need to override this variable, so I have removed it.

brianhenn: Thanks! I always wondered why this was here, glad to know that it didn't need to be.

brianhenn (commented):
Thanks @spencerkclark .

As @oliverwm1 mentioned this morning, nudge-to-fine runs are currently configured by adding a `nudging` section to the prognostic run user config. That still won't correctly set SSTs to track the fine-res run unless `use_climatological_sst: false` is also in the gfs_physics namelist. Can you update the docs to reflect this requirement in `fv3net/workflows/prognostic_c48_run/docs/config-usage.rst`?

workflows/prognostic_c48_run/docs/config-usage.rst
```diff
@@ -63,6 +63,15 @@ Notice how these configurations correspond with
...
 reference.
 
+In addition to the above settings, to prescribe the sea surface temperatures to
+the values in the reference restart files, the
+``gfs_physics_nml.use_climatological_sst`` namelist parameter must be set to
+``false``::
+
+    namelist:
+        gfs_physics_nml:
+            use_climatological_sst: false
```
oliverwm1, line 73: Does this also need to be set for making nudge-to-obs runs follow the analysis SSTs? I know the `use_analysis_sst` flag exists and needs to be set appropriately for those runs, but not sure how it interacts with this flag.

spencerkclark: It's a good question.

This flag, as currently written, toggles between the behavior before or after VulcanClimateModeling/fv3gfs-fortran#93.  It is possible that VulcanClimateModeling/fv3gfs-fortran#93 broke setting the SST to the analysis via the `use_analysis_sst` flag too, in which case you would want to set `use_climatological_sst = .false.` in those cases as well.  I'm currently running a couple tests to see.

mcgibbon: It's possible the flags act as "use_climatological_sst: use prescribed sst", and "use_analysis_sst: prescribed sst is climatology (false) or analysis (true)". Does use_analysis_sst make it so `tsfco` is the analysis sst?

spencerkclark: > Does use_analysis_sst make it so `tsfco` is the analysis sst?

Yes, this is why I am concerned.

spencerkclark: This turned out to basically be a non-issue; see analysis for the nudged-to-obs case in [this notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-03-23-test-sst-update/2021-03-25-sst-setting-n2o.ipynb).  When using `use_analysis_sst = .true.` both `Sfcprop%tsfco` and `Sfcprop%tsfc` are set to the analysis values.

*Technically* I think the most correct approach would be to set the new `use_climatological_sst` flag to `.false.` and only set the value of `Sfcprop%tsfco`; however, one can still achieve almost the same results leaving `use_climatological_sst = .true.` and setting both `Sfcprop%tsfco` and `Sfcprop%tsfc` (the results are not bitwise identical, but are close enough).  My understanding is that the surface cycling in the fortran model is weak, and only acts to slowly nudge `Sfcprop%tsfc` toward the climatology; if it is constantly overridden, then it stays very close to its overridden value.

oliverwm1 (approved):
Thanks for your careful work checking the behavior with these various flags!

This looks good to me. Can you confirm that the submodules are pointing to commits on master branch of respective repos? it's hard to tell from GitHub interface.

workflows/prognostic_c48_run/runtime/nudging.py
spencerkclark: The nudged-to-obs results got me thinking about this more and I tried another couple tests.  In these tests I included the update to `Sfcprop%tsfc` (`TSFC_NAME` in this context) back in, and tried cases with and without `use_climatological_sst = .true.`.  This time, in both runs, the SST was properly overridden.  

So, what this means is that, similar to the nudged-to-obs case, this was in fact not really particularly broken by VulcanClimateModeling/fv3gfs-fortran#93.  I think what VulcanClimateModeling/fv3gfs-fortran#93 meant was that you needed to override both `tsfc` and `tsfco` to get things to work, when before you only needed to override `tsfco`.  But anyway, I think only having to override one variable is clearer, so perhaps this exercise was useful in the end.

spencerkclark: I updated [a notebook](https://github.com/VulcanClimateModeling/explore/blob/master/spencerc/2021-03-23-test-sst-update/2021-03-23-verify-sst-setting.ipynb) to show the results of these experiments.  Out of the four options:

1. Set `use_climatological_sst = .true.` and only set the value of the `tsfco` variable in the wrapper/fortran.
2. Set `use_climatological_sst = .false.` and only set the value of the `tsfco` variable in the wrapper/fortran.
3. Set `use_climatological_sst = .true.` and set both `tsfco` and `tsfc` in the wrapper/fortran.
4. Set `use_climatological_sst = .false.` and set both `tsfco` and `tsfc` in the wrapper/fortran.

Only (1), which we haven't ever done anyway, clearly does not work.  (2), (3), and (4) all seem to produce reasonable results for the SST as a diagnostic.  (3) is what we did after VulcanClimateModeling/fv3gfs-fortran#93; (4) is what we did before.  (2) feels the cleanest, but is a departure from what we have done in the past.  In my message above I was sort of hopeful that (2) and (4) would have been equivalent, but that was not true.

After staring at the fortran a while, I really can't say which one of these options is "correct."  I almost wonder now whether we should just fall back to using (3) since it doesn't require making any namelist parameter changes to our configurations.  What do folks think?

Something that also comes up in this notebook are some odd SST differences from the verification data near the coast; I do not know where these are coming from, but I assume they have been in our runs from the beginning.

oliverwm1: Thanks for checking this all out. In general I think the confusions stems from there being more than one surface temperature variable in the Fortran code. I guess the reason for this is to allow for the "fractional grid" case?

Anyway, when I implemented the SST-setting in Fortran, it seemed safest to me to set both surface temperature variables, and it does seem like that's what we've discovered here. My sense is we should make a decision and move on. I agree option 3 is nice from the perspective of not requiring a new parameter.

spencerkclark:
> Can you confirm that the submodules are pointing to commits on master branch of respective repos? it's hard to tell from GitHub interface.

Yes, the submodules point to commits on the master branch of each repo.  For future reference you can look at the commit referenced in the compare side of the diffs of the submodules; e.g. follow the "updated [5 files](https://github.com/VulcanClimateModeling/fv3gfs-fortran/compare/75605127be63132c1d16fb6da6e70f60f88e2d40...e3f2abd275f64c582681cce0d373b9a7281aa18e)" link for the fv3gfs-fortran case.

> Anyway, when I implemented the SST-setting in Fortran, it seemed safest to me to set both surface temperature variables, and it does seem like that's what we've discovered here. My sense is we should make a decision and move on. I agree option 3 is nice from the perspective of not requiring a new parameter.

I agree we should move on.  I went with (3) since it is the simplest.  I will merge once the tests pass.

workflows/prognostic_c48_run/runtime/nudging.py
spencerkclark: Yeah, I think the changes that were made to accommodate the fractional grid case make things tricky to follow (the same reason it took a while to settle on VulcanClimateModeling/fv3gfs-fortran#93).

workflows/prognostic_c48_run/docs/config-usage.rst
spencerkclark: Two years later in refreshing myself on the discussion in this thread, I believe I linked to the wrong notebook in the above comment.  [This one](https://github.com/ai2cm/explore/blob/master/spencerc/2021-03-23-test-sst-update/2021-03-24-verify-sst-setting-n2o.ipynb) appears to be more polished and relevant to the discussion.

PR #1102, opened 2021-03-24
Add verification entries to catalog

AnnaKwa:
- Adds pressure interpolated 3D entries for `40day_may2020`, `free_plus_4K_ssts_c384_fv3gfs_20170801_20190801`, and `free_unpertubed_ssts_c384_fv3gfs_20170801_20190801`

- Update url for `40day_may2020` dycore variables to dataset with additional 2D variables saved (TMP200 etc.)

Tested with the 40 d verification, verification line for new 2D variables shows up ok and bias is calculated using the new verification data for pressure interpolated variables: https://storage.googleapis.com/vcm-ml-public/annak/2021-03-24-debug-prog-diags-3d/test-new-verif/index.html

### DIFF
--- a/external/vcm/vcm/catalog.yaml
+++ b/external/vcm/vcm/catalog.yaml
@@ -116,7 +116,9 @@ sources:
       consolidated: True
 
   40day_c48_pressure_interpolated_may2020:
-    description: 3D dynamical core diagnostics variables from 40-day nudged simulation (May 27 2020 SHiELD), coarsened to C48 resolution and interpolated to pressure levels.
+    description: >-
+      3D dynamical core diagnostics variables from 40-day nudged simulation (May 27 2020 SHiELD), coarsened to C48 resolution and interpolated to pressure levels. 
+      Generated with https://github.com/VulcanClimateModeling/explore/blob/master/annak/2021-03-23-save-c3072-verification.ipynb
     driver: zarr
     metadata:
       grid: c48
@@ -163,6 +165,23 @@ sources:
         - u200
         - v200
         - w500
+    args:
+      storage_options:
+        project: 'vcm-ml'
+        access: read_only
+      urlpath: "gs://vcm-ml-intermediate/2020-05-coarsen-c384-diagnostics/coarsen_diagnostics/atmos_8xdaily_C3072_to_C384.zarr"
+      consolidated: True
+
+  40day_c48_atmos_8xdaily_additional_vars_may2020:
+    description: >-
+      Additional pressure level 2D dynamical core diagnostics variables from 40-day nudged simulation (May 27 2020 SHiELD), coarsened to C48 resolution.
+      Generated with https://github.com/VulcanClimateModeling/explore/blob/master/annak/2021-03-23-save-c3072-2d-vars.ipynb
+    driver: zarr
+    metadata:
+      grid: c48
+      simulation: 40day_may2020
+      category: dycore
+      variables:
         - q1000
         - q850
         - q500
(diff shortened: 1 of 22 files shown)


external/vcm/vcm/catalog.yaml
```diff
@@ -142,11 +163,19 @@ sources:
...
+        - TMP500
+        - TMP200
+        - UGRD850
+        - UGRD200
     args:
       storage_options:
         project: 'vcm-ml'
         access: read_only
-      urlpath: "gs://vcm-ml-intermediate/2020-05-coarsen-c384-diagnostics/coarsen_diagnostics/atmos_8xdaily_C3072_to_C384.zarr"
+      urlpath: "gs://vcm-ml-intermediate/2020-05-coarsen-c384-diagnostics/coarsen_diagnostics/atmos_8xdaily_add_vars.zarr"
```
oliverwm1, line 178: I think it's confusing to change the url for a given catalog entry.

The prognostic run diagnostics allow having multiple catalog entries for a single "category" of diagnostics. So I would suggest just making a new zarr that only has the new derived variables, and then it will get merged with the existing one.

See:

https://github.com/VulcanClimateModeling/fv3net/blob/149b545c5dd633cae746f938e84fb82906623bc0/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/config.py#L5-L17

and 

https://github.com/VulcanClimateModeling/fv3net/blob/149b545c5dd633cae746f938e84fb82906623bc0/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/load_diagnostic_data.py#L21-L33

oliverwm1 (changes requested):
Thanks for generating these datasets! Will be great for our reports. One comment: I don't think we should be changing url's for existing catalog entries if possible.

external/vcm/vcm/catalog.yaml
AnnaKwa: Oh, neat, I didn't know it could do that.

oliverwm1:
Oh also, it would be great if you could provide a link to the notebook that generated these datasets in the catalog somewhere (in "description" is probably fine).

AnnaKwa:
Thanks @oliverwm1 the comments have been addressed. I won't merge this till the potential issue with the below-surface interpolation vs. extrapolation has been checked out.

oliverwm1 (approved):
Thanks for the changes!

I think ideally the dataset at `gs://vcm-ml-intermediate/2020-05-coarsen-c384-diagnostics/coarsen_diagnostics/atmos_8xdaily_add_vars.zarr` would only include the new variables and not the ones in the existing dataset. I'm actually not sure what xarray will do when we try to merge this dataset with the other one if there are overlapping variables.

AnnaKwa:
> Thanks for the changes!
> 
> I think ideally the dataset at `gs://vcm-ml-intermediate/2020-05-coarsen-c384-diagnostics/coarsen_diagnostics/atmos_8xdaily_add_vars.zarr` would only include the new variables and not the ones in the existing dataset. I'm actually not sure what xarray will do when we try to merge this dataset with the other one if there are overlapping variables.

Good catch, I meant to upload the version without the original variables.

AnnaKwa:
Tested with the 40 d verification, verification line for new 2D variables shows up ok and bias is calculated using the new verification data for pressure interpolated variables: https://storage.googleapis.com/vcm-ml-public/annak/2021-03-24-debug-prog-diags-3d/test-new-verif/index.html

I guess the merging of the dycore verifications with overlapping variables worked ok here, but I will fix the newer set to remove the redundant data.

PR #1103, opened 2021-03-25
Add training with backpropagation through time

mcgibbon:
This PR adds code necessary to train a model using backpropagation through time, and use it in our offline report and prognostic runs.

Please closely review the 3D support added for the packer and scaler, as the APIs have some limitations and I want to make sure their usage is clear enough.

Reviewers please note that I would suggest I move all files stored under `workflows/bptt` into the explore repo. I am keeping it in this PR only for the initial review, so there is some opportunity to get review feedback on this code.

Added public API:
- Added `BPTTModel`, `AllKerasModel`, and `StepwiseModel` to `fv3fit.keras`
- Added `ArrayPacker` and `StandardScaler` to `fv3fit`
- Added `Preloaded` class to `vcm`, which runs the next call to a function in a thread. Useful for loading data asynchronously while only loading one more result into memory at a time.

Internal changes:
- Added `unpack_layer(feature_dim: int)` and `pack_layer()` to `fv3fit.ArrayPacker`
- Normalizer now explicitly casts `self.mean` and `self.std` to be doubles, this is not a breaking change because we were already passing in doubles anywhere we used normalizers but adds safety
- if xgcm fails to import, the ImportError or numba.errors.TypingError will only be raised when xgcm is used, instead of when vcm is imported
- `storage_options` values for `project` and `access` are removed from several entries in catalog, since these were causing access to fail on my local machine. These options are only included because they were used in a reference example, and do not appear to be required.
- Fixed an error where `predict_columnwise` method of `Predictor` did not work if `predictor.sample_dim_name` was not `"sample"`

Refactored public API:
- fv3fit.load now supports recurrent-keras models.
- StandardScaler now supports 3D arrays (e.g. [sample, time, height])
- ArrayPacker now supports 3D arrays for packing, but not for unpacking

- [x] Tests added


external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -0,0 +1,547 @@
...
+            features = [packer.feature_counts[name] for name in packer.pack_names]
+            if series:
+                input_layers = [
+                    tf.keras.layers.Input(shape=[n_window, n_features])
+                    for n_features in features
+                ]
+            else:
+                input_layers = [
+                    tf.keras.layers.Input(shape=[n_features]) for n_features in features
+                ]
```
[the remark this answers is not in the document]
mcgibbon: Can't do this because `n_features` is not defined at the time of the if statement.

external/fv3fit/tests/test_io.py
```diff
@@ -29,6 +29,21 @@ class Mock2:
...
 
+def test_registering_subclass_get_name():
+    register = _Register()
+
+    @register("mock")
+    class Mock:
+        pass
+
+    @register("mock-subclass")
+    class MockSubclass:
```
frodre, line 40: `class MockSubclass(Mock)`?

The case I worked through was that there are two registered subclasses of `Mock` , `MockSubclassA` and `MockSubclassB`, would it return the name for `MockSubclassA` if asked? (It would)

I figured that would be a hairy debug process if there was weirdness with the loop logic and registering multiple subclasses of the same parent.

nbren12:
Can you split this PR up into some smaller pieces? It touches most of the repo.

mcgibbon:
> Can you split this PR up into some smaller pieces? It touches most of the repo.

This is effectively ready to merge, please no. Almost all major code changes are in fv3fit. The only other major change is adding Preloaded to `vcm`. All other changes are documentation updates and bugfixes to get various tools to work locally, or avoid crashes/failures if compiled components aren't available when they aren't being used.

The purpose of splitting it up would be to make reviewing easier, but @frodre has already put in the work to review this.

I am presently removing all code under `workflows/bptt`, and putting it in `explore`.

frodre (commented):
It looks like something in the linting in commit https://github.com/VulcanClimateModeling/fv3net/pull/1103/commits/4a6e33d32d50d80e3cac3582af9940ee8bd2e240 and the merge of the suggested code for the input layer building.

nbren12:
> This is effectively ready to merge, please no.

This is hard for me to evaluate given how large it is.

nbren12:
Where is vcm.Preloaded?

external/fv3fit/tests/test_io.py
mcgibbon: Ha good catch

mcgibbon:
Superceded by #1112

PR #1104, opened 2021-03-25
Add `-m mpi4py` to runfv3 python call

oliverwm1:
This change should ensure prognostic runs will properly exit if an error is raised on one rank but not on others.

see: https://mpi4py.readthedocs.io/en/stable/mpi4py.run.html


mcgibbon (approved):
LGTM, thanks for the fix!

PR #1106, opened 2021-03-25
Mask pressure level values if extrapolated below surface

AnnaKwa:
The fortran diagnostics and `vcm.interpolate_to_pressure_levels` handle pressure levels below the surface differently: the fortran diags extrapolates if surface pressure is less than the variable's pressure level (i.e. TMP850 has a finite value even if surface pressure < 850 hPa) while `vcm.interpolate_to_pressure_levels` leaves the value as NaN. Since the recently added 2D dycore verification variables in `vcm.catalog["40day_c48_atmos_8xdaily_additional_vars_may2020"]` use the vcm interpolation to save TMP850, etc. but the prognostic run uses the fortran diagnostics, this means the prognostic run report will be comparing different sets of points between the verification and the prognostic run for these variables.

This PR adds a `mask_values_below_sfc` transform step, which checks whether a subset of lower atmosphere variables at pressure levels P has values in columns with Psfc < P. Only columns that have Psfc >= P are kept for computing values in the report.

### DIFF
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
@@ -328,6 +328,7 @@ for variable_set in ["dycore", "physics"]:
     @add_to_diags(variable_set)
     @diag_finalizer("zonal_mean_value")
     @transform.apply("resample_time", "3H", inner_join=True)
+    @transform.apply("mask_values_below_sfc")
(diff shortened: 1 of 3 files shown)


AnnaKwa:
Closing without merging- the only python-interpolated variable that is significantly affected by this masking is `q1000`, which we do not currently use much when analyzing prognostic report outputs. To do the cell masking correctly, the area averaging would also have to be corrected to only use areas of unmasked cells, which change depending on the variables' pressure levels.

PR #1113, opened 2021-03-31
Prognostic run: allow prephysics state updates to be applied at postphysics

brianhenn:
State updates computed by the prephysics step, if not applied in the prephysics step, were previously ovewritten and could not applied in the postphysics step. This behavior was not intended as the steppers' state updates are intended to be independent of the step in which they are applied. It prevented setting `total_precipitation` via prescribed states or ML because that variable is set in the postphysics step. This PR fixes this issue by allowing state updates from the prephysics to be applied in the the postphysics step.


nbren12 (approved):
Thanks for this. I think all steppers should be able to add to state_updates, and this can be viewed as a bug fix.

PR #1114, opened 2021-03-31
prognostic_run: Make UserConfig.scikit_learn optional, simplify init

### LINKED ISSUE #1110: Documented way to set ML model urls is incorrect
The documentation generally assumes that the prepare_config will leave the minimal.yaml's ML configurations intact if no `--model_url` flag is provided:
1. https://vulcanclimatemodeling.com/docs/prognostic_c48_run/config-usage.html#machine-learning
2. https://github.com/VulcanClimateModeling/fv3net/blob/0f0b606d4e322ffda98cc4d29818d2688c796985/workflows/prognostic_c48_run/prepare_config.py#L69

But this is not the case: https://github.com/VulcanClimateModeling/fv3net/blob/0f0b606d4e322ffda98cc4d29818d2688c796985/workflows/prognostic_c48_run/prepare_config.py#L119

oliverwm1:
Make `UserConfig.scikit_learn` optional. Required some changes to check `if config.scikit_learn` instead of `if config.scikit_learn.model`.

This allowed simplifying the `user_config_from_dict_and_args` function by calling `dacite.from_dict(UserConfig, config_dict)` and then inserting command-line argument overrides and custom defaults.

Resolves #1110.

### DIFF
--- a/workflows/prognostic_c48_run/prepare_config.py
+++ b/workflows/prognostic_c48_run/prepare_config.py
@@ -100,63 +100,39 @@ def user_config_from_dict_and_args(config_dict: dict, args) -> UserConfig:
         config_dict.get("namelist", {}).get("fv_core_nml", {}).get("nudge", False)
     )
 
-    prephysics: Optional[MachineLearningConfig]
-    if "prephysics" in config_dict:
-        prephysics = dacite.from_dict(MachineLearningConfig, config_dict["prephysics"])
-    else:
-        prephysics = None
-
-    nudging: Optional[NudgingConfig]
     if "nudging" in config_dict:
         config_dict["nudging"]["restarts_path"] = config_dict["nudging"].get(
             "restarts_path", args.initial_condition_url
         )
-        nudging = dacite.from_dict(NudgingConfig, config_dict["nudging"])
-    else:
-        nudging = None
 
-    scikit_learn = MachineLearningConfig(
-        model=list(args.model_url or []), diagnostic_ml=args.diagnostic_ml
-    )
+    user_config = dacite.from_dict(UserConfig, config_dict)
 
-    if "diagnostics" in config_dict:
-        diagnostics = [
-            dacite.from_dict(DiagnosticFileConfig, diag)
-            for diag in config_dict["diagnostics"]
-        ]
-    else:
-        diagnostics = _default_diagnostics(
-            nudging, scikit_learn, nudge_to_observations, args.output_frequency,
+    # insert defaults and command line option overrides
+    if len(user_config.diagnostics) == 0:
+        user_config.diagnostics = _default_diagnostics(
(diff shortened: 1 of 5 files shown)


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -200,7 +200,9 @@ def __init__(
         self._prephysics_only_diagnostic_ml: bool = getattr(
             getattr(config, "prephysics"), "diagnostic_ml", False
         )
-        self._postphysics_only_diagnostic_ml: bool = config.scikit_learn.diagnostic_ml
+        self._postphysics_only_diagnostic_ml: bool = getattr(
```
nbren12, line 203: This is awkward. This is the downside of making MachineLearningConfig an optional. It essentially moves the conditional logic from `prepare_config.py` to here, although I think the simplifications to prepare_config are worth it in this case. Ideally this logic could be moved to where the `Steppers` are constructed https://github.com/VulcanClimateModeling/fv3net/blob/4fc74a307e6c88e12c738df9d8484b8f53499e0d/workflows/prognostic_c48_run/runtime/loop.py#L234-L234

workflows/prognostic_c48_run/prepare_config.py
```diff
@@ -101,80 +100,55 @@ def user_config_from_dict_and_args(config_dict: dict, args) -> UserConfig:
...
     nudging: Optional[NudgingConfig],
-    scikit_learn: MachineLearningConfig,
+    scikit_learn: Optional[MachineLearningConfig],
     nudge_to_obs: bool,
     frequency_minutes: int,
 ) -> List[DiagnosticFileConfig]:
     diagnostic_files: List[DiagnosticFileConfig] = []
 
-    if scikit_learn.model:
+    if scikit_learn:
```
nbren12, line 151: I always get confused about the true value of complex data structures. Can you use this instead?
```suggestion
    if scikit_learn is not None:
```

workflows/prognostic_c48_run/runtime/loop.py
nbren12: Not a required change, but something which could be addressed more elegantly. e.g. the MLStepper could be passed `do_diagnostic_ml` and change whether the ML outputs are put into "diagnostics" or "tendencies".

nbren12 (approved):
Approved, but I have a couple suggestions.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: Okay, definitely agree it would be cleaner to have the `do_diagnostic_ml` passed to the relevant Stepper, but this is a non-trivial change and I'm not sure our tests would catch a bug introduced by the change so I'm going to do it in a separate PR.

PR #1130, opened 2021-04-04
Interpolate sfc flux overrides

AnnaKwa:
This adds time interpolation to the state updates in the prognostic run `Prescriber` using the existing `time_interpolation_func` that was written for interpolating nudging reference states. Since this function is now used by multiple parts of the prognostic run, I moved it out of `runtime.nudging` and into a new `runtime.interpolation`.

Example longwave downward surface flux in prescribed vs nonprescribed runs with 450 second timesteps and 900 second reference data:
image
There is a bit of an offset because I forgot to save the diagnostics at 450s intervals, but you can see the prescriber is overriding the specified field.


workflows/prognostic_c48_run/runtime/interpolate.py
```diff
@@ -0,0 +1,56 @@
+import xarray as xr
```
AnnaKwa, line 1: No new code, this was moved from `nudging.py`

workflows/prognostic_c48_run/runtime/nudging.py
```diff
@@ -262,4 +221,4 @@ def _sst_from_reference(
         land_sea_mask.values.round().astype("int") == 0,
         reference_surface_temperature,
         surface_temperature,
-    ).assign_attrs(units=reference_surface_temperature.units)
+    ).assign_attrs(units=surface_temperature.units)
```
AnnaKwa, line 224: Changed because I ran into an instance where the C384 reference had no units.

workflows/prognostic_c48_run/tests/test_interpolate.py
```diff
@@ -0,0 +1,43 @@
+import pytest
```
AnnaKwa, line 1: Again, no new code here, this was all moved from `test_nudging.py`.

AnnaKwa:
@brianhenn I'm going to close this PR and reopen a new one to just use xarray as we discussed earlier.

PR #1132, opened 2021-04-05
Use inner join when merging nudge-to-fine state and tendencies

AnnaKwa:
See issue here: https://github.com/VulcanClimateModeling/fv3net/issues/1004

When the state and tendencies datasets from nudging runs are saved at different frequencies, the nudged-to-fine mapper keys can have empty data for some of the variables because the datasets are outer joined on time. This could lead to overestimating the size of training and test datasets, because some fraction of the timesteps used may be discarded due to NaNs existing in the samples. Since we are now using averaged nudging tendencies for all the nudged runs, this is a common issue in our data.

This PR changes the merge to use an inner join, so that mapper keys only exist for timesteps that the state, nudging tendencies, and physics tendencies are all saved at.

Resolves #<https://github.com/VulcanClimateModeling/fv3net/issues/1004>

### DIFF
--- a/external/loaders/loaders/mappers/_nudged/_nudged.py
+++ b/external/loaders/loaders/mappers/_nudged/_nudged.py
@@ -145,6 +145,7 @@ def open_nudge_to_fine(
                 "state_after_timestep.zarr",
             ],
             consolidated=consolidated,
+            join="inner",
         ).values()
     )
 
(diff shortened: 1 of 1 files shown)


brianhenn (approved):
Thanks for dealing with this @AnnaKwa

PR #1133, opened 2021-04-06
Add validation of time chunk size in runfv3 create

oliverwm1:
Configuring chunk and run length sizes that are valid for multi-segment runs can be challenging. This PR adds a validation of chunk sizes which ensures simulations will be able to run for multiple segments. The validation is done during the `runfv3 create` step, so only applies for segmented runs, not for `runfv3 run-native` runs.

Breaking changes:
- all segmented runs (even one-segment runs which don't currently have a restriction on chunk sizes) must pass this validation.

Significant internal changes:
- add `segmented_run` sub-package to `runtime` with `validate.py` and `cli.py` modules
- Define a python CLI in `runtime.segmented_run.cli` that is used within `runfv3 create`

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/docs/config-usage.rst
+++ b/workflows/prognostic_c48_run/docs/config-usage.rst
@@ -167,5 +167,15 @@ Diagnostics to be output by the Fortran model are specified in the
 to the Fortran ``diag_table`` representation of diagnostics (see fv3config_ docs).
 
 
+Chunking
+^^^^^^^^
+
+The desired chunking can be specified for each diagnostic file to be output. For
+segmented runs (:ref:`segmented-run-cli`) there is a requirement that the chunk size
+along the time dimension evenly divide the length of the time dimension for the
+given output. Segmented runs will raise an exception during initialization if this
+requirement is not met.
+
+
 .. _fv3config: https://fv3config.readthedocs.io/en/latest/
 .. _fv3fit: https://vulcanclimatemodeling.com/docs/fv3fit/
--- a/workflows/prognostic_c48_run/runfv3
+++ b/workflows/prognostic_c48_run/runfv3
@@ -42,7 +42,7 @@ function createRun {
     fv3config="$2"
     runfile="$3"
 
-    create_run $url $fv3config $runfile
+    segmented-run-create $url $fv3config $runfile
 }
 
 function readRunConfig {
--- a/workflows/prognostic_c48_run/runfv3
+++ b/workflows/prognostic_c48_run/runfv3
@@ -106,7 +106,11 @@ function appendSegment {
     fv3ExitCode=$?
     set -e
 
-    post_process_run --chunks "$rundir/chunks.yaml" "$rundir" "$postProcessedOut"
+    echo "Skipping upload of following files:"
+    cat "$rundir/preexisting_files.txt"
+    post_process_run --chunks "$rundir/chunks.yaml" \
+        --skip "$rundir/preexisting_files.txt" \
+        "$rundir" "$postProcessedOut"
     append_run --no-copy "$postProcessedOut" "$runURL"
 
     echo "Cleaning up working directory"
@@ -136,6 +140,7 @@ function runSegment {
     cp "$runfile" "$rundir/runfile.py"
     (
         cd "$rundir"
+        find . > preexisting_files.txt
         NUM_PROC=$(yq '.namelist.fv_core_nml.layout | .[0] *.[1] * 6' "$fv3config")
         mpirun -n "$NUM_PROC" python3 -m mpi4py "$runfile" |& tee -a "$rundir/logs.txt"
     )
--- a/workflows/prognostic_c48_run/runfv3
+++ b/workflows/prognostic_c48_run/runfv3
@@ -42,18 +42,7 @@ function createRun {
     fv3config="$2"
     runfile="$3"
 
-
-    if [[ $# != 3 ]]
-    then 
-        usage
-        exit 1
-    fi
-
-    verifyUrlEmpty "$url"
-
-    echo "Setting up iterations"
-    gsutil cp "$fv3config" "$url/fv3config.yml"
-    gsutil cp "$runfile" "$url/runfile.py"
+    create_run $url $fv3config $runfile
 }
 
 function readRunConfig {
--- a/workflows/prognostic_c48_run/runtime/diagnostics/manager.py
+++ b/workflows/prognostic_c48_run/runtime/diagnostics/manager.py
@@ -38,7 +38,7 @@ class DiagnosticFileConfig:
     name: str
     variables: Optional[Container] = None
     times: TimeConfig = dataclasses.field(default_factory=lambda: TimeConfig())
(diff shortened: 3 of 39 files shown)


workflows/prognostic_c48_run/runtime/diagnostics/manager.py
```diff
@@ -38,7 +38,7 @@ class DiagnosticFileConfig:
     name: str
     variables: Optional[Container] = None
     times: TimeConfig = dataclasses.field(default_factory=lambda: TimeConfig())
-    chunks: Optional[Mapping[str, int]] = None
+    chunks: Mapping[str, int] = dataclasses.field(default_factory=dict)
```
oliverwm1, line 41: empty dict as a default is better than having the chunks be optional

workflows/prognostic_c48_run/runfv3
```diff
@@ -42,18 +42,7 @@ function createRun {
...
-        usage
-        exit 1
-    fi
-
-    verifyUrlEmpty "$url"
-
-    echo "Setting up iterations"
-    gsutil cp "$fv3config" "$url/fv3config.yml"
-    gsutil cp "$runfile" "$url/runfile.py"
+    create_run $url $fv3config $runfile
```
oliverwm1, line 45: if/when we refactor this script to python, I imagine we would define a python CLI `runfv3` which has subparsers `create`, `append` etc. That will be backwards compatible. For now, the bash script is still the public entrypoint and this `create_run` CLI is mean to be private.

workflows/prognostic_c48_run/docs/config-usage.rst
```diff
@@ -167,5 +167,15 @@ Diagnostics to be output by the Fortran model are specified in the
 to the Fortran ``diag_table`` representation of diagnostics (see fv3config_ docs).
 
 
+Chunking
+^^^^^^^^
+
+The desired chunking can be specified for each diagnostic file to be output. For
```
nbren12, line 173: Usage docs are great, but more likely to become out of date since someone editing the code won't see them.

It doesn't look like this is mentioned in the reference API. Please update the docstrings of FortranFileConfig and DiagnosticsFileConfig, and defer as much of this discussion as possible to the reference APIs.

workflows/prognostic_c48_run/runfv3
nbren12, line 45: I suggest prefixing the command name with  `segmented-run` since "run" is overloaded, and I envision the segmented run actually being a python class with methods at some point. Feel free to change to underscore, but I think it's uglier and required shift when typing, so is harder on my pinkies.

```suggestion
    segmented-run-create $url $fv3config $runfile
```

workflows/prognostic_c48_run/runtime/validate.py
```diff
@@ -0,0 +1,51 @@
+from datetime import datetime, timedelta
```
nbren12, line 1: Since this validation is unique to segmented runs, perhaps we should scope this under a subpackage `segmented_run` containing this module and `segmented_run.py`.

workflows/prognostic_c48_run/runtime/validate.py
```diff
@@ -0,0 +1,51 @@
...
+from .config import UserConfig, DiagnosticFileConfig
+from .diagnostics.fortran import FortranFileConfig
+
+
+class ConfigValidationError(ValueError):
+    pass
+
+
+def validate_config(config_dict: Mapping[str, Any]):
+    """Validate configuration for segmented runs. Currently just validates chunks."""
```
nbren12, line 15: YAGNI? Please delete this unused layer.

workflows/prognostic_c48_run/runtime/validate.py
```diff
@@ -0,0 +1,51 @@
...
+    validate_chunks(config_dict)
+
+
+def validate_chunks(config_dict: Mapping[str, Any]):
+    """Ensure that the time chunk size evenly divides the time dimension size for all
+    output diagnostics. Raise ConfigValidationError if not."""
+    user_config = dacite.from_dict(UserConfig, config_dict)
```
nbren12, line 22: Fine for now, but it would be nice if validate_chunks took in a `UserConfig`. Ideally `dacite` would only be used once in the code base. We should think about formalizing the distinction between fv3config and python configuration better, although maybe this feature is unique in needing to know about both.

workflows/prognostic_c48_run/tests/test_validate.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.fixture
+def timestep():
+    return timedelta(minutes=15)
+
+
+@pytest.fixture
+def run_duration():
+    return timedelta(hours=12)
```
nbren12, line 20: It seems unnecessary to use fixtures for these simple constants. Can you just hardcode in the function for now, and parameterize later if you need to?

nbren12 (approved):
It's been a long time coming. I'm sure this feature will be helpful once we start working with run lengths other than 40 days (as with the new SHieLd run). 

Some requested changes below, but hitting improve since they are pretty minor. Please check in with me if you don't agree on any point.

workflows/prognostic_c48_run/docs/config-usage.rst
oliverwm1: Hmm. I don't think the FortranFileConfig etc should know about requirements for the segmented run, which is the main point I'm trying to make here. Ideally, at some point there might be a "segmented run" class which could have this info in its reference API, but for now I think it needs to be in the usage docs.

workflows/prognostic_c48_run/runtime/validate.py
oliverwm1: I think it's pretty typical to need to know about the namelist for runtime configuration.

workflows/prognostic_c48_run/runtime/validate.py
oliverwm1: Done, made `segmented_run.cli` and `segmented_run.validate`

workflows/prognostic_c48_run/docs/config-usage.rst
nbren12: Ok. I agree this is scoped to the segmented run rather than the runtime requirements. It just feels off to have these specific details listed in this general config usage section, while leaving it undocumented here: https://github.com/VulcanClimateModeling/fv3net/blob/29a3fa839b8ee90f9dc94bc95b22cde6d7796704/workflows/prognostic_c48_run/docs/local-execution.rst#segmented-runs.
 
You already have a cross reference to `segmented-run-cli`. Can you move these details there, and just provide a cross reference without repeating the info? 
```suggestion
The desired chunking can be specified for each diagnostic file to be output. For
.. note :: Not all chunks are valid for segmented runs. See :ref:`segmented-run-cli` for more details.
```

workflows/prognostic_c48_run/runtime/validate.py
nbren12: Maybe open an issue then? It would be shame to start passing around the whole config dict more places.

PR #1140, opened 2021-04-08
Save offline diags to subdir with item name

AnnaKwa:
When multiple models are trained using the train-diags-prog template, their individual offline reports should get saved to subdirs in the report output dir. Right now the subdirectories aren't specified in the template offline diags public output step, so the first report to complete is saved to the top level of the public output and the rest are uploaded as tmpdirs with random names into that directory.


spencerkclark (approved):
Good catch

PR #1141, opened 2021-04-08
delete redundant entries in environment.yml

nbren12:
A lot of the packages in the environment.yml are already managed by pip (and present in the constraints.txt). This increases the size of the fv3net image since these packages are installed twice. `pip` should be favored when it works.


AnnaKwa (approved):
LGTM

PR #1142, opened 2021-04-09
prog run report: separate loading and rendering

nbren12:
the main method of the report generation was long.

- Refactor all code involved in loading data from the diags.nc
  files into a new module `prognostic_run/rundirs.json`.
- group plot generation and html generation into new "render" functions
- divide the html link generation into `find` and `render` phases

The output appears unchanged:  [output.tar.gz](https://github.com/VulcanClimateModeling/fv3net/files/6282462/output.tar.gz)


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/rundirs.py
```diff
@@ -0,0 +1,223 @@
...
+
+PUBLIC_GCS_DOMAIN = "https://storage.googleapis.com"
+
+
+@dataclass
+class ComputedDiagnosticsList:
+    """Represents a list of computed diagnostics
+
+    Attrs:
+        url: URL to directory containing rundirs as subdirectories.
```
oliverwm1, line 26: I know this was the preexisting language used, but `rundirs` feels like the wrong name... these are directories with reduced diagnostics/metrics/movies in them, not the actual directories with model run output. Importantly, these paths are different than the ones specified in `rundirs.json`, which has confused me in the past.

Maybe `diagdirs`?

oliverwm1 (commented):
Very useful refactor, thanks!

Can we change the `rundirs` name? I know it was already in place, but naming a new python module after this misleading name doesn't seem great.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/rundirs.py
nbren12: Renamed to "computed_diagnostics.py" based on the class name.

oliverwm1 (approved):

PR #1144, opened 2021-04-09
fix minor prognostic_run docs issues

oliverwm1:
Update incorrect docstring and include a missing dataclass `FortranTimeConfig` in the docs. Also remove some redundant info mistakenly introduced in #1133 .


AnnaKwa (approved):
LGTM

PR #1145, opened 2021-04-09
Ensure units of prescribed states in postphysics

brianhenn:
Setting states from ML/prescribed sources in postphysics did not check to ensure that predicted arrays had units. This prevented predicted quantities from being set in postphysics because our ML models don't provide this. This PR resolves this by moving the units/attrs updating logic to the derived state update method, so that it doesn't have to repeated in different parts of the loop. Also renamed the state updates from predicted tendencies variable to help distinguish them.

### DIFF
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -130,15 +130,6 @@ def add_tendency(state: Any, tendency: State, dt: float) -> State:
     return updated  # type: ignore
 
 
-def assign_attrs_from(src: Any, dst: State) -> State:
-    """Given src state and a dst state, return dst state with src attrs
-    """
-    updated = {}
-    for name in dst:
-        updated[name] = dst[name].assign_attrs(src[name].attrs)
-    return updated  # type: ignore
-
-
 class LoggingMixin:
 
     rank: int
@@ -383,8 +374,7 @@ class TimeLoop(Iterable[Tuple[cftime.DatetimeJulian, Diagnostics]], LoggingMixin
         self._log_debug(
             f"Applying prephysics state updates for: {list(state_updates.keys())}"
         )
-        updated_state = assign_attrs_from(self._state, state_updates)
-        self._state.update_mass_conserving(updated_state)
+        self._state.update_mass_conserving(state_updates)
 
         return diagnostics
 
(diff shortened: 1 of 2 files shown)


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -449,15 +449,18 @@ def _apply_postphysics_to_dycore_state(self) -> Diagnostics:
...
+                )
+                updated_state_from_tendency[TOTAL_PRECIP] = precipitation_sum(
                     self._state[TOTAL_PRECIP],
                     diagnostics[self._postphysics_stepper.net_moistening],
                     self._timestep,
                 )
-                diagnostics[TOTAL_PRECIP] = updated_state[TOTAL_PRECIP]
+                diagnostics[TOTAL_PRECIP] = updated_state_from_tendency[TOTAL_PRECIP]
+                self._state.update_mass_conserving(updated_state_from_tendency)
+                updated_state = assign_attrs_from(self._state, self._state_updates)
```
nbren12, line 462: This fix would be more explicit if it only assigned the units rather than other attrs. I would also consider pushing this logic into `self._state.update_mass_conserving` so we don't need to copy paste it again.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -449,15 +449,18 @@ def _apply_postphysics_to_dycore_state(self) -> Diagnostics:
...
+                    self._state, tendency, dt=self._timestep
+                )
+                updated_state_from_tendency[TOTAL_PRECIP] = precipitation_sum(
                     self._state[TOTAL_PRECIP],
                     diagnostics[self._postphysics_stepper.net_moistening],
                     self._timestep,
                 )
-                diagnostics[TOTAL_PRECIP] = updated_state[TOTAL_PRECIP]
+                diagnostics[TOTAL_PRECIP] = updated_state_from_tendency[TOTAL_PRECIP]
+                self._state.update_mass_conserving(updated_state_from_tendency)
```
nbren12, line 461: Can you use one call to `self._state.update_mass_conserving`? The way it is currently written could imply it is important to split into two calls (e.g. it is relying on some state inside of `self._state`. This is harder to reason about.

workflows/prognostic_c48_run/runtime/loop.py
nbren12: Honestly, `Quantity` should not require `unit` IMO. It is just a landmine to step on with no tangible benefit.

nbren12 (approved):

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Refactored to `self._state.update_mass_conserving`

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: On second thought it _is_ currently important that there are two calls: when `total_precipitation` is predicted by prephysics ML, it's in both `updated_state_from_tendency` and in `self._state_updates`, and it's the one in the latter we actually want to use, which is why there's currently two calls with that going second. This is confusing and needs to get resolved, but I'm hesitant to overhaul this RN when it's the crux of the experiment we're about to run.

PR #1146, opened 2021-04-09
prog run report: use RunDiagnostics container

nbren12:
This refactors the plotters to exclusively use RunDiagnostics containers.
This makes the interface of the plotting functions more consistent, and
enables a better separation of concerns.

Improve the performance of RunDiagnostics.get_variable to make this feasible.
My changes caused an error if any of the runs don't have a .run attribute.
This class should not be used with data that doesn't have this attribute
anyway, so I fixed the test data.


oliverwm1 (approved):
LGTM

PR #1148, opened 2021-04-12
prognostic_run: Cast ML predictions to double

AnnaKwa:
The states updates done in the python wrapper must have double precision. In the event that the ML model outputs with single precision, the prognostic run will fail (`File "lib/_wrapper.pyx", line 443, in fv3gfs.wrapper._wrapper.set_state
ValueError: Buffer dtype mismatch, expected 'REAL_t' but got 'float'`). This PR adds a step to the end of the ML steppers to cast arrays with np.float32 values into np.float64.

### DIFF
--- a/workflows/prognostic_c48_run/docker-compose.yml
+++ b/workflows/prognostic_c48_run/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - GOOGLE_APPLICATION_CREDENTIALS=/tmp/key.json
 
   fv3net_dev:
-    image: us.gcr.io/vcm-ml/prognostic_run:latest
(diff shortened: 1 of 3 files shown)


nbren12 (changes requested):
This fix applies to any python update, not just ones from the MLStepper. Can you please move it to `DerivedState__setitem__` and `DerivedState.update_mass_conserving` instead?  Thanks.

nbren12 (approved):
Please add the casting operation here as well: https://github.com/VulcanClimateModeling/fv3net/blob/a0c181c475b877841581e0f56069688be0048f1b/workflows/prognostic_c48_run/runtime/derived_state.py#L89.

`update_state_mass_conserving` is not the only public method that can be used to set state.

Thanks.

PR #1149, opened 2021-04-13
Add fv3fit requirements to inputs for `make lock_pip`

### LINKED ISSUE #1118: Adding numba to pip-requirements.txt
It would be good to add numba to pip-requirements, since it sometimes gets installed locally either directly or as a dependency of manually-installed tools. This would lock the version which is good because it is an optional dependency of some of our dependencies.

When running `make lock_deps` with numba added to pip-requirements, however, several dependencies disappear from `constraints.txt`, most notably `scikit-learn`. It is not clear why this happens, and it should be fixed before re-building the constraints file.

oliverwm1:
Use `external/fv3fit/setup.py` when building `constraints.txt` since this is where our ML dependencies are listed.

Requirement changes:
- Added `numba` to pip-requirements.txt so that a consistent version will be installed when necessary
- Some additional changes to the `constraints.txt` occur, possibly related to my environment or possibly related to other changes in listed dependencies since the last time a rebuilt version of `constraints.txt` was checked in

Resolves #1118


constraints.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --no-annotate --output-file=constraints.txt docker/prognostic_run/requirements/base.txt docker/prognostic_run/requirements/notebook.txt external/fv3fit/requirements.txt external/fv3kube/setup.py external/vcm/setup.py pip-requirements.txt workflows/post_process_run/requirements.txt workflows/prognostic_c48_run/requirements.in
```
oliverwm1, line 5: I don't think this existing version of `constraints.txt` was generated with the current fv3net master since the Makefile `make lock_pip` rule there does not list `external/fv3fit/requirements.txt`

mcgibbon (approved):
LGTM

PR #1151, opened 2021-04-13
Prognostic run: Improve test coverage of diagnostics

nbren12:
These changes should make it easier to modify the diagnostics code with confidence.


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -447,10 +447,8 @@ def _apply_postphysics_to_dycore_state(self) -> Diagnostics:
                     diagnostics[self._postphysics_stepper.net_moistening],
                     self._timestep,
                 )
-                diagnostics[TOTAL_PRECIP] = updated_state_from_tendency[TOTAL_PRECIP]
```
brianhenn, line 450: grepping `total_precipitation` in fv3net shows that this diagnostic is weakly (non-breaking probably, but potentially confusing) assumed to be in the diagnostics by various other routines, e.g., it's in at least `workflows/prognostic_c48_run/runtime/diagnostics/default_diagnostics.py` and the end-to-end test diags specification. 

Seems reasonably in scope of this PR to make a clean break of this, presumably in favor of `total_precipitation_rate`. At least in the code -- the various test fixture data schema are inherently going to end up out of date so probably not worth it there.

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -485,29 +489,21 @@ def test_chunks_present(completed_rundir):
...
-        "net_heating",
-        "net_moistening",
-        "physics_precip",
-        "water_vapor_path",
-        "column_integrated_dQu",
-        "column_integrated_dQv",
-    ]:
-        assert diagnostics[variable].dims == dims
+    for variable in sorted(diagnostics):
         assert np.sum(np.isnan(diagnostics[variable].values)) == 0
```
brianhenn, line 499: Is this still needed if we have checksums now?

brianhenn (commented):
Thanks this is useful. I had one suggestion about cleaning up some of the `total_precipitation` cruft elsewhere in the repo

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out
```diff
@@ -0,0 +1,37 @@
+air_temperature: 41ccaafb84a9f1b3e2195eb703974c5c
```
Comment (brianhenn) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out:1]: How does the list of variables specified in the test script to be written to `diags.zarr` get expanded to this larger list?

workflows/prognostic_c48_run/runtime/loop.py
nbren12: This diagnostics was deleted below, and so is never used. The tests would have revealed an error like this. In the first commit, I wrote out all the available diagnostics and check-summed them, before implementing this change in the second commit.

workflows/prognostic_c48_run/tests/test_regression.py
nbren12: The checksums don't provide much guidance about what the potential difference is so it's nice to leave this check.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out
Comment (nbren12) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out:1]: I modified "diags.zarr" to save all the available diagnostics, and then looped through them in the test. If we somehow delete one by accident it will be missing in the regression report and the test will fail.

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -32,6 +33,9 @@
             "physics_precip",
             "column_integrated_dQu",
             "column_integrated_dQv",
+            "cnvprcp_after_python",
+            "area",
+            "total_precipitation_rate",
```
brianhenn, line 38: Does this list actually do anything then since it's called `output_variables` instead of `variables`? Visually I assumed it was doing what it would do if it were called `variables`. If that's actually the case then please delete this list.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Sure, that's fine. I guess this is another reminder why specifying default config in code is not a very maintainable approach.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out
Comment (brianhenn) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[keras].out:1]: OK thanks. See my comment about the "other_variables" list in the test script then.

workflows/prognostic_c48_run/tests/test_regression.py
brianhenn: OK makes sense

brianhenn (approved):
OK looks good, see my one request below about the (I think) distracting list of diags variables in the test script.

workflows/prognostic_c48_run/tests/test_regression.py
nbren12: Good point. I deleted it.

PR #1152, opened 2021-04-14
Delete unused 'assets', 'configurations', regression_test.sh

oliverwm1:
Delete all files in "assets" and "configurations" directories. These have not been touched in >17 months and are not used by any current code. They have been effectively replaced by the "vcm-workflow-control" repo.

Also delete `regression_tests.sh` file in root level of fv3net.


nbren12 (approved):
Looks good. the prog run report test error is unrelated.

PR #1154, opened 2021-04-15
Prognostic run report: hardcode color limits for zonal avg - pressure bias figures

AnnaKwa:
This PR changes the matplotlib figure code in the prognostic run diags to group runs for the same variable together so they can be plotted with a single color bar in a more condensed figure.

Right now the bias plots are contour plots and the Hovmoller and time averaged lat-pressure profiles are pcolormesh. I'm open to preferences on what plot to use for different quantities.

I left the original matplotlib plotting function in there in case it is better suited for some figures, or if people feel the table of individual images is preferable it is easy to revert.

Example:
https://storage.googleapis.com/vcm-ml-public/annak/2021-04-13-test-prog-diags-contourf-6/hovmoller.html
https://storage.googleapis.com/vcm-ml-public/annak/2021-04-13-test-prog-diags-contourf-6/zonal_pressure.html


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/matplotlib.py
```diff
@@ -50,8 +53,104 @@ def __repr__(self):
...
+    values, run_coords = [], []
+    for run in run_diags.runs:
+        da = run_diags.get_variable(run, var)
+        if not np.isnan(da.mean(skipna=True)):
+            values.append(da)
+            run_coords.append(run)
+    return xr.concat(values, dim="run").assign_coords({"run": run_coords})
+
+
+def plot_2d_matplotlib_groupby_run(
```
AnnaKwa, line 86: I just realized this is doing the opposite of groupby... I will change the name.

nbren12:
The panels look better than before, but I worry that grouping multiples runs into one png will be less flexible and scalable to many runs. It is nice to keep run/variables combinations in separate .pngs for so that an individual run can be copy-pasted, or so that we could group runs on the fly using javascript. 

Would you mind exploring doing the compositing in html/css? Something like this would probably come close to what you did before:
```
<div class="runs", style="float: left;">
<center>{variable name}</center>
<table><tr><td>....</table>
</div>
```
I think this also conflicts with some of the changes Oli just did.

nbren12:
I would also use `contourf` rather than`contour`.

nbren12 (approved):
Looks great, and thanks for keeping the diff small.

PR #1155, opened 2021-04-15
add vcm.cubedsphere.to_cross

nbren12:
This code was previously in my explore repo, and is pretty handy
for quickly visualizing cubed sphere data.

I used regression testing/hashing since it is hard to verify this works
without the grid data. You'll have to just trust me that this does
correctly unrotate the fields or try it out yourself.

Added public API:
- `vcm.cubedsphere.to_cross`

- [x] Tests added

### DIFF
--- a/external/fv3viz/examples/plot_cross.py
+++ b/external/fv3viz/examples/plot_cross.py
@@ -3,7 +3,8 @@
 ==========================
 
 The routine ``vcm.cubedsphere.to_cross`` is useful for quick plots when
-cartopy may not be installed.
+cartopy may not be installed and/or latitude and longitude coordinates
+are not easily available.
 """
 
 import vcm
--- a/external/fv3viz/examples/plot_cross.py
+++ b/external/fv3viz/examples/plot_cross.py
@@ -1,0 +1,20 @@
+"""
+"Cross" plot
+==========================
+
+The routine ``vcm.cubedsphere.to_cross`` is useful for quick plots when
+cartopy may not be installed.
+"""
+
+import vcm
+import fsspec
+
+url = "https://github.com/VulcanClimateModeling/vcm-ml-example-data/blob/main/fv3net/fv3viz/plot_2_plot_cube_prognostic_diags.nc?raw=true"  # noqa
+fs = fsspec.get_fs_token_paths(url)[0]
+ds = vcm.open_remote_nc(fs, url).load()
+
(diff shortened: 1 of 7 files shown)


external/vcm/vcm/cubedsphere/cross.py
```diff
@@ -0,0 +1,97 @@
...
+            for tile_num, spec in TOPOLOGY.items():
+                if spec.y == j and spec.x == i:
+                    arr = rotate(
+                        data.isel({tile: tile_num}), spec.origin, SW, dims=dims
+                    )
+                    row.append(arr)
+                    break
+            else:
+                arr = null
+                row.append(arr)
```
spencerkclark, line 94: Would it be cleaner for the `Tile` class to have attributes of just `tile` and `origin`, and the `TOPOLOGY` dictionary to have tuple keys of the column and row indices?  Then I think this logic could be somewhat simpler:
```python
if (i, j) in TOPOLOGY:
    spec = TOPOLOGY[(i, j)]
    arr = rotate(data.isel({tile: spec.tile}), spec.origin, SW, dims=dims)
else:
    arr = null
row.append(arr)
```

external/fv3viz/examples/plot_cross.py
```diff
@@ -0,0 +1,21 @@
+"""
+"Cross" plot
+==========================
+
+The routine ``vcm.cubedsphere.to_cross`` is useful for quick plots when
+cartopy may not be installed.
```
spencerkclark, line 6: Having to load and merge the grid coordinates from another datasource is often a pain too:
```suggestion
cartopy may not be installed and/or latitude and longitude coordinates
are not easily available.
```

spencerkclark (commented):
I've seen enough examples of this in use to trust that it works.  This is a very nice addition -- I know going forward I'll be using it in favor of my current quick-plot approach of faceting over the tiles.

I just had a small implementation suggestion.

external/vcm/vcm/cubedsphere/cross.py
nbren12: I agree it is more convenient to use `(i,j)` as the key in this case, but I do like that `Tile(i, j, origin)` is independent of the particular coordinate value of `tile` that it is assigned to. I parse the TOPOLOGY datastructure into a new `rotation_plan` and used your suggestion otherwise.

spencerkclark (approved):

PR #1156, opened 2021-04-15
prog_run: make run-native work for relative local paths

### LINKED ISSUE #1098: runfv3 run-native cannot parse diag_table overlay in fv3config.yml
After some updates to the config preparation in https://github.com/VulcanClimateModeling/fv3net/pull/1051, I've run into issues in local prognostic runs  when reading in an `fv3config.yml` with a diag_table overlay entry. @oliverwm1 

Can be replicated in the prognostic run dev environment with
```
# this is an up to date version of an fv3config.yml after the diag table change
gsutil cp gs://vcm-ml-scratch/test-end-to-end-integration/integration-test-fd43a7fe5cb1/prognostic_run/fv3config.yml .
runfv3 run-native fv3config.yml rundir sklearn_runfile.py 
```

Results in
```
yq: Error running jq: TypeError: Object of type 'bytes' is not JSON serializable.
parse error: Unfinished JSON term at EOF at line 1, column 44128
```
I think it might be the formatting of the `base_time` key in the diag_table.

oliverwm1:
Currently, `runfv3 run-native` fails if the provided paths to the `fv3config` and `runfile` files are relative instead of absolute. 

In a roundabout way, this causes the issued raised in #1098: if the user supplied a local path that happened to be exactly `fv3config.yml`, then `run-native` would end up using the config yaml that gets written to the run directory by `write_run_directory`, which includes the `diag_table` as bytes, which causes `yq` to choke. A separate fv3config issue is that the `fv3config.yml` that gets written to the run directory is different than the input one. See https://github.com/VulcanClimateModeling/fv3config/issues/130.

Significant internal changes:
- allow `runSegment` (and therefore `run-native`) to work if the provided `fv3config` and `runfile` arguments are relative paths instead of absolute paths

- [ ] Tests added

Resolves #1098

### DIFF
--- a/workflows/prognostic_c48_run/runfv3
+++ b/workflows/prognostic_c48_run/runfv3
@@ -138,11 +138,11 @@ function runSegment {
     write_run_directory "$fv3config" "$rundir"
     mkdir -p "$rundir"
     cp "$runfile" "$rundir/runfile.py"
+    NUM_PROC=$(yq '.namelist.fv_core_nml.layout | .[0] *.[1] * 6' "$fv3config")
     (
         cd "$rundir"
         find . > preexisting_files.txt
-        NUM_PROC=$(yq '.namelist.fv_core_nml.layout | .[0] *.[1] * 6' "$fv3config")
-        mpirun -n "$NUM_PROC" python3 -m mpi4py "$runfile" |& tee -a "$rundir/logs.txt"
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
LGTM. Thanks. Now I can stop putting `$(pwd)/runfile.py` in all of my scripts.

brianhenn:
Just seeing this and it is appreciated :)

PR #1163, opened 2021-04-28
Prognostic run: use xr.Dataset.interp for the prescriber

AnnaKwa:
This uses xarray's interpolation to interpolate for timesteps that don't have reference data when prescribing state updates in the nudged runs.


brianhenn (approved):
Thanks @AnnaKwa LGTM

PR #1165, opened 2021-04-28
Enforce experiment workflows output pattern

brianhenn:
Enforces an output location pattern for experiment argo workflows. The pattern is `gs://{bucket}/{project}/$(date +%F)/{tag}/{step}`, where `bucket` defaults to 'vcm-ml-experiments', `project` defaults to 'default', `tag` is supplied by the user, and `step` is defined by the workflow.

Refactored public API:
- For the `prognostic-run` and `train-diags-prog` workflows, users no longer specify the output path directly, instead specifying the experiment tag, and the bucket and project if need be.
- All workflows using the `prognostic-run` workflow template now write out to a step directory called `fv3gfs_run`.
- The e2e test now writes to a directory in `vcm-ml-scratch` whose path contains the date and commit plus a random string; the nudge-to-fine run and train-diags-prog portions are written under different tags in this directory.

### DIFF
--- a/docker/fv3net/Dockerfile
+++ b/docker/fv3net/Dockerfile
@@ -1,7 +1,7 @@
 FROM jupyter/base-notebook as build
 
 # Manually increment this string to invalidate the cache
-ENV MANUAL_CACHE_STRING 1
+ENV MANUAL_CACHE_STRING 2
 
 # Install dependencies (slow)
 USER root
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -374,8 +374,6 @@ spec:
             value: "{{workflow.parameters.tag}}"
           - name: step
             value: "nudge_to_fine_run"
-#           - name: output
-#             value: "{{workflow.parameters.root}}/nudge_to_fine_run"
           - name: config
             value: "{{workflow.parameters.nudge-to-fine-config}}"
           - name: segment-count
@@ -397,8 +395,6 @@ spec:
             value: "{{workflow.parameters.project}}"
           - name: tag
             value: "{{workflow.parameters.tag}}"
-#           - name: root
-#             value: "{{workflow.parameters.root}}"
           - name: train-test-data
             value: "{{tasks.resolve-output-url.outputs.result}}/nudge_to_fine_run"
           - name: training-configs
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -342,6 +342,18 @@ spec:
   - name: main
     dag:
       tasks:
+      - name: resolve-output-url
+        templateRef:
+          name: resolve-output-url
+          template: resolve-output-url
+        arguments:
+          parameters:
+            - name: bucket
+              value: "{{workflow.parameters.bucket}}"
+            - name: project
+              value: "{{workflow.parameters.project}}"
+            - name: tag
+              value: "{{workflow.parameters.tag}}"
       - name: nudge-to-fine-run
         templateRef:
           name: prognostic-run
@@ -354,8 +366,16 @@ spec:
             value: "{{workflow.parameters.initial-condition}}"
           - name: flags
             value: "--output-frequency {{workflow.parameters.nudge-to-fine-output-frequency}}"
-          - name: output
-            value: "{{workflow.parameters.root}}/nudge_to_fine_run"
+          - name: bucket
+            value: "{{workflow.parameters.bucket}}"
+          - name: project
+            value: "{{workflow.parameters.project}}"
+          - name: tag
+            value: "{{workflow.parameters.tag}}"
+          - name: step
+            value: "nudge_to_fine_run"
+#           - name: output
+#             value: "{{workflow.parameters.root}}/nudge_to_fine_run"
           - name: config
             value: "{{workflow.parameters.nudge-to-fine-config}}"
           - name: segment-count
@@ -371,10 +391,16 @@ spec:
           template: train-diags-prog
         arguments:
           parameters:
-          - name: root
-            value: "{{workflow.parameters.root}}"
+          - name: bucket
+            value: "{{workflow.parameters.bucket}}"
+          - name: project
+            value: "{{workflow.parameters.project}}"
+          - name: tag
+            value: "{{workflow.parameters.tag}}"
+#           - name: root
+#             value: "{{workflow.parameters.root}}"
           - name: train-test-data
-            value: "{{workflow.parameters.root}}/nudge_to_fine_run"
+            value: "{{tasks.resolve-output-url.outputs.result}}/nudge_to_fine_run"
           - name: training-configs
             value: "{{workflow.parameters.training-configs}}"
           - name: train-times
@@ -388,7 +414,7 @@ spec:
           - name: test-times
             value: "{{workflow.parameters.test-times}}"
           - name: public-report-output
-            value: "{{workflow.parameters.root}}/offline-report"
+            value: "{{tasks.resolve-output-url.outputs.result}}/offline-report"
           - name: cpu-prog
             value: "24"
           - name: memory-prog
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -381,7 +381,7 @@ spec:
           - name: memory
             value: 6Gi
       - name: train-offline-diags-prognostic
-        dependencies: [nudge-to-fine-run]
+        dependencies: [nudge-to-fine-run, resolve-output-url]
         templateRef:
           name: train-diags-prog
           template: train-diags-prog
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -376,6 +376,8 @@ spec:
             value: "{{workflow.parameters.nudge-to-fine-config}}"
           - name: segment-count
             value: "{{workflow.parameters.nudge-to-fine-segment-count}}"
+          - name: image-tags
+            value: "{{workflow.parameters.image-tags}}"
           - name: cpu
             value: 6000m
           - name: memory
@@ -409,6 +411,8 @@ spec:
             value: "{{workflow.parameters.test-times}}"
           - name: public-report-output
             value: "{{tasks.resolve-output-url.outputs.result}}/offline-report"
+          - name: image-tags
+            value: "{{workflow.parameters.image-tags}}"
           - name: cpu-prog
             value: "24"
           - name: memory-prog
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -376,8 +376,6 @@ spec:
             value: "{{workflow.parameters.nudge-to-fine-config}}"
           - name: segment-count
             value: "{{workflow.parameters.nudge-to-fine-segment-count}}"
-          - name: image-tags
-            value: "{{workflow.parameters.image-tags}}"
           - name: cpu
             value: 6000m
           - name: memory
@@ -411,8 +409,6 @@ spec:
             value: "{{workflow.parameters.test-times}}"
           - name: public-report-output
             value: "{{tasks.resolve-output-url.outputs.result}}/offline-report"
-          - name: image-tags
-            value: "{{workflow.parameters.image-tags}}"
           - name: cpu-prog
             value: "24"
           - name: memory-prog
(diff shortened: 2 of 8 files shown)


workflows/argo/resolve-output-url.yaml
```diff
@@ -0,0 +1,22 @@
+# vim: set sts=2 ts=2 tw=2 sw=2 :
```
nbren12, line 1: Nice!

tests/end_to_end_integration/argo.yaml
```diff
@@ -371,10 +391,16 @@ spec:
...
-          - name: root
-            value: "{{workflow.parameters.root}}"
+          - name: bucket
+            value: "{{workflow.parameters.bucket}}"
+          - name: project
+            value: "{{workflow.parameters.project}}"
+          - name: tag
+            value: "{{workflow.parameters.tag}}"
+#           - name: root
+#             value: "{{workflow.parameters.root}}"
```
nbren12, line 401: Can you delete this commented code?

nbren12 (changes requested):
Great! I am excited to check back in a month and list all the prognostic runs/train-diags-progs experiments.

I think you should update the docs though.

tests/end_to_end_integration/run_test.sh
```diff
@@ -75,13 +75,17 @@ EOF
 registry="$1"
 tag="$2"
 fv3net="$PWD"
-random="$(openssl rand --hex 6)"
-name="integration-test-$random"
-GCS_OUTPUT_URL="gs://vcm-ml-scratch/test-end-to-end-integration/$name"
+# random="$(openssl rand --hex 6)"
```
nbren12, line 78: commented code.

tests/end_to_end_integration/run_test.sh
```diff
@@ -75,13 +75,17 @@ EOF
 registry="$1"
 tag="$2"
```
nbren12, line 76: This would be more clear IMO if it were named "docker_tag" or "commit".

tests/end_to_end_integration/run_test.sh
nbren12, line 78: I think we may still want to add some randomness to this string since the tests can be invoked several times in one day on a single commit. The timestamp only includes the day.

nbren12 (commented):
Some other suggestions.

workflows/argo/prognostic-run.yaml
```diff
@@ -22,12 +22,25 @@ spec:
           - {name: initial-condition}
           - {name: config}
           - {name: reference-restarts}
+          - {name: tag}
           - {name: flags, value: ""}
-          - {name: output}
+          - {name: bucket, value: "vcm-ml-experiments"}
+          - {name: project, value: "default"}
+          - {name: step, value: "prognostic_run"}
```
oliverwm1, line 29: is there a particular reason that the `step` name is a parameter for the prognostic run? I thought we were just gonna hard-code it for all the workflow steps.

brianhenn: Yeah, because the `prognostic-run` workflow is also the N2F and N2O workflow, and presumably we want to give those steps different labels than the prognostic run.

docker/fv3net/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM jupyter/base-notebook as build
 
 # Manually increment this string to invalidate the cache
-ENV MANUAL_CACHE_STRING 1
+ENV MANUAL_CACHE_STRING 2
```
brianhenn, line 4: build was failing for a reason that seemed unrelated to this PR, and this did resolve it

workflows/argo/prognostic-run.yaml
oliverwm1: Do we want to give them different labels though? I think that's an important point to discuss. At the end of the day, they're all FV3GFS runs and all being executed with the same argo workflow. It does mean that we could not have multiple runs under the same tag (which is currently done for the e2e test, for example).

nbren12: I agree with @brianhenn. To me the relevant point is that nudge to fine runs contain the same outputs as other nudge to fine runs (and have their own example folder in vcm-workflow-control). The process that generated the data is less important to capture in the path name. As a user, I want to do queries like "What are all nudge to fine runs that can be loaded with `open_nudge_to_fine`.

brianhenn: Fair point about the baseline case, I think I've been inconsistent between 'prognostic_run' and 'baseline_run' and we don't have a vcm-workflow-control example to guide us.

nbren12: It's a good point. It's hard to capture the polymorphic nature of data produced by the prognostic run in a URL. In theory the outputs available in a run directory can be queried by looking up the fv3config.yml or listing the zarrs available, all segmented runs fit this basic format. Should we call it "segmented-run"?

To enable more efficient querying of arbitrary metadata (e.g. all C48 prog runs that had nudging on for X Y and Z), we could maintain a separate index.

I'm with @oliverwm1  now.

brianhenn: How does an e2e type experiment get written out under this approach?

oliverwm1: The e2e test is unique in that it's our only argo workflow that submits multiple prognostic runs. Under this framework, I guess you'd have to have a different `tag` for the nudge-to-fine run as you do for the `train-evaluate...` workflow.

nbren12: right..some workflows run a single step multiple times. One could see the same problem with a training an ensemble of nets. Is it important to provide this grouping in the experiment tag? If not then we could it append info to the tag? 

```
gs://bucket/default/2021-01-01/segmented-run/{tag}-nudge-to-fine/
```

This would still allow fast querying by the prefix `gs://bucket/default/2021-01-01/*/{tag}`

nbren12: I think I'll leave it up to @brianhenn for what to do on this. It's more of a breaking change from our previous approach. I think the prognostic run vs nudged approach is meaningful enough to warrant being in separate directories.

nbren12: Honestly, my preferred naming nomenclature would be

```
gs://{bucket}/{step}/d1f6fa7e-a9f4-11eb-b831-3ff9a969d9d6
```

with a `metadata.yaml` in each folder.

And we could write a small library to do things like list in order of time, experiment, config values. etc.

brianhenn: Building the query tools seems worthwhile but also pretty sure there's value to the group in terms of human-readable dates and tags :)

What about calling all fv3gfs runs just that -- `fv3gfs_run`? segmented doesn't seem like the salient point about them, and the nudged runs aren't really prognostic. And then yeah just relying on the tags to differentiate them.

So for an e2e type experiment we'd have:
`gs://{bucket}/{project}/2021-04-30/{tag}/fv3gfs_run` where tag is something like `n2f-3km-nudge-to-fine`
and:
`gs://bucket/project/2021-04-30/{tag}/fv3gfs_run`
`gs://bucket/project/2021-04-30/{tag}/trained_model`
`gs://bucket/project/2021-04-30/{tag}/offline_diags`
`gs://bucket/project/2021-04-30/{tag}/fv3gfs_run`
...
where tag is something like `n2f-3km-train-diags-prog` .

Model ensembles would have to be handled with tags also, e.g.:
`gs://bucket/project/2021-04-30/n2f-3km-nn-random-seed1/trained_model`
`gs://bucket/project/2021-04-30/n2f-3km-nn-random-seed2/trained_model`
...
But that seems to be basically what Anna has been doing anyway.

Any objections?

brianhenn: If we chose one name for all of the runs I think my preference order is
1a) fv3gfs_run
1b) prognostic_run (nudging runs are kind of prognostic I guess and we already use this)
2) segmented_run

nbren12: I agree with your suggestions excepting this one:

> 'segmented' doesn't seem like the salient point about them, and the nudged runs aren't really prognostic. 

To me, segmented runs are special compared to the output of `runfv3 run-native` for a few reasons:
1. RESTART and log files are in the `artifacts` sub directory
2. The metadata on the .zarr files is consolidated. They contain some metadata at the root level that is shared by all the runs in `artifacts`.

nbren12: but I am fine with any of your proposed names.

oliverwm1: I would be +1 on `fv3gfs_run`. Although I agree the segmented runs are special because of their output format, it's not possible for the argo workflows to output data in the `run-native` format anyway, so not sure the distinction is that useful. Anyway I'm also fine with any of the proposed names.

workflows/argo/prognostic-run.yaml
```diff
@@ -61,6 +81,7 @@ spec:
           - {name: fv3config, path: /tmp/fv3config.yaml}
           - name: runfile
             path: /tmp/runfile.py
+          - {name: output-url, path: /tmp/}
```
brianhenn, line 84: leftover will delete

workflows/argo/prognostic-run.yaml
brianhenn: Implemented using `fv3gfs_run`.

tests/end_to_end_integration/argo.yaml
```diff
@@ -376,6 +376,8 @@ spec:
             value: "{{workflow.parameters.nudge-to-fine-config}}"
           - name: segment-count
             value: "{{workflow.parameters.nudge-to-fine-segment-count}}"
+          - name: image-tags
```
nbren12, line 379: I feel some reluctance to add a workflow parameter for this, since it possible that this parameter could be inconsistent with the actual tag. `COMMIT_SHA` is an environmental variable in both the fv3net and prognostic run images. Could you use that instead?

There are also mechanisms to get the container name/tag using k8s. https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/.

In the interest of moving forward with this PR, it might make more sense to drop this git-sha functionality and discuss it at a later point. It's a somewhat independent feature.

workflows/argo/train-diags-prog.yaml
```diff
@@ -149,9 +149,9 @@ spec:
         arguments:
           parameters:
               - name: run
-                value: "{{tasks.resolve-output-url.outputs.result}}/prognostic_run"
+                value: "{{tasks.resolve-output-url.outputs.result}}/fv3gfs_run"
               - name: output
-                value: "{{tasks.resolve-output-url.outputs.result}}/prognostic_run_diags"
+                value: "{{tasks.resolve-output-url.outputs.result}}/fv3gfs_run_diags"
```
nbren12, line 154: Prefer "diagnostics" over "diags".

workflows/argo/resolve-output-url.yaml
```diff
@@ -16,9 +16,14 @@ spec:
         command: [bash]
         source: |
           set -e
+          remove_slashes () {
+              echo $(echo $1 | sed -e 's~/~-~')
```
nbren12, line 20: I think it's probably better to fail if there is a `/` rather than transform the user inputs. One nice thing about vcm-workflow-control is that we can use `git-grep` to find commits matching a specific url. This feature won't work if we transform the user inputs.

nbren12 (changes requested):
Thanks. I'm really sorry about the continued back and forth it would be great to get this PR out of the way, but I noticed a coupled of things.

workflows/argo/resolve-output-url.yaml
brianhenn: OK

workflows/argo/train-diags-prog.yaml
brianhenn: OK

tests/end_to_end_integration/argo.yaml
brianhenn: Will separate that out

workflows/argo/resolve-output-url.yaml
```diff
@@ -16,29 +16,15 @@ spec:
         command: [bash]
         source: |
           set -e
-          remove_slashes () {
-              echo $(echo $1 | sed -e 's~/~-~')
+          fail_on_slash () {
```
nbren12, line 19: I tested in my shell that this logic works.

nbren12 (approved):
Thanks!

workflows/argo/resolve-output-url.yaml
brianhenn: Yah, I tested it as well.

Issue #1164, opened 2021-04-28
prognostic_run: sklearn runfile separation of concerns improvements


oliverwm1:
The logic of the prognostic run runfile is difficult to follow and extend. Here are a few particular issues with possible solutions. Taken from a slack thread with @nbren12.

1. ~remove the `tendency_variable` and `storage_variables` configurations and infer them from the diagnostics configs~
1. ~Rectify the weird coupling between `MonitoredPhysicsTimeLoop`, `monitor`, and `Loop` class. `monitor` is passed the entire time loop class which allows for weirdness.~
1. De-duplicate the copy/pasted code here: 
https://github.com/VulcanClimateModeling/fv3net/blob/149b545c5dd633cae746f938e84fb82906623bc0/workflows/prognostic_c48_run/runtime/diagnostics/machine_learning.py#L14-L17
1. ~refactor the specific humidity logging to the sklearn_runfile main. It is currently implemented here:~
https://github.com/VulcanClimateModeling/fv3net/blob/95a60a4fd7ac361ed57207a275a786050ea55f7f/workflows/prognostic_c48_run/runtime/loop.py#L413-L426
~but should be here:~
https://github.com/VulcanClimateModeling/fv3net/blob/74d9ba08e7496abcbb314b5d1258d7a700eb7c9e/workflows/prognostic_c48_run/sklearn_runfile.py#L55-L57
1. blow up the `get_diagnostics` and `get_momentum_diagnostics` methods of the Stepper classes
https://github.com/VulcanClimateModeling/fv3net/blob/149b545c5dd633cae746f938e84fb82906623bc0/workflows/prognostic_c48_run/runtime/steppers/machine_learning.py#L209-L214

Related: #1159, #896

oliverwm1:
These points have all been resolved by above-referenced PRs.

nbren12:
Thanks Oli! This was a great push to prep the prognostic run code for future development.

PR #1167, opened 2021-04-28
Infer tendency and storage variables from config.diagnostics

oliverwm1:
Currently one must redundantly specify certain variable names in `step_tendency_variables` or `step_storage_variables` sections of the user config as well as list the relevant name in the appropriate diagnostics file config. This PR makes the former be inferred from the latter.

Refactored public API:
- `step_storage_variables` and `step_tendency_variables` deleted from `UserConfig`
- the default for `DiagnosticsFileConfig.variables` is now an empty list (instead of `All` container)

Significant internal changes:
- the internal `self._storage_variables` and `self._tendency_variables` are now inferred from the provided diagnostics configurations

- [ ] Tests added

Resolves first bullet point in #1164

### DIFF
--- a/workflows/prognostic_c48_run/tests/test_regression.py
+++ b/workflows/prognostic_c48_run/tests/test_regression.py
@@ -28,6 +28,24 @@ DIAGNOSTICS = [
         "name": "diags.zarr",
         "times": {"kind": "interval", "frequency": 900, "times": None},
     },
+    {
+        "name": "monitored_tendencies_and_storages.zarr",
+        "times": {"kind": "interval", "frequency": 900, "times": None},
+        "variables": [
+            "storage_of_specific_humidity_path_due_to_fv3_physics",
+            "storage_of_specific_humidity_path_due_to_python",
+            "storage_of_total_water_path_due_to_fv3_physics",
+            "storage_of_total_water_path_due_to_python",
+            "tendency_of_air_temperature_due_to_fv3_physics",
+            "tendency_of_air_temperature_due_to_python",
+            "tendency_of_eastward_wind_due_to_fv3_physics",
+            "tendency_of_eastward_wind_due_to_python",
+            "tendency_of_northward_wind_due_to_fv3_physics",
+            "tendency_of_northward_wind_due_to_python",
+            "tendency_of_specific_humidity_due_to_fv3_physics",
+            "tendency_of_specific_humidity_due_to_python",
+        ],
+    },
 ]
 
 
@@ -414,8 +432,7 @@ def get_ml_config(model_path):
     config = yaml.safe_load(default_fv3config)
     config["diagnostics"] = DIAGNOSTICS
     config["fortran_diagnostics"] = []
-    config["scikit_learn"] = {"model": [model_path], "zarr_output": "diags.zarr"}
-    config["step_storage_variables"] = ["specific_humidity", "total_water"]
+    config["scikit_learn"] = {"model": [model_path]}
     # use local paths in prognostic_run image. fv3config
     # downloads data. We should change this once the fixes in
     # https://github.com/VulcanClimateModeling/fv3gfs-python/pull/78 propagates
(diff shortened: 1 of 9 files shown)


workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -414,8 +432,7 @@ def get_ml_config(model_path):
     config = yaml.safe_load(default_fv3config)
     config["diagnostics"] = DIAGNOSTICS
     config["fortran_diagnostics"] = []
-    config["scikit_learn"] = {"model": [model_path], "zarr_output": "diags.zarr"}
```
oliverwm1, line 417: `zarr_output` is a deprecated configuration option

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -28,6 +28,24 @@
...
+            "storage_of_total_water_path_due_to_fv3_physics",
+            "storage_of_total_water_path_due_to_python",
+            "tendency_of_air_temperature_due_to_fv3_physics",
+            "tendency_of_air_temperature_due_to_python",
+            "tendency_of_eastward_wind_due_to_fv3_physics",
+            "tendency_of_eastward_wind_due_to_python",
+            "tendency_of_northward_wind_due_to_fv3_physics",
+            "tendency_of_northward_wind_due_to_python",
+            "tendency_of_specific_humidity_due_to_fv3_physics",
+            "tendency_of_specific_humidity_due_to_python",
```
oliverwm1, line 46: these variables also end up getting output in `diags.zarr` since that diagnostic file outputs all variables.

nbren12:
Yah the changed default is tricky. I think your changes to the test-suite make sense, but we will either need to
1. change the default behavior from using `All` to require the diagnostics to be a Set/List with a default value of no outputs.
2. judicious use of Union types and `isinstance`.
I think 1 is the cleaner approach, and won't really break anything except the tests.

The main goal of the test is to cover all available diagnostics. Can we save/log this list as the model runs, and then load it in a test to ensure that no untested diagnostics were available?

nbren12:
Another approach would be to store a global table of available diagnostic names someplace in the source code and raise a runtime error in the sklearn_runfile if any new diagnostics show up. A global table would also be a nice piece of documentation.

oliverwm1:
Okay, yeah I agree with 1. I like the idea of having a table of all available diagnostics. Will implement.

oliverwm1:
In the end, just manually specified the diagnostics to get tested instead of creating a registry/table of available diagnostics. If new diagnostics are introduced, we should consider adding them to the tested set of diagnostics.

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -23,11 +23,88 @@
 LOG_PATH = "statistics.txt"
 RUNFILE_PATH = "runfile.py"
 CHUNKS_PATH = "chunks.yaml"
-DIAGNOSTICS = [
+DIAGNOSTICS_ML = [
```
nbren12, line 26: I suggest scoping this within `get_nudging_config` to reduce the amount of module level constants (especially since this is a mutable data structure).

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -23,11 +23,88 @@
...
+            "tendency_of_northward_wind_due_to_python",
+            "tendency_of_specific_humidity_due_to_fv3_physics",
+            "tendency_of_specific_humidity_due_to_python",
+            "total_precip_after_physics",
+            "total_precipitation_rate",
+            "water_vapor_path",
+        ],
+    }
+]
+DIAGNOSTICS_NUDGING = [
```
nbren12, line 71: I suggest scoping this within get_ml_config to discourage use elsewhere.

nbren12 (approved):
Thanks! I think hardcoding configs in the tests makes sense. Just a couple minor comments below:

workflows/prognostic_c48_run/tests/test_regression.py
oliverwm1: sounds good

workflows/prognostic_c48_run/tests/test_regression.py
oliverwm1: yup

PR #1170, opened 2021-05-04
Log reduced vertical profile diagnostics to profiles.txt

oliverwm1:
Currently there is some ad-hoc logging done within `_compute_postphysics`. This PR moves the calculation of the relevant metric to a more general function, and adds a new text output from the prognostic run, `profiles.txt` which contains vertical profiles saved using json.dumps. Currently, the total number of grid points where the specific humidity limiter is active is the only profile saved.

Added public API:
- new `profiles.txt` output saved from `log_profiles`

Significant internal changes:
- move calculation/logging of specific humidity limiter out of `_compute_postphysics`
- renamed `capture.py` to `logs.py` and moved other logging functionality from `loop.py` to `logs.py`
- refactored diagnostic reduction code from `loop.py` to `metrics.py`

- [x] Tests added

Resolves bullet point 4 of #1164


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -49,35 +48,6 @@
 gravity = 9.81
 
 
-def setup_metrics_logger():
```
oliverwm1, line 52: logging functions moved to `logs.py` (which also includes stuff that was previously in `capture.py`)

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -49,35 +48,6 @@
...
-
-def log_scalar(time, scalars):
-    dt = datetime.datetime(
-        time.year, time.month, time.day, time.hour, time.minute, time.second
-    )
-    msg = json.dumps({"time": dt.isoformat(), **scalars})
-    logging.getLogger("statistics").info(msg)
-
-
-def global_average(comm, array: xr.DataArray, area: xr.DataArray) -> float:
```
oliverwm1, line 72: moved to `metrics.py`

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -198,7 +198,9 @@ def __call__(self, time, state):
             )
             tendency.update({"dQ2": dQ2_updated})
 
-        diagnostics["rank_updated_points"] = xr.where(dQ2_initial != dQ2_updated, 1, 0)
+        diagnostics["specific_humidity_limiter_active"] = xr.where(
```
oliverwm1, line 201: changed name of diagnostic to `specific_humidity_limiter_active` (which is valued either 0 or 1)

workflows/prognostic_c48_run/runtime/logs.py
```diff
@@ -83,6 +85,42 @@ def capture_fv3gfs_funcs():
...
+    ch.setLevel(logging.INFO)
+    ch.setFormatter(logging.Formatter(fmt="%(levelname)s:%(name)s:%(message)s"))
+
+    logger.addHandler(fh)
+    logger.addHandler(ch)
+
+
+def setup_loggers():
+    setup_logger("statistics")
+    setup_logger("profiles")
```
frodre, line 102: Maybe there's a way to reduce the definition/reference separation of these loggers or else warn/error when the user requests a logger not already setup.  Currently if the logger is misspelled or not specified up here, it will initialize with defaults, which is probably okay, but maybe we don't want that for consistency's sake.

Maybe we add an error when the user requests an unitialized logger (by `setup_loggers`), or just use a module-level get_logger function to ensure consistency and get rid of the `setup_loggers` function.

```
LOGGERS = {}

def setup_logger(name):
    logger = ...
    LOGGERS[name] = logger
    return logger

def get_logger(name):
    if name not in LOGGERS:
        # either error here if we want to notify user of unspecified logger
       raise Error( "you didn't add this to setup_loggers")
       # or do away with the "setup_loggers" and just trigger setup here, e.g.,
        logger = setup_logger(name)
    else:
        logger = LOGGERS[name]

    return logger

def log_values(...)
    logger = get_logger(name)
    logger.info(msg)
    ....
```

frodre: Something like this could also be extended to the stream capturing function above to standardize w/in this module

workflows/prognostic_c48_run/runtime/loop.py
frodre: Thank you, this is a great cleanup!

workflows/prognostic_c48_run/runtime/logs.py
frodre: With this suggestion we should have a few tests to make sure the setup/get_logger are working

frodre (approved):
Thank you for this! A good start on cleaning up some aspects of the loop.  My one comment is about reducing the setup/request separation for specific loggers.  I think the only necessary change is to enforce the logging setup standard defined in `setup_logger`.  

Other than that. Looks good to me.

workflows/prognostic_c48_run/runtime/logs.py
oliverwm1: Nice, yah the disconnect between the setup and logging functions is not ideal, although it is already existing before this PR.

I do think the captured logs are distinct, since they just go to stdout and don't have a file handler associated with them (the piping to `logs.txt` is defined by the parent `runfv3` script).

Perhaps if the `log_scalar` and `log_profiles` functions take a logger name, this will be cleaner.

frodre: > I do think the captured logs are distinct, since they just go to stdout and don't have a file handler associated with them (the piping to logs.txt is defined by the parent runfv3 script).

Got it, thanks for clarifying this.

workflows/prognostic_c48_run/sklearn_runfile.py
```diff
@@ -47,11 +43,15 @@
...
+        averages = runtime.globally_average_2d_diagnostics(
             comm, diagnostics, exclude=loop._states_to_output
         )
+        profiles = runtime.globally_sum_3d_diagnostics(
+            comm, diagnostics, ["specific_humidity_limiter_active"]
+        )
         if comm.rank == 0:
-            log_scalar(time, averages)
+            runtime.log_mapping(time, averages, "statistics")
+            runtime.log_mapping(time, profiles, "profiles")
```
frodre, line 54: I guess the reason I was thinking it would be nice to hide the need for the logger filehandling setup from the runfile is that it might not be immediately obvious based calls to logging here that it's necessary to adjust line 30 to get a file output.  It will still work, but only log to error, which could be confusing.   Up to you!

frodre: Could add a keyword_argument like `to_file=True` to remove the separation.

workflows/prognostic_c48_run/runtime/logs.py
```diff
@@ -83,6 +85,32 @@ def capture_fv3gfs_funcs():
...
+    fh.setLevel(logging.INFO)
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.INFO)
+    ch.setFormatter(logging.Formatter(fmt="%(levelname)s:%(name)s:%(message)s"))
+
+    logger.addHandler(fh)
+    logger.addHandler(ch)
+
+
+def log_mapping(time, content, logger_name):
```
frodre, line 101: I'd add some typing information to the function since this is the main contact point with the runfile.

`time: "datetime-like"`??, `content: Mapping[str, Any]` (I don't think here's a way to specify "serializable"), `logger_name: str`

workflows/prognostic_c48_run/sklearn_runfile.py
frodre: Just to add, I think it's fine as is since the usage is not sprawling, yet.  Just thinking if this potentially gets used in more than one place.

frodre (commented):
Thanks for the update. I added my comments!

workflows/prognostic_c48_run/sklearn_runfile.py
```diff
@@ -32,7 +27,8 @@
 
     config = runtime.get_config()
     partitioner = util.CubedSpherePartitioner.from_namelist(runtime.get_namelist())
-    setup_metrics_logger()
+    for name in ["statistics", "profiles"]:
+        runtime.setup_file_logger(name)
```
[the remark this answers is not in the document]
oliverwm1: In principle someone might want to log from all ranks (say, compute max on each rank and log) in which case it's probably safer to setup the logger on all ranks although yes, currently they're only used on rank 0.

workflows/prognostic_c48_run/sklearn_runfile.py
oliverwm1: Okay, I think for now I'd rather leave it as it is, since 1) silently doing a logger setup the first time one calls `log_mapping` is also not obvious, and 2) setting up a logger registry feels a bit overkill for the two loggers we have now. Agreed if/when usage expands we should reconsider!

workflows/prognostic_c48_run/runtime/logs.py
oliverwm1: Good idea.

PR #1172, opened 2021-05-04
fine res budget: fix I/O errors

nbren12:
I was getting many I/O errors when trying to write the data from a dataflow
job.

Requirement changes:
- added gcsfs as a subtree temporarily. This results in a big diff, so it's best to review this PR one commit at at time (the first two commits are both in upstream).


spencerkclark (approved):
Nice, glad this is working now.  It looks like there is just an issue installing the subtree version of gcsfs in CI, but once that's resolved this looks ready to go.

oliverwm1:
What about just pip installing latest directly from git instead of using a subtree? Could just temporarily put this in one of the install env scripts until there is a pypi release for desired gcsfs version.

`pip install git+https://github.com/dask/gcsfs.git@b792f65282ae041b8e719e429d0b0bca17f40980`

spencerkclark:
I assumed @nbren12 had a reason not to do that, but I agree that came to my mind as well.

nbren12:
>I think he may have made some additional changes that are not in master.

That is true, but I eventually reverted all of those, so i think I can use @oliverwm1's suggestion.

nbren12:
I remember now that Dataflow doesn't support using git+ urls in a setup.py file.

PR #1177, opened 2021-05-05
Reformat in pre-commit checks

nbren12:
92f22e7f8a5ca14b3f0affba32b09d2dfc8f7bd5 broke the reformat makefile rule.
Also, now that some time has passed, I think we can make the reformatting
happen during the commit hook. This will make it easier to reformat the code
without changing directories to the root.


AnnaKwa:
LGTM, thanks!

AnnaKwa (approved):

PR #1178, opened 2021-05-05
Delete MonitoredTimeLoop class, propagate units

oliverwm1:
The MonitoredPhysicsTimeLoop class introduces another layer within `loop.py` which makes code execution arguably harder to follow (this is an opinion...). This PR removes the `MonitoredPhysicsTimeLoop` class and directly applies the `monitor` decorator to the relevant substeps within the `TimeLoop` class. It also adds output units where possible to diagnostics generated by `monitor`.

Refactored public API:
- Delete `MonitoredPhysicsTimeLoop`

Significant internal changes:
- Move `monitor` decorator to be defined within `TimeLoop`
- Add units to diagnostics generated by `monitor`

Resolves second bullet point in #1164


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -427,76 +432,58 @@ def __iter__(self):
                     diagnostics.update(substep())
             yield self._state.time, diagnostics
 
+    def monitor(self, name: str, func):
```
oliverwm1, line 435: GitHub had trouble with this diff... I just added a tab to move the `monitor` function inside the `TimeLoop`. I also added units attributes to the `tendency` and `storage` outputs.

oliverwm1: `monitor` already used the `self` of TimeLoop/MonitoredTimeLoop, so putting the function inside the class just makes this explicit.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -314,9 +319,9 @@ def _substeps(self) -> Sequence[Callable[..., Diagnostics]]:
             self._step_prephysics,
             self._compute_physics,
             self._apply_postphysics_to_physics_state,
-            self._apply_physics,
+            self.monitor("fv3_physics", self._apply_physics),
```
oliverwm1, line 322: feels not quite right to be hard-coding these `fv3_physics` and `python` names here, but at the same time they shouldn't be configuration options. It would be nice if the "name" used for monitoring was just the actual function name.

nbren12: I think it's okay. I don't see a way around hardcoding diagnostics names. Tying them to function names doesn't allow us to rename those functions when appropriate.

nbren12 (approved):
LGTM. I think this is much clearer without the inheritance over-complicating things. Might be worth pulling the monitoring functionality into a new object with more minimal data at a future point e.g.
```
self.monitor = Monitor(state, tendency_variables, storage_variables)
self.monitor("a func", some_func)
```

PR #1186, opened 2021-05-07
Ensure consistent dim order in ensemble model

AnnaKwa:
Previously, using an `EnsembleModel` in a prognostic run would sometimes result in strange dimension errors that did not occur when the individual models in the ensemble were used (ex. https://gist.github.com/AnnaKwa/5b6a2e75e1086f85f26f9bde25419b9d). The wrong dimension (48, 48, 48) indicates that there was some inconsistency in how the prediction arrays were ordering their dimensions. It turns out the issue arose in the part of the `Predictor.predict_columnwise` that used the input data to determine order dimension; it was possible for different intializations of `EnsembleModel` to have different orderings of `EnsembleModel.input_variables` that could lead to `Predictor.predict_columnwise` having different dimension ordering.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -1,4 +1,5 @@
 from typing import Iterable, Set, Hashable
+import fsspec
 import yaml
 import os
 import xarray as xr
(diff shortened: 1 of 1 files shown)


brianhenn:
The fix looks good. Would it make sense to add a test for this?

brianhenn (approved):
Thanks @AnnaKwa for the sleuthing, this fix looks good. As discussed no test for this change make sense but the PR points at another problem in `Predictor`'s `_infer_dimension_order` (dependence on input variable order) which should be resolved as a separate issue.

PR #1190, opened 2021-05-07
allow one-based indexing in model training

nbren12:
Some datasets we use (e.g. the fine res) use with SHiELD's tile indexing
conventions (which are 1-based), but the grid inserted by the model
training routines is 0-based. The code did an outer join, so the
resulting data had a size of 7 tiles.

* add a failing test
* and fix it


AnnaKwa (approved):
LGTM, thanks for fixing this!

PR #1194, opened 2021-05-11
derived_mapping: add {eastward,westward}_wind

nbren12:
eastward/westwards winds were not in the derived mapping, but the dQu and dQv
were.


AnnaKwa (approved):
LGTM!

PR #1195, opened 2021-05-11
vcm.catalog: Corrected wind rotation matrices for c48 and c96

AnnaKwa:
The previous entries for the c48 and c96 wind rotation matrices had a factor of 0.5 error in the coefficients, resulting in incorrect magnitudes when they were used to rotate winds. This PR (https://github.com/VulcanClimateModeling/fv3gfs-fortran/pull/191) fixed the outputs from fv3gfs-fortran to automatically save the correct values (the fortran model introduces this factor of 0.5 to the coefficients and undoes it later with a multiplication by 2 when computing A grid lat/lon winds, but saves the lower values).

Verified in this notebook that the factor of 2 fixes the issue:
https://github.com/VulcanClimateModeling/explore/blob/master/annak/2021-05-11-wind-debug.ipynb


nbren12 (approved):
Great use of the catalog!

spencerkclark (approved):
Thanks again for figuring this out!

PR #1196, opened 2021-05-11
Offline diags: allow 1 based tile indexing

AnnaKwa:
This change avoids issues with outer joins that arise when the training data has tile coords starting at 1 and the grid has tile coords that start at 0.


nbren12 (approved):
LGTM. Thanks for the fix.

PR #1199, opened 2021-05-12
loaders: hybrid mapper for .zarr budget data

nbren12:
The fine resolution budget workflow now saves a single .zarr file and avoids changing the variable names of the fv3gfs model. This allows a much simpler implementation.

Re-code the hybrid mapper to use this new format.

Note that the code in `_fine_resolution` mapper has not been changed, will no longer work with the new format and is largely super-ceded by this PR. I would like to delete it in a follow up PR.


nbren12:
Supercedes #1197

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -1,88 +1,179 @@
...
+    merged["dQ2"] = (
+        merged["Q2"] - merged["tendency_of_specific_humidity_due_to_fv3_physics"]
+    )
+    merged["dQxwind"] = nudge_tends.x_wind_tendency_due_to_nudging
+    merged["dQywind"] = nudge_tends.y_wind_tendency_due_to_nudging
+
+    # Select the data we want to return
+    return (
+        xarray.merge(
+            [
```
nbren12, line 141: Not all input variables are in this list, but there should be enough to compute cos zenith angle and the surface geo-potential.

external/loaders/tests/test__nudged/prepare_schema.py
```diff
@@ -0,0 +1,43 @@
+"""Prepares mock data schema for the fine-res data
```
nbren12, line 1: This is a throwaway script, but I figured it might be useful later on.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -1,88 +1,190 @@
...
         )
+        .rename("Q2")
+    )
 
-    def __len__(self):
-        return len(self.keys())
 
-    def __iter__(self):
-        return iter(self.keys())
+def _open_fine_res_dataset(
```
oliverwm1, line 110: I found this function name confusing given that it also opens a nudged run and takes some differences.

external/loaders/loaders/mappers/_hybrid.py
oliverwm1: why not just return `merged`?

external/loaders/tests/test__nudged.py
```diff
@@ -231,10 +238,8 @@ def fine_url(tmpdir):
     return fine_url
```
oliverwm1, line 238: can this `fine_url` fixture be deleted now?

oliverwm1 (approved):
LGTM, just a few minor comments.

external/loaders/loaders/mappers/_hybrid.py
nbren12: Renamed to `open_fine_resolution_nudging_hybrid_dataset`.

external/loaders/tests/test__nudged.py
nbren12: yes. done.

PR #1201, opened 2021-05-12
Delete duplicated code in compute_diagnostics functions

oliverwm1:
This refactor PR merges the calculations in `compute_ml_diagnostics` and `compute_nudging_diagnostics` to a single `compute_diagnostics` function.

Refactored public API:
- To make diagnostic names more uniform between ML/nudging, we rename "net_heating" to "net_heating_due_to_machine_learning" and same for moistening

Significant internal changes:
- `get_diagnostics` method of Steppers now returns `Tuple[Diagnostics, xr.DataArray]` where the array is the column integrated moistening, which can then be used to update the precipitation in the TimeLoop.
- delete `net_moistening` property from Steppers
- add `label` property to Steppers
- unify `compute_ml_diagnostics` and `compute_nudging_diagnostics` to a single function

Resolves points 3 and 5 of #1164


workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -164,7 +163,7 @@ def predict(model: MultiModelAdapter, state: State) -> State:
 
 class PureMLStepper:
 
-    net_moistening = "net_moistening"
+    label = "ml"
```
oliverwm1, line 166: maybe should be more verbose, i.e. `machine_learning`?

oliverwm1: I changed it to machine learning.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/add_derived.py
```diff
@@ -71,7 +71,10 @@ def _column_pq2(ds: xr.Dataset) -> xr.DataArray:
 
 
 def _column_dq1(ds: xr.Dataset) -> xr.DataArray:
-    if "net_heating" in ds:
+    if "net_heating_due_to_machine_learning" in ds:
+        column_dq1 = ds.net_heating_due_to_machine_learning
+    elif "net_heating" in ds:
```
nbren12, line 76: This conditional logic is buried pretty deep in the call stack. If we need to maintain backwards compatibility, I would recommend renaming "net_heating" to "net_heating_due_to_machine_learning" somewhere more obvious (e.g. closer to where the data is loaded/queried).

Also, it would be nice to not maintain this backwards compatibility forever, so I would add some kind of deprecation warning, and/or making a github issue to delete it in a couple of months.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -390,12 +372,12 @@ def _compute_postphysics(self) -> Diagnostics:
...
         tendency = dissoc(self._tendencies, "dQu", "dQv")
+        diagnostics = compute_baseline_diagnostics(self._state)
 
-        if self._postphysics_stepper is None:
-            diagnostics = compute_baseline_diagnostics(self._state)
-        else:
-            diagnostics = self._postphysics_stepper.get_diagnostics(
-                self._state, tendency
+        if self._postphysics_stepper is not None:
+            stepper_label = self._postphysics_stepper.label
```
nbren12, line 378: I think the refactor of `compute_diagnostics` is positive, but I'm not sure that dispatching on a string "label" is much improvement over having a polymorphic classes with `get_diagnostics` methods. You could use `compute_diagnostics` inside of these `get_diagnostics` methods.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -405,7 +387,7 @@ def _apply_postphysics_to_dycore_state(self) -> Diagnostics:
                 )
                 updated_state_from_tendency[TOTAL_PRECIP] = precipitation_sum(
                     self._state[TOTAL_PRECIP],
-                    diagnostics[self._postphysics_stepper.net_moistening],
```
nbren12, line 408: I just had the idea that the steppers could return the actual `net_moistening` dataarray in `__call__` so that `TimeLoop` doesn't need to know about `.stepper.label` or `.stepper.net_moistening`.

nbren12 (approved):
Thanks. It's great to concentrate most of this logic in one function, although we might still want to expose it to the `TimeLoop` via a  polymorphic `get_diagnostics` function so that `TimeLoop` doesn't need to know about `label`.

I'm approving though.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/add_derived.py
oliverwm1: I think this an appropriate place for this logic because there is a lot of other similar logic in this module. E.g. for getting the correct <nQx> variable from both nudge-to-fine or nudge-to-obs runs. For example:

https://github.com/VulcanClimateModeling/fv3net/blob/d65e0dc97d3f3696d65e352e52a7e57a008c0c8b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/add_derived.py#L160-L169

I'll make a GitHub issue to remember to delete it in the future.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: Yeah, I agree it's not really that different, although having a single `compute_diagnostics` function definitely helped get rid of some duplicated code. You have a preference for leaving it as in this PR, versus reverting to having `get_diagnostics` methods within the Stepper classes? It feels a bit odd to be encoding the assumptions about the tendency key names within the compute diagnostics function and outside of the steppers.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: That could work, although we would need to add a new `TimeLoop` variable to hold this quantity from the step where the tendencies are computed to where they are applied. And it feels a bit odd to add a new output to `__call__` that is already contained within one of its other outputs. My takeaway from all this is that it would be useful to enforce consistent naming of tendencies across the Stepper classes. Currently, the `PureMLStepper` outputs tendencies whose keys are `dQ1`, `dQ2`, etc. while the `Nudging` stepper outputs tendencies whose keys are `air_temperature`, `specific_humidity` etc. That would be a bigger refactor though.

nbren12: Overall with this, I am advocating that if the time-loop does a specific computation with some output of the Steppers, that it be explicitly passed that data rather than looking it up in some dict. With the former approach, the typechecker can find bugs related to moistening not being defined. Currently this would only show up as a runtime `KeyError`. 

> although we would need to add a new TimeLoop variable to hold this quantity from the step where the tendencies are computed to where they are applied.

This piece of data is already managed by the TimeLoop, so this change would only make the complexity more explicit IMO, rather than increase it overall.

>And it feels a bit odd to add a new output to __call__ that is already contained within one of its other outputs.

I agree, we would want to remove it from `diags` so that it is only in one place.

workflows/prognostic_c48_run/runtime/loop.py
nbren12: To be clear, I think the current implementation of `compute_diagnostics` is fine. I'm proposing moving this logic in loop.py into the Steppers like this:
```
class MLStepper:
    def get_diagnostics(self, state, tendenecy):
        return compute_diagnostics(state, tendency, "ml")
```

The advantage is the Stepper would expose a method with rich behavior versus a "label" string to the `Loop` class.

oliverwm1: I implemented this change.

PR #1203, opened 2021-05-13
remove pin of fsspec to 0.* in synth

mcgibbon:
Fsspec moved from 0.X.X versioning to YYYY.MM.DD versioning, which breaks our fsspec==0.* constraint. This PR removes that constraint.

Requirement changes:
- synth no longer requires `fsspec==0.*`, keeping only the `>=` version requirement


nbren12 (approved):
LGTM. I reworded "dependency" to "constraint"

PR #1206, opened 2021-05-13
vcm.catalog: Update c3072 verification for corrected pressure level U200, U850 

AnnaKwa:
The previous catalog entry for `40day_c48_atmos_8xdaily_additional_vars_may2020`  had incorrect U200/U850 data because the wind rotation matrix at the time was off by a factor of 0.5. This update points to a corrected dataset.


oliverwm1 (approved):
LGTM! Maybe update link in description to notebook you used? (or nevermind if you just used the same notebook).

AnnaKwa:
Yup, same notebook.

PR #1207, opened 2021-05-13
Always use setup.cfg in precommit flake8 hook 

AnnaKwa:
For some reason, the flake8 precommit hook was not consistently applying the config settings in `setup.cfg`, which resulted in commit failures due to it using the default max-line-length of 79 chars (`black` autoreformats to a max of 88 chars). Telling the flake8 hook explicitly to use `setup.cfg` seems to fix this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
       name: flake8
       language_version: python3
       exclude: "__init__.py"
+      args: ["--config=setup.cfg"]
     - id: flake8
       name: flake8 __init__.py files
       files: "__init__.py"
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -19,6 +19,7 @@ repos:
       name: flake8
       language_version: python3
       exclude: "__init__.py"
+      args: ["--config=setup.cfg"]
```
mcgibbon, line 22: Feels more yaml-ic? This is what I have been using on my end.
```suggestion
      args: [--config, setup.cfg]
```

mcgibbon (approved):

PR #1208, opened 2021-05-14
Split fv3fit.ModelTrainingConfig into new objects for data and training config

mcgibbon:
Currently ModelTrainingConfig is responsible for holding configuration options for training, training data loading, and validation data loading. This PR splits those responsibilities into two separate classes, and three separate instances (training and validation data configuration use the same class).

In this PR this has been implemented to be backwards-compatible.

Added public API:
- Added classes `TrainingConfig`, `SklearnTrainingConfig`, `KerasTrainingConfig`, and `DataConfig` to `fv3fit`

Refactored public API:

Significant internal changes:
- Added `register_keras_trainer` decorator for registering new keras training classes
- load_data_sequence now takes in only one `DataConfig` argument, instead of a data path and ModelTrainingConfig
- a routine legacy_config_to_data_config is added to be used temporarily by test code to convert ModelTrainingConfig objects into DataConfig objects. The tests using this should be later refactored to create TrainingConfig and DataConfig objects directly.
- Moved tightly coupled code involving validation data into `_shared/config.py`. This code is quite prone to error, only being enabled for Keras training. It should be deleted in a follow-on PR, instead using a `DataConfig` for validation data in a way that works the same for Keras and Sklearn training.

- [x] Tests added


nbren12:
This is looking real nice @mcgibbon! Would it be worth grouping the objects into a top-level configuration like the prognostic run does: https://github.com/VulcanClimateModeling/fv3net/blob/a24d15ef763a2a157e62738dbf09738cb85fe8ad/workflows/prognostic_c48_run/runtime/config.py#L21.

It's a nice way to collect the interface and centralize the documentation.

mcgibbon:
> This is looking real nice @mcgibbon! Would it be worth grouping the objects into a top-level configuration like the prognostic run does

Yeah, that looks pretty nice!

I'll have to see depending on the backwards-compatibility-breaking change we go for. If we want to and can use a central object to contain all the configs in one yaml file, that is probably the way to go. It looks nice for external API presentation. The data configs basically immediately get transformed into batches, and then only the training config gets persisted, so there's no reason/need currently to group the configs internally in the application.

This will also depend on some things like, I want to refactor the configuration saving so that the configuration is attached to the model rather than being logged. Right now we duplicate "input variables" info between the model and the train_config.yaml, which makes it confusing when using the offline diagnostics on their own (in my experience). This may mean we want to provide the DataConfig to the training just so the model can save it, it depends on what people think is important.

I'll leave this potential change for now until those needs are more clear.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -68,15 +181,165 @@ def dump(self, path: str, filename: str = None) -> None:
...
+
+
+def validation_dataset(train_config: _ModelTrainingConfig,) -> Optional[xr.Dataset]:
+    if len(train_config.validation_timesteps) > 0:
+        check_validation_train_overlap(
+            train_config.batch_kwargs["timesteps"], train_config.validation_timesteps,
+        )
+        validation_config = validation_timesteps_config(train_config)
+        # validation config puts all data in one batch
+        print(validation_config)
```
AnnaKwa, line 327: Leftover print

external/fv3fit/fv3fit/train.py
```diff
@@ -88,47 +78,55 @@ def _get_model(config: ModelTrainingConfig) -> Estimator:
...
+        args.config_file,
+        data_path=data_path,
+        output_data_path=args.output_data_path,
+        timesteps_file=args.timesteps_file,
+        validation_timesteps_file=args.validation_timesteps_file,
+    )
     set_random_seed(train_config.random_seed)
 
-    batches = load_data_sequence(data_path, train_config)
+    train_config.dump(os.path.join(args.output_data_path, "train.yaml"))
```
AnnaKwa, line 106:
```suggestion
    train_config.dump(os.path.join(args.output_data_path, "train.yml"))
```

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,15 +2,128 @@
...
+    """Convenience wrapper for model training data
+
+    Attrs:
+        variables: names of variables to include in dataset
+        batch_function: name of function from `fv3fit.batches` to use for
+            loading batched data
+        batch_kwargs: keyword arguments to pass to batch function
+        data_path: location of training data to be loaded by batch function
+        validation_timesteps: timestamps to use as validation samples
+        timesteps_source: one of "timesteps_file",
```
AnnaKwa, line 61: `timesteps_source` and `validation_timesteps` deleted

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,15 +2,128 @@
...
+        scaler_type: scaler to use for training
+        scaler_kwargs: keyword arguments to pass to scaler initialization
+        additional_variables: list of needed variables which are not inputs
+            or outputs (e.g. pressure thickness if needed for scaling)
+        random_seed: value to use to initialize randomness
+        validation_timesteps: timestamps to use as validation samples
+        save_model_checkpoints: whether to save a copy of the model at
+            each epoch
+        model_path: output location for final model
+        timesteps_source: one of "timesteps_file",
```
AnnaKwa, line 94: Remove extra args that are no longer in this dataclass

AnnaKwa (changes requested):
Thanks, the splitting up of the big config looks good! Just one main request regarding the dump/load we talked about.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,15 +2,128 @@
...
+    save_model_checkpoints: bool = False
+
+
+@dataclasses.dataclass
+class SklearnTrainingConfig(TrainingConfig):
+
+    scaler_type: str = "standard"
+    scaler_kwargs: dict = dataclasses.field(default_factory=dict)
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
mcgibbon: I think it could mislead someone who sees this in the Keras config into thinking keras training supports this, so I'd like to leave it out until we make the support uniform.

AnnaKwa (approved):
Actually, I'm pretty sure I don't need to rereview for the requested change, so I'm approving with that request about removing dump/load. Could you run the integration test before merging just to make sure no breaking changes were missed?

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,15 +2,128 @@
...
+
+def register_keras_trainer(name):
+    def decorator(cls):
+        KERAS_MODELS[name] = cls
+        return cls
+
+    return decorator
+
+
+class Dumpable:
```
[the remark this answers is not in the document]
mcgibbon: Done. I refactored the tests to ones that ensure dataclasses.asdict, cls(**as_dict), and yaml.safe_dump all work on the config classes.

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: Removed.

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: Thanks, done. Added docstrings to the subclasses which include their additional options.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -68,15 +181,165 @@ def dump(self, path: str, filename: str = None) -> None:
...
+    overlap = set(train) & set(validate)
+    if overlap:
+        raise ValueError(
+            f"Timestep(s) {overlap} are in both train and validation sets."
+        )
+
+
+def validation_timesteps_config(train_config):
+    val_config = legacy_config_to_data_config(train_config)
+    print(val_config.data_path)
```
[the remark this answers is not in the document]
mcgibbon: Removed, thanks!

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: Removed.

external/fv3fit/fv3fit/train.py
```diff
@@ -88,47 +78,55 @@ def _get_model(config: ModelTrainingConfig) -> Estimator:
...
-
-    if args.validation_timesteps_file:
-        with open(args.validation_timesteps_file, "r") as f:
-            val_timesteps = yaml.safe_load(f)
-        train_config.validation_timesteps = val_timesteps
-
-    train_config.dump(args.output_data_path)
+    (
+        legacy_config,
+        train_config,
```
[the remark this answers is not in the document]
mcgibbon: I'm not sure what you mean, this is a configuration object, not a model object?

external/fv3fit/fv3fit/train.py
mcgibbon: I changed these new ones to use `.yaml` since that's what's [officially recommended](https://yaml.org/faq.html).

workflows/offline_ml_diags/tests/test_compute_diags.py
```diff
@@ -10,7 +10,10 @@
     dataset_fixtures_dir,
 )
 from fv3fit._shared import load_data_sequence
-from fv3fit._shared.config import ModelTrainingConfig
+from fv3fit._shared.config import (
+    _ModelTrainingConfig as ModelTrainingConfig,
```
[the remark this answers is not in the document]
mcgibbon: Yeah I did it just as a way to tell myself "this is bad, fix/delete it", I'll make this more explicit by adding TODO comments just above these imports. My next task is switching this out for the new config classes.

PR #1210, opened 2021-05-14
use a registration-based API to store keras training classes

mcgibbon:
Currently keras training classes (Estimators) are stored as imports in `fv3fit.keras._models`. This causes a circular import issue where configuration code wants to use fv3fit.keras._models, but model code wants to use configuration APIs. By having model code register itself into the configuration code, we can keep the dependency in one direction.

Significant internal changes:
- keras Estimator registration now works by using the decorator `@register_keras_estimator(name)` instead of importing into a particular module

- [x] Tests added


.pre-commit-config.yaml
```diff
@@ -24,4 +24,4 @@ repos:
       name: flake8 __init__.py files
       files: "__init__.py"
       # ignore unused import error in __init__.py files
-      args: ["--ignore=F401,E203"]
+      args: ["--ignore=F401,E203", --config, setup.cfg]
```
nbren12, line 27: Isn't this in master already: 4e001df7f3759389a9c84a6bcab392eca34c50cf?

external/fv3fit/fv3fit/_shared/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .config import ModelTrainingConfig, load_training_config
+from .config import ModelTrainingConfig, load_training_config, register_keras_trainer
```
nbren12, line 1: Should it be called "keras" if it's in shared? It seems broader in scope.  I think it's okay to name it more broadly, and then refactor the sklearn wrapper to use it later on.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,13 +2,39 @@
...
 
+KERAS_MODELS: Dict[str, type] = {}
+SKLEARN_MODEL_TYPES = ["sklearn", "rf", "random_forest", "sklearn_random_forest"]
+
+
+def get_keras_model(name):
+    return KERAS_MODELS[name]
+
+
+def register_keras_trainer(name: str):
```
nbren12, line 20: "trainer"  is a synonym of "estimator". Since the latter is an existing convention should you use that instead?

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -2,13 +2,39 @@
...
 import os
-from typing import Optional, Union, Sequence, List
+from typing import Dict, Optional, Union, Sequence, List
 
 
 DELP = "pressure_thickness_of_atmospheric_layer"
 MODEL_CONFIG_FILENAME = "training_config.yml"
 
 
+KERAS_MODELS: Dict[str, type] = {}
```
nbren12, line 12: I think you may be able to flesh this out more:

```suggestion
KERAS_MODELS: Dict[str, Type[Estimator]] = {}
```

nbren12 (approved):
I like this pattern as a way to circular dependencies!  I had some comments below about naming, placement of the feature in `_shared`.

Note that this pattern only provides a handle to the "constructor" of the class, and doesn't provide any static information about the arguments this class may have. Can it be extended in future PRs to support registering configurations?

.pre-commit-config.yaml
mcgibbon: Only for non-init files. This adds it to the init.py target.

external/fv3fit/fv3fit/_shared/__init__.py
mcgibbon: The issue right now is this is relied on to distinguish between needing a SklearnTrainingConfig or KerasTrainingConfig in my other PR. This can be generalized if/when I add support for model-level training configuration class definitions.

This currently exists in shared because it must exist in `config.py`. I could move that file to the top level outside of shared, instead?

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: OK, I can do that for now. I didn't worry about this too much because I'm definitely renaming "estimator" soon, which is a synonym of "predictor".

mcgibbon:
> Note that this pattern only provides a handle to the "constructor" of the class, and doesn't provide any static information about the arguments this class may have. Can it be extended in future PRs to support registering configurations?

That's a great idea, I'll try to use that as a way of generalizing the framework between Keras and Sklearn models. Currently the two have separate configuration classes in my other PR, but the classes could be defined at the model level instead of using this binary split.

external/fv3fit/fv3fit/_shared/config.py
nbren12: Huh? Estimate and predict are not synonyms. Predictors only predict, they don't have to be estimated/trained based on some dataset.

external/fv3fit/fv3fit/_shared/__init__.py
nbren12: The top-level would also be shared too IMO.

I think the easy move is to rename it to `register_estimator`, and then use it for the sklearn models too in a later PR once you merge the keras and sklearn factories.

It's okay to if sklearn doesn't use some general purpose shared code, but it is confusing to put code that can only be used for to constructing keras models in a shared location.

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: Huh? Why do you say `estimated/trained`, those aren't synonyms? Are you saying that "estimator" means it's estimating what the model should look like? One can equally take the class name to mean that it "estimates" the tendencies, and that's how I've understood the name up until now.

Thesaurus.com puts "predict" as one of the most relevant synonyms of "estimate". Googling "predict synonyms" and "estimate synonyms" says they're synonyms of one another.

nbren12: Maybe trainer and predictor is clearer. I always associated "estimation" with estimating the parameters of some model (aka training/learning), but I guess the trained/estimated model also "estimates" the predictands.

PR #1214, opened 2021-05-14
Light cleanup

AnnaKwa:
Fix indexing in logging of which batch is being fit in SklearnWrapper  https://github.com/VulcanClimateModeling/fv3net/issues/1200

Remove empty subplots from Jacobian matrices panels


nbren12 (approved):
Thanks!

PR #1216, opened 2021-05-18
Compute net heating with isochoric assumption if nonhydrostatic

oliverwm1:
Temperature tendencies are applied using isochoric assumption if using non-hydrostatic mode. Our heating diagnostics should reflect this. This PR implements this and uses vcm functions to standardize the computation of diagnostics.

Public API changes:
- add `isochoric` argument to `vcm.column_integrated_heat` and `vcm.column_integrated_heating`

Significant internal changes:
- add `hydrostatic` argument to `compute_diagnostics` in runtime
- certain column integrated diagnostics are modified due to change in the specific heat used (cv instead of cp for non-hydrostatic)
- other diagnostics are also changed because the `gravity` constant used to compute column integrals is now the version from vcm (which is consistent with the model's constant) instead of just 9.81
- restart checksum for the nudged run changes due to new `gravity` constant for net_moistening


external/vcm/vcm/calc/calc.py
```diff
@@ -1,17 +1,13 @@
...
-from vcm.cubedsphere.constants import INIT_TIME_DIM, COORD_Z_CENTER, VAR_LON_CENTER
+from vcm.cubedsphere.constants import INIT_TIME_DIM, VAR_LON_CENTER
 
 gravity = 9.81
 specific_heat = 1004
 
 HOUR_PER_DEG_LONGITUDE = 1.0 / 15
 
 
-def mass_integrate(phi, dp, dim=COORD_Z_CENTER):
```
oliverwm1, line 11: moved this function to `thermo.py` so that it uses a consistent value for `gravity`

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out
```diff
@@ -1 +1 @@
-566da3dda9be0fd64285cec8f3b93bed
+62a2754f455f3bd8764fb74f78b4770e
```
Comment (oliverwm1) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out:1]: because the value of `gravity` used to compute net moistening has changed, the surface precipitation in a nudged or prognostic run changes, and I think this is why the checksum for the nudging restart file changes. I'm not sure why the restart checksum does not change for the sklearn test...

Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out:1]: It could be the case that the mock sklearn model only acts to moisten the atmosphere, in which case the surface precipitation change due to ML would be zero.

external/vcm/vcm/__init__.py
```diff
@@ -25,6 +25,8 @@
     potential_temperature,
     pressure_at_interface,
     surface_pressure_from_delp,
+    column_integrated_heat,
```
nbren12, line 28: This would be a good opportunity to use a clearer name than "heat". E.g. `column_enthalpy` or `column_internal_energy`.

external/vcm/vcm/calc/thermo.py
```diff
@@ -396,22 +399,25 @@ def liquid_ice_temperature(
 
 
 def column_integrated_heat(
-    temperature: xr.DataArray, delp: xr.DataArray, vertical_dim: str = "z"
+    temperature: xr.DataArray,
+    delp: xr.DataArray,
+    isochoric: bool = True,
```
nbren12, line 404: I think this function is ambiguous and could lead to more confusion later on. 

- "heat" is not very specific.
- "isochoric" refers to holding volume constant while the other phase variables change, but this function does not involve any thermodynamic transitions. It is a function of state.

I suggest making two functions: `column_enthalpy` and/or `column_internal_energy` to disambiguate.

nbren12 (changes requested):
Thanks for correcting this bug. Unfortunately, I think the naming of these functions is still confusing and doesn't line up well with the first law of thermodynamics.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out
Comment (nbren12) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out:1]: Would be nice to move the constant change to another PR, so I can be certain that the refactors didn't break anything.

external/vcm/vcm/calc/thermo.py
```diff
@@ -420,20 +426,24 @@ def column_integrated_heat(
 
 
 def column_integrated_heating(
-    dtemperature_dt: xr.DataArray, delp: xr.DataArray, vertical_dim: str = "z"
```
nbren12, line 423: `isochoric=False` does not imply that it is isobaric. Instead of using the flag I suggest two functions

```
column_integrated_heating_from_isobaric_transition(temperature_change_isobaric, ...)
column_integrated_heating_from_isochoric_transition(temperature_change_isochoric, ...)
```

nbren12 (changes requested):
Some more thermo nits.

nbren12:
Personally, i think we should remove these functions and come up with new ones based on the two forms of the first law:

- `cp dT = Heating + V dp`
- `cv dT = Heating - p dV`

nbren12:
Also the PR description is not 100% accurate
> Temperature tendencies are applied using isochoric assumption if using non-hydrostatic mode. 
should be modified to "Temperature updates applied outside of the physics package are assumed to occur at constant volume (e.g. isochoric) if using non-hydrostatic mode."

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -69,7 +68,9 @@ def label(self) -> str:
     def __call__(self, time, state) -> Tuple[Tendencies, Diagnostics, State]:
         return {}, {}, {}
 
-    def get_diagnostics(self, state, tendency) -> Tuple[Diagnostics, xr.DataArray]:
+    def get_diagnostics(
+        self, state, tendency, hydrostatic: bool
```
nbren12, line 72: New argument here.

this is another sign that formulating our ML in terms of non-conserved variables results in undesirable coupling between the ML and the FV3GFS model physics/configurations.

I know you are looking for the quick fix, but would it really be that hard to switch our ML procedures to use "heating" rather than temperature changes? For N2F, it seems like that would result in a smaller diff than this PR. Not sure about N2O.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out
Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_checksum_restarts[nudging].out:1]: Yeah, that's a good idea, I can isolate the gravity constant change to it's own PR.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: This isn't a new coupling... even our coarsening procedure is tied to the FV3GFS model physics via the gravity constant used.

I think switching to using heating is interesting, but not sure I agree it's a minimal change since all of our training datasets would need to change.

external/vcm/vcm/calc/thermo.py
oliverwm1: I'll probably just undo the changes (or delete this function) since it's not used anywhere in our codebase as far as I can tell. The only difference between this function and `column_integrated_heating` is the attributes attached to the output array.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: It would be nice if we could grab the value of the relevant constants from the Fortran state within runtime though.

nbren12: > This isn't a new coupling

The `hydrostatic` argument is a new piece of data being passed to the stepper, and is a breaking change to the Stepper interface, as we envision adding new steppers, this would be a step in the wrong direction. I would feel better about it if it were passed to the `__init__` of the Stepper, rather than changing the signature of Stepper.get_diagnostics.

oliverwm1:
Closing in favor of #1223

PR #1220, opened 2021-05-19
Delete older loaders/tests/test_mappers

AnnaKwa:
Originally this PR was going to refactor the tests in this subdir of the loaders' tests to use the `XarrayMapper` instead of the obsolete `FineResolutionSources` (https://github.com/VulcanClimateModeling/fv3net/issues/1198), but I instead opted to remove the directory. The tests in `tests/test_mappers`
- test that mappers in list are subclasses of GeoMapper or LongRunMapper, but we don’t appear to be adding new mappers here
- test that required training variables are in the mapper, dimensions of data are correct, but this feels like these tests that should be done on the individual mappers if they have their own tests and synth data in the top level of tests. In test_mappers.py the only mapper tested would be the XarrayMapper which would trivially pass these tests because it’s initialized with a manually created dataset. It already is tested in `test_xarray.py`, so there is not much additional coverage in this directory to warrant all the fixtures set up here.


nbren12 (approved):
Thanks! It should be easier to delete the old fine_res mappers now.

PR #1223, opened 2021-05-20
Isochoric/isobaric transition based on hydrostatic flag

oliverwm1:
This PR ensures that the column integrated heating diagnostics output by prognostic runs and in offline diagnostics respect the isochoric vs isobaric assumption in the dycore. For now, the offline diagnostics hard-code the isochoric assumption (a breaking change). The prognostic runs choose based on the `hydrostatic` flag (also a breaking change).

Added public API:
- added `vcm.calc.thermo.column_integrated_heating_from_isobaric_transition`
- added `vcm.calc.thermo.column_integrated_heating_from_isochoric_transition`

Refactored public API:
- deleted `vcm.calc.thermo.column_integrated_heat`

Significant internal changes:
- pass `fv_core_nml.hydrostatic` flag to ML and nudging stepper initializations so they can compute column integrated heating appropriately


workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out
```diff
@@ -4,7 +4,7 @@ area: f672e2d555d61777324a730a427f4c22
 cnvprcp_after_physics: a0279aa038e7f90e25d8828bb1fc853b
 cnvprcp_after_python: a0279aa038e7f90e25d8828bb1fc853b
 evaporation: 5daea270430161b4f40aac01c110be63
-net_heating_due_to_nudging: 9ea773b10d712efc2f135505fdc81dd7
+net_heating_due_to_nudging: 3c30e5e85a39c89e346f171c04eb1380
```
Comment (oliverwm1) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:7]: because both the `sklearn` and `keras` mock ML models predict 0 for all the heatings, the nudging run is the only that catches this difference in net heating diagnostic.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -185,7 +186,14 @@ def __call__(self, time, state):
         )
 
         if "dQ1" in tendency:
-            diag = thermo.column_integrated_heating(dQ1_updated - tendency["dQ1"], delp)
+            if self.hydrostatic:
+                diag = thermo.column_integrated_heating_from_isobaric_transition(
```
nbren12, line 190: Can you use a more specific name than "diag"?

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/add_derived.py
```diff
@@ -72,14 +72,16 @@ def _column_pq2(ds: xr.Dataset) -> xr.DataArray:
 
 def _column_dq1(ds: xr.Dataset) -> xr.DataArray:
     if "net_heating_due_to_machine_learning" in ds:
-        column_dq1 = ds.net_heating_due_to_machine_learning
+        # fix isochoric vs isobaric transition issue
+        column_dq1 = 716.95 / 1004 * ds.net_heating_due_to_machine_learning
```
nbren12, line 76: Should this use the same physical constants as the prognostic run?

Also, what is the roadmap for removing support for the old names? It would be nice to at least raise a deprecation warning if the old names are encountered.

nbren12 (approved):
Thanks. Some small suggestions below.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/add_derived.py
oliverwm1: These are the same constants used in VCM for the column heating functions: 1004 for cp and 287.05 for Rd and therefore 716.95 for cv. We might want to consider computing using a moist version of cp/cv, which is what the Fortran model does (e.g. for total energy) but that is a separate change IMO.

I added deprecation warnings.

PR #1225, opened 2021-05-20
Offline diags: Separate page for model sensitivity figures

AnnaKwa:
This PR moves the section containing the Jacobian figures (if applicable) to a separate page linked from the main report page. A future PR will add figures for feature importance if the model in the report is a RF.

There is a minor change to how the model sensitivity figures are saved in the `compute_diags` step; instead of being saved in the top level of the output directory they go into a subdir. This is so that if future pages are added to the report they can use the same pattern of using `copy_pngs_to_report(figures_dir, output)` to put the relevant figures on the page.

Example: https://storage.googleapis.com/vcm-ml-public/annak/2021-05-20-offline-report-add-pg/index.html

### DIFF
--- a/workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
+++ b/workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
@@ -415,9 +415,12 @@ def main(args):
     ds_diagnostics["time"] = times_used
     ds_diurnal["time"] = times_used
 
-    # save jacobian
+    # save model senstivity figures: jacobian (TODO: RF feature sensitivity)
     try:
-        plot_jacobian(model, args.output_path)  # type: ignore
+        plot_jacobian(
+            model,
+            os.path.join(args.output_path, "model_sensitivity_figures"),  # type: ignore
+        )
     except AttributeError:
         pass
 
--- a/workflows/offline_ml_diags/offline_ml_diags/create_report.py
+++ b/workflows/offline_ml_diags/offline_ml_diags/create_report.py
@@ -8,13 +8,12 @@ from typing import MutableMapping, Sequence, List
 
 import fv3viz
 import numpy as np
-from report import insert_report_figure
+import report
 import vcm
 from vcm.cloud import get_fs
 import diagnostics_utils.plot as diagplot
 from ._helpers import (
     get_metric_string,
-    write_report,
     open_diagnostics_outputs,
     copy_outputs,
     tidy_title,
@@ -123,6 +122,8 @@ if __name__ == "__main__":
 
     report_sections: MutableMapping[str, Sequence[str]] = {}
 
+    # Links
+
     # histogram of timesteps used for testing
     try:
         timesteps = ds_diurnal["time"]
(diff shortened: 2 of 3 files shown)


workflows/offline_ml_diags/offline_ml_diags/create_report.py
```diff
@@ -255,16 +238,60 @@ def _create_arg_parser() -> argparse.Namespace:
...
+        ds_diags, config["output_variables"]
+    )
+    ds_diurnal = drop_temperature_humidity_tendencies_if_not_predicted(
+        ds_diurnal, config["output_variables"]
+    )
+
+    config.pop("mapping_kwargs", None)  # this item clutters the report
+    if args.commit_sha:
+        config["commit"] = args.commit_sha
+
```
AnnaKwa, line 282: Everything in this block is the same as before up to line 283. Github just marked the diff like this because I moved a large block of`__main__` into the `render_index` function.

workflows/offline_ml_diags/offline_ml_diags/create_report.py
```diff
@@ -86,43 +87,25 @@ def _create_arg_parser() -> argparse.Namespace:
...
-        ds_diurnal, config["output_variables"]
+    return report.create_html(
+        sections=report_sections, title="Model sensitivity to inputs",
     )
 
-    config.pop("mapping_kwargs", None)  # this item clutters the report
-    if args.commit_sha:
-        config["commit"] = args.commit_sha
 
+def render_index(metrics, ds_diags, ds_diurnal, ds_transect, output_dir) -> str:
```
AnnaKwa, line 101: This function is the original code from `__main__` refactored out into its own function.

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
```diff
@@ -415,9 +415,12 @@ def main(args):
     ds_diagnostics["time"] = times_used
     ds_diurnal["time"] = times_used
 
-    # save jacobian
+    # save model senstivity figures: jacobian (TODO: RF feature sensitivity)
     try:
-        plot_jacobian(model, args.output_path)  # type: ignore
+        plot_jacobian(
```
oliverwm1, line 420: not new to this PR, so this is just a clarification question, but why does this plot get created as part of "compute diags" instead of as part of the report generation like everything else? I was confused for a while since `render_model_sensitivity` wasn't actually generating any figures.

oliverwm1 (approved):
Thanks, it's helpful to have a bit more organization in the offline report and code! we might want to consider doing the same for other sections of the report.

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
AnnaKwa: Since the jacobian calculation is part of the keras model, doing this here avoids having to read in the model in the `create_report` step.

oliverwm1: right, I guess what I mean is, shouldn't "compute diags" calculate the Jacobian and then save it to one of the netCDFs to then get loaded/plotted later?

PR #1228, opened 2021-05-24
Offline report: RF feature importance figures

AnnaKwa:
This adds figures for the random forest's feature importances to the offline report, if applicable.

- [x] Tests added

Example: https://storage.googleapis.com/vcm-ml-public/annak/2021-05-24-rf-feature-importance/model_sensitivity.html


workflows/offline_ml_diags/offline_ml_diags/_helpers.py
```diff
@@ -46,6 +48,23 @@
...
 
+def get_variable_indices(
+    data: xr.Dataset, variables: Sequence[str]
+) -> Mapping[str, Tuple[int, int]]:
+    if "time" in data.dims:
+        data = data.isel(time=0).squeeze(drop=True)
+    stacked = data.stack(sample=["tile", "x", "y"])
+    variable_dims = count_features_2d(
+        variables, stacked.transpose("sample", ...), "sample"
+    )
```
frodre, line 59: Instead of using the previously private function, maybe use a Packer and then infer feature indices from there?  Otherwise, I'd suggest writing the logic here to detect features.  I just worry about it always being specific to the packer without intentionally and clearly linking to a packer here.  Also, I don't know if this overlaps with any of the fv3fit changes Jeremy is planning on making.

workflows/offline_ml_diags/offline_ml_diags/_helpers.py
```diff
@@ -46,6 +48,23 @@
...
+    start = 0
+    variable_indices = {}
+    for var in variables:
+        variable_indices[var] = (start, start + variable_dims[var])
+        start += variable_dims[var]
```
frodre, line 64: I assume variable_dims retains the same order as variables so you could use the `.items()` instead of two dictionary accesses.

workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py
```diff
@@ -78,3 +81,88 @@ def plot_jacobian(model: fv3fit.keras._models.DenseModel, output_dir: str):
...
+    mean_importances = np.array(importances).mean(axis=0)
+    std_importances = np.array(importances).std(axis=0)
+
+    n_vector_features = len(
+        [
+            var
+            for var in input_feature_indices
+            if ((input_feature_indices[var][1] - input_feature_indices[var][0]) > 1)
+        ]
+    )
```
frodre, line 104: Might be simpler to grok with a simpler loop
```
n_vector_features = 0
for var in ...:
    feature_len = 
    if feature_len > 1:
        n_vecture_features += 1
```

workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py
```diff
@@ -78,3 +81,88 @@ def plot_jacobian(model: fv3fit.keras._models.DenseModel, output_dir: str):
...
+
+
+def _subplot_vector_feature_importances(
+    axs, variable_indices, mean_importances, std_importances
+):
+    vector_features = [
+        var
+        for var in variable_indices
+        if (variable_indices[var][1] - variable_indices[var][0] > 1)
+    ]
```
frodre, line 130: Repeated from above.  Good target for a smaller helper function for consolidation.

workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py
```diff
@@ -78,3 +81,88 @@ def plot_jacobian(model: fv3fit.keras._models.DenseModel, output_dir: str):
...
+    for i, feature in enumerate(vector_features):
+        dim_length = variable_indices[feature][1] - variable_indices[feature][0]
+        axs[0, i].errorbar(
+            range(dim_length),
+            mean_importances[
+                variable_indices[feature][0] : variable_indices[feature][1]
```
frodre, line 136: Multiple deep accesses to the same item means it should probably get its own variable assignment.  Good for readability too!

frodre: e.g.,
```
feature_start = ...[1]
feature_end = ...[0]
selection = slice(feature_start, feature_end)
mean_importances[selection]
....

File: workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py (hunk 3 above)
Comment (frodre) [reply workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py:104]: Since we're already grabbing the lengths and making a determination of scalar or vector, we could separate things out up here and pass them into their respective functions instead of redoing the separation in each individual helper.

File: workflows/offline_ml_diags/offline_ml_diags/_plot_input_sensitivity.py [hunk 6]
```diff
@@ -78,3 +81,88 @@ def plot_jacobian(model: fv3fit.keras._models.DenseModel, output_dir: str):
...
+        )
+        axs[0, i].set_xlabel(f"{feature}, model level")
+    return axs
+
+
+def _subplot_scalar_feature_importances(
+    axs, variable_indices, mean_importances, std_importances
+):
+    # Plot the scalar feature importances together in the last axes object
+    scalar_features = [
```
frodre, line 150: In general, I think these two helpers could be simplified to just the plotting with passed-in importances.

workflows/offline_ml_diags/tests/test_helpers.py
```diff
@@ -5,9 +5,43 @@
...
 
 
+def data_array_with_feature_dim(feature_dim: int, has_time_dim=True):
+    da = xr.DataArray(
+        [
+            [
+                [[[1 for x in range(2)] for y in range(2)] for z in range(feature_dim)]
+                for tile in range(2)
+            ]
+        ],
```
frodre, line 19: Gave a shot at a simpler statement
```
data = np.arange(8*feature_dim).reshape(1, 2, 2, feature_dim, 2)
```

frodre (changes requested):
Thanks, @AnnaKwa for the update!  There is a good start but there are a few things I think could improve this addition.

I worry about grabbing deeply nested private functions and just making them public without a solid need.  Mainly since it bakes in whatever `ArrayPacker` needs are for feature determination.   Since we don't use the packer in `get_variable_indices` and since you wrote tests for that function, I would probably leave the ArrayPacker function private and write whatever feature counting logic necessary.

Additionally, I felt parts of the plotting code were fairly complex.  I think there are some good opportunities for consolidating operations and simplifying the syntax. Happy to chat tomorrow about anything here.

workflows/offline_ml_diags/offline_ml_diags/_helpers.py
AnnaKwa: That make sense, I'll write a copy of the logic in this module as opposed to importing the packer.

workflows/offline_ml_diags/tests/test_helpers.py
AnnaKwa: So much saner!

AnnaKwa:
Thanks Andre, the plotting code is much tidier now with the suggestions implemented! This is ready for re-review.

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
```diff
@@ -422,7 +424,17 @@ def main(args):
             os.path.join(args.output_path, "model_sensitivity_figures"),  # type: ignore
         )
     except AttributeError:
-        pass
+        try:
+            input_feature_indices = get_variable_indices(
+                data=pred_mapper[timesteps[0]], variables=model.input_variables
+            )
+            plot_rf_feature_importance(
```
oliverwm1, line 431: Ideally we shouldn't be creating plots during the `compute_diags.py` script. Can we just compute the data necessary to make this plot and save it to be loaded/plotted in the report script? This would be really helpful if, for example, we want to load and compare the diagnostic for a couple different models.

AnnaKwa: For these cases since they require the model (which the report step doesn't load) and the calculation of the input sensitivities is a one liner for the keras models (`model.jacobian`) and will be a one liner for the RF as well after Jeremy's fv3fit refactoring. In the case where we want compare these for multiple models in a notebook it is currently not too bad to just load the models and directly calculate the input sensitivities in the notebook rather than save the sensitivity diagnostic in an intermediate format that has to be opened and parsed by the report script. I do agree that ideally these should be treated like the other diagnostics. That would require a fair amount of changes here as well as the report, so to keep this PR scope limited I will leave the plot saving as it is here, and open an issue to resolve in a follow on PR. Let me know if that sounds ok @oliverwm1 @frodre

oliverwm1: Okay sounds good. I don't really know what the code is like to compute the sensitivities, so thought it might be nice to be able to directly open the results in a notebook. Definitely agree no changes required for this PR! But if you agree it's worth the effort, would be nice to address in the future.

workflows/offline_ml_diags/offline_ml_diags/_helpers.py
```diff
@@ -46,6 +47,52 @@
...
+) -> Mapping[str, int]:
+    """
+    count features for (sample[, z]) arrays.
+    Copied from fv3fit._shared.packer, as this logic is pretty robust.
+    """
+    for name in quantity_names:
+        if len(dataset[name].dims) > 2:
+            value = dataset[name]
+            raise ValueError(
+                "can only pack 1D/2D (sample[, z]) "
```
frodre, line 61: I'd consider switching "pack" to "count" for these errors since it might be confusing w/ no packing happening.

frodre (approved):
Thanks for the updates, @AnnaKwa.  LGTM!

PR #1229, opened 2021-05-25
Remove output path from keras fit logic

mcgibbon:
Currently, keras' fit routine is told where to dump checkpoint files. This makes it impossible to define a fit API based on a yaml configuration, because the output directory is a command-line argument.

This PR refactors keras's fit routine so checkpoints are only saved to the output directory when `model.dump` is called.

Significant internal changes:
- When checkpoint saving is enabled for keras fitting, it now saves the checkpoints to a temporary directory which gets copied to the output directory when the model dump routine is called.

- [ ] Tests added


nbren12 (approved):
LGTM. I agree the old way put the cart before the horse, and made the final "dump" procedure less atomic.

BTW, I just noticed that PackedKerasModel.fit and  `SklearnWrapper.fit` do not have the same interface. The former has many key word arguments which should probably be moved into its `__init__`.

PR #1230, opened 2021-05-25
prog diags: allow save to save remote file

nbren12:
Unlike many of our CLIs, prognostic_run_diag save local/path gs://remote/output will fail.


frodre (approved):
Nice catch.  Thanks for the quick update.

PR #1236, opened 2021-05-27
Add hyperparameter config, use generic training interfaces

mcgibbon:
Currently, sklearn and keras training have separate pre-processing logic before their fit routines are called. Their configuration also loosely types the keyword arguments, by accepting kwargs. This opens the possibility for typos in configuration to go unnoticed, and makes it hard to document what options are available. It also puts model-specific concerns into the top-level routines, instead of being able to dispatch those concerns to Estimator classes.

This PR adds the concept of a hyperparameter configuration class for an Estimator, which contains all of the init arguments for that estimator. This is integrated into the TrainingConfig, and is the only part of the configuration that differs between different estimators.

Added public API:
- Added OptimizerConfig and RegularizerConfig which replace the current arguments for optimizer and regularizer on DenseModel

Refactored public API:
- Removed DenseTrainingConfig and SklearnTrainingConfig, added DenseHyperparameters and RandomForestHyperparameters which strongly type the hyperparameter configuration
- Added hyperparameter class to the API for `register_estimator`
- Renamed `get_config_class` to `get_hyperparameter_class`
- Predictor now takes `**kwargs` in its API to satisfy mypy, but raises a TypeError if any are passed

Significant internal changes:
- Modified Estimator classes to take a single "hyperparameters" dataclass argument, instead of multiple keyword arguments
- added RandomForest estimator which wraps the existing SklearnWrapper estimator, and also provides functionality from now-deleted `sklearn/_train.py`
- Removed target_scaler and n_jobs from the SklearnWrapper init, as there was no way to set these from configuration

- [ ] Tests added


external/fv3fit/fv3fit/train.py
```diff
@@ -51,28 +51,13 @@ def get_parser():
...
-            input_variables=config.input_variables,
-            output_variables=config.output_variables,
-            optimizer=get_optimizer(config.hyperparameters),
-            kernel_regularizer=get_regularizer(config.hyperparameters),
-            save_model_checkpoints=config.save_model_checkpoints,
-            **config.hyperparameters,
-        )
-    else:
-        raise NotImplementedError(f"Model type {config.model_type} is not implemented.")
+    cls = get_estimator_class(config.model_type)
```
mcgibbon, line 54: Note the central purpose of this PR is to be able to refactor this function like this.

external/fv3fit/fv3fit/sklearn/_wrapper.py
```diff
@@ -32,6 +34,66 @@ def _tuple_to_multiindex(d: tuple) -> pd.MultiIndex:
     return pd.MultiIndex.from_tuples(list_, names=names)
 
 
+@_shared.io.register("sklearn")
+@register_estimator("sklearn", RandomForestHyperparameters)
+@register_estimator("rf", RandomForestHyperparameters)
+@register_estimator("random_forest", RandomForestHyperparameters)
+@register_estimator("sklearn_random_forest", RandomForestHyperparameters)
+class RandomForest(Estimator):
```
nbren12, line 42: Nice. I like this way of providing a clean interface to the overly complex sklearn code. It might be worth renaming this module to `_random_forest.py` and making `RegressorEnsemble` and `SklearnWrapper` private APIs.

nbren12 (commented):
Looking good! I'll let someone else review more closely. Some general comments/requests:

-  can you shorten the PR title if you plan to make it a commit message?
-  There are a lot of `Optional` configurations, and each one of these requires some conditional logic to disambiguate. You could use more concrete defaults in places to avoid this.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -358,12 +357,16 @@ def load(cls, path: str) -> "PackedKerasModel":
                 y_scaler = LayerStandardScaler.load(f_binary)
             with open(os.path.join(path, cls._OPTIONS_FILENAME), "r") as f:
                 options = yaml.safe_load(f)
+            print(options)
```
brianhenn, line 360: Leftover?

external/fv3fit/fv3fit/keras/_training.py
```diff
@@ -19,7 +23,7 @@ def get_model(
     sample_dim_name: str,
```
brianhenn, line 23: Haven't looked through the tests enough to see why this function is still needed, but it would be nice to be able to get rid of it since it's no longer used by the training routine. OTOH this PR is pretty big already...

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -48,25 +54,24 @@ class TrainingConfig:
...
-            result = subclass.from_dict(kwargs)
-        else:
-            result = cls(**kwargs)
-        return result
+        kwargs = {**kwargs}  # make a copy to avoid mutating the input
+        hyperparameter_class = get_hyperparameter_class(kwargs["model_type"])
+        kwargs["hyperparameters"] = dacite.from_dict(
+            data_class=hyperparameter_class, data=kwargs["hyperparameters"]
+        )
+        return cls(**kwargs)
```
brianhenn, line 68: Couldn't this be `dacite.from_dict(cls, **kwargs)`?

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -317,9 +318,7 @@ def dump(self, path: str) -> None:
             with open(os.path.join(path, self._Y_SCALER_FILENAME), "wb") as f_binary:
                 self.y_scaler.dump(f_binary)
             with open(os.path.join(path, self._OPTIONS_FILENAME), "w") as f:
-                yaml.safe_dump(
-                    {"normalize_loss": self._normalize_loss, "loss": self._loss}, f
-                )
+                yaml.safe_dump(dataclasses.asdict(self._hyperparameters), f)
```
brianhenn, line 321: This is great to be able to dump this estimator's hyperparameters in a self-defining way.

brianhenn (commented):
Thanks @mcgibbon this is going to be great. I found a number of small issues, and obviously need to make sure the e2e tests can pass since there is some coupling between fv3fit and the offline diags.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -113,53 +118,94 @@ class DataConfig:
...
+class RandomForestHyperparameters:
+    scaler_type: str = "standard"
+    scaler_kwargs: Optional[Mapping] = None
+    n_jobs: int = -1
+    random_state: int = 0
+    n_estimators: int = 100
+    max_depth: Optional[int] = None
+    min_samples_split: Union[int, float] = 2
+    min_samples_leaf: Union[int, float] = 1
+    max_features: Union[Literal["auto", "sqrt", "log2"], int, float] = "auto"
```
AnnaKwa, line 208: For flexibility in the future could you also add defaults `{"max_samples": Union[int, float]=None, "bootstrap": bool=True, "criterion": str="mse", "warm_start": bool=True}

The last especially is needed for removing `RegressorEnsemble`.

external/fv3fit/fv3fit/sklearn/_wrapper.py
```diff
@@ -32,6 +33,58 @@ def _tuple_to_multiindex(d: tuple) -> pd.MultiIndex:
...
+    def __init__(
+        self,
+        sample_dim_name: str,
+        input_variables: Iterable[str],
+        output_variables: Iterable[str],
+        hyperparameters: RandomForestHyperparameters,
+    ):
+        batch_regressor = RegressorEnsemble(
+            sklearn.ensemble.RandomForestRegressor(
+                n_jobs=hyperparameters.n_jobs,
```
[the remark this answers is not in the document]
AnnaKwa: I think in the current use case (using `RegressorEnsemble` as a hacky forest and 1 tree per sklearn RF) neither the higher-up `SklearnWrapper` n_jobs nor the sklearn model hyperparameter n_jobs are doing any parallelization of subsetted data after splits. Since there's no difference between picking one of those to set right now I think it would be easier for the future PR that removes the `RegressorEnsemble` to have `n_jobs` set as the `RandomForestRegressor` hyperparameter in this PR, so that this code doesn't have to get adjusted later. At least I think that is how the parallelization is currently working (or not working).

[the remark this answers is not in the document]
brianhenn: Isn't the current parallel backend context in `predict` parallelizing calls over the RFs in the `RegressorEnsemble` if `n_jobs = -1`?

external/fv3fit/fv3fit/sklearn/_wrapper.py
```diff
@@ -237,13 +282,8 @@ def load(cls, path: str) -> "SklearnWrapper":
 
         output_features_ = _tuple_to_multiindex(output_features_dict_)
 
-        obj = cls(
-            sample_dim_name,
-            input_variables,
-            output_variables,
-            model,
-            target_scaler=scaler_obj,
```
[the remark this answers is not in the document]
AnnaKwa: Oh, good catch. I think this will cause an error when trying to use `.predict` from a loaded model.

external/fv3fit/fv3fit/typing.py
```diff
@@ -0,0 +1,6 @@
+from typing_extensions import Protocol
+from typing import Dict
+
+
+class Dataclass(Protocol):
```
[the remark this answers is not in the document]
mcgibbon: Yes, there's no built-in type so you have to make a Protocol object manually for hinting. I added a link to SO.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -113,53 +118,94 @@ class DataConfig:
...
-        if self.scaler_type == "mass":
-            if DELP not in self.additional_variables:
-                self.additional_variables.append(DELP)
+@dataclasses.dataclass
+class RandomForestHyperparameters:
+    scaler_type: str = "standard"
+    scaler_kwargs: Optional[Mapping] = None
+    n_jobs: int = -1
+    random_state: int = 0
+    n_estimators: int = 100
```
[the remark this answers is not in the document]
brianhenn: OK as discussed that's fine

external/fv3fit/fv3fit/sklearn/_wrapper.py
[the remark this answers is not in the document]
brianhenn: OK as discussed it's not actually doing any parallelism currently so we can delete that context and follow @AnnaKwa's suggestion

AnnaKwa:
Thanks Jeremy, the new interface is much cleaner! The responses to the couple of comments I left LGTM.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -380,7 +367,7 @@ def load_configs(
     else:
         validation_data_config = None
 
-    return legacy_config, training_config, train_data_config, validation_data_config
+    return training_config, train_data_config, validation_data_config
```
brianhenn, line 370: update function type signature to reflect

brianhenn (commented):
Thanks I think my issues have been addressed. Happy to approve once tests are passing.

external/fv3fit/fv3fit/sklearn/_wrapper.py
mcgibbon: Also, this has exposed that offline_ml_diags/_plot_input_sensitivity.py is tightly coupled to fv3fit's internals, because it type hints specific model types rather than having a jacobian/feature importances method defined. I'll work on that in another PR.

mcgibbon:
Integration tests passed https://app.circleci.com/pipelines/github/VulcanClimateModeling/fv3net/8551/workflows/c46866c3-1f89-467c-b87b-457ab89c69a1

brianhenn (approved):
Thanks Jeremy!

PR #1240, opened 2021-06-02
Decouple prognostic run tests using constant predictor

mcgibbon:
Currently the prognostic run tests depend on implementation details of fv3fit estimators. This PR decouples this by providing a predictor for testing in fv3fit, and using this predictor in all prognostic run tests.

Added public API:
- Added fv3fit.testing.ConstantOutputPredictor for use in integration tests

Significant internal changes:
- prognostic_c48_run tests now use fv3fit.testing.ConstantOutputPredictor instead of mocks of keras and sklearn-based models

- [x] Tests added


nbren12 (commented):
I think breaking the tests was somewhat expected here,  I wouldn't over-index on it since the regression tests are designed to detect changes in the prognostic runtime code not the mocks/fv3fit. I do wonder if `ConstantOutputPredictor` could be simplified somewhat now that we don't need to construct the more complex DummyRegressor objects any longer. In principle it only needs a `.load` which does nothing and a .predict function which could return `eps * sin(t)` or something.

nbren12:
You should delete the old `.out` files and regenerate the regression data with `--regtest-reset` btw.

mcgibbon:
I'm up for any additional ways to simplify the mock class, but I think this is as simple as it can be for the moment.

For the `test_regression` tests we need to be able to dump the model to a run directory for use through subprocess. That means it needs to be registered and write at least something to disk, because that's the only way the prognostic run knows to initialize this class. Past that, ideally we need to be able to set the predicted values from the regression test framework, which is done through npz dump/load. And we need to know the input/output variable names in the prognostic run, which can again only be done through disk because of subprocess.

mcgibbon:
Biggest debatable change here is adding `dump` to predictor.

With the current API it’s technically not possible to load a class that implements Predictor unless it’s an Estimator, because only Estimator can be written to disk. In my BPTT work I just lied in the API and said my class was an Estimator when it isn’t. For this PR this change means being able to define a dummy predictor without needing to define `fit` (or similarly lie about being an Estimator).

This lines up with some plans later to turn `Estimator` into a class (or even just a function) that returns a `Predictor` to be dumped, instead of dumping/loading itself.

nbren12 (changes requested):
JM and I spoke over zoom, and agreed to 
1. revert the changes to the the `Predictor` and `Estimator` base classes in this PR, and then
2. consider a more sensible grouping of .fit/.load/.dump and .predict in the future. This may look like making a plain `fit(config, data)` function which returns a object with .load/.dump/.predict.

nbren12 (approved):

external/fv3fit/fv3fit/testing.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        super().__init__(
+            sample_dim_name=sample_dim_name,
+            input_variables=input_variables,
+            output_variables=output_variables,
+        )
+        self._outputs: Dict[Hashable, Union[np.ndarray, float]] = {}
+
+    def set_outputs(self, **outputs: Union[np.ndarray, float]):
+        """
```
brianhenn, line 40: Slightly more info here would be helpful in writing a test using this., e.g., that the output variables will have dimensions (sample_dim_name, "z") if a 1-D array is passed and (sample_dim_name) if a scalar is passed.

external/fv3fit/fv3fit/testing.py
```diff
@@ -0,0 +1,89 @@
...
+            output = self._outputs[name]
+            if isinstance(output, np.ndarray):
+                array = np.repeat(output[None, :], repeats=n_samples, axis=0)
+                data_vars[name] = xr.DataArray(
+                    data=array, dims=[self.sample_dim_name, "z"]
+                )
+            else:
+                array = np.full([n_samples], float(output))
+                data_vars[name] = xr.DataArray(data=array, dims=[self.sample_dim_name])
+        if self.sample_dim_name in X:
```
brianhenn, line 61: Nit:
```suggestion
        if self.sample_dim_name in X.coords:
```

external/fv3fit/tests/test_keras_testing.py
```diff
@@ -63,3 +71,101 @@ def test_dummy_model(dummy_model, gridded_dataset):
...
+    predictor.set_outputs(**outputs)
+    ds_stacked = safe.stack_once(
+        gridded_dataset, "sample", [dim for dim in gridded_dataset.dims if dim != "z"]
+    ).transpose("sample", "z")
+
+    ds_pred = predictor.predict(ds_stacked)
+
+    assert sorted(list(ds_pred.data_vars.keys())) == sorted(output_variables)
+    for name in output_variables:
+        assert np.all(ds_pred[name].values == outputs[name][None, :])
```
brianhenn, line 101: Wasn't obvious to me that this relies on broadcasting for the equality, maybe make it more so by using `xr.broadcast_like` or `np.repeat`?

external/fv3fit/tests/test_keras_testing.py
```diff
@@ -63,3 +71,101 @@ def test_dummy_model(dummy_model, gridded_dataset):
...
+
+@pytest.mark.parametrize(
+    "input_variables, output_variables",
+    [
+        pytest.param(["feature0", "feature1"], ["pred0"], id="2_1"),
+        pytest.param(["feature0", "feature1"], ["pred0", "pred1"], id="2_2"),
+        pytest.param(["feature0"], ["pred0", "pred1"], id="1_2"),
+    ],
+)
+def test_constant_model_predict_columnwise(input_variables, output_variables, nz):
```
brianhenn, line 112: These three tests repeat a lot of code, could you make the predictor with set outputs as fixture?

brianhenn (approved):
Thanks @mcgibbon this is nice. Approving with a few minor fixes/simplifications requested.

external/fv3fit/fv3fit/testing.py
mcgibbon: Updating alongside using `X.coords` below for consistency.

external/fv3fit/tests/test_keras_testing.py
mcgibbon: Avoiding fixtures for this to avoid too much magic coupling, but I reduced duplicated code a little bit using a helper function.

external/fv3fit/fv3fit/testing.py
mcgibbon: Added more description.

external/fv3fit/tests/test_keras_testing.py
mcgibbon: Added a test that the sample dimension length is correct following our slack discussion.

PR #1242, opened 2021-06-03
prognostic run: save plots to tensorboard

nbren12:
Tensorboard is a great way to monitor a long-running computation like an FV3
run with online learning. This PR adds an simple switch to allow visualizing 2D
fields in tensorboard.

Added public API:
- new configuration option `user_config.diagnostics[].tensorboard`

- [x] Tests added


workflows/prognostic_c48_run/runtime/diagnostics/tensorboard.py
```diff
@@ -0,0 +1,41 @@
...
+
+def plot_to_image(figure):
+    """Converts the matplotlib plot specified by 'figure' to a PNG image and
+    returns it. The supplied figure is closed and inaccessible after this call."""
+    # Save the plot to a PNG in memory.
+    buf = io.BytesIO()
+    plt.savefig(buf, format="png")
+    # Closing the figure prevents it from being displayed directly inside
+    # the notebook.
+    plt.close(figure)
```
frodre, line 21: Why does the close appear here and not in the context where the figure is created?  Maybe adjusting `savefig` above means it doesn't have to be closed here?

frodre (approved):
Thanks, @nbren12 for the update and showing me what it looks like in action.  Minor comments relating to the figure saving/closing, but overall looks good!

workflows/prognostic_c48_run/sklearn_runfile.py
```diff
@@ -38,23 +38,26 @@
...
+                comm, diagnostics, ["specific_humidity_limiter_active"]
+            )
+            if comm.rank == 0:
+                runtime.log_mapping(time, averages, "statistics")
+                runtime.log_mapping(time, profiles, "profiles")
 
-        for diag_file in diag_files:
-            diag_file.observe(time, diagnostics)
+            for diag_file in diag_files:
+                diag_file.observe(time, diagnostics)
```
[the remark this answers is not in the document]
nbren12: That's correct. The variable and averaging/time selection is handled by `DiagnosticFile`, which could probably be renamed now that it works with different sinks. Keeping with the plumbing metaphor, it's more of a "faucet" now.

workflows/prognostic_c48_run/runtime/diagnostics/tensorboard.py
nbren12: I'll remove it.

PR #1244, opened 2021-06-03
Fix argo syntax in prognostic run workflow

### LINKED ISSUE #1243: end-to-end integration test is broken
I tried launching the e2e locally and get `Invalid 'when' expression '{{input.parameters.online-diags}} == true': Invalid token: '{{' (hint: try wrapping the affected expression in quotes ("))` after the nudge-to-fine run. Looking at the [logs](https://app.circleci.com/pipelines/github/VulcanClimateModeling/fv3net/8553/workflows/9d7c3d89-b61e-435b-8c3b-90757222b8b9/jobs/69046) for the e2e test after #790, it seems it was also exiting after this step (but for some reason not given a failing signal).

Probably should just be `{{inputs.parameter.online-diags}}` instead of `{{input.parameter.online-diags}}`? I hope we will still get a failure warning if there is a runtime error (versus argo syntax error).

cc @nbren12

oliverwm1:
Resolves #1243

### DIFF
--- a/workflows/argo/prognostic-run.yaml
+++ b/workflows/argo/prognostic-run.yaml
@@ -65,7 +65,7 @@ spec:
               - {name: cpu, value: "{{inputs.parameters.cpu}}"}
               - {name: memory, value: "{{inputs.parameters.memory}}"}
       - - name: online-diags
-          when: "{{input.parameters.online-diags}} == true"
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
LGTM. Thanks.

PR #1245, opened 2021-06-04
Generate report after every prognostic run

oliverwm1:
This PR saves manual effort by automatically generating a report for every prognostic run. For example, the e2e test generated one here: https://storage.googleapis.com/vcm-ml-public/argo/2021-06-04-8a2953298004/index.html. The location it is saved to is `gs://vcm-ml-public/argo/{date}-{uuid}`. The logs display the saved location.

Added public API:
- new `report-single-run` argo template

Significant internal changes:
- update e2e test configuration to work for report
- stop removing the first 24 timesteps of data for diurnal cycle computation


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -487,7 +487,7 @@ def time_mean_biases_dycore(prognostic, verification, grid):
     @add_to_diags("physics")
     @diag_finalizer(f"diurnal_{mask_type}")
     @transform.apply("mask_to_sfc_type", mask_type)
-    @transform.apply("resample_time", "1H", time_slice=slice(24, -1), inner_join=True)
+    @transform.apply("resample_time", "1H", inner_join=True)
```
oliverwm1, line 490: the report failed in the e2e test without this since the prognostic run goes for only 8 hours so there were no diurnal cycle diagnostics. We could also make the report generation more robust to missing diagnostics. But this time slicing is pretty arbitrary and I think a holdover from some specific plots we wanted last spring.

nbren12 (approved):
Thanks. This will save a lot of time. So your plan for now is to look up this information in the logs? These get purged pretty quickly, so I am not sure how well that will support the use case of getting a link to the report after running a experiment overnight. Any thoughts about that? I had thought an email or plugging into `artifacts` would make sense, both would require additional effort.

oliverwm1:
> These get purged pretty quickly
they are always still available through the google web console even after the pods are gone (I usually just search by the pod name), but I agree this is a bit annoying.

> I had thought an email or plugging into artifacts would make sense, both would require additional effort.

I looked into email a bit, would take me some work so figured best as a separate PR. Was going to work today on adding a report search-by-url functionality to artifacts.

PR #1250, opened 2021-06-09
Offline report: add time mean error maps

AnnaKwa:
This adds time mean error (prediction - target) map figures to the offline report. Previously, the time mean maps were limited to showing two-column ML outputs' target and prediction maps. This worked alright for estimating the spatial errors in tendency outputs which could be positive or negative, but now that we are predicting quantities that are strictly postive, e.g. downward fluxes, it is hard to tell where the predictions are doing a bad job by comparing maps of the full values rather than their differences.

I split the time mean maps into their own page, since the section is getting long. A future PR will reorganize the rest of the index page so all sections can go into their own page.

example: https://storage.googleapis.com/vcm-ml-public/annak/2021-06-08-error-maps/index.html


external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -186,7 +186,7 @@ def plot_cube(
             cb_ax = ax.inset_axes([1.05, 0, 0.02, 1])
         cbar = plt.colorbar(handles[0], cax=cb_ax, extend="both")
         cbar.set_label(
-            _get_var_label(plottable_variable[var_name].attrs, cbar_label or var_name)
+            cbar_label or _get_var_label(plottable_variable[var_name].attrs, var_name)
```
AnnaKwa, line 189: Fixes an old bug that prevented the colorbar label from being set manually.

workflows/offline_ml_diags/offline_ml_diags/create_report.py
```diff
@@ -122,6 +162,7 @@ def render_index(metrics, ds_diags, ds_diurnal, ds_transect, output_dir) -> str:
             section_name="Timesteps used for testing",
             output_dir=output_dir,
         )
+    column_integrated_metrics = column_integrated_metric_names(metrics)
```
brianhenn, line 165: clearer name?
```suggestion
    column_integrated_variable_names = column_integrated_metric_names(metrics)
```

workflows/offline_ml_diags/offline_ml_diags/create_report.py
```diff
@@ -293,6 +319,15 @@ def render_index(metrics, ds_diags, ds_diurnal, ds_transect, output_dir) -> str:
     with open(os.path.join(temp_output_dir.name, MODEL_SENSITIVITY_HTML), "w") as f:
         f.write(html_model_sensitivity)
 
+    html_time_mean_maps = render_time_mean_maps(
+        os.path.join(args.input_path, "time_mean_maps_figures"),
+        temp_output_dir.name,
+        ds_diags,
+        column_integrated_vars=column_integrated_metric_names(metrics),
```
brianhenn, line 326:
```suggestion
        column_integrated_variable_names
```

workflows/offline_ml_diags/offline_ml_diags/create_report.py
```diff
@@ -98,12 +100,50 @@ def render_model_sensitivity(figures_dir, output_dir) -> str:
...
 
 
+def render_time_mean_maps(
+    figures_dir, output_dir, ds_diags, column_integrated_vars
+) -> str:
+    report_sections: MutableMapping[str, Sequence[str]] = {}
+
+    # time averaged column integrated quantity maps
+    for var in column_integrated_vars:
+        ds_diags[f"error_in_{var}"] = (
```
brianhenn, line 110: To avoid having to manage a case where `derivation_plot_coords=None` and the ambiguity about faceting in the subsequent plotting function, you could instead assign a coord `derivation=['error']` to the `var` DataArray for this case. Up to you but might be simpler.

brianhenn (approved):
Thanks @AnnaKwa, a couple optional suggestions below

PR #1251, opened 2021-06-09
Add C48 May 2020 restarts zarr to catalog

brianhenn:
Adds a C48 restarts zarr of the May 2020 SHiELD run to the vcm catalog. Also adds a couple of non-breaking changes to the dataflow pipeline to make it possible to reproduce how the zarr was created.

Added public API:
- a `--no-coarse-suffix` arg to `fv3net.pipelines.restarts_to_zarr` to make it possible to point that pipeline at the output of the `coarsen_restarts` pipeline


oliverwm1 (approved):
LGTM. A couple things I noticed that are particular about the dataset are that 1. the time coordinate consists of string timestamps e.g. `'20160801.001500'` and 2. the tile dimension has a coordinate which is 1-based. This can cause issues if merging with datasets that use our more typical 0-based convention. Anyway, maybe worth mentioning in catalog metadata.

PR #1254, opened 2021-06-10
prognostic_run: add ability to monitor dynamics step

oliverwm1:
This will allow easily computing tendencies and storages due to the FV3 dynamics.


spencerkclark (approved):
Looks good to me.

PR #1257, opened 2021-06-11
Give create-run step in run-fv3gfs more memory

oliverwm1:
End-to-end integration test has been failing on the `create-run` step, often with OOM. Try increasing memory limit. Also removes two unused parameters to the step.

### DIFF
--- a/workflows/argo/run-fv3gfs.yaml
+++ b/workflows/argo/run-fv3gfs.yaml
@@ -21,8 +21,6 @@ spec:
           arguments:
             parameters:
             - {name: runURL, value: "{{inputs.parameters.output-url}}"}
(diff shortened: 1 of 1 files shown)


workflows/argo/run-fv3gfs.yaml
```diff
@@ -21,8 +21,6 @@ spec:
           arguments:
             parameters:
             - {name: runURL, value: "{{inputs.parameters.output-url}}"}
-            - {name: cpu, value: "{{inputs.parameters.cpu}}"}
-            - {name: memory, value: "{{inputs.parameters.memory}}"}
```
oliverwm1, line 25: the `create-run` step doesn't need to know about these parameters.

AnnaKwa (approved):
LGTM thanks!

PR #1258, opened 2021-06-14
Remove gcsfs subtree and update

nbren12:
The latest releases of gcsfs include the bug commits we contributed, so we no
longer need the subtree

Requirement changes:
- updated gcsfs. no new dependencies


workflows/dataflow/setup.py
```diff
@@ -3,13 +3,14 @@
 dependencies = [
     "dask==2.13.0",
     "f90nml==1.1.2",
+    "fsspec==0.7.1",
+    "gcsfs==0.6.1",
```
oliverwm1, line 7: are these the versions we want to pin here?

oliverwm1 (approved):
Approving, but don't we want to update the pinned versions of gcsfs/fsspec for the dataflow jobs?

workflows/dataflow/setup.py
nbren12: Good catch. I just reverted these to the old values. I will fix.

Issue #896, opened 2021-01-12
Runtime configuration of the prognostic run?


nbren12:
We currently have essentially two configuration interfaces for the prognostic run. The first is the script `prepare_config.py` and the second is the sections of the `fv3config.yml` file that are referred to at runtime in the scikit-learn runfile. This is confusing because defaults and behaviors are defined in both places. 

In particular maintaining two sets of "defaults" seems problematic. For example, this code defines some default behavior for the diagnostics for backwards compatibility reasons: https://github.com/VulcanClimateModeling/fv3net/blob/2c7d698ea2d3f061a8ea13c823104504300db03b/workflows/prognostic_c48_run/runtime/diagnostics/manager.py#L210 

OTOH, the diagnostics overlay defines a different set of defaults: https://github.com/VulcanClimateModeling/fv3net/blob/2c7d698ea2d3f061a8ea13c823104504300db03b/workflows/prognostic_c48_run/prepare_config.py#L106

Would it be possible to unify these different configuration interfaces? What small improvements are possible?

cc @oliverwm1 @brianhenn

nbren12:
somewhat addressed by the HighLevelConfig object in runtime/segmented_run/prepare_config.py

PR #1259, opened 2021-06-15
Delete default diagnostics insertion from prepare_config.py

oliverwm1:
The "default diagnostics" which are inserted by the `prepare_config.py` script are no longer needed now that we have well-defined examples in the vcm-workflow-control repo. Deleting these defaults simplifies code substantially, since we now just have a single default (no diagnostics) and this lets us delete a CLI option to the `prepare_config.py` script.

Refactored public API:
- now if the user-supplied minimal `fv3config.yaml` contains no diagnostics specification, the compiled `fv3config.yaml` output by `prepare_config.py` will also contain no diagnostics
- delete `output-frequency` option from `prepare_config.py`

- [x] Tests added

Related issue: #896


nbren12 (approved):
Excellent deletion! This is great.

PR #1262, opened 2021-06-15
Use standard diagnostics location for memoize cache

### LINKED ISSUE #1212: Adjust "memoized" diagnostics to match the new directory scheme
The computed diagnostics are stored in folders according to this scheme:

https://github.com/VulcanClimateModeling/fv3net/blob/c816c5f86e55e8654ad7d36769c6939bd7059193/workflows/prognostic_run_diags/scripts/memoized_compute_diagnostics.sh#L16-L16

Can this be made more compatible with the new directory scheme introduced in f3c770bb0361758abec30de9d5378c1415849e81?

oliverwm1:
This PR adjusts the script that memoizes the computation of prognostic run diagnostics to use the standard output location `gs://bucket/project/date/tag/fv3gfs_run_diagnostics` as the cache. This is useful, since we can easily use the `artifacts ls` tool to search for available prog run diagnostics.

Resolves #1212.


nbren12 (approved):
Looks great!

PR #1264, opened 2021-06-16
Unify fv3fit training tests using functional interface

mcgibbon:
Fv3fit currently has separate files to test sklearn and keras training. This can make it difficult to add new models, because the model-specific code in these tests means brand new tests must be made. This PR creates tests that work for all sklearn and keras models, and cover the test cases currently covered by the existing tests (reproducibility, dumping, loading) and more (ability to fit the identity function). Tests for new model types are added automatically, using the default hyperparameter values for the new model training type.

Currently fv3fit uses an Estimator class for training. This class is difficult to test because of the existence of mutable state. To avoid testing the dump functionality of Estimator and the behavior of the class when multiple method calls are involved, this PR refactors training to use a "training function", which returns a dumpable/loadable Predictor.

Removing Estimator also aligns better with the current I/O interface, which associates a single class to a string. It is already not possible to save a loadable Predictor to disk unless either it has a dump function (as we do with the sklearn and keras predictor/estimators), or you make use of internal implementation details of `fv3fit.dump` (as is done by EnsemblePredictor).

Refactored public API:
- removed DummyModel, use ConstantOutputPredictor instead
- removed Estimator
- renamed and refactored routines to use a training Callable instead of Estimator, `get_estimator_class` is now `get_training_function` and `register_estimator` is now `register_training_function`
- added training functions for existing Estimator classes
- added `dump` to `Predictor` API
- ConstantOutputPredictor now returns `0.` for outputs whose value has not been set, instead of raising a `KeyError`.

Significant internal changes:
- Refactored and renamed `_RegressorEnsemble` to `_RandomForestEnsemble`
- Replaced keras and sklearn training test files with unified training tests that parameterize better over model type, and use the new configuration classes

- [x] Tests added


external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -85,15 +87,18 @@ def load(cls, path: str) -> "SklearnWrapper":
...
     """
 
     def __init__(
-        self, base_regressor, regressors: Sequence[sklearn.base.BaseEstimator] = None,
+        self,
+        hyperparameters: RandomForestHyperparameters,
+        regressors: Sequence[sklearn.base.BaseEstimator] = None,
     ) -> None:
-        self.base_regressor = base_regressor
+        self.hyperparameters = hyperparameters
```
mcgibbon, line 101: I did this refactor when I suspected this was the cause of non-reproducible runs, I can revert it if desired. However it does move in the direction we want to go with these classes (merging them, making them random forest-specific).

external/fv3fit/fv3fit/train.py
```diff
@@ -64,10 +68,7 @@ def dump_dataclass(obj, yaml_filename):
         yaml.safe_dump(dataclasses.asdict(obj), f)
 
 
-if __name__ == "__main__":
-    logging.basicConfig(level=logging.INFO)
-    parser = get_parser()
-    args = parser.parse_args()
+def main(args):
```
mcgibbon, line 71: Can also revert this if you like, I'm planning to use a refactor like this to add an in-process integration test but the test wasn't added in this PR.

external/fv3fit/tests/_regtest_outputs/test_sklearn_wrapper.test_predict_columnwise_is_deterministic.out
```diff
@@ -1 +1 @@
-fe1edd06a4cdb3bbca2652fc980a4d65
+d8844c0d03138cb2434e3756e62907ca
```
mcgibbon, line 1: Test was refactored to use a RandomForestRegressor instead of a dummy class.

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -52,28 +55,21 @@ def test_flatten_same_order():
     np.testing.assert_allclose(a, b)
 
 
-@pytest.fixture
-def test_regressor_ensemble():
```
mcgibbon, line 56: This fixture was only used by one test and its name added no additional information, so I merged the fixture into the test.

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -52,28 +55,21 @@ def test_flatten_same_order():
...
+        regressor_ensemble.fit(X, y)
     assert regressor_ensemble.n_estimators == 3
     X = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
     y = np.dot(X, np.array([1, 2])) + 3
     regressor_ensemble.fit(X, y)
     # test that .fit appends a new regressor
     assert regressor_ensemble.n_estimators == 4
     # test that new regressors are actually fit and not empty base regressor
-    assert len(regressor_ensemble.regressors[-1].coef_) > 0
+    assert len(regressor_ensemble.regressors[-1].estimators_) > 0
```
mcgibbon, line 72: The test was refactored to use a random forest instead of a DummyRegressor as part of making the sklearn code random-forest-specific.

mcgibbon:
Integration tests passed: https://app.circleci.com/pipelines/github/VulcanClimateModeling/fv3net/8655/workflows/13030e81-f5c5-4cfb-a158-3ad009258b5d/jobs/70334

pip-requirements.txt
```diff
@@ -1,5 +1,4 @@
 apache-beam[gcp]
-zarr
```
mcgibbon, line 2: This is a duplicate entry in the pip-requirements file, the other entry has a minimum version. With this line here, you (or at least I) can't `pip install -r` this file.

external/fv3fit/docker-compose.yml
```diff
@@ -1,14 +0,0 @@
-# this is only for development purposes and integration with visual studio code
```
mcgibbon, line 1: This file is unused and refers to a deleted dockerfile.

external/fv3fit/fv3fit/train.py
```diff
@@ -112,6 +113,19 @@ def dump_dataclass(obj, yaml_filename):
             os.path.join(args.local_download_path, "validation")
         )
 
-    model = _get_model(train_config)
-    model.fit(train_batches)
+    model = fit_model(
```
nbren12, line 116: Would it be possible reduce the number of arguments on this `fit_model` function e.g. `fit_model(some_config, train_batches, val_batches)` and therefore make it easier to use. It seems like at minimum `model_type` and `hyperparamaters` could be grouped into a single object, and probably `input_variables` and `output_variables` too.

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -86,15 +100,20 @@ def load(cls, path: str) -> "SklearnWrapper":
...
+    """
+    Ensemble of RandomForestRegressor objects that are each trained on a separate
+    batch of data.
     """
 
     def __init__(
-        self, base_regressor, regressors: Sequence[sklearn.base.BaseEstimator] = None,
+        self,
+        base_regressor,
+        n_jobs,
```
nbren12, line 111: How does this new parameter interact with the serialization? Does a trained model have this value baked in, in such a way that it cannot be trained/applied with a different number of jobs than it was trained with?

nbren12 (commented):
I didn't review too closely but overall I think these changes seem to simplify the code and remove some layers in the model construction chain.

At some point I could see wanting to restore a stateful "Trainer" object that can e.g. train a model sequentially online or start training from an old training checkpoint.

external/fv3fit/fv3fit/train.py
mcgibbon: I'm leaning a bit towards deleting `fit_model` entirely and putting its two lines here, after thinking about this.

If we enforce a one-to-one relationship, we can derive `model_type` from `type(hyperparameters)`. In the present API, many `model_type`s are allowed to share the same `hyperparameters` class (e.g. when we had aliases for the "random forest" model), which maybe doesn't make sense.

I excluded input/output variables because of the training function API, where I wanted to enforce input and output variables as clear required arguments. A flaw of the last design was grouping infrastructure-enforced, top-level arguments in the same class as model-specific hyperparameters, which led to model-specific hyperparameters getting used in top-level code. I think it's worthwhile limiting attributes on `hyperparameters` to model-specific details to avoid this in the future.

external/fv3fit/fv3fit/sklearn/_random_forest.py
mcgibbon: The parameter is only used in the training. The value we actually pass to the RandomForestRegressor is None, which says "use the joblib.parallel context you're in, or one core if you're not in one". I assume that means if you predicted inside a joblib.parallel context it would use it, but I don't know what the API is.

The parameter gets saved and loaded during serialization, but only for the reason base_regressor is - because this is an Estimator and Predictor. Both base_regresor and n_jobs should not be used when this class is used only as a Predictor.

mcgibbon:
> At some point I could see wanting to restore a stateful "Trainer" object that can e.g. train a model sequentially online or start training from an old training checkpoint.

This interface could definitely make use of such a trainer internally, so I'm not too worried about forward compatibility. For now I'll go with YAGNI and say we should implement stateful trainers when we want to use them so we have all the contextual details. For example, in-model training would likely be a script completely separate from `fv3fit.train` that produces a `Predictor` object, and doesn't necessarily need to use training functions.

external/fv3fit/tests/training/test_train_keras.py
```diff
@@ -1,149 +0,0 @@
...
-    return fv3fit.keras.get_model(
-        train_config.model_type,
-        loaders.SAMPLE_DIM_NAME,
-        train_config.input_variables,
-        train_config.output_variables,
-        train_config,
-    )
-
-
-def test_reproducibility(
```
AnnaKwa, line 47: I know the keras reproducibility isn't on the same level as the sklearn tests added in this PR, where the same hash ought to be able to be reproduced, but I think we should keep a test that the keras model predictions stay within some tolerance as long as you are training with the same random seed.

AnnaKwa: Actually while looking into this just now I realize that the original test was too lenient because the testing data coming from the tests/training/conftest was using zero arrays, which were trivial to match predictions for with different seeds... the test works as intended when it uses np.random.choice.

AnnaKwa (changes requested):
Thanks! I think things are clearer now that there's just the Predictor and no Estimator. 

My request is to keep a test of the keras model 'reproducibility' for the same seed.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -192,7 +207,7 @@ def fit(
 
         fit_kwargs = copy.copy(self._fit_kwargs)
         fit_kwargs = _fill_default(fit_kwargs, batch_size, "batch_size", None)
-        fit_kwargs = _fill_default(fit_kwargs, epochs, "epochs", 1)
+        fit_kwargs = _fill_default(fit_kwargs, epochs, "epochs", self._epochs)
```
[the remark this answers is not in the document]
mcgibbon: I think it defaulted to 1, but we always set it anyways.

external/fv3fit/tests/training/test_train.py
```diff
@@ -0,0 +1,157 @@
...
+        output_variables,
+        hyperparameters,
+        train_batches,
+        val_batches,
+    )
+    data_array = sample_func()
+    test_dataset = xr.Dataset(data_vars={"var_in": data_array, "var_out": data_array})
+    hash_before_predict = vcm.testing.checksum_dataarray_mapping(test_dataset)
+    out_dataset = model.predict(test_dataset)
+    # TODO: move this to a separate test
```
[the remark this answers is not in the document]
mcgibbon: Done. The not great part about these tests is they require calling the training to get a predictor to test, so they take a while. Adding more tests makes it worse.

external/fv3fit/fv3fit/train.py
mcgibbon: OK, `fit_model` is inlined now which removes the `model_type` argument, and makes it clear that the training function and hyperparameters should both be derived from `model_type`.

external/fv3fit/fv3fit/keras/_training.py
```diff
@@ -24,7 +23,7 @@ def get_model(
     input_variables: Iterable[str],
     output_variables: Iterable[str],
     legacy_config: _ModelTrainingConfig,
-) -> Estimator:
+):
     """Initialize and return a Estimator instance.
```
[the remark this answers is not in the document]
mcgibbon: I deleted this file instead.

external/fv3fit/fv3fit/keras/_training.py
```diff
@@ -43,7 +42,7 @@ def get_model(
 
 
 # TODO: merge this helper function with get_keras_model
-def get_model_class(model_type: str) -> Type[Estimator]:
+def get_model_class(model_type: str):
     """Returns a class implementing the Estimator interface corresponding to the model type.
```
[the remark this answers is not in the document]
mcgibbon: File deleted.

external/fv3fit/tests/training/test_train_keras.py
mcgibbon: Added a test for this.

external/fv3fit/fv3fit/train.py
nbren12: Inlining seems like a good solution.

> In the present API, many model_types are allowed to share the same hyperparameters class (e.g. when we had aliases for the "random forest" model), 

I hadn't realized this. This makes sense to me.

nbren12:
>  For now I'll go with YAGNI
For sure. It's just food for thought about making an extensible design.

AnnaKwa (approved):
LGTM!

PR #1269, opened 2021-06-17
Catalog: update 40day_c48_pressure_interpolated_may2020 for wind bug fix

AnnaKwa:
Previously, the 2D c3072 verification dataset was corrected for the wind bug, but I forgot to correct the 3D pressure interpolated dataset (forgot the original winds were not A grid).


oliverwm1 (approved):
Thanks!

brianhenn:
Just a heads up that this meant the wind biases in the prognostic run reports were incorrect (e.g., did not match pattern of wind nudging tendencies for the nudged training runs).

PR #1271, opened 2021-06-18
Add precipitation histogram to prognostic run report

oliverwm1:
Useful to show a histogram of precipitation on prognostic run reports. Example report [HERE](https://storage.googleapis.com/vcm-ml-public/oliwm/test/new-report/index.html).

Significant internal changes:
- Added function to compute histograms of arbitrary variables. For now, just doing surface precipitation rate.
- Create new `process_diagnostics.html` page to the prog run report. Moved diurnal cycle there, and added plot of precipitation histogram.
- Added metric to compute percentiles from above histograms. For now, computing 25th, 50th, 75th, 90, 99, and 99.9th percentiles. e.g.:

```
"percentile_25/total_precip_to_surface": {
      "value": 0.3747434426709828,
      "units": "mm/day"
}
```

Sped up report test by only creating report for a single run (since no longer require >1 run for report generation).

- [x] Tests added

### DIFF
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
@@ -502,11 +502,13 @@ def compute_histogram(prognostic, verification, grid):
         count, bins = np.histogram(
             prognostic[varname], bins=HISTOGRAM_BINS[varname], density=True
         )
-        bin_midpoints = 0.5 * (bins[:-1] + bins[1:])
-        coords = {f"{varname}_bins": bin_midpoints}
+        coords = {f"{varname}_bins": bins[:-1]}
+        width = bins[1:] - bins[:-1]
+        width_da = xr.DataArray(width, coords=coords, dims=list(coords.keys()))
         count_da = xr.DataArray(count, coords=coords, dims=list(coords.keys()))
         count_da[f"{varname}_bins"].attrs["units"] = prognostic[varname].units
         counts[varname] = count_da
+        counts[f"{varname}_bin_width"] = width_da
     return _assign_diagnostic_time_attrs(counts, prognostic)
 
 
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
@@ -23,7 +23,7 @@ import fsspec
 
 from toolz import curry
 from collections import defaultdict
-from typing import Dict, Callable, Mapping, Tuple, Union
+from typing import Dict, Callable, Mapping, Union
 
 from joblib import Parallel, delayed
 
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
@@ -37,6 +37,7 @@ from fv3net.diagnostics.prognostic_run import config
 from fv3net.diagnostics.prognostic_run import diurnal_cycle
 from fv3net.diagnostics.prognostic_run import transform
 from fv3net.diagnostics.prognostic_run.constants import (
+    HISTOGRAM_BINS,
     HORIZONTAL_DIMS,
     DiagArg,
     GLOBAL_AVERAGE_DYCORE_VARS,
@@ -501,6 +502,23 @@ for mask_type in ["global", "land", "sea"]:
         else:
             diag = diurnal_cycle.calc_diagnostics(prognostic, verification, grid).load()
             return _assign_diagnostic_time_attrs(diag, prognostic)
+
+
+@add_to_diags("physics")
+@diag_finalizer("histogram")
+@transform.apply("resample_time", "3H", inner_join=True)
+@transform.apply("subset_variables", HISTOGRAM_BINS.keys())
+def compute_precip_histogram(prognostic, verification, grid):
+    logger.info("Computing precipitation histogram")
+    count_ds = xr.Dataset()
+    for varname in prognostic:
+        count, bins = np.histogram(
+            prognostic[varname], bins=HISTOGRAM_BINS[varname], density=True
+        )
+        count_da = xr.DataArray(count, coords={f"{varname}_bin": bins[:-1]})
+        count_da[f"{varname}_bin"].attrs["units"] = prognostic[varname].units
+        count_ds = xr.merge([count_ds, count_da.rename(varname)])
+    return _assign_diagnostic_time_attrs(count_ds, prognostic)
 
 
 def register_parser(subparsers):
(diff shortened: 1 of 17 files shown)


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -229,16 +227,6 @@ def _assign_diagnostic_time_attrs(
     return diagnostics_ds
 
 
-def dump_nc(ds: xr.Dataset, f):
```
oliverwm1, line 232: We switched to using the `vcm.dump_nc` version of this a while ago, so this func is unused.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -503,6 +491,25 @@ def _diurnal_func(
...
 
+@add_to_diags("physics")
+@diag_finalizer("histogram")
+@transform.apply("resample_time", "3H", inner_join=True)
+@transform.apply("subset_variables", list(HISTOGRAM_BINS.keys()))
+def compute_histogram(prognostic, verification, grid):
+    logger.info("Computing histograms for physics diagnostics")
+    counts = xr.Dataset()
+    for varname in prognostic.data_vars:
+        count, bins = np.histogram(
```
oliverwm1, line 502: I found using `np.histogram` much faster (and simpler code-wise) than doing an `xarray` groupby.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -75,9 +80,7 @@ def make_plots(self, data) -> Iterable:
             yield func(data)
 
 
-def plot_1d(
-    run_diags: RunDiagnostics, varfilter: str, run_attr_name: str = "run",
-) -> HVPlot:
+def plot_1d(run_diags: RunDiagnostics, varfilter: str) -> HVPlot:
```
oliverwm1, line 83: deleted unused argument `run_attr_name`

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -503,6 +491,27 @@ def _diurnal_func(
             return _assign_diagnostic_time_attrs(diag, prognostic)
 
 
+@add_to_diags("physics")
+@diag_finalizer("histogram")
+@transform.apply("resample_time", "3H", inner_join=True)
+@transform.apply("subset_variables", list(HISTOGRAM_BINS.keys()))
+def compute_histogram(prognostic, verification, grid):
```
nbren12, line 498: Can you add a test for this function and consider refactoring it to vcm? histograms are pretty general purpose functionality. BTW there is https://github.com/xgcm/xhistogram.git, but I don't see the need to add the dependency unless we really need to daskify this computation.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -138,6 +138,25 @@ def rmse_time_mean(diags):
...
+        percentiles = xr.Dataset()
+        data_vars = [v for v in histogram.data_vars if not v.endswith("bin_width")]
+        for varname in data_vars:
+            bins = histogram[f"{varname}_bins"].values
+            bin_width = histogram[f"{varname}_bin_width"]
+            bin_midpoints = bins + 0.5 * bin_width
+            cumulative_distribution = np.cumsum(histogram[varname][:-1] * bin_width)
+            percentiles[varname] = bin_midpoints[
+                np.argmin(np.abs(cumulative_distribution.values - p / 100))
+            ]
```
nbren12, line 155: This bit seems like it should be refactored to a helper and tested.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -138,6 +138,25 @@ def rmse_time_mean(diags):
...
+        restore_units(histogram, percentiles)
```
nbren12, line 156: "restore"? Sounds like we use a lot of verbs for this operation of grabbing metadata from one xarray object and applying it to another. I think "assign" is more widely used. E.g.
```
assign_units(from_=histogram, to=percentiles)
```

nbren12 (approved):
Looks great. Apart from some minor comments below, I had some thoughts
1. I was pretty confused about why the metric functions returned `xr.Dataset` rather than floats, before realizing that the `to_dict` function looks at the units and values. It would be clearer if the metrics functions returned dataclasses with `.value` and `.units` attrs, so that we are passing the minimal amount of data around. 
2. It would be great to enhance the visibility of the scalar metrics, so that they are amongst the first things we see. e.g. by showing a table instead of the histograms. The histograms are kind of useless and ugly IMO.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
oliverwm1: yes, I saw that package and think @spencerkclark has used it to good effect. but seems like it's under active/early development, so would rather avoid the dep.

I can refactor to VCM—but need to think a bit about the right format for the output. Specifically, it's helpful to know the bin widths. Right now I'm using a separate DataArray for them (so basically the output is a dataset w/ two arrays) but it might be better to encode it in the coordinates somehow.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
oliverwm1: okay.. but it is an existing function so I think I'll just leave it.

oliverwm1:
I was pretty confused about why the metric functions returned xr.Dataset rather than floats, before realizing that the to_dict function looks at the units and values. It would be clearer if the metrics functions returned dataclasses with .value and .units attrs, so that we are passing the minimal amount of data around.

> yeah I agree the `xr.Dataset` is a bit overkill. We could improve this in another PR.

It would be great to enhance the visibility of the scalar metrics, so that they are amongst the first things we see. e.g. by showing a table instead of the histograms. The histograms are kind of useless and ugly IMO, especially for a single run report.

> ouch, poor histogram ;) It would be helpful to have the verification data on there, I agree. The example is particularly noisy since it's a short run. But yeah, a table showing the percentiles for each run would be quite useful.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
spencerkclark: There is effort ongoing to refine and move the functionality of xhistogram to xarray (https://github.com/pydata/xarray/issues/4610, https://github.com/pydata/xarray/pull/5400), so I agree it probably would make sense to wait until that occurs.

nbren12:
> ouch, poor histogram ;)

sorry. I didn't mean your histograms, i meant the bar charts with the metrics on the main page. The histograms are great!

oliverwm1:
> > ouch, poor histogram ;)
> 
> sorry. I didn't mean your histograms, i meant the bar charts with the metrics on the main page. The histograms are great!

haha sounds good. I agree the bar charts are hard to parse. I'll work on adding a metrics table with the things we actually care about.

PR #1272, opened 2021-06-18
Don't use quotes around $flags in memoized_compute_diagnostics.sh

oliverwm1:
Fixes bug introduced in #1262 when satisfying shellcheck (if `$flags` is surrounded by quotes the script fails if `$flags` is an empty variable, e.g. when the default verification data is used).

### DIFF
--- a/workflows/prognostic_run_diags/scripts/memoized_compute_diagnostics.sh
+++ b/workflows/prognostic_run_diags/scripts/memoized_compute_diagnostics.sh
@@ -31,7 +31,7 @@ if [[ $diagsExitCode -eq 0 && $metricsExitCode -eq 0 ]]; then
     echo "Prognostic run diagnostics detected in cache for given run. Using cached diagnostics."
 else
     echo "No prognostic run diagnostics detected in cache for given run. Computing diagnostics and adding to cache."	
-    prognostic_run_diags save $flags $run diags.nc
+    prognostic_run_diags save $flags "$run" diags.nc
     prognostic_run_diags metrics diags.nc > metrics.json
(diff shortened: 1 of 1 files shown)


workflows/prognostic_run_diags/scripts/memoized_compute_diagnostics.sh
```diff
@@ -31,7 +31,7 @@ if [[ $diagsExitCode -eq 0 && $metricsExitCode -eq 0 ]]; then
     echo "Prognostic run diagnostics detected in cache for given run. Using cached diagnostics."
 else
     echo "No prognostic run diagnostics detected in cache for given run. Computing diagnostics and adding to cache."	
-    prognostic_run_diags save "$flags" "$run" diags.nc
+    prognostic_run_diags save $flags $run diags.nc
```
nbren12, line 34: We can at least leave the quotes around run in case that has a space in it. I agree that `$flags` should not be quoted. I don't know if there is a way to tell shellcheck that. Maybe there is no way in bash to safely handle variadic behavior?

```suggestion
    prognostic_run_diags save $flags "$run" diags.nc
```

nbren12 (approved):
I had one small suggestion so I disabled auto merge.

workflows/prognostic_run_diags/scripts/memoized_compute_diagnostics.sh
oliverwm1: Sounds good. I'm not sure, but I think we'll be able to delete this script soon anyway.

PR #1274, opened 2021-06-21
Don't call gsutil cp if output is equal to cacheURL

oliverwm1:
`gsutil cp SOURCE TARGET` fails if TARGET==SOURCE, so adjust `memoized_compute_diagnostics.sh` script to handle this case.


AnnaKwa (approved):
LGTM

PR #1276, opened 2021-06-23
Prognostic report: add 3-7 RMSE avg and land/sea domain RMSE and bias

AnnaKwa:
This adds the following the the metrics json file:
- 3-7 day RMSE average (previously only saved day 3 RMSE)
- RMSE and bias for land/sea/global domains

### DIFF
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
@@ -17,6 +17,7 @@ from .constants import HORIZONTAL_DIMS
 import json
 
 _METRICS = []
+
 GRID_VARS = ["lon", "lat", "lonb", "latb", "area"]
 
 
@@ -88,6 +89,25 @@ def rmse_3day(diags):
 
     restore_units(rms_global, rms_at_day_3)
     return rms_at_day_3
+
+
+@add_to_metrics("rmse_days_3to7_avg")
+def rmse_days_3to7_avg(diags):
+    rms_global = grab_diag(diags, "rms_global").drop(GRID_VARS, errors="ignore")
+    time_since_start = rms_global.time.values - rms_global.time.isel(time=0).item()
+    ds = rms_global.assign_coords(
+        {"days_since_start": ("time", [t.days for t in time_since_start])}
+    )
+
+    if max(ds["days_since_start"] > 7):
+        rmse_days_3to7_avg = (
+            ds.where(ds["days_since_start"] >= 3).where(ds["days_since_start"] <= 7)
+        ).mean(skipna=True)
+    else:  # don't compute metric if run didn't make it to 7 days
+        rmse_days_3to7_avg = xr.Dataset()
+
+    restore_units(rms_global, rmse_days_3to7_avg)
+    return rmse_days_3to7_avg
 
 
 @add_to_metrics("drift_3day")
--- a/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
+++ b/workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
@@ -19,6 +19,7 @@ import json
 _METRICS = []
 
 GRID_VARS = ["lon", "lat", "lonb", "latb", "area"]
+ns_per_day = 1e9 * 60 * 60 * 24
 
 
 def grab_diag(ds, name):
@@ -95,8 +96,14 @@ def rmse_3day(diags):
 def rmse_days_3to7_avg(diags):
     rms_global = grab_diag(diags, "rms_global").drop(GRID_VARS, errors="ignore")
     time_since_start = rms_global.time.values - rms_global.time.isel(time=0).item()
+
     ds = rms_global.assign_coords(
-        {"days_since_start": ("time", [t.days for t in time_since_start])}
+        {
+            "days_since_start": (
+                "time",
+                [t.item() / ns_per_day for t in time_since_start],
+            )
+        }
     )
 
     if max(ds["days_since_start"] > 7):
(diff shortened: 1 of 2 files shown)


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -90,6 +110,31 @@ def rmse_3day(diags):
...
+            )
+        }
+    )
+
+    if max(ds["days_since_start"] > 8):
+        rmse_days_3to7_avg = (
+            ds.where(ds["days_since_start"] >= 3).where(ds["days_since_start"] <= 8)
+        ).mean(skipna=True)
+    else:  # don't compute metric if run didn't make it to 7 days
+        rmse_days_3to7_avg = xr.Dataset()
```
oliverwm1, line 132:
```suggestion
    rms_global_daily = rms_global.resample(time="1D").mean()

    if rms_global_daily.sizes["time"] >= 7:
        rmse_days_3to7_avg = rms_global_daily.isel(time=slice(2, 7)).mean("time")
    else:  # don't compute metric if run didn't make it to 7 days
        rmse_days_3to7_avg = xr.Dataset()
```
following the pattern in the `rmse_day3` function makes this simpler, although I guess if the start time is not at 0 UTC it would not be quite correct. But I think it's best to be consistent with the other function.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -118,24 +163,30 @@ def time_and_global_mean_value(diags):
...
 
 
-@add_to_metrics("time_and_global_mean_bias")
-def time_and_global_mean_bias(diags):
-    time_mean_bias = grab_diag(diags, "time_mean_bias")
-    area = diags["area"]
-    time_and_global_mean_bias = weighted_mean(time_mean_bias, area, HORIZONTAL_DIMS)
-    restore_units(time_mean_bias, time_and_global_mean_bias)
-    return time_and_global_mean_bias
+for mask_type in ["global", "land", "sea"]:
```
oliverwm1, line 166: Could you add the `land` and `sea` metrics for the `time_and_global_mean_value` above? Nice to have the absolute value in addition to the bias.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -118,24 +163,30 @@ def time_and_global_mean_value(diags):
...
-    area = diags["area"]
-    rms_of_time_mean_bias = np.sqrt(
-        weighted_mean(time_mean_bias ** 2, area, HORIZONTAL_DIMS)
-    )
-    restore_units(time_mean_bias, rms_of_time_mean_bias)
-    return rms_of_time_mean_bias
+
+for mask_type in ["global", "land", "sea"]:
+
+    @add_to_metrics(f"rmse_of_time_mean_{mask_type}")
```
oliverwm1, line 181: This is a breaking change for the name of the "global" `rmse_of_time_mean` metric (which is used in the report maps' titles). Could we leave the name of the global one unchanged? E.g. don't add the mask type suffix for that case. Or alternatively, we could include the metric under both names and deprecate the one without the suffix for future removal.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -46,7 +50,23 @@ def prepend_to_key(d, prefix):
 
 
 def weighted_mean(ds, w, dims):
-    return (ds * w).sum(dims) / w.sum(dims)
+    return (ds * w).sum(dims, skipna=True) / w.sum(dims, skipna=True)
```
oliverwm1, line 53: do we need this change? the masking sets the area (ie weights) to 0 outside the region of interest, so don't think we need to worry about nans.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -46,7 +50,23 @@ def prepend_to_key(d, prefix):
...
+
+
+def _mask_array(
```
oliverwm1, line 56: I understand not wanting to couple too much between this and the `transform.py` module, but might be nice to not have this code duplicated. I'll leave it up to you. Could move the `mask_array` function to a new module and import it in both places.

oliverwm1 (changes requested):
Thanks for adding these metrics! Main requested change is to make sure we don't have a breaking change in the name of the global time-mean RMSE metric. Have some other suggestions too. Do you think we should add a day-5 RMSE? That was one of the metrics discussed for the n2f ablation study.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
oliverwm1: Would be good to double-check my slice integers here.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
AnnaKwa: I'll keep this copy here, since I think the diags dataset that is used in `metrics.py` doesn't have a latitude data array. I guess it would still work if I moved the function to a shared location and passed some dummy arg for `latitude` when it's called here, but that could be confusing in the future if someone thinks they can use the `tropics` mask key for a metric.

oliverwm1: The dataset should have latitude (I think under `lat`) since I know it's used for making maps in the report.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
AnnaKwa: I'll go with the latter option and add a comment about deprecating it.

AnnaKwa:
Thanks @oliverwm1 , this is ready for re-review.

oliverwm1 (approved):
LGTM, thanks for the changes!

PR #1277, opened 2021-06-23
Add mypy checking to loaders

mcgibbon:
This PR fixes mypy errors in the `loaders` package and enables mypy tests for that package.

### DIFF
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -1,7 +1,16 @@
 import logging
 from numpy.random import RandomState
 import pandas as pd
-from typing import Iterable, Sequence, Mapping, Any, Optional, Union, List
+from typing import (
+    Iterable,
+    Sequence,
+    Mapping,
+    Any,
+    Optional,
+    Union,
+    List,
+    no_type_check,
+)
 import xarray as xr
 from vcm import safe, parse_datetime_from_str
 from toolz import partition_all, curry, compose_left
@@ -30,6 +39,8 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
+# TODO: remove this decorator or delete this function
+@no_type_check
 def batches_from_geodata(
     data_path: Union[str, List, tuple],
     variable_names: Iterable[str],
@@ -91,7 +102,7 @@ def _create_mapper(
 
 def batches_from_mapper(
     data_mapping: Mapping[str, xr.Dataset],
-    variable_names: Iterable[str],
+    variable_names: Sequence[str],
     timesteps_per_batch: int = 1,
     random_seed: int = 0,
     timesteps: Optional[Sequence[str]] = None,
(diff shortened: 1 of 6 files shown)


external/loaders/loaders/batches/_batch.py
```diff
@@ -30,6 +39,8 @@
 logger.setLevel(logging.INFO)
 
 
+# TODO: remove this decorator or delete this function
```
mcgibbon, line 42: Ignored type checking on the two functions in this file because the PR I will be merging immediately after this may delete these functions.

mcgibbon:
Integration tests passed.

external/loaders/loaders/batches/_batch.py
```diff
@@ -104,11 +115,11 @@ def batches_from_mapper(
     iterated over in  ..sklearn.train.
 
     Args:
-        data_mapping (Mapping[str, xr.Dataset]): Interface to select data for
+        data_mapping: Interface to select data for
             given timestep keys.
-        variable_names (Iterable[str]): data variables to select
+        variable_names: data variables to select
         timesteps_per_batch (int, optional): Defaults to 1.
```
oliverwm1, line 121:
```suggestion
        timesteps_per_batch: Defaults to 1.
```

external/loaders/loaders/mappers/_fine_resolution_budget.py
```diff
@@ -155,14 +167,18 @@ def __init__(self, url):
...
+        if match is None:
+            raise ValueError(
+                f"invalid url, must contain pattern {pattern} but received {url}"
+            )
         date = vcm.parse_timestep_str_from_path(url)
         tile = match.group(1)
         return date, int(tile)
 
-    def __getitem__(self, key: str) -> xr.Dataset:
+    def __getitem__(self, key):
```
oliverwm1, line 181: why do we need to remove the type hints here?

external/loaders/loaders/mappers/_fine_resolution_budget.py
```diff
@@ -194,13 +216,11 @@ class FineResolutionSources(GeoMapper):
     def __init__(
         self,
         fine_resolution_time_mapping: Mapping[Time, xr.Dataset],
-        offset_seconds: Union[int, float] = 0,
```
oliverwm1, line 197: I'm guessing you removed this arg because it's not used? You sure the change has no effects?

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -38,6 +38,7 @@ def __getitem__(self, key: str) -> xr.Dataset:
 T_Mapper = FineResolutionResidual
 
 
+@no_type_check
```
oliverwm1, line 41: Any particular reason to skip this one? I don't think it's pending deletion.

oliverwm1 (commented):
Generally looks good. The API change to FineResolutionSources caught my eye as a non-type-check-related change, but probably fine. Why skip the hybrid mapper?

external/loaders/loaders/mappers/_fine_resolution_budget.py
mcgibbon: These are already type hinted (according to mypy) due to the inheritance from `Mapping`. When I added manual type hints to some routines (but probably not this one), I got issues with things like mypy saying AbstractSet is incompatible with Iterable, so I just removed the method-level type hints.

external/loaders/loaders/mappers/_fine_resolution_budget.py
mcgibbon: In this case, if you look in the code below, you can see `self._offset_seconds` is set and then never used in the code. If someone is using it, they have a bug.

external/loaders/loaders/mappers/_hybrid.py
mcgibbon: It's substantially refactored in my next PR, where I've removed all the remaining no_type_checks added in this one.

oliverwm1 (approved):
LGTM

PR #1278, opened 2021-06-23
Refactor data configuration as needed for use in offline_diags

mcgibbon:
Configuration for data loading is currently very loosely typed, and has some inconsistencies. In particular, the offline diagnostics currently assume that the batches function used has keyword arguments `mapper_function` and `mapper_kwargs`, despite these not being part of the "batch function" API.

This PR creates formal types which will allow refactoring the offline diagnostics to make use of them. The central feature of this PR is the classes `BatchesConfig` and `BatchesFromMapperConfig`. Both of these provide `load_batches` for use in `fv3fit.train`. I will be refactoring offline diags to mainly depend on `load_batches`, but to provide snapshot data if `load_mapper` is available on the BatchesLoader instance used in the configuration. `loaders` is significantly refactored so that a `MapperConfig` class used by `BatchesFromMapperConfig` can be used to formalize the use of a `data_path`.

Added public API:
- Added BatchesLoader abstract base class used by fv3fit, which provides a way to retrieve a sequence of batches
- Added BatchesFromMapperConfig, which formalizes the pattern existing in many batches functions where a mapper is produced and then converted to batches
- Added BatchesConfig which reflects the existing batch function API with no reference to or requirement of a mapper function
- Added `loaders.typing` with Mapper, Batches, and functions which produce them

Refactored public API:
- DataConfig is removed from`fv3fit`, and is replaced by several classes in `loaders`
- consolidated mapper functions to all take in a `data_path` argument, so they can be configured and called consistently by new MapperConfig class
- Removed arguments `rename_vars`, `dim_order`, and `drop_vars` from `open_fine_res_apparent_sources` mapper function
- Refactored call signature of `open_fine_resolution_nudging_hybrid` to avoid Mapping in favor of stronger typing

Significant internal changes:
- Removed `variables` from data configuration classes, instead passing it when batches are retrieved as is required for use of these classes in the offline diagnostics workflow. Added `additional_variables` attribute to `TrainingConfig` to enable this.
- Deleted load_data_sequence routine in fv3fit, instead using `config.load_batches(variables)`
- Replaced many instances of "batch_function" or "batch_kwargs" with "batches_function" and "batches_kwargs", so that "mapper_function" returning "mapper" reflects "batches_function" returning "batches" (not just a single batch).

- [x] Tests added


external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -356,7 +341,7 @@ def load_configs(
     output_data_path: str,
     timesteps_file=None,
     validation_timesteps_file=None,
-) -> Tuple[TrainingConfig, DataConfig, Optional[DataConfig]]:
+) -> Tuple[TrainingConfig, loaders.BatchesConfig, Optional[loaders.BatchesConfig]]:
```
mcgibbon, line 344: Note this is using the more restrictive type because it's analogous to what the old training config does, this function will be deleted completely when we can remove the old yaml format.

external/loaders/loaders/mappers/_fine_resolution_budget.py
```diff
@@ -427,17 +427,17 @@ def open_fine_res_apparent_sources(
...
-        rename_vars = {
-            "grid_xt": "x",
-            "grid_yt": "y",
-            "pfull": "z",
-            "delp": "pressure_thickness_of_atmospheric_layer",
-        }
-
-    fine_resolution_sources_mapper: GeoMapper = FineResolutionSources(
-        open_fine_resolution_budget(fine_res_url),
+    rename_vars: Mapping[Hashable, Hashable] = {
```
mcgibbon, line 430: I'm under the impression these are never used apart from their defaults. Please let me know if that's not the case.

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -171,22 +175,31 @@ def open_nudge_to_fine(
     return XarrayMapper(ds.rename(rename_vars))
 
 
+@register_mapper_function
 def open_nudge_to_fine_multiple_datasets(
-    urls: Sequence[str], names: Optional[Sequence[Hashable]] = None, **kwargs
-):
+    data_path: str,
```
mcgibbon, line 180: Please let me know if this signature will cause issues with argo integration.

external/loaders/tests/test__config.py
```diff
@@ -0,0 +1,255 @@
...
+
+def test_load_mapper():
+    with mapper_context():
+        mock_mapper = {"key": xr.Dataset()}
+        mock_mapper_function = unittest.mock.MagicMock(return_value=mock_mapper)
+        mock_mapper_function.__name__ = "mock_mapper_function"
+        loaders._config.register_mapper_function(mock_mapper_function)
+        assert "mock_mapper_function" in loaders._config.MAPPER_FUNCTIONS
+        data_path = "test/data/path"
+        mapper_kwargs = {"arg1": "value1", "arg2": 2}
```
nbren12, line 52: This code block is shared by some tests below. Can you refactor to a helper?

external/loaders/tests/test__protocol.py
```diff
@@ -0,0 +1,25 @@
+from typing_extensions import runtime_checkable, Protocol
+
+
+def test_extended_kwargs_signature_is_still_instance():
```
nbren12, line 4: did you end up using the `Protocol` or the `Callable` for the Mappingfunction? If the latter, then these tests can be removed.

nbren12 (approved):
Looks good. I reviewed this commit-by-commit before realizing some of these commits were on master already. I'll dismiss my out of date comments.

external/loaders/tests/test__config.py
```diff
@@ -0,0 +1,176 @@
...
+import contextlib
+import xarray as xr
+import tempfile
+import os
+import yaml
+import dataclasses
+
+
+@contextlib.contextmanager
+def mapper_context():
```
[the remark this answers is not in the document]
mcgibbon: pytest-xdist actually doesn't need thread safety, it's pytest-parallel that does.

Regardless I'll check out the pattern and see about refactoring the code here to use it. Removing globals is usually a good thing.

external/loaders/loaders/_config.py
```diff
@@ -0,0 +1,117 @@
...
+    BATCHES_FUNCTIONS[func.__name__] = func
+    return func
+
+
+@dataclasses.dataclass
+class MapperConfig:
+
+    data_path: str
+    mapper_function: str
+    mapper_kwargs: dict
```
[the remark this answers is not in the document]
mcgibbon: Yes, this would enable us to use stricter type hinting on the Functions as well. I'm going to leave this for future work but +1.

external/loaders/loaders/_config.py
```diff
@@ -0,0 +1,117 @@
+import abc
+from typing import Dict
+from loaders.typing import Mapper, MapperFunction, Batches, BatchesFunction
+import dataclasses
+import dacite
+
+
+MAPPER_FUNCTIONS: Dict[str, MapperFunction] = {}
```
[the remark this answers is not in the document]
mcgibbon: I'll check it out.

external/loaders/loaders/_config.py
```diff
@@ -0,0 +1,117 @@
...
+            Sequence of datasets according to configuration
+        """
+        mapper = self.mapper_config.load_mapper()
+        batches_function = BATCHES_FUNCTIONS[self.batches_function]
+        ds_batches = batches_function(mapper, list(variables), **self.batches_kwargs,)
+        return ds_batches
+
+
+@dataclasses.dataclass
+class BatchesConfig(BatchesLoader):
```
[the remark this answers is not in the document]
mcgibbon: I'll see how hard it is to get sphinx to automatically detect and display the available options.

external/loaders/loaders/typing.py
```diff
@@ -1,17 +1,26 @@
...
 import xarray as xr
 
 
 Mapper = Mapping[str, xr.Dataset]
 Batches = Sequence[xr.Dataset]
 
+# to strictly type the call signature, we would need to refactor these types
+# to have consistent signatures, e.g. by combining the flexible arguments
+# in a dataclass similarly to what is done in fv3fit for models
+MapperFunction = Callable[..., Mapper]
```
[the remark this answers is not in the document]
mcgibbon: Thanks for the reminder.

external/loaders/loaders/mappers/_fine_resolution_budget.py
frodre: Yeah, it's a good bet that this is tied pretty concretely to the underlying data, which I assume is mostly static for now.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -1,103 +1,88 @@
...
-        self.fine_res = fine_res
-
-    def keys(self):
-        return list(set(self.physics_mapper.keys()) & set(self.fine_res.keys()))
-
-
-class FineResolutionResidual(ResidualMapper):
-    """Define the fine resolution dQ as a residual from the physics tendencies in
-    another mapper
+class PhysicsResidual(Mapper):
```
frodre, line 10: Nice simplification work here!

external/loaders/loaders/mappers/_nudged/_nudged.py
frodre: I tried to look around for experiments that might use this mapper function, but I don't know where all those experiment configs reside now, if anywhere.

frodre: Maybe get @brianhenn, @AnnaKwa, or @spencerkclark to chime in on this.

external/loaders/setup.py
```diff
@@ -18,6 +18,7 @@
         "xarray>=0.15.1",
         "zarr>=2.4.0",
         "joblib>=0.16.0",
+        "typing_extensions",
```
frodre, line 21: deletion referenced by 
https://github.com/VulcanClimateModeling/fv3net/pull/1278/files?diff=split&w=1#r658117991

frodre (approved):
Nice work!  Mostly just minor comments other than agreement that there can be some consolidation in the written tests.  I ticked the integration tests to run, so hopefully, everything looks good there.  Otherwise, LGTM.

oliverwm1 (commented):
Perhaps there has already been discussion on this issue that I'm not aware of, but it there is an assumption in many places that `data_path` is of type `str` even though it can also be `Sequence[str]`. Is this intentional? do we no longer support supplying multiple data paths as positional arguments to the CLI?

I guess it's not strictly necessary, since for the cases where we need multiple paths (e.g. training on multiple climates), you could specify the extra paths in the yaml. But that's not ideal.

external/loaders/loaders/_config.py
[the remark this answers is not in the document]
mcgibbon: Added. It's a little simpler of a use case than the IO one so I only needed to add a `register` decorator to `dict`.

external/loaders/loaders/_config.py
```diff
@@ -0,0 +1,117 @@
...
+        for subclass in cls.__subclasses__():
+            try:
+                return dacite.from_dict(data_class=subclass, data=kwargs)
+            except (TypeError, AttributeError, dacite.exceptions.MissingValueError):
+                pass
+        raise ValueError("invalid BatchesLoader dictionary")
+
+
+@dataclasses.dataclass
+class BatchesFromMapperConfig(BatchesLoader):
```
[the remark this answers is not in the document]
mcgibbon: I added docstrings, and expanded the sphinx documentation for these to include the new configuration and hook into the registration system. I also fixed the docstring formatting by adding `sphinx.ext.napoleon` to the doc build. yaml documentation should exist in the form of vcm_workflow_control examples when that use case is ready.

external/loaders/loaders/_config.py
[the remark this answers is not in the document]
mcgibbon: See other comment, this is now auto-documented based on the registration state.

external/loaders/loaders/mappers/_fine_resolution_budget.py
```diff
@@ -155,14 +169,18 @@ def __init__(self, url):
...
         match = re.search(pattern, url)
+        if match is None:
+            raise ValueError(
+                f"invalid url, must contain pattern {pattern} but received {url}"
+            )
         date = vcm.parse_timestep_str_from_path(url)
         tile = match.group(1)
         return date, int(tile)
 
-    def __getitem__(self, key: str) -> xr.Dataset:
```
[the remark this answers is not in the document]
mcgibbon: This now gets type hinted for mypy via the `Mapping` entry in the class inheritance. The original entry here was incorrect, the key is not a string.

external/loaders/loaders/mappers/_nudged/_nudged.py
mcgibbon: We talked about this and decided this can be fine as part of moving to use configuration yamls to specify the data path, instead of CLI arguments (which we also discussed was fine)

external/loaders/loaders/typing.py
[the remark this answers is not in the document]
mcgibbon: Done.

external/loaders/setup.py
mcgibbon: Done.

external/loaders/tests/test__config.py
```diff
@@ -0,0 +1,176 @@
...
+def test_load_batches():
+    with batches_context():
+        mock_batches = [xr.Dataset()]
+        mock_batches_function = unittest.mock.MagicMock(return_value=mock_batches)
+        mock_batches_function.__name__ = "mock_batches_function"
+        loaders._config.register_batches_function(mock_batches_function)
+        assert "mock_batches_function" in loaders._config.BATCHES_FUNCTIONS
+        data_path = "test/data/path"
+        variables = ["var1"]
+        batches_kwargs = {"arg1": "value1", "arg2": 2}
```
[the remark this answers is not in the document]
mcgibbon: I reduced duplicated code here quite a bit by pulling the mocking step into the helper.

I was at least able to do this for the two "batches from mapper" test functions, since there are two functionalities (load_batches and load_mapper) that require exactly the same init. For the other two tests, the inits need slight modifications so I left them in place, rather than make the helper too complicated.

external/loaders/tests/test__config.py
```diff
@@ -128,13 +150,40 @@ def test_load_batches_from_mapper():
...
+        mock_batches_function.__name__ = "mock_batches_function"
+        loaders._config.register_batches_function(mock_batches_function)
+        mock_mapper = {"key": xr.Dataset()}
+        mock_mapper_function = unittest.mock.MagicMock(return_value=mock_mapper)
+        mock_mapper_function.__name__ = "mock_mapper_function"
+        loaders._config.register_mapper_function(mock_mapper_function)
+        data_path = "test/data/path"
+        variables = ["var1"]
+        batches_kwargs = {"arg1": "value1", "arg2": 2}
+        mapper_kwargs = {"arg3": 3}
```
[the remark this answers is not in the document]
mcgibbon: See comment above.

mcgibbon:
Updated, but I'll be doing my best not to look at any of these comments until I get back.

oliverwm1 (approved):
My concern about the data path issue was addressed since we agreed to move path specification to within the config yamls. Still more work to actually implement that, but I think this PR is a good step.

mcgibbon:
Waiting on integration tests to run then will merge/auto-merge

external/loaders/tests/test__config.py
```diff
@@ -0,0 +1,262 @@
...
+        loaders._config.batches_from_mapper_functions.register(mock_batches_function)
+        yield mock_batches_function
+
+
+def test_load_mapper():
+    with mapper_context():
+        mock_mapper = {"key": xr.Dataset()}
+        mock_mapper_function = unittest.mock.MagicMock(return_value=mock_mapper)
+        mock_mapper_function.__name__ = "mock_mapper_function"
+        loaders._config.mapper_functions.register(mock_mapper_function)
```
frodre, line 61: Duplicated in the `mapper_context` creation?

external/loaders/tests/test__config.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+@contextlib.contextmanager
+def mapper_context():
+    with registration_context(loaders._config.mapper_functions):
+        mock_mapper = {"key": xr.Dataset()}
+        mock_mapper_function = unittest.mock.MagicMock(return_value=mock_mapper)
+        mock_mapper_function.__name__ = "mock_mapper_function"
+        loaders._config.mapper_functions.register(mock_mapper_function)
+        yield mock_mapper_function
```
frodre, line 33: The layered contexts for adding in mocked functions still seems overly-complex but maybe it is the best answer.  I wonder if we could hide some of this by using pytest's `monkeypatch` to temporarily set the module registers.

e.g.:

```
class MockRegister(loader._config.Register()
@pytest.fixture
def mock_mapper():
    return {"key": xr.Dataset()}

@pytest.fixture
def mock_batches():
    return [xr.Dataset()]

def get_magic_mock(name, return_val):
    mock = unittest.moc.MagicMock(return_value=return_val)
    mock.__name__ = name
    return mock

@pytest.fixture
def mock_mapper_func(mock_mapper, monkeypatch):

    # anything using this fixture will have the temporarily cleared register
    monkeypatch.setattr(loaders._config, "mapper_functions", loaders._config.Register())
    mocked_func = get_magic_mock("mock_mapper_function", mock_mapper)

    return loaders._config.mapper_functions.register(mocked_func)

def test_load_mapper(mock_mapper, mock_mapper_func)
    data_path = "test/data/path"
    mapper_kwargs = {"arg1": "value1", "arg2": 2}
    config = loaders._config.MapperConfig(
        data_path=data_path,
        mapper_function="mock_mapper_function",
        mapper_kwargs=mapper_kwargs,
    )
    result = config.load_mapper()
    assert result is mock_mapper
    mock_mapper_function.assert_called_once_with(data_path, **mapper_kwargs)
```

Something like this would let you load the expected return value and the function with adjusted registry all just using fixtures instead of context adjustments.

frodre (commented):
Thanks for the updates!  Would love to see the outcome of the autogenerated documentation from the registered functions.  

I think the configuration tests could still use some simplification to help ease diving back into the code for any of us in the future.  I proposed a different strategy than using the contexts, so we'll see where that conversation goes.

external/loaders/tests/test__config.py
mcgibbon: The only issue with true monkeypatching (e.g. with monkeypatch) is that these objects are generally not used as `import module; module.object` but rather as `from module import object`, to reduce line length. Monkeypatching would not replace occurences of these registry dictionaries that have been imported in this manner, which could lead to bugs (including bugs from future changes). Since I already had things working via a clear-and-restore-dictionary context manager, I opted to keep it that way instead.

> Something like this would let you load the expected return value and the function with adjusted registry all just using fixtures instead of context adjustments.

This isn't strictly true - monkeypatch uses a context manager to achieve the same effect, just under the hood. If the issue is the `with` statements, I could turn these into fixtures instead if that's desirable. I just thought it's clearer that it's manipulating things external to the tests when it's a context manager.

frodre: > This isn't strictly true - monkeypatch uses a context manager to achieve the same effect, just under the hood. If the issue is the with statements, I could turn these into fixtures instead if that's desirable. I just thought it's clearer that it's manipulating things external to the tests when it's a context manager.

Sorry, you're right about what I wrote at the end.  What I had in mind was that it might be better to hide the complexity of the layered context management innards from future us.  But I can see the argument for making it clear that the test is modifying external state while running.


> Monkeypatching would not replace occurrences of these registry dictionaries that have been imported in this mannerMonkeypatching would not replace occurrences of these registry dictionaries that have been imported in this manner

The `setitem` monkeypatch for could be used instead of replacing the module object, but that gives less lee-way to other manual adjustments to the register.

The reason I looked into this was mostly that the code looked quite complex and still has a number of repeats.  When that happens it may make future us more hesitant to touch/work on it.  Moving contexts into fixtures would reduce the nesting a bit and make for a standard way of getting testing info to multiple places, but then obscure the contexts switching.  Although, obscuring could also be a point of confusion for further test development. 

Maybe just the context naming could be more illustrative like `clean_batches_register` and then a `_get_magic_mock` helper would be good.

PR #1279, opened 2021-06-24
artifacts generate CLI to resolve output urls

oliverwm1:
Move definition of experiment directory structure to `artifacts` package by defining new `artifacts generate` CLI. Note that the `step` is still added directly within each relevant argo workflow call.

Added public API:
- `artifacts generate BUCKET PROJECT TAG`

Refactored public API:
- `resolve-output-url.yaml` argo workflow now uses above CLI

- [x] Tests added


oliverwm1:
I'm confused by the integration test error. On my local machine, the `artifacts generate` CLI works fine. Is it somehow using an old version of the fv3net image?

oliverwm1:
I pulled `us.gcr.io/vcm-ml/fv3net:3e415ee16bbe188903906493739d8e436374db2e` and `artifacts generate` works fine in there. Does seem like the wrong image is being used somehow.

oliverwm1:
Turns out `kustomize` does not play nicely with an image being specified within an `argo` `script`.

workflows/argo/resolve-output-url.yaml
```diff
@@ -11,20 +11,15 @@ spec:
...
+        parameters:
+        - name: url
+          valueFrom:
+            path: /tmp/url.txt
+      container:
+        image: us.gcr.io/vcm-ml/fv3net
+        command: ["bash", "-c", "-x", "-e"]
+        args:
+          - |
+            artifacts generate "{{inputs.parameters.bucket}}" "{{inputs.parameters.project}}" "{{inputs.parameters.tag}}" > /tmp/url.txt
```
nbren12, line 25: I think you can access the stdout via `outputs.result` as the original code did and can simplify this a bit:
```
 ```suggestion
        command:
          - artifacts
          - generate
          - "{{inputs.parameters.bucket}}"
          - "{{inputs.parameters.project}}"
          - "{{inputs.parameters.tag}}"
          
```

See these argo docs: https://argoproj.github.io/argo-workflows/fields/#fields_23.

workflows/argo/train-diags-prog.yaml
```diff
@@ -68,7 +68,7 @@ spec:
             - name: times
               value: "{{inputs.parameters.train-times}}"
             - name: output
-              value: "{{tasks.resolve-output-url.outputs.result}}/trained_models/{{item.name}}"
```
nbren12, line 71: I think you can revert this changes.

workflows/prognostic_run_diags/fv3net/artifacts/generate.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def main(args):
+    url = generate(args.bucket, args.project, args.tag)
+    print(url)
+
+
+def register_parser(subparsers):
+    parser = subparsers.add_parser(
+        "generate", help="Generate a URL from provided bucket, project, and tag."
```
nbren12, line 20: To me "generate" implies creation of a persistent resource, which is not what this CLI really does even though it does support this task. Should we stick to the existing convention "resolve-url"?

workflows/prognostic_run_diags/fv3net/artifacts/report_search.py
```diff
@@ -104,3 +104,34 @@ def _insert_public_domain(url) -> str:
...
+
+
+def main(args):
+    if args.write:
+        index = ReportIndex()
+        index.compute(args.reports_url)
+        index.dump(os.path.join(args.reports_url, "index.json"))
+    index = ReportIndex.from_json(os.path.join(args.reports_url, "index.json"))
+    for link in index.public_links(args.url):
+        print(link)
```
nbren12, line 116: To enhance discoverability, it would be nice to refactor this as a function in `query`. That is where users (me) tend to look for functions like this.

workflows/prognostic_run_diags/fv3net/artifacts/tests/test_generate.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+from fv3net.artifacts.generate import generate
+import pytest
+
+FAKE_TIME = datetime.datetime(2021, 6, 3, 3, 2, 4)
```
nbren12, line 5: Not used anywhere else, so I suggest scoping within class `mydatetime`.

nbren12 (approved):
Looks good. Great to see this functionality concentrated in one place. I think the argo workflow can be a bit simpler, and also have some nits about the naming.

For inspiration, this is the approach that weights and biases takes to artifacts management: https://wandb.ai/wandb/arttest/reports/Intro-to-W-B-Artifacts--VmlldzozNTAzMDM.

workflows/argo/resolve-output-url.yaml
oliverwm1: Oh I thought the `outputs.result` was only available for `script` templates. I'll give this a shot.

workflows/prognostic_run_diags/fv3net/artifacts/generate.py
oliverwm1: Yeah that sounds good.

workflows/prognostic_run_diags/fv3net/artifacts/report_search.py
oliverwm1: That seems a bit odd to me since the parser which points to this function is defined in `report_search.py`. I'm going to leave it as is.

PR #1281, opened 2021-06-25
Delete compute_diagnostics_memoized.sh script

### LINKED ISSUE #1267: Movie links in prognostic report broken
For some reason the base part of the URL for movie links is repeated twice in a recent report I made.

e.g.: `https://storage.googleapis.com/https://storage.googleapis.com/vcm-ml-public/argo/2021-06-116-n2f-sponge-layer-tapering-with-baseline/prog-control/column_ML_wind_tendencies.mp4`

oliverwm1:
The memoization of the prognostic run diagnostics is slow (because a pod has to spin up for each run, even if the diagnostic ultimately don't have to be computed) and mostly unnecessary since diagnostics are now computed for every prognostic run. This PR deletes the `compute_diagnostics_memoized.sh` script and changes the report generation argo workflow to NOT computed any diagnostics by default. There is still an option to recompute all the diagnostics, which is useful for cases where you used the wrong verification data or you want to recompute diagnostics with a more recent fv3net version.

New public API:
- `report-from-json` CLI for generating a report from a JSON with an array of dicts with `name` and `url` items, where the url points to a folder of run diagnostics

Refactored public API:
- report generation entrypoint for `prognostic-run-diags` argo workflow now does NOT compute any diagnostics by default, and assumes they have already been computed. An option is provided to recompute the diagnostics.

Significant internal changes:
- argo `diagnostics` template now includes movie generation
- now the movies are also automatically generated after prognostic runs as well as the `diags.nc` and `metrics.json` files
- delete unnecessary argo parameters related to memory requests (report generation shouldn't vary much, and diagnostic/movie generation both use ~8cpu so they may as well use all the memory that the med-sim-pool nodes provide)

- [x] Tests added. I manually verified that the report argo workflow works as expected.

Resolves #1267


workflows/argo/prognostic_run_diags.yaml
```diff
@@ -6,76 +6,62 @@ spec:
...
+        limits:
+          memory: "8Gi"
+          cpu: "1000m"
+      command: ["bash", "-c", "-x", "-e"]
+      args:
+      - |
+        cat << EOFRUNDIRSJSON > rundirs.json
+        {{inputs.parameters.runs}}
+        EOFRUNDIRSJSON
+        jq '.[].url |= . + "_diagnostics"' rundirs.json > diagnostic_dirs.json
```
oliverwm1, line 63: this `_diagnostics` suffix is already used elsewhere in this argo workflow (and in the prognostic run one) so I thought it was okay to hard-code here. Ideally we move all the logic into one place, but that is not totally trivial for the many argo parameters.

nbren12, line 63: Would prefer to see this logic embedded in the `report-from-json` CLI maybe as a flag `--url-is-rundir` or something. This will make it easier to use the same rundirs.json locally and in argo.

workflows/prognostic_run_diags/docs/usage.rst
```diff
@@ -51,4 +51,7 @@ To make a report from runs that are not in a combined folder like
       path/to/baseline_run_diags \
       path/to/prognostic_run_diags
 
+Or you may encode the paths to URLs in a JSON and use the
+`prognostic_run_diags report-from-json` entrypoint.
```
nbren12, line 55:
```suggestion
`prognostic_run_diags report-from-json` CLI.
```

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/computed_diagnostics.py
```diff
@@ -52,6 +52,21 @@ def url_to_folder(url):
             {str(k): url_to_folder(url) for k, url in enumerate(urls)}
         )
 
+    @staticmethod
+    def from_json(url: str) -> "ComputedDiagnosticsList":
+        """Open labeled computed diagnostics at urls specified in given JSON."""
+
+        def url_to_folder(url):
+            fs, _, path = fsspec.get_fs_token_paths(url)
```
nbren12, line 60: nit: Since this code depends on `vcm` anyway, I would lean towards using the `vcm.cloud.get_fs` function.

nbren12 (approved):
Looks great. improves the coverage and deletes code.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/computed_diagnostics.py
oliverwm1: I am going to leave it as, since we due use the `path` output of the function here. I copied this from the existing `from_urls` method.

PR #1283, opened 2021-06-30
Prognostic report: fix contour cbar levels across rows

AnnaKwa:
Currently the zonal-pressure bias plots are hard to compare across experiments because matplotlib is inconsistent in selecting color levels for a fixed vmax. This PR fixes the cbar levels across the row.

before:
image

after:
image


nbren12 (approved):
LGTM

nbren12:
It still isn't symmetric in the case when there is data beyond only one of the limits though.

oliverwm1:
> It still isn't symmetric in the case when there is data beyond only one of the limits though.

setting `extend='both'` would take take of this I think

PR #1285, opened 2021-07-01
prog_run_diags: metrics and new derived diagnostics on the fly

oliverwm1:
Since the metrics are fast to compute, it is nice to directly compute them from the `diags.nc` on load. This is nice, since we then get the metrics for the verification data for free. In addition, being able to compute some additional derived diagnostics from the `diags.nc` is helpful. This PR implements both, and includes the mass stream function as an example derived diagnostic. See example report [here](https://storage.googleapis.com/vcm-ml-public/oliwm/test-new-report/example2/index.html).

Added public API:
- `fv3net.diagnostics.prognostic_run.derived_diagnostics.derived_registry` which computes new diagnostics from a given `diags.nc` dataset

Refactored public API:

Significant internal changes:
- added new `Registry` class which is now used for registering functions for computing metrics, diagnostics and derived diagnostics
- renamed some modules for better clarity between "derived variables" and "derived diagnostics"
- added option to compute metrics from loaded diagnostics and now the report does this
- now compute precip histogram bias (used for getting histogram for verification data)
- within argo workflow, just generate 960 frames for movies (10 days if 15-minute data) to speed generation

- [x] Tests added


workflows/argo/prognostic_run_diags.yaml
```diff
@@ -184,7 +184,7 @@ spec:
       command: ["/bin/bash", "-x", "-e", "-c"]
       args:
       - |
-        prognostic_run_diags movie {{inputs.parameters.run}} {{inputs.parameters.output}}
+        prognostic_run_diags movie --n_timesteps 960 {{inputs.parameters.run}} {{inputs.parameters.output}}
```
oliverwm1, line 187: this is just to speed up movies, since they are now computed by default. This will give us 10-day movies for 15-minute data, which is plenty I think.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -30,7 +30,7 @@ def grab_diag(ds, name):
             replace_dict[var] = var[: -len(match)]
 
     if len(replace_dict) == 0:
-        raise ValueError(f"No diagnostics with name {name} found.")
+        return xr.Dataset()
```
oliverwm1, line 33: in particular since the metrics are now computed for the verification dataset which lacks some diagnostics, it is necessary to make sure the code works for the case of missing diagnostics.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/movies.py
```diff
@@ -18,7 +18,7 @@
 import vcm.catalog
 
 from fv3net.diagnostics.prognostic_run import config
-import fv3net.diagnostics.prognostic_run.load_diagnostic_data as load_diags
+import fv3net.diagnostics.prognostic_run.load_run_data as load_diags
```
oliverwm1, line 21: refactored `load_diagnostic_data` to `load_run_data` since the `diagnostic` word is overloaded and here mostly means reduced diagnostics that are computed from the raw "run" output

nbren12 (changes requested):
Looking good, but needs some tests and possibly refactoring the register pattern to avoid the third copy-paste of this pattern in the code.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/registry.py
```diff
@@ -0,0 +1,23 @@
+from collections import defaultdict
+from typing import Any, Callable, Mapping
+
+import xarray as xr
+from toolz import curry
+
+
+class Registry:
```
nbren12, line 8: Can you add some tests for this?

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/registry.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, merge_func: Callable[[Mapping[str, xr.Dataset]], Any]):
```
nbren12, line 9: Nice job parameterizing the merge function! I would just call it "merge" since that is a good verb for a function name.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -67,38 +65,23 @@ def _mask_array(
     return masked_arr
 
 
-@curry
-def add_to_metrics(metricname: str, func: Callable[[xr.Dataset], xr.Dataset]):
-    """Register a function to be used for computing metrics
-
-    This function will be passed the diagnostics xarray dataset,
```
AnnaKwa, line 74: The refactor to have a common `Registry` class is nice and streamlined, though a side effect is that the information formerly in the `add_to_x` docstrings about what each registry expects and outputs is gone. Please update the documentation section "Developing this workflow" to include this information.

AnnaKwa (changes requested):
Thanks, this looks good! The consolidation into a single `Registry` class is nice, however it's less clear from the code now how to add to a given registry since the docstrings that used to detail this information are gone with the individual registries. Could you also update the docs to include this? It might be helpful to future contributors to have some inline comments to this effect as well where the registries are defined in their modules.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -45,148 +40,44 @@
...
-            )
-            src_attrs = src_ds[variable].attrs
-            da = da.assign_attrs(src_attrs)
-        else:
-            logger.debug(
-                f"Not transferring attributes from source dataset for {variable}. This "
-                "may cause issues with automated report generation."
-            )
-
+        da = ds[variable]
```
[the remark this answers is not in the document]
oliverwm1: I made it redundant by setting `xr.set_options(keep_attrs=True)` in this module and in `transform.py`. The histogram one needs to be done manually, since the `vcm` function uses numpy.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/registry.py
```diff
@@ -0,0 +1,52 @@
...
+    @curry
+    def register(
+        self, name: str, func: Callable[[Any], Union[xr.Dataset, xr.DataArray]]
+    ):
+        if name in self._funcs:
+            raise ValueError(f"Function {name} has already been added to registry.")
+        self._funcs[name] = func
+
+    def compute(self, *args, n_jobs=-1, **kwargs) -> Any:
+        computed_outputs = Parallel(n_jobs=n_jobs, verbose=True)(
```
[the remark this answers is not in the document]
oliverwm1: yeah, thanks for reminder about the special behavior for the diurnal cycle. I agree it should be fine.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
oliverwm1: Good point. Maybe each registry's merge function is a good place for this docstring, since that function is what distinguishes each registry. Will also update the docs.

oliverwm1: In the end I just added some comments next to the registry definition. Maybe the right way to do this is to require the developer to specify the function signatures when they initialize the registry but I'm not sure if it's worth the complexity to implement.

AnnaKwa (approved):
Documentation updates LGTM, thanks!

workflows/prognostic_run_diags/tests/test_registry.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+import xarray as xr
+import pytest
+from typing import Sequence, Tuple
+
+from fv3net.diagnostics.prognostic_run.registry import Registry
+
+
+@pytest.fixture()
```
nbren12, line 9: Would prefer a `_get_registry()` function to this fixture to make it easier to parameterize in the future without leading the fixture-house-of-cards phenomena which has plagued us in the past.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -134,7 +145,7 @@ def rms_errors(prognostic, verification_c48, grid):
 def zonal_means_dycore(prognostic, verification, grid):
     logger.info("Preparing zonal+time means (dycore)")
     zonal_means = zonal_mean(prognostic, grid.lat)
-    return time_mean(zonal_means)
+    return _assign_source_attrs(time_mean(zonal_means), prognostic)
```
nbren12: Can you push this attribute assignment into the `time_mean` and `rms` functions? It seems the RMS function should define the fact that it doesn't change the units for example.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -291,39 +304,39 @@ def global_biases_dycore(prognostic, verification, grid, mask_type=mask_type):
         logger.info(f"Preparing average biases for dycore variables ({mask_type})")
         bias_errors = bias(verification, prognostic, grid.area, HORIZONTAL_DIMS)
```
nbren12: likewise, push attribute assignment into `bias`.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/transform.py
```diff
@@ -17,6 +17,8 @@
 
 from .constants import HORIZONTAL_DIMS, DiagArg
 
+xr.set_options(keep_attrs=True)
```
nbren12, line 20: I get worried about this kind of global state/configuration making things magical. Can you scope this setting to the specific computing functions `time_mean`, `rms`, `bias`, etc?

nbren12 (approved):
I like the `Registry`, but have some concerns about using `xr.set_options(attrs=True)`, I preferred the more explicit approach in your earlier commit.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/compute.py
oliverwm1: Yeah I considered that. Currently the function works for either Dataset or DataArray inputs, so will need to restrict to only Datasets or do some type checking to properly assign attributes in both cases.

oliverwm1: or perhaps better, still use the xarray keep_attrs=True option, but just within particular contexts? I think that would be simplest.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/transform.py
oliverwm1: Done.

PR #1287, opened 2021-07-06
fv3fit: add DerivedModel predictor

AnnaKwa:
This adds a `DerivedModel` class which wraps a "base" ML model and allows additonal derived variables to be returned along with the base ML predicted variables. The immediate use case for this would be to use the ML-predicted downward surface shortwave flux to derive the net surface shortwave flux for updating in the prognostic run. The base ML model can be any other trained fv3fit model. Like the existing `EnsembleModel`, `DerivedModel` has no training function or dump method and instead must be manually created in a model dir with a config file (see added docs). 

The configuration is pretty simple, ex. the yaml below will add albedo to the list of input variables and merge net surface shortwave into the rest of the prediction dataset from the trained ML.
```
    model: gs://vcm-ml-scratch/annak/2021-07-07-nn-no-net-sw/trained_model
    additional_input_variables:
        - surface_diffused_shortwave_albedo
    derived_output_variables:
        - net_shortwave_sfc_flux_derived
```

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -1,10 +1,135 @@
-from typing import Iterable, Set, Hashable
+from typing import Iterable, Set, Hashable, Sequence
 import fsspec
 import yaml
 import os
 import xarray as xr
+import vcm
+import dataclasses
+
 from . import io
 from .predictor import Predictor
+from .config import (
+    DerivedModelHyperparameters,
+    register_training_function,
+    get_training_function,
+)
+
+
+@register_training_function("DerivedModel", DerivedModelHyperparameters)
+def train_derived_model(
+    input_variables: Iterable[str],
+    output_variables: Iterable[str],
+    hyperparameters: DerivedModelHyperparameters,
+    train_batches: Sequence[xr.Dataset],
+    validation_batches: Sequence[xr.Dataset],
+):
+
+    model = DerivedModel(
+        "sample",
+        base_model_input_variables=input_variables,
+        output_variables=output_variables,
+        hyperparameters=hyperparameters,
+    )
+    # TODO: make use of validation_batches, currently validation dataset is
+    # passed through hyperparameters.fit_kwargs
+    model.fit(train_batches, validation_batches)
+    return model
+
+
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -2,11 +2,97 @@
...
+import vcm
+
 from . import io
 from .predictor import Predictor
 
 
+@io.register("derived_model")
+class DerivedModel(Predictor):
+    _CONFIG_FILENAME = "derived_model.yaml"
+    _BASE_MODEL_DIRECTORY = "base_model_data"
```
mcgibbon, line 16: This attribute isn't used right now, you may want to delete it and add it when you implement `dump`.

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -2,11 +2,97 @@
...
+
+    def __init__(
+        self,
+        base_model: Predictor,
+        additional_input_variables: Iterable[Hashable],
+        derived_output_variables: Iterable[Hashable],
```
mcgibbon, line 22: Is it possible to do a sanity check on these? e.g. by checking if the derived variables are supported or by automatically getting `additional_input_variables` from the derived variables via some kind of call to `vcm.DerivedMapping`?

mcgibbon (approved):
LGTM, thanks for the docs!

For next steps, we can remove the manual "model" creation step by adding it as a configurable automatic step in `fv3fit.train`. I would suggest implementing the `dump` method and adding a toggle-able code block in the train routine to wrap this as we discussed on zoom. You also may be able to add more safety or reduce the work needed to specify the input/output variables.

external/fv3fit/fv3fit/_shared/models.py
AnnaKwa: The former (check if derived variables are supported) could be done by checking if they exist in `DerivedMapping._VARIABLES`, I can add this. The latter (atuomatically get `additional_input_variables`) is hard to implement with how `vcm.DerivedMapping` currently works, which is that it assumes it's the user's responsibility to initialize with an appropriate dataset and will just return KeyErrors if that isn't the case.

PR #1289, opened 2021-07-09
fv3fit: dump DerivedModel if derived_output_variables in config

AnnaKwa:
The previous PR to add the `DerivedModel` class did not include a dump method; it relied on the user manually creating the DerivedModel's save directory and yaml. This PR adds a dump method and config entry`derived_output_variables` to the `TrainingConfig`. If any `derived_output_variables` are specified in the config, the training script will wrap the trained ML model as a DerivedModel before saving. The `DerivedModel` will check to see if any additional input variables need to be added to its `input_variables` on top of the ML features to derive these variables.

Added the option to register a list of  `required_input_variables` to derived variables in `vcm.DerivedMapping`. The `DerivedModel` checks against these to see if it requires additional input variables to be present in the dataset used as input to its predict method on top of what is used as the base ML feature set. 

Added public API:
- optional config arg `TrainingConfig.derived_output_variables`. This has also been added to the legacy `_ModelTrainingConfig` since it's still being used.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -78,7 +78,6 @@ class TrainingConfig:
     sample_dim_name: str = "sample"
     random_seed: Union[float, int] = 0
     derived_output_variables: List[str] = dataclasses.field(default_factory=list)
-    nonfeature_input_variables: List[str] = dataclasses.field(default_factory=list)
 
     @classmethod
     def from_dict(cls, kwargs) -> "TrainingConfig":
@@ -299,7 +298,6 @@ class _ModelTrainingConfig:
     model_path: str = ""
     timesteps_source: str = "timesteps_file"
     derived_output_variables: List[str] = dataclasses.field(default_factory=list)
-    nonfeature_input_variables: List[str] = dataclasses.field(default_factory=list)
 
     def __post_init__(self):
         if self.scaler_type == "mass":
@@ -333,7 +331,6 @@ def legacy_config_to_new_config(legacy_config: _ModelTrainingConfig) -> Training
         "random_seed",
         "sample_dim_name",
         "derived_output_variables",
-        "nonfeature_input_variables",
     ]
     if config_class is RandomForestHyperparameters:
         for key in ("scaler_type", "scaler_kwargs"):
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -298,8 +298,8 @@ class _ModelTrainingConfig:
     save_model_checkpoints: bool = False
     model_path: str = ""
     timesteps_source: str = "timesteps_file"
-    derived_output_variables: List[str] = None
-    nonfeature_input_variables: List[str] = None
+    derived_output_variables: List[str] = dataclasses.field(default_factory=list)
+    nonfeature_input_variables: List[str] = dataclasses.field(default_factory=list)
 
     def __post_init__(self):
         if self.scaler_type == "mass":
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -69,6 +69,9 @@ class TrainingConfig:
         sample_dim_name: deprecated, internal name used for sample dimension
             when training and predicting
         random_seed: value to use to initialize randomness
+        derived_output_variables: optional list of prediction variables that
+            are not directly predicted by the ML model but instead are derived
+            using the ML-predicted output_variables
     """
 
     model_type: str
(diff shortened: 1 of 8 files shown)


external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -77,6 +77,7 @@ class TrainingConfig:
     hyperparameters: Dataclass
     sample_dim_name: str = "sample"
     random_seed: Union[float, int] = 0
+    derived_output_variables: List[str] = dataclasses.field(default_factory=list)
```
mcgibbon, line 80: Please add a line to the docstring about this.

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -62,24 +65,33 @@ def predict(self, X: xr.Dataset) -> xr.Dataset:
...
-            "using instructions at "
-            "http://vulcanclimatemodeling.com/docs/fv3fit/derived_model.html"
-        )
+        base_model_path = os.path.join(path, self._BASE_MODEL_SUBDIR)
+        options = {
+            "additional_input_variables": self._additional_input_variables,
+            "derived_output_variables": self._derived_output_variables,
+            "model": base_model_path,
+        }
+        self._base_model.dump(base_model_path)
```
mcgibbon, line 73: I think you need to use `fv3fit.dump` here so that it includes the name label. If you add a dump/load test, it should catch this.

external/fv3fit/fv3fit/train.py
```diff
@@ -105,6 +105,12 @@ def main(args):
         train_batches=train_batches,
         validation_batches=val_batches,
     )
+    if len(train_config.derived_output_variables) > 0:
+        model = DerivedModel(
+            model,
+            train_config.nonfeature_input_variables,
```
mcgibbon, line 111: I think this call needs to be updated to remove the `nonfeature_input_variables` attribute you made implicit.

external/fv3fit/tests/test_derived_model.py
```diff
@@ -20,11 +20,18 @@
 )
 
 
+def test_get_additional_inputs():
```
mcgibbon, line 23: Please add tests for dumping and loading DerivedModel. Specifically a test that if you dump and load and then predict, you get the same result as before you dumped.

external/vcm/vcm/derived_mapping.py
```diff
@@ -105,27 +113,37 @@ def northward_wind(self):
...
+)
 def horizontal_wind_tendency_parallel_to_horizontal_wind(self):
     tendency_projection_onto_wind = (
         self["eastward_wind"] * self["dQu"] + self["northward_wind"] * self["dQv"]
     ) / np.linalg.norm((self["eastward_wind"], self["northward_wind"]))
     return tendency_projection_onto_wind
 
 
-@DerivedMapping.register("net_shortwave_sfc_flux_derived")
+@DerivedMapping.register(
```
mcgibbon, line 143: (very) Optional suggestion: You could consider adding automatic unit tests for these routines that check `required_inputs` is accurate, or even automaticAlly determine them. This would require A mock "self" object which returns some AppropriAtely-shAped zero ArrAys And logs cAlls to __getitem__, then pAssing it to these functions when they get registered.

This is Also fine though, just thought I'd suggest.

(Apologies for my cApitAl A's, my keyboArd is bugged)

mcgibbon (changes requested):
One bug that should be caught by a dump/load test. Really cool to see the additional inputs derived automatically.

external/fv3fit/tests/test_derived_model.py
```diff
@@ -54,7 +59,29 @@ def test_derived_alert_to_missing_additional_input():
...
+    input_data = ds_in.stack(sample=["x", "y"])
+    prediction = derived_model.predict(input_data)
+
+    fv3fit.dump(derived_model, str(tmpdir))
+    loaded_model = fv3fit.load(str(tmpdir))
+
+    prediction_after_load = loaded_model.predict(input_data)
+    for var in loaded_model.output_variables:
+        assert var in derived_model.output_variables
+        assert np.allclose(prediction[var].values, prediction_after_load[var].values)
```
mcgibbon, line 87: I think we have a utility function in vcm (or just in xarray) that will compare datasets including metadata. You may want to switch to that since a common issue with dumping and loading models is the loaded model can't restore metadata properly.

mcgibbon (approved):
Great work! LGTM. One optional suggestion to improve a test's coverage.

external/fv3fit/tests/test_derived_model.py
AnnaKwa: Thanks, switched to use `xr.Dataset.identical`

PR #1290, opened 2021-07-09
prog_run_diags: Add top-level metrics

oliverwm1:
Add some 'top level' metrics to index page of prognostic run report. Also, add metrics for 1) strength of ITCZ, as measured by upward mass transport in ascending part of Hadley cell, and 2) the mean value of various diagnostics in the region of ascent.

[Here is an example report](https://storage.googleapis.com/vcm-ml-public/argo/2021-07-09t215809-6bbbf0e08b4d/index.html) from some 40-day runs. (these were done with 13a2e6f, before the pQ1 fix, so the Q1 is off for the prognostic runs)

Significant internal changes:
- add table with "top level" metrics to `index.html`
- replace metric bar plots on `index.html` with an HTML table of all the metrics
- properly compute `pQ1` for case of overriding surface radiative fluxes (requires user to have output `_from_RRTMG` fluxes)

- [x] Tests added


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -222,6 +253,13 @@ def compute_percentile(
     return bin_midpoints[closest_index]
 
 
+def itcz_edges(psi: xr.DataArray, lat: str = "latitude",) -> Tuple[float, float]:
+    """Compute latitude of ITCZ edges given mass streamfunction at particular level."""
+    lat_min = psi.sel({lat: slice(-30, 10)}).idxmin(lat).item()
+    lat_max = psi.sel({lat: slice(-10, 30)}).idxmax(lat).item()
```
oliverwm1, line 259: sometimes people will fit a polynomial near the region of the min/max in order to identify the min/max positions on a scale finer than the grid resolution (in this case 2°).

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -238,8 +276,7 @@ def register_parser(subparsers):
 
 
 def main(args):
-    diags = xr.open_dataset(args.input)
-    diags["time"] = diags.time - diags.time[0]
```
oliverwm1, line 242: none of the metrics actually depend on the time coordinate starting at zero, so I'm removing this.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -238,8 +276,7 @@ def register_parser(subparsers):
...
+    diags = _add_derived_diagnostics(xr.open_dataset(args.input))
```
oliverwm1, line 279: with this change, the `metrics.json` saved will include metrics computed from derived diagnostics, which is helpful if we want to use this in places other than the static HTML report.

external/report/report/create_report.py
```diff
@@ -99,6 +99,14 @@ def __repr__(self) -> str:
         return "<ol>\n" + "\n".join(items_li) + "\n</ol>"
 
 
+class RawHTML:
```
nbren12, line 102: I thought we already had an object that did this. I could be wrong though.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/computed_diagnostics.py
```diff
@@ -200,6 +200,49 @@ def get_metric_all_runs(self, metric_type: str, variable: str) -> pd.Series:
         metric_name = self.metric_name(metric_type, variable)
         return self.metrics[self.metrics.metric == metric_name]
 
+    def get_metric_type_table(self, metric_type: str) -> pd.DataFrame:
```
nbren12, line 203: Rather than increasing the surface area of `RunMetrics`, I would move make this function into the "view" code (static_report.py). "nicely labeled" seems to imply this is related more to visualization than data loading. Adding more methods to the RunMetrics interface will make it harder to implement for new backends (e.g. a database or wandb).

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/computed_diagnostics.py
```diff
@@ -200,6 +200,49 @@ def get_metric_all_runs(self, metric_type: str, variable: str) -> pd.Series:
...
+        data = {}
+        for varname in variables:
+            units = self.get_metric_units(metric_type, varname, runs[0])
+            column_name = f"{varname} [{units}]"
+            data[column_name] = [
+                self.get_metric_value(metric_type, varname, run) for run in runs
+            ]
+        return pd.DataFrame(data, index=runs)
+
+    def get_table(self, metric_names: Mapping[str, Sequence[str]]) -> pd.DataFrame:
```
nbren12, line 224: Same comment as above. I would make this private API in static_report.py.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -197,6 +203,31 @@ def percentile_metric(diags, percentile=percentile):
         return percentiles
 
 
+@metrics_registry.register("tropics_max_minus_min")
+def itcz_mass_transport(diags):
+    psi_mid_troposphere_name = "mass_streamfunction_300_700_zonal_and_time_mean"
```
nbren12, line 208: Use `constants.MASS_STREAMFUNCTION_MID_TROPOSPHERE`?

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -197,6 +203,31 @@ def percentile_metric(diags, percentile=percentile):
...
+        return xr.Dataset()
+    psi = diags[psi_mid_troposphere_name]
+    lat_min, lat_max = itcz_edges(psi)
+    max_minus_min = psi.sel(latitude=lat_max) - psi.sel(latitude=lat_min)
+    return xr.Dataset({psi_mid_troposphere_name: max_minus_min.assign_attrs(psi.attrs)})
+
+
+@metrics_registry.register("tropical_ascent_region_mean")
+def tropical_ascent_region_mean(diags):
+    psi_mid_troposphere_name = "mass_streamfunction_300_700_zonal_and_time_mean"
```
nbren12, line 219:
```suggestion
    psi_mid_troposphere_name = constants.MASS_STREAMFUNCTION_MID_TROPOSPHERE
```

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -300,50 +300,37 @@ def histogram_plots(diagnostics: Iterable[xr.Dataset]) -> HVPlot:
...
-            }
-    return table
+    return metric_type_table(metrics, "drift")
+
+
+def metric_type_table(metrics: RunMetrics, metric_type: str) -> hv.HoloMap:
+    """Return HTML for metrics table of all variables of metric_type"""
+    df = metrics.get_metric_type_table(metric_type).transpose()
+    header = f"<h3>{metric_type}</h3>"
+    return RawHTML(header + df.to_html(justify="center", float_format="{:.2f}".format))
```
nbren12, line 325: Would it make sense to make a class `Pandas` in report that hides this conversion logic?

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -300,50 +300,37 @@ def histogram_plots(diagnostics: Iterable[xr.Dataset]) -> HVPlot:
...
-            type_label = f"{name} {metric_type} [{units}]"
-            table[type_label] = {
-                run: f"{metrics.get_metric_value(metric_type, name, run):.2f}"
-                for run in metrics.runs
-            }
-    return table
+    return metric_type_table(metrics, "drift")
+
+
+def metric_type_table(metrics: RunMetrics, metric_type: str) -> hv.HoloMap:
```
nbren12, line 321:
```suggestion
def metric_type_table(metrics: RunMetrics, metric_type: str) -> RawHTML:
```

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -300,50 +300,37 @@ def histogram_plots(diagnostics: Iterable[xr.Dataset]) -> HVPlot:
...
+    """Return HTML for metrics table of all variables of metric_type"""
+    df = metrics.get_metric_type_table(metric_type).transpose()
+    header = f"<h3>{metric_type}</h3>"
+    return RawHTML(header + df.to_html(justify="center", float_format="{:.2f}".format))
+
+
+def metric_table(
+    metrics: RunMetrics, metric_names: Mapping[str, Sequence[str]]
+) -> hv.HoloMap:
```
nbren12, line 330:
```suggestion
) -> RawHTML:
```

nbren12 (changes requested):
Looks good to me! My main request is to turn the `RunMetrics.*_table` methods into functions to avoid increasing the surface area of the RunMetrics class.

I will need some more time to digest the meaning of these new metrics. Why does the baseline run have the lowest "column_integrated_q1 tropical_ascent_region_mean" but the highest "mass_streamfunction_300_700_zonal_and_time_mean tropics_max_minus_min"? I would tropical heating and mass transport to be correlated.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
oliverwm1: I'm going to leave the conversion here to avoid adding the `pandas` dependency to the report package.

external/report/report/create_report.py
oliverwm1: You're right, there was one defined in `views/matplotlib.py`. I refactored the functions that used it to use the new `RawHTML` from `report`.

oliverwm1:
> I will need some more time to digest the meaning of these new metrics. Why does the baseline run have the lowest "column_integrated_q1 tropical_ascent_region_mean" but the highest "mass_streamfunction_300_700_zonal_and_time_mean tropics_max_minus_min"? I would expect tropical heating and mass transport to be correlated.

The example report I linked above still had the bug in the computation of pQ1 for the prognostic runs (the surface radiative fluxes used to compute the implied pQ1 were not the ones relevant for the atmospheric budget due to the overriding). I am generating a new example with this issue fixed, can link you once it's done.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/static_report.py
nbren12: ok.

nbren12 (approved):

PR #1292, opened 2021-07-12
fix bug from a5805ad8

nbren12:


frodre (approved):
Fix looks good.  Integration test says it failed, but looks like it did fine based on output.

PR #1298, opened 2021-07-15
package sklearn_runfile.py in `runtime` 

nbren12:
There is only one runfile now so we can get rid of some special handling of this file by bundling it in the `runtime` directory.

Public API changes:
- remove `runfile` argument from `runfv3` subcommands and the argo runfv3 workflow.


brianhenn (approved):
LGTM thanks

PR #1299, opened 2021-07-15
vcm.DerivedMapping: add one-hot encoding for sea/land/seaice sfc types

AnnaKwa:
This adds derived variables in `vcm.DerivedMapping` for sea / land / sea ice surfaces. Currently the neural networks only use `sfc_geopotential` as a proxy for surface type since the surface type information is contained in the 0/1/2 categorical variable `land_sea_mask`.

- [x] Tests added

### DIFF
--- a/external/vcm/vcm/derived_mapping.py
+++ b/external/vcm/vcm/derived_mapping.py
@@ -151,3 +151,27 @@ def net_shortwave_sfc_flux_derived(self):
         "override_for_time_adjusted_total_sky_downward_shortwave_flux_at_surface"
     ]
     return (1 - albedo) * downward_sfc_shortwave_flux
+
+
+@DerivedMapping.register(
+    "is_land", required_inputs=["land_sea_mask"],
+)
+def is_land(self):
+    # one hot encoding for land / (sea or sea ice) surface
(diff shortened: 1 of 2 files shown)


external/vcm/vcm/derived_mapping.py
```diff
@@ -151,3 +151,27 @@ def net_shortwave_sfc_flux_derived(self):
...
     ]
     return (1 - albedo) * downward_sfc_shortwave_flux
+
+
+@DerivedMapping.register(
+    "is_land", required_inputs=["land_sea_mask"],
+)
+def is_land(self):
+    # one hot encoding for land / (sea or sea ice) surface
+    return xr.where(self["land_sea_mask"] == 1, 1.0, 0.0)
```
spencerkclark, line 161: In general I think this typically works, but if we want to be totally safe we could use `vcm.xarray_utils.isclose` for these checks since the land surface type is often encoded as a float.
```suggestion
    return xr.where(vcm.xarray_utils.isclose(self["land_sea_mask"], 1), 1.0, 0.0)
```

spencerkclark (approved):

Issue #1296, opened 2021-07-12
Enhance test coverage of runfv3 and fv3post


nbren12:
Currently, the test coverage of fv3post and runfv3 outside of the E2E tests is pretty low. I have encountered numerous bugs (e.g. the fallout of a5805ad,  #1294 ) that are only caught by E2E test. Refactoring the prognostic run tests to use the append segment workflow should catch a lot of these errors at an earlier point and speedup development cycles.

cc @AnnaKwa @oliverwm1

PR #1300, opened 2021-07-17
Prognostic run improvements

nbren12:
Improve the packaging and tests of the prognostic run workflow. See the commit descriptions for motivation.

Added public API:
- prepare_config.py is now installed as a script called prepare-config on the PATH

Related to #1296


workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -516,27 +517,32 @@ def completed_rundir(configuration, tmpdir_factory):
     else:
         raise NotImplementedError()
 
-    rundir = tmpdir_factory.mktemp("rundir")
+    rundir = tmpdir_factory.mktemp("rundir").join("subdir")
```
brianhenn, line 520: What's the reason for making the rundir `subdir`?

brianhenn (approved):
LGTM thanks

workflows/prognostic_c48_run/tests/test_regression.py
nbren12: createRun complains about the folder existing if I use an already existing directory.

PR #1301, opened 2021-07-20
fv3fit: move stacking transform into single column models' .fit method

AnnaKwa:
This PR moves the stacking of training data for single-column models out of the `loaders` batch functions and into the models' `.fit` method. This will allow future models to ingest data in whatever format they need. In the more immediate future it lets us get rid of the `training` boolean arg in the batch function that used to toggle the stacking transforms, which was getting in the way of fv3fit config refactors in the offline diags.

Models now take in unstacked data and do the stacking transform within their `.fit`. A future PR will also change `.predict` methods to take in and return unstacked data, but for now `.predict` expects stacked data.

Added public API:
- Deleted `training` and `subsample_size` args from the `loaders.batches_from_x` functions. The subsample arg assumed the now-deleted stacking transform had been done on the data. It was formerly used by Andre for emulation data but he said it was ok to remove.

Tests modified to use unstacked data as `.fit` input. They still use stacked data in `.predict` tests.


external/fv3fit/fv3fit/_shared/utils.py
```diff
@@ -8,3 +18,99 @@ def parse_data_path(data_path: Union[List, str]):
...
+        return len(self._batches)
+
+    def _stack_batch(self, ds_unstacked: xr.Dataset) -> xr.Dataset:
+        ds = stack_non_vertical(ds_unstacked).load().dropna(dim=SAMPLE_DIM_NAME)
+        ds = _check_empty(ds)
+        ds = _preserve_samples_per_batch(ds)
+        return _shuffled(self._random_state, ds)
+
+
+def stack_non_vertical(ds: xr.Dataset,) -> xr.Dataset:
```
AnnaKwa, line 41: These are copied directly from their previous location in `.loaders`

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -216,9 +216,9 @@ def fit(
         fit_kwargs = _fill_default(
             fit_kwargs, use_last_batch_to_validate, "use_last_batch_to_validate", False
         )
-
-        Xy = _XyArraySequence(self.X_packer, self.y_packer, batches)
-
+        random_state = np.random.RandomState(np.random.get_state()[1][0])
```
AnnaKwa, line 219: Now that the shuffle method is inside `.fit` it doesn't have easy access to the `TrainingConfig.random_seed`. My solution here was to assume the random state at this point in time will be reproducible if the starting random state that is set at the beginning of `train.py` is the same. LMK if that is not the case. The reproducibility tests seem ok.

external/fv3fit/tests/test_utils.py
```diff
@@ -0,0 +1,111 @@
...
+from fv3fit._shared.utils import (
+    _shuffled,
+    _get_chunk_indices,
+    _check_empty,
+    stack_non_vertical,
+    _preserve_samples_per_batch,
+)
+
+
+@pytest.fixture
```
AnnaKwa, line 13: Nothing new here, copied from `loaders`.

external/loaders/loaders/_utils.py
```diff
@@ -81,7 +79,7 @@ def add_wind_rotation_info(res: str, ds: xr.Dataset) -> xr.Dataset:
         res: grid resolution, format as f'c{number cells in tile}'
     """
 
-    rotation = _load_wind_rotation_matrix(res)
+    rotation = _load_wind_rotation_matrix(res).drop("tile")
```
AnnaKwa, line 82: Unrelated to this PR, but fixes the bug that appears if you merge datasets that disagree on starting tile index.

external/fv3fit/fv3fit/keras/_models/models.py
AnnaKwa: > assume the random state at this point in time will be reproducible if the starting random state that is set at the beginning of train.py is the same.

Looks like this is not true, since the reproducibility regtests that passed locally failed on CircleCI. I think the random state used here needs to also be directly set with the same seed to keep the same behavior. I think we could remove the `TrainingConfig.random_seed` and instead make it part of the models' `Hyperparameters` so that it's available to `.fit` if it needs to set a random state, and call `fv3fit.set_random_seed(random_seed)` in the model training function instead of `train.py`. Let me know if you see any issues with that.

AnnaKwa:
Copying this comment here because the original is in an outdated part of the code and might not show up in the file view.

Now that the shuffle method is inside .fit it doesn't have easy access to the TrainingConfig.random_seed. My solution in this PR is to remove the TrainingConfig.random_seed and instead make it part of the relevant models' Hyperparameters so that it's available to .fit if it needs to set a random state, and call fv3fit.set_random_seed(random_seed) in the model training function instead of train.py. I did this in commit c2984c1. Let me know if you see any issues with that.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -81,7 +80,6 @@ class TrainingConfig:
     hyperparameters: Dataclass
     additional_variables: List[str] = dataclasses.field(default_factory=list)
     sample_dim_name: str = "sample"
-    random_seed: Union[float, int] = 0
```
mcgibbon, line 84: Why is this moved? The logic to set random seeds should be transferrable across all training codes, and is done in `fv3fit.train`. sklearn has an additional way of setting random seeds, but I think even if we didn't set it the top-level logic would make it deterministic.

Speaking of which, separate from this PR we could probably use a test that runs `fv3fit.train` twice and tests that the answer is the same.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -42,6 +46,7 @@ def train_dense_model(
     train_batches: Sequence[xr.Dataset],
     validation_batches: Sequence[xr.Dataset],
 ):
+    set_random_seed(hyperparameters.random_seed)
```
mcgibbon, line 49: This shouldn't be needed since it's already done in `fv3fit.train`?

external/fv3fit/tests/test_models.py
```diff
@@ -66,11 +67,11 @@ def test_nonnegative_model_outputs():
...
-            "input": (["sample"], np.arange(100)),
+            "input": (["x"], np.arange(100)),
             # even with negative targets, trained model should be nonnegative
-            "output": (["sample"], np.full((100,), -1e4)),
+            "output": (["x"], np.full((100,), -1e4)),
         }
     )
     model.fit([batch])
-    prediction = model.predict(batch)
+    prediction = model.predict(stack_non_vertical(batch))
```
mcgibbon, line 76: You should be able to revert the changes in this test and it should still pass. Alternatively, shouldn't `predict` take in an un-stacked dataset at least in one test?

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -100,11 +101,11 @@ def _get_sklearn_wrapper(scale_factor=None, dumps_returns: bytes = b"HEY!"):
...
 def test_SklearnWrapper_fit_predict_scaler(scale=2.0):
     wrapper = _get_sklearn_wrapper(scale)
-    dims = ["sample", "z"]
+    dims = ["unstacked_dim", "z"]
     data = xr.Dataset({"x": (dims, np.ones((1, 1))), "y": (dims, np.ones((1, 1)))})
     wrapper.fit([data])
-
-    output = wrapper.predict(data)
+    stacked_data = stack_non_vertical(data)
+    output = wrapper.predict(stacked_data)
```
mcgibbon, line 108: Here as well, do we need to test predicting on unstacked data?

external/fv3fit/fv3fit/_shared/utils.py
```diff
@@ -8,3 +18,107 @@ def parse_data_path(data_path: Union[List, str]):
         return data_path[0]
     else:
         return data_path
+
+
+class StackedBatches(Sequence[xr.Dataset]):
```
mcgibbon, line 23: Please put these in a new file called `stacking.py` or something similar, to avoid putting it in a `utils` file.

external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out
```diff
@@ -1,3 +1,3 @@
 ('var_in', '98a134e67c4a2b9a2da5d9744988191f')
 ('var_out', '98a134e67c4a2b9a2da5d9744988191f')
-('var_out', 'f92a68c6d69eb995dbf2375d3de9bc9f')
+('var_out', '848a1d247289e7ef7e698fdf4622a9c4')
```
Comment (mcgibbon) [inline external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out:3]: Do you have some idea of why this changed? It's OK for it to change, but ideally we know why.

external/fv3fit/tests/training/test_train.py
```diff
@@ -59,7 +60,7 @@ def assert_can_learn_identity(
     model, test_dataset = train_identity_model(
         model_type, sample_func=sample_func, hyperparameters=hyperparameters
     )
-    out_dataset = model.predict(test_dataset)
+    out_dataset = model.predict(stack_non_vertical(test_dataset)).unstack("sample")
```
mcgibbon, line 63: Similar question about predict, it seems like we should test two modes - stacked to stacked, and unstacked to unstacked. Thorough tests only need to be done for unstacked to unstacked.

external/loaders/loaders/batches/_batch.py
```diff
@@ -66,8 +60,6 @@ def batches_from_geodata(
         timesteps_per_batch (int, optional): Defaults to 1.
         random_seed (int, optional): Defaults to 0.
         res: grid resolution, format as f'c{number cells in tile}'
-        subsample_size: draw a random subsample from the batch of the
-            specified size along the sampling dimension
```
mcgibbon, line 70: Is it necessary to transfer this feature into the training code, or is this feature unused?

external/loaders/tests/test__batch.py
```diff
@@ -68,14 +68,13 @@ def test_batches_from_mapper(mapper):
...
+    original_dims = set(batched_data_sequence[0].dims)
     assert len(batched_data_sequence) == 2
-    expected_num_samples = 6 * 48 * 48 * 2
     for i, batch in enumerate(batched_data_sequence):
         assert len(batch["z"]) == Z_DIM_SIZE
         assert set(batch.data_vars) == set(DATA_VARS)
         for name in batch.data_vars.keys():
-            assert batch[name].dims[0] == loaders.SAMPLE_DIM_NAME
-            assert batch[name].sizes[loaders.SAMPLE_DIM_NAME] == expected_num_samples
+            assert set(batch[name].dims) == original_dims
```
mcgibbon, line 77: Can you add some kind of assert about dimension lengths?

mcgibbon (commented):
Looking good, main question is around tests for `predict` using stacked data. A couple other minor points.

mcgibbon:
Ignore my comments/questions about testing predict on stacked data, I misunderstood the APIs and the future-PR change.

external/loaders/loaders/batches/_batch.py
AnnaKwa: This is unused, I checked with Andre who was the only past user and he said it was fine to remove.

external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out
Comment (AnnaKwa) [reply external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out]: This is outdated, previously the regtest output changed because I had a commit that set the random states in the training function rather than at the top level in train.py.

external/fv3fit/fv3fit/keras/_models/models.py
mcgibbon: Looks like you've reverted this now.

external/fv3fit/tests/_regtest_outputs/test_sklearn.test_random_state_uniform_reproducibility.out
```diff
@@ -1,2 +1,2 @@
-1.19.4
+1.19.5
```
mcgibbon, line 1: This says 1.19.5 but the test_numpy_version says 1.19.4 - what's up?

external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out
```diff
@@ -1,3 +0,0 @@
-('var_in', '98a134e67c4a2b9a2da5d9744988191f')
```
Comment (mcgibbon) [inline external/fv3fit/tests/training/_regtest_outputs/test_train.test_train_default_model_on_identity[sklearn_random_forest].out:1]: You probably already have, but make sure to delete all of these whose tests have been deleted.

mcgibbon (approved):
LGTM. Thanks! This will help make the data configuration really simple across our tools.

external/fv3fit/tests/_regtest_outputs/test_sklearn.test_random_state_uniform_reproducibility.out
AnnaKwa: Oh, this must be old from before I moved the numpy version test out of this test function. I'll delete.

PR #1303, opened 2021-07-22
Prog report: do all computations in parallel

AnnaKwa:
This PR adds a step to combine all computations across diags registries into a flat list before parallelizing over them. This increases the parallelism for the cases iterating over individual registries' computations didn't fully utilize all the available cores, which could be somewhat of a slow down for ~year-long prognostic runs.

Added public API:
- made `Registry.funcs` and `Registry.load(...)` public so that computations for multiple registries can be combined in parallel


oliverwm1 (approved):
I'm approving, but it would be cool to have a solution that doesn't require special handling of the registries/duplicating the Parallel call. One option would be splitting up the `.load()`s that operate on each `Dataset` and having them operate on a sequence of `DataArray`s instead. I think it would have a similar effect of increasing CPU usage although you might still get some (now much shorter) periods of low usage.

Anyway up to you if you think it's worth implementing. Would need to handle the two cases of the `func` outputs being either `Dataset` (split up computation) or `DataArray` (don't).

PR #1310, opened 2021-07-27
Update e2e test for new model training configs

oliverwm1:
PR on top of #1304


.circleci/config.yml
```diff
@@ -147,8 +147,7 @@ jobs:
           name: Install dependencies
           command: |
             apt-get install -y wget jq make gettext gzip
-            wget -qO /bin/yq https://github.com/mikefarah/yq/releases/download/3.2.1/yq_linux_386
-            chmod +x /bin/yq
+            pip3 install yq
```
mcgibbon, line 150: Do we need/want a version pin on this?

tests/end_to_end_integration/.gitignore
```diff
@@ -1,2 +1,4 @@
 config.yaml
 kustomization.yaml
+training-data-config-compiled.yaml
+validation-data-config-compiled.yaml
```
mcgibbon, line 4: nit: We normally name our files with underscores and argo variables with dashes. You may want to refactor these to use underscores.

mcgibbon (approved):
LGTM. Minor nits to consider.

tests/end_to_end_integration/.gitignore
oliverwm1: for better or worse, that has not been the standard practice for config/argo yamls. E.g. see the workflows/argo directory or the various config yamls in vcm-workflow-control. I'm going to leave this as is.

.circleci/config.yml
oliverwm1: Okay I pinned it at a recent version.

PR #1311, opened 2021-07-28
Fix emulation function "get_nc_files"

frodre:
There was a bug i noticed when using `nc_dir_to_tf_dataset` that stems from a problem with `get_nc_files`.  

If provided with a path to a GCS directory of netCDF paths, the `get_nc_files` function used the `glob` method of a filesystem object, which would strip the protocol (`gs://`) from the list of files.  The removal of this information breaks the loading of the remote netCDF downstream because we a) don't provide initiated filesystem objects (gets messy fast) and b) without any protocol info in the path `get_remote_nc` thinks the file is local.  I adjusted `get_nc_files` to keep the path if it detects a GCS protocol for now.

- [x] Test added

### DIFF
--- a/external/fv3fit/tests/emulation/test_data_io.py
+++ b/external/fv3fit/tests/emulation/test_data_io.py
@@ -48,7 +48,6 @@ class MockGCSFilesystem:
         ]
 
 
-@pytest.mark.network
 def test_get_nc_files_remote_protocol_prepend():
 
     fs = MockGCSFilesystem()
--- a/external/fv3fit/tests/emulation/test_data_io.py
+++ b/external/fv3fit/tests/emulation/test_data_io.py
@@ -1,4 +1,3 @@
-import pytest
 import os
 import tempfile
 import numpy as np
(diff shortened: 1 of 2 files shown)


oliverwm1 (commented):
Would be great to come up with a general purpose solution to this issue, as I've always struggled with it in various places. But I think this looks fine for here!

Was the fv3gfs-fortran submodule update intentional? looks like it broke prognostic run tests.

external/fv3fit/tests/emulation/test_data_io.py
```diff
@@ -19,13 +20,40 @@ def test_get_nc_files():
...
+    protocol = ("gs", "gcs")
+
+    def glob(*args):
+        return [
+            "fake-bucket/file1.nc",
+            "fake-bucket/file2.nc",
+        ]
+
+
+@pytest.mark.network
```
oliverwm1, line 51: since you've mocked the filesystem, does this really need network access?

frodre: Nope! I forgot to delete this after I switched to a mock.

frodre:
I must've branched this fix off of the wrong branch on my end.  I'll see if I can fix that

oliverwm1 (approved):
LGTM!

PR #1314, opened 2021-07-29
fv3fit: change .predict methods to use unstacked data

AnnaKwa:
We are moving away from requiring fv3fit models to take in stacked column data. This PR follows up on https://github.com/VulcanClimateModeling/fv3net/pull/1301 , which changed `.fit` methods to take in unstacked data. Changes here move the stacking into the `.predict` methods and unstack before returning the prediction dataset. Each model's `.predict` method now takes in and returns unstacked data (effectively I have just relocated the transformations and dimension checks from `predict_columnwise` in to the individual models' `predict`).

This change deprecates`Predictor.predict_columnwise`, as this is now done by `.predict`. For now I have left the `predict_columnwise` method but it simply calls `.predict`. In a followup PR I will replace usages of `predict_columnwise` in other workflows with `.predict` and then delete this method. I kept those changes out of this PR because there are a lot of files that would make the diff big. I did delete a couple tests of `predict_columnwise` in the tests if they duplicated tests of `predict` since those files were already being updated for this PR.

There are a lot of changed files but most of them are tests that have been amended to have calls of `.predict` use unstacked data. 

This PR adds an internal hard coded `_fv3fit_sample_dim` used in stacking data, which is introduced to avoid stack name conflicts when allowing for input data which is already stacked using the widely-used `sample` dimension name. A future PR that hard codes the sample dim name and removes it from training function args will remove this attribute.

Refactored public API:
- `Predictor.predict` methods changed to take in and return unstacked datasets
- `vcm.safe.stack_once` changed to allow data arrays to be stacked. In fv3fit usage it is always datasets being stacked, but I ran into an issue in a test where `stack_once` couldn't be used on a data array and it seemed like these functions ought to be able to be used on either.


external/fv3fit/fv3fit/_shared/stacking.py
```diff
@@ -113,3 +113,27 @@ def _get_chunk_indices(chunks):
         indices.append(list(range(start, start + chunk)))
         start += chunk
     return indices
+
+
+def infer_dimension_order(ds: xr.Dataset) -> Tuple:
```
AnnaKwa, line 118: These are copied from their previous location in `predictor.py` and made public for use in models' `.predict` methods.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -308,11 +316,23 @@ def _fit_loop(
...
+            self.predict_array(self.X_packer.to_array(inputs_stacked))
         )
-        return ds_pred.assign_coords({self.sample_dim_name: sample_coord})
+        ds_pred = ds_pred_stacked.assign_coords(
+            {self.sample_dim_name: sample_coord}
+        ).unstack(self.sample_dim_name)
+        ds_pred = match_prediction_to_input_coords(X, ds_pred)
+        dim_order = [
+            dim for dim in infer_dimension_order(inputs_) if dim in ds_pred.dims
+        ]
```
mcgibbon, line 334: It feels like there are a lot of low level details here. Is it possible to split this into three parts? Stacking the sample dim, prediction on stacked data, and unstacking the sample dim? Ideally these would each be function/method calls that self-document what they're doing, instead of needing the 3-line comment above.

If it's possible to put the stacking/unstacking into a helper function that's used the same by _random_forest.py, testing.py, and here, that would be ideal.

external/vcm/vcm/safe.py
```diff
@@ -30,7 +31,7 @@ def _validate_stack_dims(ds, dims, allowed_broadcast_dims=()):
 
 
 def stack_once(
-    ds: xr.Dataset,
+    ds: Union[xr.DataArray, xr.Dataset],
```
mcgibbon, line 34: This change may cause a lot of mypy headaches, because the return type is now `Union[xr.DataArray, xr.Dataset]`. Can you use [generic typing](https://docs.python.org/3/library/typing.html#typing.Generic) to say the return type is the type of `ds`? I think you'll specifically want to use `T = TypeVar('T', xr.DataArray, xr.Dataset)`

mcgibbon (commented):
I did not look in detail yet at the refactored tests, but I had the general comment to make sure the two use cases are appropriately covered by the tests. One is where predict is given and returns stacked data (probably with "sample" as the dim name), and the other is where predict is given and returns unstacked data. This should mostly work because a stacked dataset is a special case of an unstacked dataset where there is one unstacked dimension. As we discussed, the code probably breaks right now when given data with a "sample" dimension.

external/fv3fit/fv3fit/keras/_models/models.py
AnnaKwa: I moved the "predict on stacked data" parts that were unique to each model into their own method, which is called between the common 1st/3rd parts in `.predict`. It seemed more clear to leave most of the 1st/3rd parts outside of functions but lmk if you disagree. I consolidated the dimension ordering into the `match_prediction_to_input_coords` helper function to tidy up the last part.

external/fv3fit/fv3fit/_shared/predictor.py
```diff
@@ -75,53 +78,4 @@ def predict_columnwise(
         Returns:
             the predictions defined on the same dimensions as X
         """
-        if len(sample_dims) == 0 and feature_dim is None:
```
mcgibbon, line 78: You could consider emitting a [deprecation warning](https://docs.python.org/3/library/warnings.html) here.

external/fv3fit/fv3fit/_shared/stacking.py
```diff
@@ -113,3 +132,28 @@ def _get_chunk_indices(chunks):
...
+    # add check here for cases when the dimension order is inconsistent between arrays?
+    dim_order = []
+    for variable in ds:
+        for dim in ds[variable].dims:
+            if dim not in dim_order:
+                dim_order.append(dim)
+    return tuple(dim_order)
+
+
+def match_prediction_to_input_coords(
```
mcgibbon, line 147: Nice!

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -307,15 +318,25 @@ def _fit_loop(
...
+        stacked_output_array = self.model.predict(stacked_input_array)
+        return self.y_packer.to_dataset(stacked_output_array)
+
     def predict(self, X: xr.Dataset) -> xr.Dataset:
-        sample_coord = X[self.sample_dim_name]
-        ds_pred = self.y_packer.to_dataset(
-            self.predict_array(self.X_packer.to_array(X))
+        # Takes unstacked data, stacks into sample dimension before
+        # keras model prediction, and returns unstacked prediction
+        # ensure dimension order is the same
```
mcgibbon, line 329:
```suggestion
```
Your code does a great job of communicating this process.

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -236,29 +243,40 @@ def _init_target_scaler(self, batch):
...
+        y = self.model.predict(X)
         if self.target_scaler is not None:
             y = self.target_scaler.denormalize(y)
         else:
             raise ValueError("Target scaler not present.")
+        return unpack(y, self._fv3fit_sample_dim, self.output_features_)
+
+    def predict(self, data):
+        # Takes unstacked data, stacks into sample dimension before
+        # sklearn model prediction, and returns unstacked prediction
```
mcgibbon, line 269:
```suggestion
```
This code is also clear.

external/fv3fit/fv3fit/testing.py
```diff
@@ -67,13 +73,14 @@ def predict(self, X: xr.Dataset) -> xr.Dataset:
...
-        if self.sample_dim_name in X.coords:
-            coords: Optional[Mapping[Hashable, Any]] = {
-                self.sample_dim_name: X.coords[self.sample_dim_name]
-            }
-        else:
-            coords = None
-        return xr.Dataset(data_vars=data_vars, coords=coords)
+        coords: Optional[Mapping[Hashable, Any]] = {
+            self.sample_dim_name: stacked_X.coords[self.sample_dim_name]
+        }
```
mcgibbon, line 78: There's nothing here to guarantee that self.sample_dim_name is in stacked_X.coords, but this should be fixed if you use a global SAMPLE_DIM_NAME here that is also used by `stack_non_vertical`.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -99,13 +106,15 @@ def __init__(
         self._spectral_normalization = hyperparameters.spectral_normalization
         self._gaussian_noise = hyperparameters.gaussian_noise
         self._nonnegative_outputs = hyperparameters.nonnegative_outputs
+        # TODO: remove internal sample dim name once sample dim is hardcoded everywhere
+        self._fv3fit_sample_dim = f"_sample_fv3fit"
```
mcgibbon, line 110: Instead of changing SAMPLE_DIM_NAME to be an instance variable, you should be able to update SAMPLE_DIM_NAME to be "_sample_fv3fit".

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -211,8 +204,24 @@ def test_predict_columnwise_is_deterministic(regtest):
...
+    If this fails, look for non-deterministic logic (e.g. converting sets to lists)
+    """
+    input_data, wrapper = fit_wrapper_with_columnar_data()
+    output = wrapper.predict(input_data)
     print(joblib.hash(np.asarray(output["b"])), file=regtest)
+
+
+def test_predict_returns_unstacked_dims():
+    # 2D output dims same as input dims
+    input_data, wrapper = fit_wrapper_with_columnar_data()
```
mcgibbon, line 224: In case fit_wrapper_with_columnar_data gets updated?
```suggestion
    input_data, wrapper = fit_wrapper_with_columnar_data()
    assert len(input_data.dims) > 2
```

mcgibbon (commented):
Much more readable with your refactors, thanks! Comments are mainly on new changes.

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -205,24 +212,22 @@ def __init__(
...
-        self._sample_dim_name = sample_dim_name
-        self._input_variables = input_variables
-        self._output_variables = output_variables
         self.model = model
-
         self.scaler_type = scaler_type
         self.scaler_kwargs = scaler_kwargs or {}
         self.target_scaler: Optional[scaler.NormalizeTransform] = None
+        # TODO: remove internal sample dim name once sample dim is hardcoded everywhere
+        self._sample_dim_name = sample_dim_name
```
mcgibbon, line 220: I think you can remove this `self._sample_dim_name` variable now, and remove it from the arguments in your later PR which will change the config.

mcgibbon (approved):
LGTM! A couple minor points remaining, but the PR is good even without them.

PR #1315, opened 2021-07-30
remove out of date fine resolution loader

nbren12:
this loader is no longer used, and is largely replaced by the new functionality in 4b20ebf. To compute Q1 and Q2 use `loaders.mappers._hybrid.{apparent_heating,apparent_moistening}`.

Refactored public API:
- deleted fine resolution loaders


brianhenn (approved):
LGTM thanks

PR #1316, opened 2021-07-30
Change units of column ML momentum tendencies to Pa

oliverwm1:
For column-integrated budgets of momentum, it is useful to have the column-integrated version of dQu and dQv in units of Pa. For example, this way they are directly comparable to the surface momentum flux output by the physics diagnostics. Also makes online diagnostics consistent with the offline diagnostics (which use units of Pa).

Significant internal changes:
- use `vcm.mass_integrate` instead of `_mass_average`
- modify prognostic run diagnostics so that loaded data will always be in units of `Pa` (even with old runs)

- [x] Tests added


workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/derived_variables.py
```diff
@@ -131,28 +131,44 @@ def _column_dq2(ds: xr.Dataset) -> xr.DataArray:
...
     if "column_integrated_dQu" in ds:
-        column_dqu = SECONDS_PER_DAY * ds.column_integrated_dQu
+        warnings.warn(
+            "'column_integrated_dQu' is a deprecated variable name. "
+            "It will not be supported in future versions of fv3net. Use "
+            "'column_integrated_dQu_stress' (units of Pa) instead.",
+            DeprecationWarning,
+        )
+        # convert from m/s/s to Pa by multiplying by surface pressure divided by gravity
+        column_dqu = 100000 / 9.8065 * ds.column_integrated_dQu
```
oliverwm1, line 141: need to use a reference surface pressure here, since don't have the actual (lat-lon dependent) surface pressure available. But since we are deprecating this version of the diagnostic, I think it's okay.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/derived_variables.py
```diff
@@ -131,28 +131,44 @@ def _column_dq2(ds: xr.Dataset) -> xr.DataArray:
...
         # assume given dataset has no ML prediction of momentum tendencies
         column_dqu = xr.zeros_like(ds.PRATEsfc)
     column_dqu.attrs = {
-        "long_name": "<dQu> vertical mean eastward wind tendency from ML",
-        "units": "m/s/day",
+        "long_name": "<dQu> column integrated eastward wind tendency from ML",
+        "units": "Pa",
     }
-    return column_dqu.rename("vertical_mean_dQu")
+    return column_dqu.rename("column_int_dQu")
```
oliverwm1, line 151: need to use a name different than `column_integrated_dQu` to distinguish from the previous diagnostic that had this name.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/derived_variables.py
brianhenn: I guess over Tibet this might be a little weird but yeah not sure how to do that without a lot of work to get the varying pressure in here

brianhenn (approved):
Thanks looks good

Issue #1318, opened 2021-08-03
Prognostic run doesn't output all requested outputs


nbren12:
The prognostic run will not fail as expected when an unknown variable is requested AND `kind: interval` is used. For example, this diagnostic configuration will not result in a crash:
```
            {
                "name": "diags_3d.zarr",
                "times": {"kind": "interval", "frequency": 900},
                "chunks": {"time": 12},
                "variables": ["not an available input variable"],
            }
```

This happens because of this line of code:

https://github.com/VulcanClimateModeling/fv3net/blob/1861596078da06665862aa05080b58bc1e5be359/workflows/prognostic_c48_run/runtime/diagnostics/manager.py#L153

nbren12:
Fixing this bug breaks some of the tests. The existing behavior of `All()` is not compatible with error checking.

PR #1319, opened 2021-08-03
prog run: raise error for undefined diagnostics

nbren12:
Previously if the user requested saving a diagnostic field not computed
    by the model, the model execution would continue. The test suite also
    used the no-longer supported `All` variable selection configuration.
    
    Raising an exception will provide the user quicker feedback when they
    misconfigure a run or if there is a bug.


nbren12:
resolves #1318

oliverwm1 (approved):
LGTM

PR #1321, opened 2021-08-04
Refactor monitoring to separate class

nbren12:
The diagnostics in the prognsostic run are confusing and TimeLoop is a very complicated class.

Enable future refactors/improvements by splitting the monitoring capabilities into a new object with unit tests.


workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,78 @@
...
+)
+import vcm
+from runtime.types import Diagnostics, State
+
+from .names import DELP
+
+logger = logging.getLogger(__name__)
+
+
+class Monitor:
```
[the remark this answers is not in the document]
mcgibbon: Realized the request here is really to move the docstrings that existed over to this class.

nit: I'd also suggest calling this a `MonitoringWrapper`, since it's still not clear that a "Monitor" wraps functions.

mcgibbon (commented):
Very positive change! Needs some docstrings. Looking forward to seeing this in.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -106,7 +98,9 @@ def _print(self, message: str):
             print(message)
 
 
-class TimeLoop(Iterable[Tuple[cftime.DatetimeJulian, Diagnostics]], LoggingMixin):
+class TimeLoop(
+    Iterable[Tuple[cftime.DatetimeJulian, Dict[str, xr.DataArray]]], LoggingMixin
```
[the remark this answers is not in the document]
nbren12: to satisfy our lord of mypy and the gospel of Hashables.

workflows/prognostic_c48_run/runtime/monitor.py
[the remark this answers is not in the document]
nbren12: I left the name the same. I might add a context manager to it as well, so it might not just be used as a decorator. This functionality has been called `monitor` for a while.

workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,107 @@
...
+
+        This will add the following diagnostics:
+        - `tendency_of_{variable}_due_to_{name}`
+        - `storage_of_{variable}_path_due_to_{name}`. A pressure-integrated version
+        of the above
+        - `storage_of_mass_due_to_{name}`, the total mass tendency in Pa/s.
+
+        Args:
+            name: the name to tag the tendency diagnostics with
+            func: a stepping function
```
mcgibbon, line 50:
```suggestion
            func: a stepping function which modifies the `state` this object is monitoring
```

workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,107 @@
...
+        self.timestep = timestep
+
+    def __call__(
+        self, name: str, func: Callable[[], Diagnostics],
+    ) -> Callable[[], Diagnostics]:
+        """Decorator to add tendency monitoring to an update function
+
+        This will add the following diagnostics:
+        - `tendency_of_{variable}_due_to_{name}`
+        - `storage_of_{variable}_path_due_to_{name}`. A pressure-integrated version
```
mcgibbon, line 44:
```suggestion
        - `storage_of_{variable}_path_due_to_{name}`. A mass-integrated version
```

workflows/prognostic_c48_run/runtime/monitor.py
mcgibbon: Given this tight coupling between `func` and `state`, should `state` be passed at call time instead of on init? Otherwise this class is still tightly coupled to the owning scope (you need to track the relation between the Monitor internal `state` instance and the scope it's being used in).

workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,107 @@
...
+                if "units" in before[variable].attrs:
+                    diags[diag_name].attrs["units"] = before[variable].units + "/s"
+
+            mass_change = (delp_after - delp_before).sum("z") / self.timestep
+            mass_change.attrs["units"] = "Pa/s"
+            diags[f"storage_of_mass_due_to_{name}"] = mass_change
+            return diags
+
+        # ensure monitored function has same name as original
+        step.__name__ = func.__name__
```
mcgibbon, line 89: pre-existing: `functools.wraps` handles copying over the docstring as well, and might be a better high-level function to use here.

workflows/prognostic_c48_run/tests/test_monitor.py
```diff
@@ -0,0 +1,15 @@
+import xarray
+from runtime.monitor import Monitor
+from runtime.names import DELP
+
+
+def test_Monitor_monitor():
```
mcgibbon, line 6:
```suggestion
def test_monitor_inserts_tendency_and_storage_of_one_var():
```

workflows/prognostic_c48_run/tests/test_monitor.py
```diff
@@ -0,0 +1,15 @@
...
+    ds = xarray.Dataset({"x": ([], 0.0), "y": (["z"], [1, 2]), DELP: (["z"], [1, 1])})
+    tend = "tendency_of_x_due_to_blah"
+    storage = "storage_of_y_path_due_to_blah"
+
+    monitor = Monitor([tend, storage], ds, timestep=900)
+    out = monitor("blah", lambda: {})()
```
mcgibbon, line 12: Input to monitor is supposed to be an `Iterable[str]`. If "blah" is iterated over, it will produce the strings "b", "l", "a", "h". Please update this test/example to use a list/set/other explicit iterable.

workflows/prognostic_c48_run/tests/test_monitor.py
```diff
@@ -0,0 +1,15 @@
...
+from runtime.monitor import Monitor
+from runtime.names import DELP
+
+
+def test_Monitor_monitor():
+    ds = xarray.Dataset({"x": ([], 0.0), "y": (["z"], [1, 2]), DELP: (["z"], [1, 1])})
+    tend = "tendency_of_x_due_to_blah"
+    storage = "storage_of_y_path_due_to_blah"
+
+    monitor = Monitor([tend, storage], ds, timestep=900)
```
mcgibbon, line 11: Is this expected to pass, given that "blah" does not exist in the dataset/state? I would think this should raise an exception. If this is expected, could you add it as a special test and make the default case use an existing variable? This would help avoid confusing someone reading the tests.

workflows/prognostic_c48_run/tests/test_monitor.py
mcgibbon: Non-blocking since this is an improvement with any level of testing, but you should consider testing these cases:

- monitoring multiple variables
- monitoring with an empty set of variables

Given the inputs are so simple, you could also add assertions on the values.

workflows/prognostic_c48_run/tests/test_monitor.py
nbren12: I'm not sure what you mean. This is the signature of __call__:

    def __call__(
        self, name: str, func: Callable[[], Diagnostics],
    ) -> Callable[[], Diagnostics]

workflows/prognostic_c48_run/tests/test_monitor.py
nbren12: Yes. The name "blah" is specified below.

workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,107 @@
...
+            diagnostics inserted in place
+        """
+
+        def step() -> Diagnostics:
+
+            vars_ = self.storage_variables | self.tendency_variables
+            delp_before = self._state[DELP]
+            before = {key: self._state[key] for key in vars_}
+            diags = func()
+            delp_after = self._state[DELP]
```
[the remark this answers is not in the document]
nbren12: I think it works. If it didn't the mass change variable would always be zero, and it is not.

workflows/prognostic_c48_run/runtime/monitor.py
nbren12: possibly, but that would be a backwards incompatible change to `TimeLoop.monitor`.  In practice, `state` is a singleton anyway since it just wraps `fv3gfs.get/set_state`.

workflows/prognostic_c48_run/tests/test_monitor.py
nbren12: > Non-blocking since this is an improvement with any level of testing

I'd like to punt on this if it's okay.

workflows/prognostic_c48_run/runtime/monitor.py
nbren12: There is no coverage of the clock timing code, so I don't want to risk breaking it.

workflows/prognostic_c48_run/tests/test_monitor.py
nbren12: Hopefully f86baf3 clarifies how this works. Let me know if it is still confusing.

workflows/prognostic_c48_run/runtime/monitor.py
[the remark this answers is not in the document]
mcgibbon: It works for our particular code, but it's unsafe. This only works if the state dictionary is updated. It doesn't work if func updates the array inside the state dictionary instead of updating the state dictionary with a new array.

workflows/prognostic_c48_run/runtime/monitor.py
mcgibbon: Should `monitor` be public API? `git grep monitor .` in fv3net root shows it's only used privately.

workflows/prognostic_c48_run/tests/test_monitor.py
mcgibbon: I meant `Monitor.__init__`.

workflows/prognostic_c48_run/tests/test_monitor.py
mcgibbon: I'm still a bit confused. How can Monitor monitor a variable "blah" that is not defined?

mcgibbon: Ah my confusion was in thinking "blah" is a variable when it's not. This is the same issue as the `__init__` confusion. Resolved.

workflows/prognostic_c48_run/tests/test_monitor.py
mcgibbon: OK by me. I would still ask you update the name so it's clear this is a limited test.

workflows/prognostic_c48_run/runtime/monitor.py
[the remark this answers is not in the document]
nbren12: I see your point, but this seems like a bigger issue about whether `DerivedFV3State.__getitem__` should exist and is not specific to this code. In practice, we treat `DataArray` as immutable objects in this code.

workflows/prognostic_c48_run/runtime/monitor.py
nbren12: The only "public API" provided by the prog run is `runfv3` IMO.

workflows/prognostic_c48_run/runtime/monitor.py
```diff
@@ -0,0 +1,119 @@
...
+
+        This will add the following diagnostics:
+        - `tendency_of_{variable}_due_to_{name}`
+        - `storage_of_{variable}_path_due_to_{name}`. A mass-integrated version
+        of the above
+        - `storage_of_mass_due_to_{name}`, the total mass tendency in Pa/s.
+
+        Args:
+            name: the name to tag the tendency diagnostics with
+            func: a stepping function
```
mcgibbon, line 40:
```suggestion
            func: a stepping function which modifies the `state` dictionary this object
                is monitoring, but does not directly modify the `DataArray` objects
                it contains
```

workflows/prognostic_c48_run/runtime/monitor.py
mcgibbon: Then it should be fine to make a backwards-incompatible change to an internal function?

mcgibbon (approved):
Thanks!

workflows/prognostic_c48_run/runtime/monitor.py
nbren12: For sure...if we think that change is helpful and in-scope.

PR #1323, opened 2021-08-04
Simplify hybrid mapper function

oliverwm1:
Since #1304, we no longer need to include a placeholder positional argument in `open_fine_resolution_nudging_hybrid`. This PR removes that argument to simplify configuration. It also fixes a bug in `XarrayMapper` in which the provided `time` argument was not used within the mapper.

- [x] Tests added


nbren12 (approved):
Great! Thanks for fixing the bug.

PR #1324, opened 2021-08-05
Prognostic run: stream logs

AnnaKwa:
This resolves issue https://github.com/VulcanClimateModeling/fv3net/issues/1322, where the prognostic run stopped streaming logs. It also puts the `-m mpi4py` back into the subprocess command to prevent jobs from hanging without raising errors.

(Delete unused sections)

Resolves #<https://github.com/VulcanClimateModeling/fv3net/issues/1322>


workflows/prognostic_c48_run/runtime/segmented_run/run.py
```diff
@@ -33,17 +33,22 @@ def run_segment(config: dict, rundir: str):
...
+                    "-m",
+                    "mpi4py",
                     runfile.absolute().as_posix(),
                 ],
-                stdout=f,
-                stderr=f,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
             )
+            for c in iter(lambda: process.stdout.read(1), b""):
```
nbren12, line 49: Seems like it could be inefficient to read one byte at a time.
```suggestion
            for c in iter(lambda: process.stdout.read(32), b""):
```

workflows/prognostic_c48_run/runtime/segmented_run/run.py
```diff
@@ -33,17 +33,22 @@ def run_segment(config: dict, rundir: str):
...
+                sys.stdout.buffer.write(c)
+                f.buffer.write(c)
```
nbren12, line 51: Was there some reason you used "buffer"?
```suggestion
                sys.stdout.write(c)
                f.write(c)
```

nbren12 (approved):
LGTM, just a couple nits.

workflows/prognostic_c48_run/runtime/segmented_run/run.py
AnnaKwa: No particular reason, other than that the example I was working off of used it.

AnnaKwa: I guess the example had it because `stdout.write(), f.write()` take str args, so you can either decode then use those, or write to the buffer.

PR #1326, opened 2021-08-05
Increase default memory request for offline diags workflow

oliverwm1:
The offline diags fail with the default memory request (6Gi) even with a 'standard' ML model like a RF which predicts dQ1 and dQ2 and was trained on 130 timesteps worth of samples. These often leads to failures partway through the long-running train-diags-prog-run workflow. This PR increases the default request to 10Gi.


AnnaKwa (approved):
LGTM

PR #1329, opened 2021-08-09
Use consistent colorbar limits in prognostic run report

oliverwm1:
Even though we make separate plots for each run on the `maps.html` and `hovmoller.html` pages, it is helpful for all of the plots for a given variable to use a consistent colorbar. This makes comparing between runs a lot easier. This PR provides a public API to some fv3viz functionality that can determine useful cmap params and uses it within the prog run report.

New public API:
- `fv3viz.infer_cmap_params`

Significant internal changes:
- consistent colorbar limits across all runs
- adjust manual contour levels for zonal mean bias plots so that they are symmetric around 0

- [ ] Tests added

### DIFF
--- a/external/fv3viz/tests/test_plot_helpers.py
+++ b/external/fv3viz/tests/test_plot_helpers.py
@@ -33,6 +33,6 @@ def test__get_var_label(attrs, var_name, expected_label):
         (np.array([-0.5, 0, 1]), {"vmin": -0.6}, (-0.6, 0.6, "RdBu_r")),
     ],
 )
-def test_auto_limits_cmap(data, args, expected_result):
-    result = fv3viz.auto_limits_cmap(data, **args)
+def test_infer_cmap_params(data, args, expected_result):
+    result = fv3viz.infer_cmap_params(data, **args)
     assert result == expected_result
--- a/external/fv3viz/tests/test_plot_helpers.py
+++ b/external/fv3viz/tests/test_plot_helpers.py
@@ -1,0 +1,38 @@
+import numpy as np
+import pytest
+
+import fv3viz
+from fv3viz._plot_helpers import _get_var_label
+
+
+@pytest.mark.parametrize(
+    "attrs,var_name,expected_label",
+    [
+        ({}, "temp", "temp"),
+        ({"long_name": "air_temperature"}, "temp", "air_temperature"),
+        ({"units": "degK"}, "temp", "temp [degK]"),
+        (
+            {"long_name": "air_temperature", "units": "degK"},
+            "temp",
+            "air_temperature [degK]",
+        ),
+    ],
+)
+def test__get_var_label(attrs, var_name, expected_label):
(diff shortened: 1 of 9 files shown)


external/fv3viz/tests/test_plot_helpers.py
```diff
@@ -0,0 +1,38 @@
...
+    assert _get_var_label(attrs, var_name) == expected_label
+
+
+@pytest.mark.parametrize(
+    "data, args, expected_result",
+    [
+        (np.array([0.0, 0.5, 1.0]), {}, (0.0, 1.0, "viridis")),
+        (np.array([-0.5, 0, 1]), {}, (-1.0, 1.0, "RdBu_r")),
+        (np.array([0.0, 0.5, 1.0]), {"robust": True}, (0.02, 0.98, "viridis")),
+        (np.array([0.0, 0.5, 1.0]), {"cmap": "jet"}, (0.0, 1.0, "jet")),
```
AnnaKwa, line 31: Jet, haha

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/matplotlib.py
```diff
@@ -64,21 +64,20 @@ def fig_to_b64(fig, format="png", dpi=None):
...
 )
 
-CBAR_RANGE = {
-    "eastward_wind_pressure_level_zonal_bias": 30,
-    "northward_wind_pressure_level_zonal_bias": 3,
-    "air_temperature_pressure_level_zonal_bias": 15,
-    "specific_humidity_pressure_level_zonal_bias": 1e-3,
-    "vertical_wind_pressure_level_zonal_bias": 0.02,
-    "mass_streamfunction_pressure_level_zonal_bias": 100,
+CONTOUR_LEVELS = {
```
AnnaKwa, line 67: Thanks, optional but now that the levels are all centered about zero it might be nice to add a test that all items in this dict have that property (since it may not be obvious to someone adding a new key that the aranges are centered like this).

brianhenn (approved):
Thanks @oliverwm1. I have an optional suggestion.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/views/matplotlib.py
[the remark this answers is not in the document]
oliverwm1: in practice it looks like the vmin/vmax that are inferred by `_get_cmap_kwargs` get ignored in favor of the hard-coded contour levels, but I'm not sure that's expected behavior. I'll do your suggestion of

> not using the results of _get_cmap_kwargs for the bias contourf plots

oliverwm1: I think we might not want to enforce this in case someone adds hard-coded contour levels for variables that are not biases (and so not necessarily centered at zero).

PR #1330, opened 2021-08-10
fv3fit: Fix PureKerasModel.predict to work on unstacked data

AnnaKwa:
The last PR that changed models' `.predict` methods to work on unstacked data forgot to also include the recurrent `PureKerasModel` in this fix.

The handling of how the output metadata gets used to restore coordinates is kind of convoluted; I think the solution that was proposed in discussion (saving an object with the Predictor that stores this info at fit time) would help clean this section up. I made an issue to remind myself not to forget: https://github.com/VulcanClimateModeling/fv3net/issues/1331

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/recurrent.py
+++ b/external/fv3fit/fv3fit/keras/_models/recurrent.py
@@ -170,7 +170,11 @@ class _BPTTTrainer:
         """
         if self.train_keras_model is not None:
             raise RuntimeError("cannot build, model is already built!")
-        X_tmp = _rename_sample_dim(X, self.sample_dim_name)
+        temporary_sample_dim = self.sample_dim_name + "_tmp"
+        if self.sample_dim_name in X.dims:
+            X_tmp = X.rename({self.sample_dim_name: temporary_sample_dim})
+        else:
+            X_tmp = X
         dataset_2d = stack_non_vertical(X_tmp)
         inputs = self.input_packer.to_array(dataset_2d)
         state = self.prognostic_packer.to_array(dataset_2d)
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -165,12 +170,8 @@ def fit_statistics(self, X: xr.Dataset):
...
         if self.train_keras_model is not None:
             raise RuntimeError("cannot build, model is already built!")
-        temporary_sample_dim = self.sample_dim_name + "_tmp"
-        if self.sample_dim_name in X.dims:
-            X_tmp = X.rename({self.sample_dim_name: temporary_sample_dim})
-        else:
-            X_tmp = X
-        dataset_2d = stack_non_vertical(X_tmp, self.sample_dim_name)
+        X_tmp = _rename_sample_dim(X, self.sample_dim_name)
+        dataset_2d = stack_non_vertical(X_tmp)
```
mcgibbon, line 174: Why was this changed?

external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -542,14 +543,26 @@ def load(cls, path: str) -> "PureKerasModel":
 
     def predict(self, X: xr.Dataset) -> xr.Dataset:
         """Predict an output xarray dataset from an input xarray dataset."""
-        inputs = [X[name].values for name in self.input_variables]
+        X_tmp = _rename_sample_dim(X, self.sample_dim_name)
```
mcgibbon, line 546: This line uses `self.sample_dim_name` but some other lines use `SAMPLE_DIM_NAME`. It doesn't matter too much which one you use, but please change them to be consistent. If you use `SAMPLE_DIM_NAME`, please remove `sample_dim_name` from the `__init__`.

mcgibbon (commented):
Tests should be added that the changed PureKerasModel code works with unstacked -> unstacked predictions.

external/fv3fit/fv3fit/keras/_models/recurrent.py
AnnaKwa: The same behavior was kept but moved to a function since I added some code that also used this. But that's now unnecessary after revisions, so I'm reverting it back to how it was before.

mcgibbon (approved):
LGTM. Thanks for the test!

PR #1332, opened 2021-08-10
Offline diags: use DerivedMapping for pQ, Q if needed

AnnaKwa:
This PR does some tidying up and bug fixing in the offline diags. It will be easiest to review commit by commit since each fix/change has a corresponding commit.

- Currently, the offline diags have a lot of lines dedicated to adding pQ and Q variables to the dataset. Now that we have `vcm.DerivedMapping` a lot of this can be cleaned up.

- This also removes the requirement of the dataset to have a "net precipitation" variable, which is another derived variable that in turn relies on dQ2 existing in the data. If dQ2 is not an output variable, the computation over positive/negative net precip domains will be skipped.

- `BatchesFromMapperConfig` will raise an error if the timesteps list contains duplicate entries

- Fix a bug in fv3fit.DerivedModel where it did not work if the input data contained the prediction variables (as could happen if it was used in the offline diags)

A few bugs from stuff that was lost in translation during the recent update to the offline diags are fixed
- path to look for transect data is fixed
- bug raised if no 3D variables were in dataset passed to `compute_metrics`

Added public API:
- `DerivedMapping` keys `Q1, Q2`

- [x] Tests added

ex. 
with derived Q1, 2 added: http://storage.googleapis.com/vcm-ml-public/annak/2021-08-18-offline-diags-test/rf-dq1-dq2/index.html
without " : http://storage.googleapis.com/vcm-ml-public/annak/2021-08-18-offline-diags-test/nn-sfc-fluxes/index.html


workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
```diff
@@ -143,49 +135,30 @@ def _compute_diurnal_cycle(ds: xr.Dataset) -> xr.Dataset:
...
-    metric_vars = copy(predicted_vars)
-    if "dQ1" in predicted_vars and "dQ2" in predicted_vars:
-        metric_vars += ["Q1", "Q2"]
 
     # for each batch...
     for i, ds in enumerate(batches):
         logger.info(f"Processing batch {i+1}/{len(batches)}")
 
-        ds = _fill_empty_dQ1_dQ2(ds)
+        # ds = _fill_empty_dQ1_dQ2(ds)
```
frodre, line 159: Remove if unnecessary

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
```diff
@@ -278,27 +251,26 @@ def transform(ds):
...
         return insert_prediction(ds_derived, ds_prediction)
 
     return transform
 
 
+def _derived_heating_moistening_model(model: fv3fit.Predictor) -> fv3fit.DerivedModel:
+    # if dQ1, dQ2 are predicted, add derived Q1, Q2 variables so that the
+    # apparent sources can be inserted
+    derived_Q_vars = []
+    if "dQ1" in model.input_variables:
```
frodre, line 267: should these be `output_variables` instead?

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
```diff
@@ -316,8 +288,9 @@ def main(args):
 
     logger.info("Opening ML model")
     model = fv3fit.load(args.model_path)
+    if "dQ1" in model.output_variables or "dQ2" in model.output_variables:
+        model = _derived_heating_moistening_model(model)
```
frodre, line 292: Since you check the existence in and outside of the function, you could instead use a helper (e.g., `_check_for_derivable_outputs`) that returns a list of variable names `Q1` and/or `Q2` or  an empty list. Then if that list isn't empty, adjust the model into `fv3fit.DerivedModel`.

frodre (approved):
Thanks for the cleanup, @AnnaKwa!  I have some minor suggestions but overall this looks good.

external/vcm/vcm/derived_mapping.py
```diff
@@ -175,3 +175,19 @@ def is_sea(self):
 def is_sea_ice(self):
     # one hot encoding for sea ice surface
     return xr.where(vcm.xarray_utils.isclose(self["land_sea_mask"], 2), 1.0, 0.0)
+
+
+@DerivedMapping.register("Q1", required_inputs=["dQ1", "pQ1"])
```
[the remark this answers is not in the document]
AnnaKwa: Good point. I also changed the policy and docstring to include required variables even in the not-well-defined case (e.g. e.g. `dQu` only needs `dQxwind, dQywind` if `dQu` is not already in the mapping, but `dQxwind, dQywind` should still be listed as `required_inputs`. This is because the usage of this registry is for when an output is explicitly requested as a derived output variable and thus it is assumed the variable does not already exist and needs to be derived.

workflows/offline_ml_diags/offline_ml_diags/_helpers.py
```diff
@@ -181,7 +181,8 @@ def open_diagnostics_outputs(
         ds_diags = xr.open_dataset(f).load()
     with fsspec.open(os.path.join(data_dir, diurnal_nc_name), "rb") as f:
         ds_diurnal = xr.open_dataset(f).load()
-    if vcm.get_fs(transect_nc_name).exists(transect_nc_name):
+    fs_path_transect_nc = os.path.join(data_dir, transect_nc_name).replace("gs://", "")
```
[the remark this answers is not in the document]
AnnaKwa: Whoops, I think I mixed up some other call with fs.exists... it works fine with the protocol. I'll fix this.

workflows/offline_ml_diags/offline_ml_diags/compute_diags.py
AnnaKwa: Good idea, it's cleaner now and easy to extend if more derived outputs need to get added in the future.

AnnaKwa:
Thanks @frodre  for approving the offline diags & DerivedMapping PR last week! 

Just as a heads up, regarding your comment about the required inputs for Q1/Q2 being just pQ1/pQ2, I did make the suggested change, but it led me down a rabbit hole of considering the possible cases where the underlying data does/doesn’t have one or more of Q/dQ/pQ’s. e.g. fine res does have Q1 and Q2 so doesn’t have to derive those, in the all-physics case the models only have dQ so if it is deriving Q it should treat pQ as zero. 

The complication there is that for the usage in DerivedModel  , the pQ’s need to be added to  additional_input_variables even if they do not exist in the data, because at the time the DerivedModel is constructed it does not yet have the input data to check if this variables already exists or not. My solution was to make pQ1, pQ2 derived variables which use the underlying data if present and return 3D datasets of zeroes if not present (basically the same way the pQ’s were dealt with in the old removed part of compute_diags.py. (commit https://github.com/VulcanClimateModeling/fv3net/pull/1332/commits/1896070b4853eec2a2438135cbad00dfcf84f641) LMK if you think there’s a more preferable way to deal with this.

But after adding the derived pQ variables, it's now possible for a derived variable (Q) to have a required input (pQ) that itself is derived and has its own required inputs (delp, which is used to get x/y/z dims to fill in with zeroes). Therefore commit https://github.com/VulcanClimateModeling/fv3net/pull/1332/commits/b604dc3a93472e0a0fc91f1098897a30e09b7fa6 updates the function that gets all required input variables for `derived_output_variables` to make it a recursive check. I also moved this function out of `fv3fit.DerivedModel` into `vcm.DerivedMapping`.

external/vcm/vcm/derived_mapping.py
```diff
@@ -58,6 +58,27 @@ def _data_arrays(self, keys: Iterable[Hashable]):
...
+        # own required inputs (e.g. pQ's)
+        def _recurse_find_deps(vars, deps):
+            vars_with_deps = [var for var in vars if var in cls.REQUIRED_INPUTS]
+            if len(vars_with_deps) == 0:
+                return
+            else:
+                for var in vars_with_deps:
+                    new_deps = cls.REQUIRED_INPUTS[var]
+                    deps += new_deps
+                _recurse_find_deps(new_deps, deps)
```
frodre, line 76: I think this might be a bug lying in wait.  `new_deps` is set by the last iteration in the loop, so the recursion would miss any new dependencies from previous iterations.

This also makes me think we could also add some circular dependency protection here.  I think if we check whether any of the `new_deps` already exist in the `vars_with_deps` would work.  Might be more informative than reaching the recursion limit error.

external/vcm/tests/test_derived_mapping.py
```diff
@@ -172,3 +172,24 @@ def test_is_land():
...
     np.testing.assert_array_almost_equal(derived_state["is_sea_ice"], [0.0, 0.0, 1.0])
+
+
+@pytest.mark.parametrize(
+    "dependency_map, derived_vars, reqs",
+    [
+        ({"c": ["d"]}, ["a"], []),
+        ({"a": ["b"], "c": ["d"]}, ["a"], ["b"]),
+        ({"a": ["b"], "c": ["d"]}, ["a", "c"], ["b", "d"]),
+        ({"a": ["b"], "b": ["c"], "c": ["d"]}, ["a"], ["b", "c", "d"]),
```
frodre, line 183: I think this would fail due to the `new_deps` bug.
`({"a": ["b"], "c": ["d"], "b": ["e"]}, ["a", "c"], ["b", "d", "e"]),`

frodre (commented):
Thanks @AnnaKwa for the update and explanation around design choices here!  I think the recursion option looks nice and clean.  You should look into the bug I mentioned and make necessary changes if it needs fixing, but I'm only commenting since I'll be out tomorrow!  Looks good.

external/vcm/tests/test_derived_mapping.py
AnnaKwa: Thanks, good catch- fixed now and added your test.

PR #1334, opened 2021-08-16
Add option to include temp nudging in fine-res Q1

oliverwm1:
For fine-res runs which include nudging of temperature, we want to be able to include this term in the apparent heating Q1. This PR adds a kwarg for this option.


brianhenn (approved):
LGTM thanks @oliverwm1

PR #1335, opened 2021-08-16
Use continuumio/miniconda3 for CirceCI image base

oliverwm1:
The image we use for `test_unit` and a couple other CI tests is based off an old continuumio image that has not been updated for a couple years. This PR updates the dockerfile which built our CI image to use a newer one, and updates the CircleCI config to use the new image which I manually built and pushed.

Edit: it also moves the installation of `google-cloud-sdk` into the building of our CI image, which should save ~30s on each test.

### DIFF
--- a/.circleci/Makefile
+++ b/.circleci/Makefile
@@ -1,0 +1,5 @@
+build:
+	docker build -t us.gcr.io/vcm-ml/circleci-miniconda3-gfortran .
+
+push: build
+	docker push us.gcr.io/vcm-ml/circleci-miniconda3-gfortran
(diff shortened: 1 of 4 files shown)


frodre (approved):
Thanks for the update

I found this in the main makefile, which can be removed.  I think having closer to the dockerfile location makes more sense for visibility.

```
build_image_ci:
	docker build -t us.gcr.io/vcm-ml/circleci-miniconda-gfortran:latest - < .circleci/dockerfile
```

oliverwm1:
> Thanks for the update
> 
> I found this in the main makefile, which can be removed. I think having closer to the dockerfile location makes more sense for visibility.
> 
> ```
> build_image_ci:
> 	docker build -t us.gcr.io/vcm-ml/circleci-miniconda-gfortran:latest - < .circleci/dockerfile
> ```

oh good find. I will delete that rule.

PR #1336, opened 2021-08-20
prognostic run: Ensure initial_time used in diagnostic configs is cftime.DatetimeJulian

AnnaKwa:
The initial time arg to some of the time related functions is sometimes a datetime.datetime instead of cftime.DatetimeJulian, which raised an error when trying to do comparisions in the `SelectedTimes` time container. 

This also adds time rounding to the `SelectedTimes` container to ensure that the conversion to DatetimeJulian doesn't end up with a slight offset from start of the minute, which in turn results in no selected times getting saved.


brianhenn (approved):
Thanks @AnnaKwa

PR #1337, opened 2021-08-20
fix and test transect snapshot functionality of offline report

mcgibbon:
The PR adding loaders data configuration intended that the offline diags report includes transect snapshots if a "batches from mapper" configuration is provided, such that it can grab a mapper's single data output. However this feature was not working due to a bug in the report generation tool. This PR adds assertions that this part of the report is produced when a BatchesFromMapperConfig is used in the report generation integration test, and fixes the code to produce that part of the report.

Significant internal changes:
- Fixed bug in offline_ml_diags.create_report which would cause transect snapshot to never be reported, even if its netcdf file is present

- [*] Tests added

### DIFF
--- a/workflows/offline_ml_diags/offline_ml_diags/_helpers.py
+++ b/workflows/offline_ml_diags/offline_ml_diags/_helpers.py
@@ -1,4 +1,3 @@
-import fsspec
 import json
 import numpy as np
 import os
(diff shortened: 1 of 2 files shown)


brianhenn (approved):
Thanks @mcgibbon . How was this silently failing -- the if statement looking for the transect dataset was always false?

PR #1342, opened 2021-09-02
Move input/output variables to hyperparameters

mcgibbon:
Currently input/output variables is defined at the top level in TrainingConfig, and passed to training routines. This results in awkward assertion statements in some training routines that require certain inputs or outputs, or do not allow certain outputs. This PR modifies the handling of variables so that the hyperparameters class is responsible for telling the top level only the information it needs (what variables must be in the dataset for training), allowing each training routine to define how it determines inputs and outputs for its resulting Predictor.

Refactored public API:
- Hyperparameter classes now inherit from an ABC and must have a `variables` attribute or property method.
- input_variables and output_variables are moved from TrainingConfig onto RandomForestHyperparameters and DenseHyperparameters. When passed as top level properties in a configuration yaml, a deprecation warning will be emitted.
- `additional_variables` no longer needs to be specified when training random forests with mass scaling

### DIFF
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -234,9 +234,7 @@ class RandomForestHyperparameters(Hyperparameters):
     Args:
         input_variables: names of variables to use as inputs
         output_variables: names of variables to use as outputs
-        scaler_type: scaler to use for training, must be "standard" or "mass".
-            If set to "mass", then "pressure_thickness_of_atmospheric_layer" must
-            be included in the TrainingConfig `additional_variables` field.
+        scaler_type: scaler to use for training, must be "standard" or "mass"
         scaler_kwargs: keyword arguments to pass to scaler initialization
         n_jobs: number of jobs to run in parallel when training a single random forest
         random_state: random seed to use when building trees, will be
@@ -273,4 +271,8 @@ class RandomForestHyperparameters(Hyperparameters):
 
     @property
     def variables(self) -> Set[str]:
-        return set(self.input_variables).union(self.output_variables)
+        if self.scaler_type == "mass":
+            additional_variables = ["pressure_thickness_of_atmospheric_layer"]
+        else:
+            additional_variables = []
+        return set(self.input_variables).union(self.output_variables).union(additional_variables)
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -15,9 +15,11 @@ from typing import (
 from fv3fit.typing import Dataclass
 import xarray as xr
 from .predictor import Predictor
+from .hyperparameters import Hyperparameters
 import dacite
 import numpy as np
 import random
+import warnings
 
 # TODO: move all keras configs under fv3fit.keras
 import tensorflow as tf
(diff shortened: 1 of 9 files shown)


nbren12 (approved):
LGTM! Really positive and simple change. This will make it much easier to configure new model types.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -71,17 +71,34 @@ class TrainingConfig:
     """
 
     model_type: str
-    input_variables: List[str]
```
AnnaKwa, line 74: Also remove docstring lines for these, in addition to the `additional_variables`

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -226,6 +253,9 @@ class RandomForestHyperparameters:
             If False, the whole dataset is used to build each tree.
     """
 
+    input_variables: List[str]
+    output_variables: List[str]
+
```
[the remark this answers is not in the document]
AnnaKwa: (and relocate its docstring as well)

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -44,13 +44,16 @@
...
-    input_variables: Iterable[str],
-    output_variables: Iterable[str],
     hyperparameters: DenseHyperparameters,
     train_batches: Sequence[xr.Dataset],
     validation_batches: Sequence[xr.Dataset],
 ):
-    model = DenseModel("sample", input_variables, output_variables, hyperparameters)
+    model = DenseModel(
+        "sample",
+        hyperparameters.input_variables,
```
AnnaKwa, line 53: I know it would involve changes across a lot of files for the models and training functions, but should we eventually get rid of the input/output variable args since they should always be in the hyperparameters after this PR?

AnnaKwa (approved):
Thanks Jeremy! It is nice that these changes get rid of the mostly unused `additional_variables` arg as well in the TrainingConfig. I have a minor request to keep this as a kwarg for the `RandomForestHyperparameters` in order to avoid breaking the mass weighting feature. I don't need to rereview that as it's a pretty straightforward change.

external/fv3fit/fv3fit/_shared/hyperparameters.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Hyperparameters(abc.ABC):
+    """
+    Attributes:
+        variables: names of data variables required to train the model
+    """
+
+    @property
+    @abc.abstractmethod
+    def variables(self) -> Sequence[str]:
```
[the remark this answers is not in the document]
mcgibbon: And unordered, too. Done.

external/fv3fit/fv3fit/_shared/config.py
mcgibbon: Thanks! Done.

external/fv3fit/fv3fit/_shared/config.py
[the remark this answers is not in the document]
mcgibbon: Instead I added code that automatically adds `pressure_thickness_of_atmospheric_layer` to the variables when `scaler_type == mass`. We should never need to specify `additional_variables` again!

external/fv3fit/fv3fit/keras/_models/models.py
mcgibbon: I considered doing so in this PR, but Predictor specifies that the init should take in input and output variables as arguments. This is somewhat reasonable, since Predictor needs to report what the input and output variables are as properties. The "issue" here leading to weirdness is that DenseModel is both our training code and our Predictor (which is sort of why we moved to training functions).

I think it shouldn't be too much of an issue to leave this as the case for the Dense and RandomForest model training routines. If we ever get around to refactoring the objects into functions this problem will go away, but there's no pressing need to do so.

PR #1343, opened 2021-09-03
Replace offline metrics calculation with registry

AnnaKwa:
This PR refactors the offline diags calculation of MSE, bias, and variance. It replaces the old `metrics.py` code, which was difficult to extend, with a registry based approach. It makes use of a lot of the same code as the prognostic run diags, with the idea that in the future both offline/online diags could use the same package for computations. Currently I put the offline diags code into the `diagnostics_utils` package. I think the "summary" and "diurnal cycle" computations could also be moved here, but for now I want to limit this PR to the computations that were removed from the old `metrics.py`.

example report:
https://storage.googleapis.com/vcm-ml-public/annak/2021-08-31-offline-diags-test/new4/index.html

Refactored public API:
- Change in name formatting in the scalar metrics json file: now uses the same format as in the prog run diags, e.g. `{var}_{computation}_{sfc_type}`. 

- [x] Tests added


external/diagnostics_utils/diagnostics_utils/transform.py
```diff
@@ -0,0 +1,216 @@
+"""
+Transforms operate on diagnostic function inputs to adjust data before
+diagnostic values are calculated.
+
+A transform should take in the transform-specific arguments with a diagnostic
+function argument tuple as the final argument and return the adjusted
+diagnostic function arguments.
+"""
+
```
AnnaKwa, line 9: Most of this is directly copied from the prognostic diags, with the exception of `_is_3d`, `select_2(3)d_variables`, and `regrid_zdim_to_pressure_levels`. However, the `DiagArg` here has four args, the last being an additional `delp` data array that is passed around so that the vertical dimension can be regridded to pressure levels. I am open to suggestions of how to handle keeping this necessary delp data around if we want to make the `DiagArg` identical to the prog run diags for future refactoring to use a single package.

external/diagnostics_utils/diagnostics_utils/registry.py
```diff
@@ -0,0 +1,53 @@
+from collections import defaultdict
```
AnnaKwa, line 1: Same as the prognostic run diags registry.

external/diagnostics_utils/diagnostics_utils/compute_metrics.py
```diff
@@ -0,0 +1,223 @@
...
+            .rename(lat_bins="latitude")
+        )
+    latitude_midpoints = [x.item().mid for x in zm["latitude"]]
+    return zm.assign_coords(latitude=latitude_midpoints)
+
+
+for mask_type in ["global", "sea", "land"]:
+
+    @metrics_registry.register(f"mse_2d_{mask_type}")
+    @transform.apply("select_2d_variables")
```
nbren12, line 82: I realize this pattern is in the prognostic run, but the approach of "registering" a function with its  name as a string and then referring to it later with this same string later is circuitous. A python module is already a "registry" of functions. We can refer to a function using its name `transforms.<name>` and call it later on using normal python syntax.

Can you modify `transform.apply` to take in a function instead and modify its usage throughout?
```suggestion
    @transform.apply(transform.select_2d_variables)
```

external/diagnostics_utils/diagnostics_utils/registry.py
nbren12: Then should we avoid copy paste?

external/diagnostics_utils/diagnostics_utils/transform.py
nbren12: Would be great to avoid this copy paste by refactoring/generalizing the logic of existing prognostic run code. For example the following data structure should support both the offline/online use-cases
```
class DiagArg:
    prediction: xr.Dataset
    target: xr.Dataset
    grid: xr.Dataset
    delp: Optional[xr.DataArray]
```

nbren12 (changes requested):
Thanks anna. I took a quick look. 

This seems like a good improvement over the existing offline run code, but it should avoid copying of existing functionality (e.g. the transforms/registry from the online diags). I make some suggestions below. 

I'm not sure `diagnostics_utils` is the right home for these functions. I've never quite understood the point of `diagnostics_utils` compared to `vcm` or `fv3viz`. Surely generic operations like masking and computing biases would be nice to include in `vcm`.

nbren12:
BTW, I'd be happy to talk with you about where to put these changes other than `diagnostic_utils`.

AnnaKwa:
@nbren12 Thanks for the comments, as we discussed I moved the additional transforms into the prog run diags with the idea that both offline/online diags will eventually share the same source tree. The diffs have gotten kind of large but mostly due to the refactors to i) pass `DiagArg` object around instead of the separate dataset args (
2f007c5 ) , and ii) apply the transform functions directly instead of looking up their keys in the registry (894e539).

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/transform.py
```diff
@@ -297,3 +298,51 @@ def subset_variables(variables: Sequence, arg: DiagArg) -> DiagArg:
...
+        arg.verification,
+        arg.grid,
+        arg.delp,
+    )
+    prediction_vars = [var for var in prediction if not _is_3d(prediction[var])]
+    return DiagArg(prediction[prediction_vars], target[prediction_vars], grid, delp)
+
+
+@add_to_input_transform_fns
+def regrid_zdim_to_pressure_levels(arg: DiagArg) -> DiagArg:
```
nbren12, line 331: Can you document what pressure levels this regrids to?

Also would be nice to allow overriding the vertical dimension.
```suggestion
def regrid_zdim_to_pressure_levels(arg: DiagArg, vertical_dim=VERTICAL_DIM) -> DiagArg:
```

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/transform.py
```diff
@@ -297,3 +298,51 @@ def subset_variables(variables: Sequence, arg: DiagArg) -> DiagArg:
     prognostic_vars = [var for var in variables if var in prognostic]
     verification_vars = [var for var in variables if var in verification]
     return DiagArg(prognostic[prognostic_vars], verification[verification_vars], grid)
+
+
+def _is_3d(da: xr.DataArray):
+    return VERTICAL_DIM in da.dims
+
+
+@add_to_input_transform_fns
```
nbren12, line 306: Is this decorator needed now that you refer to this functions by symbol?

nbren12 (approved):
Looking great! It's exciting to see the offline and online diagnostics code converge.

Just a couple small nits below.

workflows/prognostic_run_diags/fv3net/diagnostics/prognostic_run/transform.py
AnnaKwa: The transform tests still use the registry to ensure any new transform has to pass tests- if we're ok with losing that automatic testing and having new additions be manually added to the list to test, `TRANSFORM_FNS` could be dropped altogether. As it stands now with the functions getting referred to directly, someone could still add a new transform that gets used in the diags but omit the decorator and also dodge the test coverage.

nbren12:
ok. Sounds fine to leave it then.

PR #1344, opened 2021-09-03
fv3viz: cartopy coastline source

brianhenn:
fv3net's fv3viz uses cartopy's coastlines feature, which relies on external data that is intermittently unavailable, causing the fv3net image and the integration tests to fail. This PR redirects fv3viz's cartopy downloads to our vcm-ml-example-data repo so that there is no dependence on outside data hosting. 

Significant internal changes:
- modifies the cartopy config when fv3viz is imported to point its downloaders to vcm-ml-example-data coastline shapefile (other external cartopy features will fail now but none are used by fv3viz, and they could be added if really needed)

### DIFF
--- a/external/fv3viz/README.md
+++ b/external/fv3viz/README.md
@@ -3,3 +3,8 @@ functions: 1) those associated with plotting cubed-sphere tile data as maps usin
 histograms of set of timesteps useful for describing ML datasets.
 
 To get started, see example images and cubed-sphere plotting routines provided in the image gallery.
+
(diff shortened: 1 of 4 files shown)


oliverwm1 (approved):
LGTM, although I guess now if users import fv3viz in a notebook, they will only have access to the particular zip you included in the data repo? I think that's okay, but could bite someone trying to make a hi-res plot or something.

nbren12:
You could make an environment variable that the user can use to override that setting. That's probably the solution cartopy should have anyway.

brianhenn:
Added the environment variable override to use cartopy's downloaders as requested

PR #1348, opened 2021-09-07
TendencyPrescriber to override fv3 physics

oliverwm1:
Add capability to override tendencies from a `Step` with tendencies saved in a dataset. The original tendencies from this step can still be saved as diagnostics.

Added public API:
- `TendencyPrescriber`
- new `get_tendency_prescriber` factory

Significant internal changes:
- if both emulation and tendency prescribing are configured, the FV3 physics will first be emulated, and then the emulated tendencies will be overridden by the prescribed ones.
- modify prognostic run regression tests so that only the `predictor` run uses the online emulator, while the `nudged` run now has overriding

- [x] Tests added


workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -24,3 +27,37 @@ def get_emulator_adapter(
...
+            diagnostic_variables=set(config.diagnostic_variables),
+        )
+
+
+def get_fv3_physics_adapter(
+    config: UserConfig,
+    state: DerivedFV3State,
+    timestep: float,
+    communicator: fv3gfs.util.CubedSphereCommunicator,
+) -> Optional[Union[PrognosticAdapter, OverriderAdapter]]:
```
nbren12, line 55: This is a very broad output type and could get confusing for readers to work out the interfaces the returned object is meant to have. Would be clearer to explicitly define the expected interface using either `typing_extensions.Protocol` or a an abstract base-class for `Adapter`. It's also possible that this leaky abstraction will result in disambiguating conditional logic in the `loop.py`, but I haven't read that far yet.

 
```suggestion
) -> Optional[Adapter]:
```

workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -24,3 +27,37 @@ def get_emulator_adapter(
...
+    if config.online_emulator is not None and config.overrider is not None:
```
nbren12, line 56: It seems it should be straightforward to both prescribe and emulate at the same time. I could envision wanting to do this for a some kind of piggy-backing setup.

Otherwise, I would use an union type for the config so that it is actually impossible to configure the model incorrectly.

workflows/prognostic_c48_run/runtime/overrider.py
```diff
@@ -0,0 +1,130 @@
...
+    url: str
+    variables: Mapping[str, str]
+
+
+@dataclasses.dataclass
+class OverriderAdapter:
+    """Wrap a Step function to override certain tendencies."""
+
+    config: OverriderConfig
+    state: DerivedFV3State
```
nbren12, line 40: Suggest against using the hard-coded `DerivedFV3State` class name which can only be made online in favor of the general purpose `runtime.types.State`. This is what I did for the emulator. Technically your tests below would not pass the type-checker.

workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -24,3 +27,37 @@ def get_emulator_adapter(
             diagnostic_variables=set(config.diagnostic_variables),
             timestep=timestep,
         )
+
+
+def get_overrider_adapter(
```
nbren12, line 32: You should make this private since it is confusing to have two "public" factories for OverriderAdpaters. One which returns just overrider_adapters and a more general one `get_fv3_physics_adapter` which can also return emulators.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -187,10 +187,10 @@ def __init__(
         self.monitor = Monitor.from_variables(
             config.diagnostic_variables, state=self._state, timestep=self._timestep,
         )
-
-        self._emulate = runtime.factories.get_emulator_adapter(
-            config, self._state, self._timestep
+        self._adapt = runtime.factories.get_fv3_physics_adapter(
```
nbren12, line 190: to do both simultaneously:

```suggestion
        self._prescribe_tendency = runtime.factories.get_fv3_physics_adapter(
            ... args
        )
        self._emulate = runtime.factories.get_emulator_adapter(
            config, self._state, self._timestep)
        
```

workflows/prognostic_c48_run/runtime/overrider.py
```diff
@@ -0,0 +1,130 @@
...
+    diagnostic_variables: Set[str] = dataclasses.field(default_factory=set)
+
+    def __post_init__(self: "OverriderAdapter"):
+        if self.communicator.rank == 0:
+            logger.debug(f"Opening tendency overriding dataset from: {self.config.url}")
+        tile = self.communicator.partitioner.tile_index(self.communicator.rank)
+        mapper = fsspec.get_mapper(self.config.url)
+        ds = xr.open_zarr(mapper, consolidated=True)
+        ds = ds[list(self.config.variables.values())].isel(tile=tile)
+        self._tendency_ds = DatasetCachedByChunk(ds, "time")
```
nbren12, line 52: Did you try using [LRUStoreCache](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.LRUStoreCache)? Seems like it could be one-liner, and if it achieves similar performance I would opt for that over rolling our own custom caching layer.
```suggestion
        mb = 2**24
        mapper = zarr.storage.LRUStoreCache(fsspec.get_mapper(self.config.url), max_size=(128 * mb)
        ds = xr.open_zarr(mapper, consolidated=True)
        ds = ds[list(self.config.variables.values())].isel(tile=tile)
        self._tendency_ds = ds
```

workflows/prognostic_c48_run/runtime/overrider.py
```diff
@@ -0,0 +1,130 @@
...
+        tendencies = self.communicator.tile.scatter_state(ds_to_quantity_state(ds))
+        return quantity_state_to_ds(tendencies)
+
+    @property
+    def monitor(self) -> Monitor:
+        return Monitor.from_variables(
+            self.diagnostic_variables, self.state, self.timestep
+        )
+
+    def override(self, name: str, func: Step) -> Diagnostics:
```
nbren12, line 67: Not used publicly so I would opt to hide it.

```suggestion
    def _override(self, name: str, func: Step) -> Diagnostics:
```

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out
```diff
@@ -1,30 +1,31 @@
...
-air_temperature_tendency_due_to_nudging: 74cc6401aef31920fc1abed6b000c5cc
+air_temperature_tendency_due_to_nudging: d461b6836c0c43116df5d661f126bfd9
 area: f31fc363daec5c8e5f76f2762412deb5
-cnvprcp_after_physics: 18dd5bdf6fb264bf9342558240cb5bb9
-cnvprcp_after_python: 18dd5bdf6fb264bf9342558240cb5bb9
-column_heating_due_to_nudging: 6c95d9c76803e619636fdf4700965a22
-evaporation: aeaa17fdb186a9bfb1cb4476f2946d78
-net_moistening_due_to_nudging: 2138071ad62572e69ae5448316d7aa32
-physics_precip: 6130e940fdc375fd5740be268e1f3cf2
+cnvprcp_after_physics: 07932d60c92186bd2093affebec0ab8d
```
Comment (nbren12) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:4]: Lots of checksum changes here, which makes it tricky to review and ensure that existing configurations aren't being broken by your new feature.  If you find something unappealing about the existing regression configurations could you isolate the changes causing this to their own PR? Thanks.

workflows/prognostic_c48_run/tests/test_overrider.py
```diff
@@ -0,0 +1,107 @@
...
+    np.arange(11 * 4 * 6).reshape((11, 4, 6)),
+    dims=["time", "x", "y"],
+    coords={"time": np.arange(2000, 2011)},
+).chunk(
+    {"time": 5, "x": 4, "y": 4}  # type: ignore
+)
+ds = xr.Dataset({"var": da})
+
+
+class MockDerivedState:
```
nbren12, line 22: Is there already a mock for this?

nbren12 (changes requested):
Thanks. Cool to see this was implemented without too many changes to `loop.py`. I think our decorator based approach is leading to clean implementations. I have several high level comments about naming and think the implementation could be simplified by not rolling our own caching layer.

First, I don't find the name `adapt` very informative in the context of loop.py and I am not sure there is a useful shared concept between emulating and tendency prescribing. Yes they share an interface but so does the `Monitor`. By allowing simultaneous emulating and overriding, we can just name things `prescribe`, `emulate`, and `monitor`, and compose them freely.

I'd suggest renaming to `TendencyPresciber` since "prescribing" is already an existing name in the coder and "prescribe" and "override" sound like synonyms in this context.

workflows/prognostic_c48_run/runtime/overrider.py
oliverwm1: good call, i tried it previously but for some reason couldn't get it working. seems to do the trick nicely now.

workflows/prognostic_c48_run/runtime/factories.py
oliverwm1: Yeah I suppose it's true we might want to do both at once. I can adjust it so they are composable.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: Sure, but we still need to decide which (or both or none) of these adapters is used to decorate `apply_physics`. The useful 'shared concept' between overriding and emulation in `loop.py` is that they are both being applied to the physics step.

workflows/prognostic_c48_run/runtime/overrider.py
oliverwm1: Using `runtime.types.State` leads to:
```
workflows/prognostic_c48_run/runtime/overrider.py:69: error: "MutableMapping[Hashable, DataArray]" has no attribute "time"```

File: workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out (hunk 8 above)
Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:4]: The fact that the `predictor` config checksums do not change is good evidence that other features (diagnostics, etc) are not breaking IMO.

File: workflows/prognostic_c48_run/tests/test_overrider.py (hunk 9 above)
Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/test_overrider.py:22]: Not that I am aware of. A search of `MockDerivedState` across fv3net does not find any examples. There is a `MockFV3GFS` in test_derived_state but does not really suit my purposes here.

File: workflows/prognostic_c48_run/runtime/factories.py (hunk 1 above)
Comment (oliverwm1) [reply workflows/prognostic_c48_run/runtime/factories.py:55]: I deleted this function.

File: workflows/prognostic_c48_run/runtime/factories.py [hunk 10]
```diff
@@ -24,3 +27,37 @@ def get_emulator_adapter(
...
+    config: UserConfig,
+    state: DerivedFV3State,
+    communicator: fv3gfs.util.CubedSphereCommunicator,
+    timestep: float,
```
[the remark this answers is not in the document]
oliverwm1: Done.

workflows/prognostic_c48_run/runtime/factories.py
oliverwm1: I deleted the `get_fv3_physics_adapter` function.

workflows/prognostic_c48_run/runtime/factories.py
oliverwm1: Now if a user provides emulation and tendency prescribing configs, both will be done. (emulation first, which is then overriden with the prescribed tendency. Seems more likely we'd want to do it in that order than the reverse).

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out
Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:4]: If you feel strongly about this I can make a separate PR. I just added tendency prescribing to the nudged run to test it in integration with our other pieces. Similar to the `emulator` getting added to the regression tests previously.

workflows/prognostic_c48_run/runtime/overrider.py
nbren12: Ah. you need the time attribute. ok...

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out
Comment (nbren12) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:4]: ok. I think the risk is low in this case given that the other tests pass, but I didn't bring it up for no reason. This is not a forgiving part of the code...a single import out of order can cause the model to crash and consume hours of debugging time. This happens regularly, and I've relied heavily on `git log/bisect` to track these problems down. Keeping the git history clean is worth the 5 minutes it would take to split the checksum breaking changes into a new PR or interactively rebase your branch to be a bit cleaner do so that you can "rebase and merge" rather than "squash and merge".

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -209,11 +212,19 @@ def _get_states_to_output(config: UserConfig) -> Sequence[str]:
...
 
-    def emulate(self, name: str, func: Step) -> Step:
-        if self._emulate is None:
-            return self.monitor(name, func)
-        else:
+    def _get_communicator(self):
+        partitioner = fv3gfs.util.CubedSpherePartitioner.from_namelist(get_namelist())
+        return fv3gfs.util.CubedSphereCommunicator(self.comm, partitioner)
+
+    def emulate_or_prescribe_tendency(self, name: str, func: Step) -> Step:
```
nbren12, line 219: Looks good to me!

workflows/prognostic_c48_run/runtime/tendency_prescriber.py
```diff
@@ -0,0 +1,99 @@
...
+        variables: mapping from state name to name of corresponding tendency in
+            provided dataset. For example: {"air_temperature": "fine_res_Q1"}.
+    """
+
+    url: str
+    variables: Mapping[str, str]
+
+
+@dataclasses.dataclass
+class TendencyPrescriberAdapter:
```
nbren12, line 36: Is there some API being "adapted"? The word adapter refers to the wrapping the ML model (which has its own class hierarchy) with a suitable class with the API expected by the loop.py. In this case there is no underlying class to adapt, so I would simply call this "TendencyPrescriber". 
```suggestion
class TendencyPrescriber:
```

workflows/prognostic_c48_run/tests/test_tendency_prescriber.py
```diff
@@ -0,0 +1,92 @@
...
+        return self._time
+
+    def __getitem__(self, key: str) -> xr.DataArray:
+        return self._state[key]
+
+    def __setitem__(self, key: str, value: xr.DataArray):
+        self._state[key] = value
+
+
+def _get_tendency_ds(time):
```
nbren12, line 32: I'd drop the "ds" abbreviation. Using type hints makes it clear what the output type is.

```suggestion
def _get_tendency(time: cftime.DatetimeJulian) -> xr.Dataset:
```

workflows/prognostic_c48_run/runtime/utils.py
```diff
@@ -0,0 +1,22 @@
+import xarray as xr
+import fv3gfs.util
+
+from runtime.types import QuantityState
+
+
+def ds_to_quantity_state(ds: xr.Dataset) -> QuantityState:
```
nbren12, line 7:
```suggestion
def dataset_to_quantity_state(ds: xr.Dataset) -> QuantityState:
```

workflows/prognostic_c48_run/runtime/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    quantity_state: QuantityState = {
+        variable: fv3gfs.util.Quantity.from_data_array(ds[variable])
+        for variable in ds.data_vars
+    }
+    return quantity_state
+
+
+def quantity_state_to_ds(quantity_state: QuantityState) -> xr.Dataset:
```
nbren12, line 15:
```suggestion
def quantity_state_to_dataset(quantity_state: QuantityState) -> xr.Dataset:
```

workflows/prognostic_c48_run/runtime/utils.py
```diff
@@ -0,0 +1,22 @@
+import xarray as xr
```
nbren12, line 1: Can you think of a more specific name than `utils? Maybe `conversions`?

nbren12 (approved):
LGTM. Just some nits below.

Could be worthwhile to name this concept of  `StepTransformer = Callable[[str, Step], Step]` in `runtime.types`.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out
Comment (oliverwm1) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_check_variables[nudging].out:4]: Thanks, yeah I agree care is required here. I just think for this PR given the small area of overlap with `loop.py` and the nudging code, we are pretty safe.

PR #1351, opened 2021-09-08
Revert prog run logging config change

### LINKED ISSUE #1350: Logging statements missing from prognostic run output
Various intended logging statements in the prognostic run do not appear in the output. I searched through a log file from a recent integration test run and found that the following are missing (this is an incomplete list):

https://github.com/VulcanClimateModeling/fv3net/blob/17b01e4fe52a01984b1978d961fd62e6f2411984/workflows/prognostic_c48_run/runtime/loop.py#L151

https://github.com/VulcanClimateModeling/fv3net/blob/17b01e4fe52a01984b1978d961fd62e6f2411984/workflows/prognostic_c48_run/runtime/loop.py#L171

https://github.com/VulcanClimateModeling/fv3net/blob/17b01e4fe52a01984b1978d961fd62e6f2411984/workflows/prognostic_c48_run/runtime/loop.py#L227

Some things that do appear in the logs:

https://github.com/VulcanClimateModeling/fv3net/blob/17b01e4fe52a01984b1978d961fd62e6f2411984/workflows/prognostic_c48_run/runtime/loop.py#L306-L309

https://github.com/VulcanClimateModeling/fv3net/blob/17b01e4fe52a01984b1978d961fd62e6f2411984/workflows/prognostic_c48_run/runtime/main.py#L59-L60

oliverwm1:
#1345 (specifically https://github.com/VulcanClimateModeling/fv3net/pull/1345/commits/652d97818677c3fb628535353a87d784bf6e9cc8) introduced a change to the prognostic run logging configuration which resulted in #1350. This PR reverts that change and resolves #1350.

Since the loggers in the prog run use the convention of being named `__name__`, setting `logging.getLogger("runtime").setLevel(logging.DEBUG)` does not work as intended (would need to set level for `runtime.loop` etc). Using basicConfig worked well, so going back to that.

### DIFF
--- a/workflows/prognostic_c48_run/runtime/main.py
+++ b/workflows/prognostic_c48_run/runtime/main.py
@@ -15,7 +15,8 @@ import runtime
 STATISTICS_LOG_NAME = "statistics"
 PROFILES_LOG_NAME = "profiles"
 
-logging.getLogger("runtime").setLevel(logging.DEBUG)
+logging.basicConfig(level=logging.DEBUG)
+
 logging.getLogger("fv3gfs.util").setLevel(logging.WARN)
 logging.getLogger("fsspec").setLevel(logging.WARN)
 logging.getLogger("urllib3").setLevel(logging.WARN)


nbren12 (approved):
I'm fine with the change, but still confused about this behavior given the python docs on logger objects: https://docs.python.org/3/library/logging.html#logger-objects.

It says

> The name is potentially a period-separated hierarchical value, like foo.bar.baz (though it could also be just plain foo, for example). Loggers that are further down in the hierarchical list are children of loggers higher up in the list. For example, given a logger with a name of foo, loggers with names of foo.bar, foo.bar.baz, and foo.bam are all descendants of foo. 

So that "runtime" is a parent logger of "runtime.loop".

Then it says,
> When a logger is created, the level is set to NOTSET (which causes all messages to be processed when the logger is the root logger, or delegation to the parent when the logger is a non-root logger). Note that the root logger is created with level WARNING.

> The term ‘delegation to the parent’ means that if a logger has a level of NOTSET, its chain of ancestor loggers is traversed until either an ancestor with a level other than NOTSET is found, or the root is reached.

Which implies that child loggers should emit based on the logging level of parents. e.g. this syntax works for `logging.getLogger("fsspec")`.

nbren12:
In the future, I'd also try to describe the error it solves in the commit message as well rather than just referring to the issue number. This allows someone looking at `git log` to understand what the error being solved is.

oliverwm1:
Yeah, I don't entirely understand why the original code did not work, and as I mentioned via slack, it did not work only in certain contexts (specifically if calling `runfv3 append` from command line, but not if calling `subprocess.check_call(['runfv3', 'append'])` from within python). Worthy of more investigation.

PR #1352, opened 2021-09-09
prognostic run: PureMLStepper can return both state and tendency updates

AnnaKwa:
This PR alters the `PureMLStepper` so that it can return predictions for both tendency and state updates. This makes the old `MLStateStepper` redundant, so it is removed.

This allows for usage of ML models in the prognostic run that predict the postphysics state directly, as opposed to a tendency that is added to the state (ex. total precipitation). There is some special handling for the case where the ML model predictions a precipitation rate; in that case the state update for precip is converted to an accumulation since the actual update that needs to occur is in the `total_precipitation`.

Since this change includes some logic for determing what is a state update vs. tendency, I also changed the nudging calculation to return the same tendency names as the ML stepper. Previously the nudging tendency in the loop had the same name as the variable and the suffix was added before saving.

I tested this with a local prognostic run of the precip model, and verified that it does separate the tendency and state updates correctly out of the postphysics stepper:
```
INFO:runtime.loop:Postphysics stepper adds tendency update to state for dict_keys(['dQ1', 'dQ2'])
INFO:runtime.loop:Postphysics stepper updates state directly for dict_keys(['total_precipitation'])
```

- [x] Tests added


workflows/prognostic_c48_run/tests/machine_learning_mocks.py
```diff
@@ -36,16 +35,20 @@ def get_mock_predictor(model_predictands: str = "tendencies") -> fv3fit.Predicto
     if model_predictands == "tendencies":
         output_variables = ["dQ1", "dQ2", "dQu", "dQv"]
     elif model_predictands == "rad_fluxes":
-        output_variables = ["downward_shortwave", "net_shortwave", "downward_longwave"]
+        output_variables = [
+            "total_sky_downward_shortwave_flux_at_surface",
```
AnnaKwa, line 39: I changed the names of the radiative flux variables in the test to match those in the state, since the MLStepper now uses the state to check which predictions are state updates.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -84,8 +83,20 @@ def add_tendency(state: Any, tendency: State, dt: float) -> State:
 
     with xr.set_options(keep_attrs=True):
         updated = {}
-        for name in tendency:
-            state_name = TENDENCY_TO_STATE_NAME.get(name, name)
```
AnnaKwa, line 88: Previously, the nudging stepper would return tendencies with the same name as the state variable.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -390,6 +406,10 @@ def _compute_postphysics(self) -> Diagnostics:
             (self._tendencies, diagnostics, state_updates,) = self._postphysics_stepper(
                 self._state.time, self._state
             )
+            state_updates = _replace_precip_rate_with_accumlation(
```
AnnaKwa, line 409: @brianhenn Can I get a quick check from you on this part of the PR? Just to make sure that correct update for 
 prescribed or ML-predicted surface precipitation is the accumulated precip as calculated here.

workflows/prognostic_c48_run/runtime/derived_state.py
```diff
@@ -22,7 +22,11 @@ class FV3StateMapper(Mapping):
 
     def __init__(self, getter, alternate_keys: Mapping[str, str] = None):
         self._getter = getter
-        self._alternate_keys = alternate_keys or {"lon": "longitude", "lat": "latitude"}
+        self._alternate_keys = alternate_keys or {
```
nbren12, line 25: Can we do without the renaming or maybe pass it via the `alternate_keys` argument?

workflows/prognostic_c48_run/runtime/names.py
```diff
@@ -20,3 +23,4 @@
     "dQu": EAST_WIND,
     "dQv": NORTH_WIND,
 }
+NUDGING_TENDENCY_SUFFIX = "tendency_due_to_nudging"
```
nbren12, line 26: I understood you did this to disambiguate between state updates and tendencies in `add_tendency`, but his new global value introduces a fair amount of coupling between `loop.py`, `diagnostics`, `nudging`, and `test_nudging` and requires more name/string modifying logic. Can you implement this instead by naming the nudging tendencies `dQu`, `dQv`, dQ1, and dQ2 as the "ML" Stepper does? It seems that the existing convention should suffice.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -162,12 +163,29 @@ def predict(model: MultiModelAdapter, state: State) -> State:
     return {key: cast(xr.DataArray, output[key]) for key in output.data_vars}
 
 
+def _is_state_update_variable(key, state: State):
```
nbren12, line 166: Seems like this function should be defined near where `State` is defined or in `runtime.names`.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -162,12 +163,29 @@ def predict(model: MultiModelAdapter, state: State) -> State:
...
+        # the second check is to exclude derived variables such as dQu,v
+        return True
+    elif key == TOTAL_PRECIP_RATE:
+        # Special case where models predict precip rate which is
+        # converted to state update on accumulated precip
+        return True
+    elif key in TENDENCY_TO_STATE_NAME:
+        return False
+    else:
+        raise ValueError(
```
nbren12, line 177: The name `_is_state_update_variable` does not suggest this function raises an error, and I'm not convinced this is something that should kill the model execution. Can you delete this error or raise/make a warning in the calling routine?

nbren12 (changes requested):
Thanks! The code in `steppers.machine_learning` looks nice and clean, but I have some suggestions about how to reduce the coupling in how the "nudging tendency variables" are represented.

workflows/prognostic_c48_run/runtime/names.py
nbren12: If that doesn't work, we can discuss other ways to remove the coupling.

workflows/prognostic_c48_run/runtime/derived_state.py
AnnaKwa: The `physics_precip` name is hardcoded in the `PrecipitativeModel` and I think using that name in there (as opposed to `surface_precipitation_rate` does help clarify how the architecture works. 

By 'pass via the `alternate_keys` arg`, do you mean pass it in when the loop's `state` is initialized in `loop.py`? If so, what's the benefit in doing that versus having it be in the default `_alternate_keys`? Just wondering since the "physics_precip"/"surface_precipitation_rate" pair is hard coded into the base diagnostics that get saved, so I assumed it'd be an ok alternate key pair to include here.

nbren12: > Do you mean pass it in when the loop's stateis initialized inloop.py

That's right. Best to be up-front with this kind of "default" behavior IMO rather than hiding it in the code.

However, I just remembered the  `runtime.machine_learning.RenamingAdapter` also provides a configurable renaming via `config.scikit_learn.input_standard_names`. It doesn't seem like any code needs to change to support this renaming.

workflows/prognostic_c48_run/runtime/names.py
AnnaKwa: I removed the suffix import, and now the nudging tendencies follow the same naming as the ML tendencies.

workflows/prognostic_c48_run/runtime/derived_state.py
AnnaKwa: As discussed in person, I'll leave this here since the renaming pair is kind of buried in the prog run code, which can be a pain to dig out when writing the config (ex. I renamed it to the wrong precip variable the first time I tested a model using  this input)

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -78,14 +81,34 @@ def get_momentum_diagnostics(self, state, tendency) -> Diagnostics:
...
 
+def _replace_precip_rate_with_accumlation(state_updates: State, dt: float) -> State:
+    # Precipitative ML models predict a rate, but the precipitation to update
+    # in the state is an accumulated total over the timestep
+    if TOTAL_PRECIP_RATE in state_updates:
+        state_updates[TOTAL_PRECIP] = precipitation_accumulation(
+            state_updates[TOTAL_PRECIP_RATE], dt
+        )
+        state_updates.pop(TOTAL_PRECIP_RATE)
+    return state_updates
```
nbren12, line 92: Since this function operates in place, I'd suggest not returning anything.
```suggestion
```

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -379,7 +402,20 @@ def _compute_postphysics(self) -> Diagnostics:
             (self._tendencies, diagnostics, state_updates,) = self._postphysics_stepper(
                 self._state.time, self._state
             )
+            state_updates = _replace_precip_rate_with_accumlation(
+                state_updates, self._timestep
+            )
+
+            self._log_info(
```
nbren12, line 409: Maybe use debug level logging for this information to avoid adding even more to the INFO level logs.

nbren12 (approved):
Looks great! Thanks. Just a couple nits below.

Either @brianhenn or @oliverwm1 will need to comment on the correctness of the precip setting procedure.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -78,14 +81,35 @@ def get_momentum_diagnostics(self, state, tendency) -> Diagnostics:
         return {}
 
 
+def _replace_precip_rate_with_accumlation(
```
brianhenn: typo in 'accumulation'

workflows/prognostic_c48_run/runtime/names.py
```diff
@@ -20,3 +25,16 @@
...
     "dQv": NORTH_WIND,
 }
+STATE_NAME_TO_TENDENCY = {value: key for key, value in TENDENCY_TO_STATE_NAME.items()}
+
+
+def is_state_update_variable(key, state: State):
+    if key in state.keys() and key not in TENDENCY_TO_STATE_NAME:
+        # the second check is to exclude derived variables such as dQu,v
+        return True
+    elif key == TOTAL_PRECIP_RATE:
```
brianhenn, line 35: Does this need a special case here and one in the derived mapping alternate keys, or could it be done with only one of those?

workflows/prognostic_c48_run/runtime/nudging.py
```diff
@@ -219,6 +219,7 @@ def get_nudging_tendency(
             variable.
     Returns:
         nudging_tendencies (dict): A dictionary whose keys are standard names
+            appended with "_tendency_due_to_nudging"
```
brianhenn, line 222: I don't think this is true anymore

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -183,7 +184,17 @@ def __call__(self, time, state):
...
+            if is_state_update_variable(key, state):
+                state_updates[key] = value
+            else:
+                tendency[key] = value
+
+        for name in state_updates.keys():
+            diagnostics[name] = state_updates[name]
 
         dQ1_initial = tendency.get("dQ1", xr.zeros_like(state[SPHUM]))
         dQ2_initial = tendency.get("dQ2", xr.zeros_like(state[SPHUM]))
```
brianhenn, line 200: The precipitative model is going to have some potentially confusing interactions with the non-negative specific humidity limiter, though this is not really any different from the biases the limiter introduces in other models and there's nothing I can think to easily be done about it.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -398,19 +433,23 @@ def _apply_postphysics_to_dycore_state(self) -> Diagnostics:
                 updated_state_from_tendency = add_tendency(
                     self._state, tendency, dt=self._timestep
                 )
+
+                # if total precip is updated directly by stepper,
```
brianhenn, line 437: OK I think I follow the logic that makes sure that that total_precipitation updates from state updates are set directly while updates from dQ2 use precipitation_sum. 

Not for this PR but it keeps coming up how useful it would be to have a loop unit test which ensure that all of the rules we've built into the loop are actually followed.

workflows/prognostic_c48_run/runtime/diagnostics/compute.py
```diff
@@ -30,6 +38,22 @@ def precipitation_sum(
     return total_precip
 
 
+def precipitation_accumulation(
```
brianhenn, line 41: Looks good

brianhenn (approved):
Thanks @AnnaKwa

workflows/prognostic_c48_run/runtime/names.py
AnnaKwa: Just here, since this PR handles the accumulation from rate conversion independently of the derived mapping. i.e. the loop explicitly replaces this prediction with its equivalent accumulation.

workflows/prognostic_c48_run/runtime/loop.py
nbren12: > which ensure that all of the rules we've built into the loop are actually followed

The regression tests are pretty good and take maybe 30 seconds to run per configuration.  IMO loop.py is not really unit-testable outside of the regression tests since it interacts so much with the wrapper. 

 I'd rather suggest moving "all of the rules we've built into the loop" into units. The regression tests and mypy are pretty useful for that kind of refactor.

nbren12: For example, the coupling between add_tendency, loop, and stepper could be resolved by having the steppers return a function that applies their internal predicted tendencies to the state. Any physicsy code can be pushed into a stepper. or something like `monitor`.

brianhenn: Makes sense. I'd be interested in working on those kinds of refactors at some point.

workflows/prognostic_c48_run/runtime/names.py
brianhenn: Oh, what I meant was that there are two special cases in this PR (the alternate keys in the derived mapper being the other), and whether we could have gotten away with just one.

PR #1353, opened 2021-09-10
Add prepare config python apis

nbren12:
Add a python API for preparing configs with simple initial conditions. I didn't modify the CLI, since I am using python APIs to build configs for my emulations runs anyway.

I suggest reviewing one commit at a time to separate refactors from the new feature.

### DIFF
--- a/workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py
+++ b/workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py
@@ -4,7 +4,7 @@ import yaml
 import logging
 import sys
 from datetime import datetime, timedelta
-from typing import Mapping, Sequence
+from typing import Mapping, Sequence, Optional
 
 import dacite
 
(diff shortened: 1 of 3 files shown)


workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py
```diff
@@ -121,55 +123,90 @@ def _diag_table_overlay(
...
+        timestep: a YYYYMMDD.HHMMSS timestamp
+    """
 
+    base_url: str
+    timestep: str
 
-def _prepare_config_from_parsed_config(
-    user_config: UserConfig, fv3_config: dict, base_version: str, args
-):
-    if not set(fv3_config) <= FV3CONFIG_KEYS:
```
nbren12, line 141: This check cannot possibly pass given the code below so I have deleted it.

oliverwm1 (approved):
LGTM. Still think it would be a good goal to delete the special initial condition handling (i.e. the `fv3kube.c48_initial_conditions_overlay` function) but this is a nice step to modularizing/generalizing the prepare config script!

PR #1355, opened 2021-09-13
Prognostic Zhao-Carr microphysics emulation

frodre:
Incorporate the Zhao-Carr microphysics emulation package and `call_py_fort` installation from `fv3gfs-fortran` into the prognostic run image.

- [ ] Tests added


workflows/prognostic_c48_run/emulation/emulation/_filesystem.py
```diff
@@ -0,0 +1,43 @@
+# keras has several routines which interact with file paths directly as opposed to
+# filesystem objects, which means we need these wrappers so we can allow remote paths
+
+import contextlib
+import tempfile
+import fsspec
+import os
+
+
+@contextlib.contextmanager
```
frodre, line 10: This file was copied out of fv3fit because I wasn't sure if we would have the dependency installed.  Could delete and link to fv3fit, or just pare down to the `get_dir`, which is the only thing used in `emulate.py`

workflows/prognostic_c48_run/emulation/emulation/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .monitor import store
```
frodre, line 1: Only the top-level entrypoints that fv3gfs-fortran expects are included here

workflows/prognostic_c48_run/.gitignore
```diff
@@ -0,0 +1,2 @@
+rundir/
```
nbren12, line 1: Can you revert this? doesn't seem related to the PR.

workflows/prognostic_c48_run/emulation/emulation/_filesystem.py
nbren12: > just pare down to the get_dir, which is the only thing used in emulate.py

This sounds like a good idea.

workflows/prognostic_c48_run/emulation/emulation/debug.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def print_location_ping(state):
+
+    logger.info("Ping reached!")
+
+
+def dump_state(state):
+
+    DUMP_PATH = str(os.environ.get("STATE_DUMP_PATH"))
```
nbren12, line 38: Needs default?
```suggestion
    DUMP_PATH = str(os.environ.get("STATE_DUMP_PATH", "."))
```

workflows/prognostic_c48_run/emulation/emulation/monitor.py
```diff
@@ -0,0 +1,242 @@
+import logging
```
nbren12, line 1: I think this module is sufficiently complex that it should have some unit tests.

nbren12, line 1: I think this module is sufficiently complex that it should have some unit tests. To simplify testing, instead of using globals, you could refactor it into a singleton class e.g.:
```
singleton = Class.from_environment(sys.environ)
store = singleton.store
...
```

workflows/prognostic_c48_run/emulation/microphysics_parameter_metadata.yaml
```diff
@@ -0,0 +1,320 @@
+air_pressure:
```
nbren12, line 1: I don't think this file is included in the `REQUIREMENTS.in` or `setup.py`. I'm kind of surprised `runfv3` works. Have you actually tried generating the data in the prognostic run image?

nbren12 (changes requested):
Thanks Andre. I've noticed a couple things to work on. I'm still working to run this code myself, so more may come up.

workflows/prognostic_c48_run/emulation/emulation/emulate.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@print_errors
+def _get_timestep(namelist):
+    return int(namelist["coupler_nml"]["dt_atmos"])
+
+
+@print_errors
+def _load_tf_model() -> tf.keras.Model:
+    logger.debug(f"Loading keras model: {TF_MODEL_PATH}")
```
nbren12, line 53: Suggest making this `info` level since it is pretty significant information (and is currently the only way to tell what model is used since the env variables are not stored in the outputs).

```suggestion
    logger.info(f"Loading keras model: {TF_MODEL_PATH}")
```

workflows/prognostic_c48_run/emulation/microphysics_parameter_metadata.yaml
frodre: The directory is copied in the dockerfile, and an then an environment variable is set in that points to it.  I haven't used the packaging route of including files before, but I think that might be a good option for a static file like this.

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -510,6 +511,19 @@ def get_ml_config(model_path):
     return config
 
 
+def get_emulation_config():
```
frodre, line 514: Moved copy pasted integration test back into the main file.  Use monkeypatch to temporarily update environment variables.

workflows/prognostic_c48_run/emulation/emulation/_emulate/microphysics.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def from_environ(cls, d: Mapping):
+
+        model_path = d["TF_MODEL_PATH"]
+
+        return cls(model_path)
+
+    def microphysics(self, state):
+
+        inputs = [state[name].T for name in self.model.input_names]
```
nbren12, line 86: could you add a comment with the dimension order?

workflows/prognostic_c48_run/emulation/emulation/_emulate/microphysics.py
```diff
@@ -0,0 +1,99 @@
...
+        self.orig_outputs = None
+
+    @classmethod
+    def from_environ(cls, d: Mapping):
+
+        model_path = d["TF_MODEL_PATH"]
+
+        return cls(model_path)
+
+    def microphysics(self, state):
```
nbren12, line 84: Can you add type hints at least for the "public APIs" `from_environ`, `microphysics`, and `store`. This is important since `State` is a different type then we use in the prognostic run code, and it could be confusing.

workflows/prognostic_c48_run/emulation/emulation/_monitor/monitor.py
```diff
@@ -0,0 +1,257 @@
...
+def _store_zarr(state, time, monitor, metadata):
+
+    logger.info(f"Storing zarr model state on rank {MPI.COMM_WORLD.Get_rank()}")
+    logger.debug(f"Model fields: {list(state.keys())}")
+    state = _convert_to_quantities(state, metadata)
+    state["time"] = time
+    monitor.store(state)
+
+
+class Config:
```
nbren12, line 178: I would rename this to `StorageHook` or something similar since it has methods like `microphysics` and `store` that seem unrelated to configuration.

workflows/prognostic_c48_run/emulation/emulation/_monitor/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .._utils import wrap_configurable_hook
+
+try:
+    config = Config.from_environ(os.environ)
+    error = None
+except Exception as e:
+    config = None
+    error = e
+
+store = wrap_configurable_hook(config, error, "store")
```
nbren12, line 13: I'm not sure `wrap_configurable_hook` is a well-factored function. The name is pretty opaque and the interface is tightly coupled to the try-except block. Seems like this would be simpler (likewise for the code in `microphysics`).

```suggestion
try:
    config = Config.from_environ(os.environ)
    store = config.store
except KeyError as e:
    logging.exception(e)
    def store(_):
        pass
```

Similar LOC but without the hard to understand function name.

workflows/prognostic_c48_run/emulation/emulation/_emulate/microphysics.py
```diff
@@ -0,0 +1,99 @@
...
+    else:
+        model_outputs = {
+            name: output.T  # transposed adjust
+            for name, output in zip(output_names, predictions)
+        }
+
+    return model_outputs
+
+
+class Config:
```
nbren12, line 60:
```suggestion
class MicrophysicsHook:
```

nbren12 (approved):
Thanks for the tests! Excited to use this code. The only "required change" is to move this to move this to `external/emulation`, but the code would be easier to follow if there were fewer nested `__init__` files and the names were clearer. Some suggestions below.

PR #1359, opened 2021-09-15
Prognostic run: fix missing nudging tendency names

AnnaKwa:
PR https://github.com/VulcanClimateModeling/fv3net/commit/06934dc6494cd282cb404b3624c084292a1329f4 broke the nudging runs because I changed the nudging to use the same tendency to state name dict as the ML stepper, but forgot to add names for nudging tendencies that aren't predicted by the ML (pressure, D grid winds). This adds those to `names.TENDENCY_TO_STATE_NAME` to avoid key errors.


oliverwm1 (approved):
LGTM.

PR #1360, opened 2021-09-15
Online emulator master fixes

nbren12:
Fix a bug requiring the `fv3fit.train_emulator` to be run with root access and allow training with any derived variable.


frodre (approved):
LGTM

PR #1363, opened 2021-09-15
fv3fit: let PrecipitativeModel use additional inputs

AnnaKwa:
This changes the type of additional_input_variables in PrecipitativeHyperparameters to list instead of iterable. When the yaml config is loaded, this field is read as a list and dacite will fail to create the hyperparameter object from the loaded dict due to the type mismatch with the old arg type (iterable).


mcgibbon (approved):
Thanks for the bug fix!

PR #1361, opened 2021-09-15
Decouple fv3net docker images from fv3gfs-fortran

nbren12:
The prognostic run dockerfile depends on ESMF and MPI from images pushed
from the fv3gfs-fortran repo. This effectively coupled the build systems
of fv3net to fv3gfs-fortran, but this repo has different concerns.
fv3net emphasize modern tools to support experimental ML, while
fv3gfs-fortran empahsize stability in bit-for-bit checksums. As a
consequence, this coupling effectively pinned this repo to ubuntu
version supported by fv3gfs-fortran (18.04), preventing fv3net from
updating to a more recent OS and software like python 3.7 (See #1083).

Building esmf, mpi, and fms in the prognostic run dockerfile effectively
decouples these repos.

The checksums remain unaltered.

PR #1366, opened 2021-09-16
check if caching works

nbren12:
Check if caching works.


nbren12:
Looks like the caching in #1361 works.

PR #1367, opened 2021-09-16
Update prognostic run image to Ubuntu 20.04 LTS

nbren12:
We have been stuck on a very old python version 3.6 that is not even supported by the latest versions of some of our dependencies (e.g. xarray). This PR updates to Ubuntu 20.04, which has python 3.8.

The checksums have changed since the fortran compiler is also on a different version in the new OS.


frodre (approved):
Thanks for making this update!  I triggered the integration tests just to be sure that everything works out okay.

I'm assuming we weren't expecting bit-for-bit with the compiler change for the checksums.

PR #1373, opened 2021-09-22
fv3viz: simpler `plot_cube` API without `mappable_var`

brianhenn:
The `fv3viz.plot_cube` function required a dataset input that was the output of `fv3viz.mappable_var`, a two-step process that also required correctly setting the kwargs for `mappable_var`. This PR changes `fv3viz` API to use a single call of `plot_cube`, in which now the grid metadata must be specified in order to accomplish the grid and plotting variable dimension alignment that was previously done by `fv3viz.mappable_var`. The default grid metadata in `fv3viz` is from the python wrapper, so for most applications it does not actually need to be provided. It also adds a new public data structure in `vcm.cubedsphere` for storing grid metadata.

Added public API:
- `vcm.cubedsphere.GridMetadata`: a data structure for grid variable name and dimensions

Refactored public API:
- `fv3viz.plot_cube` now takes two required arguments (`ds` and `var_name`), and has a new optional `grid_metadata` arg for when the grid is not the default values, i.e., GFDL FV3 Fortran diagnostic data
- the former `fv3viz.mappable_var` and `fv3viz.plot_cube_axes` are now private 
- removed `fv3viz.plot_diag_var_single_map` which largely duplicated `plot_cube`

Significant internal changes:
- `fv3viz.plot_cube` internally uses `_mappable_var` and the grid metadata to ensure plottable data

- [X] Tests added


external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -448,7 +387,7 @@ def _validate_cube_shape(lat_shape, lon_shape, latb_shape, lonb_shape, array_sha
         raise ValueError("Lon, lat, and data_var each must be 3-dimensional.")
 
 
-def plot_cube_axes(
+def _plot_cube_axes(
```
brianhenn, line 390: I couldn't find usage of this outside of notebooks, and my understanding is that everything you'd want to use this for can be done with plot_cube or pcolormesh_cube.

external/fv3viz/fv3viz/_plot_diagnostics.py
```diff
@@ -74,29 +73,8 @@ def plot_diurnal_cycle(
     return fig
 
 
-# functions below here are from the previous design and probably outdated
-# leaving for now as they might be adapted to work with new design
-
-
-def plot_diag_var_single_map(da, grid, var_name, plot_cube_kwargs=None):
```
brianhenn, line 81: Unused as far as I can tell and almost entirely duplicates `plot_cube`

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -59,15 +68,21 @@ def plot_cube(
...
+    onto a global map projection, with optional faceting of additional dims
 
     Args:
-        plottable_variable (xr.Dataset):
-            Dataset containing variable to plotted via pcolormesh, along with
-            coordinate variables (lat, latb, lon, lonb). This dataset object
-            can be created from the helper function :py:func:`mappable_var`, which takes
-            in an fv3gfs restart or diagnostic dataset along with the name of
-            the variable to be plotted.
+        ds (xr.Dataset):
```
oliverwm1, line 75: nit: docstrings shouldn't include types if there are type hints in the call signature

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -121,7 +136,8 @@ def plot_cube(
     Example:
         # plot diag winds at two times
         fig, axes, hs, cbar, facet_grid = plot_cube(
-            mappable_var(diag_ds, 'VGRD850').isel(time = slice(2, 4)),
+            diag_ds.isel(time = slice(2, 4),
```
oliverwm1, line 139:
```suggestion
            diag_ds.isel(time = slice(2, 4)),
```

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -171,6 +189,7 @@ def plot_cube(
             fig, ax = plt.subplots(1, 1, subplot_kw={"projection": projection})
         else:
             fig = ax.figure
+        print(array.shape)
```
oliverwm1, line 192: leftover print statement

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -190,125 +209,45 @@ def plot_cube(
...
-    for coord in [coord_y_center, coord_x_center, coord_y_outer, coord_x_outer]:
-        if coord in new_ds.coords:
-            new_ds = new_ds.drop(coord)
-
-    return new_ds
+    mappable_ds = xr.Dataset()
+    for var, dims in grid_metadata.coord_vars.items():
+        mappable_ds[var] = _align_grid_var_dims(ds[var], required_dims=dims)
+    var_da = _align_plot_var_dims(ds[var_name], grid_metadata.y, grid_metadata.x)
+    return mappable_ds.merge(var_da)
```
oliverwm1, line 241: Is this just a refactor or is there a more significant code change?

external/vcm/vcm/cubedsphere/grid_metadata.py
```diff
@@ -0,0 +1,24 @@
...
+class GridMetadata:
+    x: str
+    y: str
+    x_interface: str
+    y_interface: str
+    tile: str = "tile"
+    lon: str = "lon"
+    lonb: str = "lonb"
+    lat: str = "lat"
+    latb: str = "latb"
```
oliverwm1, line 14: Maybe just give everything defaults? Not really clear what distinguishes the dimension names from the lat/lon variable names. If anything, I would say the former are less likely to change than the latter.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
```diff
@@ -17,19 +17,7 @@
...
-    "coord_y_outer": "y_interface",
-}
-
-_COORD_VARS = {
-    "lonb": ["y_interface", "x_interface", "tile"],
-    "latb": ["y_interface", "x_interface", "tile"],
-    "lon": ["y", "x", "tile"],
-    "lat": ["y", "x", "tile"],
-}
+COORD_VARS = ["lon", "lat", "lonb", "latb"]
```
oliverwm1, line 20: love these simplifications.

oliverwm1 (commented):
Thanks for this change, will make life easier!

external/fv3viz/fv3viz/_plot_cube.py
brianhenn: Code change that accomplished 1) fixing the in-place modification of the coord vars bug and 2) no longer setting the grid vars as xarray coordinates since this was never actually needed by the `plot_cube` routine. Also allowed for adding more obvious error messages when the wrong grid metadata is passed and better tests.

external/vcm/vcm/cubedsphere/grid_metadata.py
brianhenn: Sure will add.

external/fv3viz/fv3viz/_plot_cube.py
oliverwm1: Okay thanks!

oliverwm1 (approved):
LGTM, one other change is to make sure to update the API docs to no longer include `mappable_var`.

brianhenn:
Thanks. I think sphinx automatically ignores private functions if that's what you mean?

oliverwm1:
> Thanks. I think sphinx automatically ignores private functions if that's what you mean?

I just noticed both `mappable_var` and `plot_cube_axes` are listed here: https://github.com/ai2cm/fv3net/blob/master/external/fv3viz/docs/fv3viz_api.rst

brianhenn:
Ah yeah good catch

brianhenn:
Thanks deleted them but interesting that sphinx will just ignore members that don't exist.

oliverwm1:
> Thanks deleted them but interesting that sphinx will just ignore members that don't exist.

yah I didn't realize that, but I guess explains why there weren't errors on docs build.

PR #1374, opened 2021-09-22
Bump xarray to 0.19.0

oliverwm1:
Now that the prognostic run image is on python 3.8, we can relax the xarray<0.17.0 version limit. Bumping to the latest version of xarray gives us some nice conveniences like being apply to pass a GCS url directly to `xr.open_zarr` (and by default it will try to open consolidated metadata!).

Note that the dimension order shown in the xarray repr changes, and this required an update to some regression test data. There are also changes in the checksums of certain arrays, possibly also because of this dimension order change, but I have not examined in detail. Notably, none of the prognostic run checksums change, only unit tests with artificial data.

Requirement changes:
- xarray>=0.19.0 in `pip-requirements.txt`
- [x] Ran `make lock_pip` following these [instructions](https://vulcanclimatemodeling.com/docs/fv3net/dependency_management.html#dependency-management)
- [x] Add PR review with license info for any additions to `constraints.txt` (n/a since no new deps)


oliverwm1:
@brianhenn FYI looks like the call signature of `xr.tutorial.open_dataset` [changed](https://github.com/pydata/xarray/pull/4102/files) so the fv3viz gallery is broken: https://app.circleci.com/pipelines/github/ai2cm/fv3net/9479/workflows/99fa1353-5555-43d1-8516-6ed949fdd528/jobs/80831

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -310,7 +309,7 @@ def main(args):
         grid = load_grid_info(args.grid_resolution)
     else:
         with fsspec.open(args.grid, "rb") as f:
-            grid = xr.open_dataset(f).load()
+            grid = xr.open_dataset(f, engine="h5netcdf").load()
```
oliverwm1, line 312: Now need to use this engine for opening a local fsspec filesystem object (maybe an xarray regression?)

constraints.txt
```diff
@@ -11,6 +11,7 @@ altair==4.1.0
 anyio==2.2.0
 apache-beam==2.27.0
 appdirs==1.4.4
+appnope==0.1.2
```
nbren12, line 14: Please verify this has an appropriate license.

nbren12 (changes requested):
Thanks for tracking down all the regressions and fixing some mypy issues.

Perhaps we should modify the regression tests to be less sensitive to small changes like this (e.g. just hash the numpy array with a certain dimension order).

constraints.txt
oliverwm1: Good catch. It is BSD 2-clause. https://github.com/minrk/appnope/blob/main/LICENSE

nbren12 (approved):

oliverwm1:
> Perhaps we should modify the regression tests to be less sensitive to small changes like this (e.g. just hash the numpy array with a certain dimension order).

Yeah, that might be good in some cases. Although we probably do want to test dimension order in other cases (I've run into issues where multiplying two arrays was slow via dask, I think because they had a different dimension order).

PR #1375, opened 2021-09-22
Offline diags: bug fix for land/sea area weighted averages

AnnaKwa:
This is a bug fix for the land/sea area weighted averaging in the offline metrics; the previous refactor used the wrong transform which did not mask the area weights.


oliverwm1 (approved):
Thanks for the fix.

PR #1386, opened 2021-10-06
Check for existence of wind diags before using them

oliverwm1:
Fixes bug in `fv3net.diagnostics.prognostic_run.derived_diagnostics` that prevents computing metrics or generating report from diagnostics that do not include time- and zonal-mean northward wind.


brianhenn (approved):
LGTM

PR #1387, opened 2021-10-06
emulator prediction masking

nbren12:
Code extracted from #1265.

Masking experiments which ignore the emulator outputs for a given level in favor the emulatee (the gfsphysics) reveal the connection between offline and online skill. If an online drift is related to offline skill at a particular level, then using the gfsphysics there should improve the online drifts. 

Adds some mask types and function `update_state_masked`, and uses them in the emulator.

Applying this mask types elsewhere is a natural follow-up.

### DIFF
--- a/workflows/prognostic_c48_run/runtime/emulator.py
+++ b/workflows/prognostic_c48_run/runtime/emulator.py
@@ -25,7 +25,9 @@ class Config:
         online: if True, the emulator will be replace fv3 physics for all
             humidities above level ``ignore_humidity_below``.
         train: if True, each timestep will be used to train the model.
-        output_mask: if provided, use the physics instead of the emulator here
+        mask_kind: the type of mask. Defers to the functions
+            ``runtime.masking.compute_{mask_kind}``. The default does not mask any
+            of the emulator predictions.
         ignore_humidity_below: if mask_kind ="default", then use the fv3 physics
             instead of the emulator above this level.
     """


workflows/prognostic_c48_run/runtime/emulator.py
```diff
@@ -33,7 +25,7 @@ class Config:
         online: if True, the emulator will be replace fv3 physics for all
             humidities above level ``ignore_humidity_below``.
         train: if True, each timestep will be used to train the model.
-        ignore_humidity_below: see ``online``.
```
frodre, line 36: Include this attribute and explain when it's used

workflows/prognostic_c48_run/runtime/emulator.py
```diff
@@ -45,6 +37,7 @@ class Config:
     # will ignore the emulator for any z larger than this value
     # remember higher z is lower in the atmosphere hence "below"
     ignore_humidity_below: Optional[int] = None
+    mask_kind: str = "default"
```
frodre, line 40: I think it would be more natural to have a `mask_kind=None` be no masking instead of `mask_kind="default", ignore_humidity_below=None`.  See other comment about the state update.

workflows/prognostic_c48_run/runtime/emulator.py
```diff
@@ -145,44 +140,3 @@ def step() -> Diagnostics:
...
-        if name == SPHUM:
-            if self.ignore_humidity_below is not None:
-                return arr.z > self.ignore_humidity_below
-            else:
-                return xr.DataArray(False)
-        else:
-            return xr.DataArray(True)
-
-
-def update_state_with_emulator(
```
frodre, line 183: I'd propose we leave `update_state_with_emulator` function in this module and within that, optionally apply the mask when specified

workflows/prognostic_c48_run/runtime/masking.py
```diff
@@ -0,0 +1,101 @@
...
+        state: the mutable state object
+        src: updates to put into state
+        compute_mask: a function returning a mask. Where this mask is True, the
+            original state array will be used.
+
+    """
+    for key in src:
+        arr = state[key]
+        mask = compute_mask(key, arr)
+        state[key] = arr.where(mask, src[key].variable)
```
frodre, line 33: Rather than in-place updates, just return a new dict of blended fields

workflows/prognostic_c48_run/runtime/masking.py
```diff
@@ -0,0 +1,101 @@
...
+    Optional,
+)
+
+from runtime.names import EAST_WIND, SPHUM, CLOUD
+
+
+__all__ = ["get_mask", "update_state_masked"]
+
+
+def update_state_masked(
```
frodre, line 17: Switch this to something like `apply_mask`

frodre (changes requested):
Thanks, @nbren12!  I think this will be a useful feature.   I had some minor adjustment requests.  I was thinking that it makes more sense to leave the update in the emulator module and call an apply from the masking module.  Also, I wanted check and see whether the fv3gfs-fortran update was a mistake.

workflows/prognostic_c48_run/runtime/emulator.py
nbren12: The default has the same effect, and always using a mask is nice to avoid adding `if mask_kind is None` anywhere.

workflows/prognostic_c48_run/runtime/masking.py
nbren12: Renamed to `where_masked` since the underlying xarray function is named `where`.

nbren12:
@frodre  This is ready for re-review I had to rewrite this history to cleanly separate the changes the modified the regression tests results (6a844c4) from the others.

workflows/prognostic_c48_run/runtime/emulator.py
frodre: That's fine.  I was just thinking if using the config, I might have the question, "how do I know masking is off?" or "how do I turn off masking?".  The current default does that, but I don't think it's clear that it's the case.  `mask_kind=None` was an attempt to remedy that, but I think if you just explain in the docstring when masking is "off" that would also suffice.

workflows/prognostic_c48_run/runtime/emulator.py
frodre: should `output_mask` also have a docstring or is that an old name for `mask_kind`?

frodre (approved):
A few suggestions on the docstring but otherwise, LGTM!  Thanks for this, I can already think of some useful places for this.

nbren12:
Thanks @frodre. I fixed the remaining docstring issues. 

I'd recommend against reusing this configuration elsewhere. I left it as is so that it is backwards compatible with my existing experiments, but I think we'll need to make a change and introduce a `masking.Config` if this feature goes any further.

PR #1388, opened 2021-10-07
emulation: Add L2 regularization to multi-level models

nbren12:
This may help online stability

(cherry picked from commit a59f9b51498afc5ffa9f1bc9bc3f3c09d59adcdf)

- `--l2` flag to `fv3fit.train_emulator` CLI

### DIFF
--- a/external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
+++ b/external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
@@ -382,6 +382,7 @@ def get_model(config: Config) -> tf.keras.Model:
             var_level=config.target.level,
             num_hidden=config.num_hidden,
             num_hidden_layers=config.num_hidden_layers,
+            regularizer=regularizer,
         )
     elif isinstance(config.target, RHLossSingleLevel):
         logging.info("Using RHScaler")
@@ -389,6 +390,7 @@ def get_model(config: Config) -> tf.keras.Model:
             var_level=config.target.level,
             num_hidden=config.num_hidden,
             num_hidden_layers=config.num_hidden_layers,
+            regularizer=regularizer,
         )
     else:
         raise NotImplementedError(f"{config}")
--- a/external/fv3fit/fv3fit/emulation/thermobasis/models.py
+++ b/external/fv3fit/fv3fit/emulation/thermobasis/models.py
@@ -22,7 +22,13 @@ def atleast_2d(x: tf.Variable) -> tf.Variable:
 
 
 class ScalarMLP(tf.keras.layers.Layer):
-    def __init__(self, num_hidden=256, num_hidden_layers=1, var_level=0):
+    def __init__(
+        self,
+        num_hidden: int = 256,
+        num_hidden_layers: int = 1,
+        var_level: int = 0,
+        regularizer: Optional[tf.keras.regularizers.Regularizer] = None,
+    ):
         super(ScalarMLP, self).__init__()
         self.scalers_fitted = False
         self.sequential = tf.keras.Sequential()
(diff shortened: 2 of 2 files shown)


nbren12:
From #1265

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
```diff
@@ -82,6 +82,7 @@ class Config:
     output_path: str = ""
     weight_sharing: bool = False
     cloud_water: bool = False
+    l2: Optional[float] = None
```
mcgibbon, line 85: nit: If you set this as type `float` with default value 0, it could simplify the code somewhat, though this is less of a concern since Keras accepts None wherever regularizers are used.

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
```diff
@@ -156,6 +157,11 @@ def register_parser(parser):
             help="comma separated list of variable names.",
         )
 
+        # other options
+        parser.add_argument(
```
mcgibbon, line 161: nit: Is there a way to generate the args from the dataclass or vice-versa? The data here is currently duplicated which can become a source of bugs.

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
```diff
@@ -252,7 +259,9 @@ def step(self, in_, out):
         return info
 
     def get_loss(self, in_, out):
-        return self.config.target.loss(self.model(in_), out)
+        prediction = self.model(in_)
+        prediction_loss, info = self.config.target.loss(prediction, out)
+        return prediction_loss + sum(self.model.losses), info
```
mcgibbon, line 264: What this function does has changed but the name didn't - could you add a docstring describing this behavior?

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
```diff
@@ -343,17 +352,22 @@ def load(cls, path: str):
         return model
 
 
+def get_regularizer(config: Config) -> Optional[tf.keras.regularizers.Regularizer]:
```
mcgibbon, line 355: nit: Elsewhere you use a pattern of having properties on Config, e.g. `config.target.loss`. Does it make sense to have `config.regularizer` for consistency? This can be always defined if you use a regularizer with default norm of zero.

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
nbren12: This is a good idea for future work.

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
nbren12: It's true this is inconsistent, but there is an existing pattern of `get_model(config)` below. I'm not sure which is better in general. In this case `get_regularizer` is mostly a helper for `get_model`, so I'm not sure I want to make it more visible by putting it on the Config directly.

nbren12:
> With this change, you can set L2 in the config when using RHScalarMLP or ScalarMLP, and there is no error to the user - it just gets ignored. Can you please add an exception in that case?

Rather than adding an exception I implemented the feature.

nbren12:
And you are right that the using one config class for everything is not ideal.

external/fv3fit/fv3fit/emulation/thermobasis/emulator.py
mcgibbon: Let's keep it like this for this PR. For your future consideration, a benefit of using `config.regularizer` instead of `get_regularizer(config)` is that the former allows injection of new configs into functions of this package, whereas the latter requires you to update `get_regularizer` whenever you update `config`.

Honestly from what you're saying it sounds like there should be a `config.get_model` method. In the top-level infrastructure there is a `get_training_function` method, but that's because it takes a configuration string as an input, which isn't an object we can control and add methods to.

mcgibbon (approved):
LGTM

PR #1389, opened 2021-10-07
loaders: open_fine_resolution mapper

oliverwm1:
Need a mapper to be able to train models to predict the full fine-res Q1/Q2 tendencies and to evaluate them in the offline report. This mapper includes the (default) option to use the state from the fine-res as inputs (only T and q are available for now) or to provide a url to another coarse-res run from which states can be used as ML inputs.

Added public API:
- `loaders.open_fine_resolution`

- [x] Tests added


external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -215,3 +216,76 @@ def open_fine_resolution_nudging_hybrid(
...
+    if "longitude" in merged:
+        merged["longitude"] = merged.longitude.isel(time=0)
+
+    return merged.astype(numpy.float32).drop("tile")
+
+
+@mapper_functions.register
+def open_fine_resolution(
+    fine_url: str = "",
+    state_url: Optional[str] = None,
```
nbren12, line 270: "state" is a little vague here. Maybe mention "input_feature" somehow?

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -215,3 +216,76 @@ def open_fine_resolution_nudging_hybrid(
...
     )
+
+
+def open_fine_resolution_dataset(
+    fine_url: str = "gs://vcm-ml-experiments/default/2021-04-27/2020-05-27-40-day-X-SHiELD-simulation/fine-res-budget.zarr",  # noqa: E501
+    state_url: Optional[str] = None,
+    include_temperature_nudging: bool = False,
+) -> xarray.Dataset:
+
+    fine = open_zarr_maybe_consolidated(fine_url)
```
nbren12, line 227: There is a lot of copy paste here with the hybrid mapper above. There seem like some clear opportunities for resuse e.g. by defining some shared helpers like `assign_apparent_sources`. 

Also mixes I/O and computations. I would pull the zarr opening to the calling routine and making a pure function `assign_fine_res_sources` for the rest:
```
fine = open_zarr_maybe_consolidated(fine_url)
ds = assign_fine_res_sources(fine, **kwargs)
return XarrayMapper(ds)
```

Then bonus points for cleanly documenting the required variable names. You can probably use the protocol I defined here to do that: https://github.com/ai2cm/fv3net/blob/b9d97ea72c296b6b0eb830d4115599a2500f2496/external/loaders/loaders/mappers/_hybrid.py#L53

e.g.

```
def assign_fine_res_sources(fine: FineResSources) -> xr.Dataset
```

external/loaders/loaders/mappers/__init__.py
```diff
@@ -7,7 +7,7 @@
 )
 from ._transformations import ValMap, KeyMap, SubsetTimes
 from ._local import LocalMapper, mapper_to_local
-from ._hybrid import open_fine_resolution_nudging_hybrid
+from ._hybrid import open_fine_resolution_nudging_hybrid, open_fine_resolution
```
nbren12, line 10: nit: generally seems like a duplication of code to do all this renaming in multiple `__init__.py`. Makes it hard to follow what is defined where.

external/loaders/loaders/mappers/_hybrid.py
oliverwm1: Okay, I have reduced the copy paste as it relates to computing the fine-res sources and standardizing that dataset. I'd prefer not to make bigger changes (e.g. modifying I/O) since I think consistency with the `open_fine_resolution_nudging_hybrid_dataset` function is important.

PR #1392, opened 2021-10-08
fv3fit: clean up sample dim arg

AnnaKwa:
This removes the `sample_dim_name` arg from the initialization of most of the models where it was just an internal implementation detail used to stack and unstack data. The one exception to this is in the BPTT `StepwiseModel` where dataset input to `integrate_stepwise` already has a sample dimension which might not match the `fv3fit._shared.SAMPLE_DIM_NAME` used in other models.

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/recurrent.py
+++ b/external/fv3fit/fv3fit/keras/_models/recurrent.py
@@ -1,4 +1,4 @@
-from typing import Dict, Any, Iterable, Optional, Sequence, Tuple, List
+from typing import Dict, Any, Iterable, Optional, Sequence, Tuple, List, Hashable
 import tensorflow as tf
 import xarray as xr
 from ..._shared import (
@@ -10,7 +10,10 @@ from .packer import get_unpack_layer
 from .normalizer import LayerStandardScaler
 import numpy as np
 import vcm.safe
+import os
+import yaml
 from fv3fit.keras._models.shared import get_input_vector, PureKerasModel
+from ._filesystem import get_dir, put_dir
 
 Z_DIMS = ["z", "z_interface"]
 
@@ -430,12 +433,12 @@ class _BPTTTrainer:
     @property
     def predictor_model(self) -> "StepwiseModel":
         predictor_model = StepwiseModel(
-            self.sample_dim_name,
             tuple(self.input_packer.pack_names)
             + tuple(self.prognostic_packer.pack_names),
             self.output_variables,
             self._stepwise_output_metadata,
             self.predict_keras_model,
+            self.sample_dim_name,
         )
         return predictor_model
 
@@ -546,6 +549,17 @@ class StepwiseModel(PureKerasModel):
     include air_temperature and specific_humidity
     """
 
+    def __init__(
+        self,
+        input_variables: Iterable[Hashable],
+        output_variables: Iterable[Hashable],
+        output_metadata: Iterable[Dict[str, Any]],
+        model: tf.keras.Model,
+        sample_dim_name: str,
+    ):
+        super().__init__(input_variables, output_variables, output_metadata, model)
+        self.sample_dim_name = sample_dim_name
+
     def integrate_stepwise(self, ds: xr.Dataset) -> xr.Dataset:
         """Perform an identical integration to the one used to train the
         model.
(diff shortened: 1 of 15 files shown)


external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -546,6 +549,17 @@ class StepwiseModel(PureKerasModel):
     include air_temperature and specific_humidity
     """
 
+    def __init__(
+        self,
+        input_variables: Iterable[Hashable],
+        output_variables: Iterable[Hashable],
+        output_metadata: Iterable[Dict[str, Any]],
+        model: tf.keras.Model,
+        sample_dim_name: str,
```
AnnaKwa, line 558: This is the one model that still has sample dim as an arg, since it looks like its method to integrate forward in time works on a dataset that can have any name for its sample dim.

external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -430,12 +433,12 @@ def get_target_state(self, X):
     @property
     def predictor_model(self) -> "StepwiseModel":
         predictor_model = StepwiseModel(
-            self.sample_dim_name,
             tuple(self.input_packer.pack_names)
             + tuple(self.prognostic_packer.pack_names),
             self.output_variables,
             self._stepwise_output_metadata,
             self.predict_keras_model,
+            self.sample_dim_name,
```
mcgibbon, line 441: suggestion: if you update these to use keyword arguments for the variables (just here, you don't need to change the call signature) we can bugs from mis-ordered arguments.

external/fv3fit/fv3fit/keras/_models/recurrent.py
mcgibbon: This is fine.

external/fv3fit/fv3fit/keras/_models/recurrent.py
```diff
@@ -588,3 +602,39 @@ def integrate_stepwise(self, ds: xr.Dataset) -> xr.Dataset:
...
+                f.write(
+                    yaml.dump(
+                        {
+                            "input_variables": self.input_variables,
+                            "output_variables": self.output_variables,
+                            "output_metadata": self._output_metadata,
+                            "sample_dim_name": self.sample_dim_name,
+                        }
+                    )
+                )
```
mcgibbon, line 640: Why is this appearing as added code? Did you add it? If so then why, it seems unrelated?

workflows/prognostic_c48_run/tests/_regtest_outputs/test_machine_learning.test_ml_steppers_regression_checksum[tendencies].out
```diff
@@ -4,9 +4,9 @@
     - - dQ2
       - 8e405c2b4f212a44d5604ece97b291e2
     - - dQu
-      - 4d9fdc296ac232d82308a2f8e3d5e444
+      - 6da7a5b20608af0b0a30bc891861e807
     - - dQv
-      - 4d9fdc296ac232d82308a2f8e3d5e444
+      - 6da7a5b20608af0b0a30bc891861e807
```
Comment (mcgibbon) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_machine_learning.test_ml_steppers_regression_checksum[tendencies].out:9]: Do you have an explanation for why these changed? I don't understand why.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_schema[nudging].out
```diff
@@ -1,10 +1,10 @@
 xarray.Dataset {
 dimensions:
-	time = 2 ;
 	tile = 6 ;
-	z = 63 ;
-	y = 12 ;
+	time = 2 ;
 	x = 12 ;
+	y = 12 ;
+	z = 63 ;
```
Comment (mcgibbon) [inline workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_schema[nudging].out:7]: Inconsequential changes like this are fine.

mcgibbon (commented):
BPTT training still uses the old behavior (uses the first dimension of air_temperature in the dataset as the sample dimension name) but that's fine - I don't know what would happen if you changed it. The main need has been to update the `fv3fit.train` endpoints to remove these args.

I need to understand why dump/load was added to StepwisePredictor and have some more idea of why the regression targets changed (or at least why it's fine) before I can approve.

external/fv3fit/fv3fit/keras/_models/recurrent.py
mcgibbon: OK.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_schema[nudging].out
Comment (AnnaKwa) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_regression.test_fv3run_diagnostic_outputs_schema[nudging].out:7]: Thank you!! I was going nuts trying to think why the ConstantOutputPredictor output would have changed. I think this was it.

mcgibbon (approved):
LGTM

PR #1393, opened 2021-10-08
Generalize runtime.emulator.StepTransformer

oliverwm1:
This is a refactor PR that generalizes the `emulator` step transformer (previously "adapter") so that it will be able to work with other ML models. This is necessary for fine-res prediction. The one new feature is that a new diagnostic `..._due_to_applied_emulator` is available, which stores the actual updates applied to the state due to the emulator+physics instead of just those predicted by the emulator model (but possibly masked or not actually applied).

Added public API:
- new tendency/storage diagnostics available

Significant internal changes:
- step transformers no longer monitor the function that they are transforming
- internal refactors to generalize emulator model usage

- [x] Tests added


workflows/prognostic_c48_run/runtime/emulator.py
```diff
@@ -81,43 +108,35 @@ def monitor(self) -> Monitor:
...
         }
         before = self.monitor.checkpoint()
         diags = func()
         change_in_func = self.monitor.compute_change(name, before, self.state)
 
-        if self.config.train:
-            self.emulator.partial_fit(inputs, self.state)
-        emulator_prediction = self.emulator.predict(inputs)
+        if hasattr(self.model, "partial_fit"):
+            self.model.partial_fit(inputs, self.state)
```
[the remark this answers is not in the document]
oliverwm1: I think this is the right approach.

workflows/prognostic_c48_run/runtime/emulator.py
```diff
@@ -45,102 +71,120 @@ class Config:
...
-                ),
-            )
-            self.state.update(updated_state)
-
-        return {**diags, **inputs_to_save, **changes, **change_in_func}
+        change_due_to_ml = self.monitor.compute_change(self.label, before, prediction)
+        self.model.apply(prediction, self.state)
+        change_due_to_applied_ml = self.monitor.compute_change(
+            f"applied_{self.label}", before, self.state
+        )
```
nbren12, line 160: We cannot promise that these are actually the tendencies applied (since we could be inside of another `StepTransfomer`).  I would use another `monitor` to do this on the higher level of the code. Here is an example of how to do this:
```
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -224,7 +224,7 @@ class TimeLoop(
         elif self._emulate is not None and self._prescribe_tendency is None:
             return self._emulate(name, func)
         else:
-            return self.monitor(name, func)
+            return func
 
     def _get_prephysics_stepper(
         self, config: UserConfig, hydrostatic: bool
@@ -471,7 +471,12 @@ class TimeLoop(
                 self._step_prephysics,
                 self._compute_physics,
                 self._apply_postphysics_to_physics_state,
-                self.emulate_or_prescribe_tendency("fv3_physics", self._apply_physics),
+                self.monitor(
+                    "applied_physics",
+                    self.emulate_or_prescribe_tendency(
+                        "fv3_physics", self._apply_physics
+                    ),
+                ),
                 self._compute_postphysics,
                 self.monitor("python", self._apply_postphysics_to_dycore_state),
             ]:
```

nbren12 (changes requested):
Looks good to me overall, but I would pull the "applied_emulator" diagnostics further up the call stack to ensure that the name is accurate.

workflows/prognostic_c48_run/runtime/emulator.py
oliverwm1: Nice, yah that looks good.

nbren12: Likewise, the `fv3_physics` monitoring could be removed from this code as well, and implemented like this
```
self.monitor("applied_physics", self.emulate("emulator", self.monitor("fv3_physics", self._apply_physics)))
```

oliverwm1: Although i would prefer to not apply
```
-            return self.monitor(name, func)
+            return func
```
in order to maintain backwards compatibility with naming. e.g. in a non-emulation run still want to be able to output `...due_to_fv3_physics`

oliverwm1: It would just be equal to `applied_physics` in this case.

nbren12 (approved):
Looks great now. Thanks! Pretty awesome that we can compose things now with such small changes to the `TimeLoop`.

nbren12:
Would be great to refactor the report to use some of these new diagnostic names.

PR #1394, opened 2021-10-08
Add ConvolutionalNetwork and LossConfig classes

mcgibbon:
We are working to add a convolutional model to fv3fit. This PR adds some building blocks to make that happen, particularly new dataclasses ConvolutionalNetwork and LossConfig which can be configured from yaml.

LossConfig duplicates functionality present in slightly different form in two places - the Dense loss, which applies per-feature scaling, and the PrecipitativeModel loss, which scales based on the standard deviation across all features after removing the feature mean. It also adds tests for these different loss options. This new configuration class is assumed to be the loss for a single variable, so the interface for weighting is greatly simplified. Ideally, we should refactor the existing code to use this new configuration class in a later PR.

Added public API:
- Added ConvolutionalNetwork class to configure a sequence of convolutional layers
- Added LossConfig class to configure a loss function for a single variable
- Added numpy_print_precision context manager to `fv3fit.testing`

Significant internal changes:
- Updated unit test code to include "x" and "y" dimensions in test data

- [x] Tests added


external/fv3fit/fv3fit/keras/_models/shared/loss.py
```diff
@@ -0,0 +1,86 @@
...
+            tf.math.reduce_mean(tf.math.square((y_pred - y_true) / std), axis=0)
+        )
+
+    return custom_loss
+
+
+def _standard_scaled_mae(std):
+    std = tf.constant(std, dtype=std.dtype)
+
+    def custom_loss(y_true, y_pred):
```
AnnaKwa, line 20: Nit- standard and uniform_standard functions handle the custom loss differently- can we just handle them both the same way for consistency?

external/fv3fit/fv3fit/keras/_models/shared/loss.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Attributes:
+        loss_type: one of "mse" or "mae"
+        scaling: "standard" corresponds to scaling each feature's lossby
+            its scale, "standard_uniform" corresponds to scaling
+            each feature's loss by the mean of all feature scales, where
+            the scale is variance for MSE loss or standard deviation
+            for MAE loss
+        weight: A scaling factor by which to modify this loss
+    """
```
AnnaKwa, line 56: The `DenseModel` has a `normalize_loss` option that defaults to True- I don't know if we've ever used the `False` option, but you may want to keep it in the `LossConfig` and add a test.

external/fv3fit/fv3fit/keras/_models/shared/convolutional_network.py
```diff
@@ -0,0 +1,118 @@
...
+            config: configuration for convolutional network
+            label: inserted into layer names, if this function is used multiple times
+                to build one network you must provide a different label each time
+        
+        Returns:
+            tensors resulting from the requested convolutional network, each
+            tensor has the same dimensionality as the input tensor but will have fewer
+            points along the x- and y-dimensions, due to convolutions
+        """
+        if self.depth < 1:
```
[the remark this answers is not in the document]
mcgibbon: Yes, I'll refactor to use that pattern.

external/fv3fit/fv3fit/keras/_models/shared/loss.py
mcgibbon: What do you mean when you say they "handle the custom loss differently"? The configuration is meant to change the behavior of the loss function so that it uses individual std values or the same std value for all levels/features.

external/fv3fit/fv3fit/keras/_models/shared/loss.py
mcgibbon: I decided not to implement "no normalization" because we've never had a use case for it or even discussed possibly using it. When I added it to DenseModel I probably shouldn't have because of YAGNI.

external/fv3fit/fv3fit/keras/_models/shared/loss.py
```diff
@@ -0,0 +1,86 @@
...
+                loss = _uniform_scaled_mse(std)
+            elif self.scaling == "standard":
+                loss = _standard_scaled_mse(std)
+        elif self.loss_type == "mae":
+            if self.scaling == "standard_uniform":
+                loss = _uniform_scaled_mae(std)
+            elif self.scaling == "standard":
+                loss = _standard_scaled_mae(std)
+        else:
+            raise NotImplementedError(f"loss_type {self.loss_type} is not implemented")
```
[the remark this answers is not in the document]
mcgibbon: I should refactor this as well.

external/fv3fit/tests/keras/test_convolutional_network.py
```diff
@@ -0,0 +1,125 @@
...
+        pytest.param(
+            (3, 11, 15, 2),
+            3,
+            3,
+            5,
+            (3, 7, 11),
+            id="two_convolutions_different_input_shape",
+        ),
+        pytest.param(
+            (3, 10, 10, 2), 3, 2, 10, (3, 8, 8), id="one_convolution_more_kernel_size"
```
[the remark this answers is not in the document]
mcgibbon: Will fix, "larger filter" above should say "larger kernel" and this one should say "more filters"

PR #1395, opened 2021-10-11
fv3fit: add NoWeightSharingSLP layer

nbren12:
Emulators can more skillfully predict a single level than multiple levels.

Add  class for using independent weights for all output variables.

Added public API:
- fv3fit.emulation.layers.NoWeightSharingSLP


oliverwm1 (approved):
LGTM

PR #1399, opened 2021-10-12
Add internal_energy to vcm.DerivedMapping to allow monitoring

oliverwm1:
Add a quantity named `internal_energy`, computed as `c_v * T` to `vcm.DerivedMapping`. This allows easy monitoring of this quantity using our usual syntax during prognostic runs.

Added public API:
- `vcm.internal_energy` function
- `internal_energy` registered to `vcm.DerivedMapping`

Significant internal changes:
- added tendency/storage of internal energy to prognostic run regression test diagnostics

- [x] Tests added


workflows/prognostic_c48_run/runtime/metrics-schema.json
```diff
@@ -234,6 +234,31 @@
             "default": 0.0,
             "units": "kg/m^2/s"
         },
+        "storage_of_internal_energy_path_due_to_python": {
+            "type": "number",
+            "default": 0.0,
+            "units": "J/kg kg/m**2/s"
```
AnnaKwa, line 240: Units typo?

workflows/prognostic_c48_run/runtime/metrics-schema.json
```diff
@@ -234,6 +234,31 @@
...
+        },
+        "storage_of_internal_energy_path_due_to_emulator": {
+            "type": "number",
+            "default": 0.0,
+            "units": "J/kg kg/m**2/s"
+        },
+        "storage_of_internal_energy_path_due_to_applied_physics": {
```
AnnaKwa, line 257: Is "storage of X path" (also appears below for specific humidity) an atmospheric physics quantity I don't know, or a a typo in the schema?

AnnaKwa (approved):
LGTM, I had a couple questions about the schema.

workflows/prognostic_c48_run/runtime/metrics-schema.json
oliverwm1: Ideally it would be written as W/m**2 (which this is equivalent to) but since the `monitor.compute_change` function is quite general it does not know to do this simplification. It just adds the kg/m**2/s units (mass integral+tendency) to whatever units the input has. We could add some units package but that seems like overkill.

PR #1400, opened 2021-10-12
Add step transformer for standard fv3fit models

oliverwm1:
Add a config and adapter to be able to use standard `fv3fit` predictors within the FV3 physics step transformer. For example, this will allow fine-res ML prediction of Q1 and Q2 while the physics parameterizations update other state variables.

Added public API:
- new `Config` and `Adapter` within runtime.transformers.fv3fit

Refactored public API:
- changed the `online_emulator` item of `UserConfig` to be called `physics_machine_learning` and it can now either be a config to adapt an emulator model or a fv3fit predictor model.
- factory function will now return appropriate `StepTransformer` depending on value of `UserConfig.physics_machine_learning`

Significant internal changes:
- moved `runtime.tendency_prescriber` to `runtime.transformers.tendency_prescriber`
- moved `runtime.emulator` to `runtime.transformers.emulator` and refactored part of it to `runtime.transformers.core`

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/tests/test_regression.py
+++ b/workflows/prognostic_c48_run/tests/test_regression.py
@@ -594,7 +594,7 @@ def test_fv3run_diagnostic_outputs_schema(regtest, completed_rundir):
     diagnostics.info(regtest)
 
 
-def test_fv3run_python_mass_conserving(completed_segment, configuration):
+def test_metrics_valid(completed_segment, configuration):
     if configuration == ConfigEnum.nudging:
         pytest.skip()
 
@@ -608,6 +608,21 @@ def test_fv3run_python_mass_conserving(completed_segment, configuration):
     for metric in lines:
         obj = json.loads(metric)
         runtime.metrics.validate(obj)
+
+
+@pytest.mark.xfail
+def test_fv3run_python_mass_conserving(completed_segment, configuration):
+    if configuration == ConfigEnum.nudging:
+        pytest.skip()
+
+    path = str(completed_segment.join(STATISTICS_PATH))
+
+    # read python mass conservation info
+    with open(path) as f:
+        lines = f.readlines()
+
+    for metric in lines:
+        obj = json.loads(metric)
 
         np.testing.assert_allclose(
             obj["storage_of_mass_due_to_python"],


workflows/prognostic_c48_run/runtime/transformers/core.py
```diff
@@ -42,64 +30,6 @@ def partial_fit(self, inputs: State, state: State):
         pass
 
 
-@dataclasses.dataclass
```
oliverwm1, line 45: following two classes refactored to `runtime.transformers.emulator`

workflows/prognostic_c48_run/examples/fine_res_ml.yml
```diff
@@ -0,0 +1,20 @@
+base_version: v0.5
```
nbren12, line 1: What's up with these "examples"? If they are only used for testing, I would bury them in the tests, since their purpose is not quite clear in the present location.

workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -43,7 +56,9 @@ class UserConfig:
     scikit_learn: Optional[MachineLearningConfig] = None
     nudging: Optional[NudgingConfig] = None
     tendency_prescriber: Optional[TendencyPrescriberConfig] = None
-    online_emulator: Optional[runtime.emulator.Config] = None
+    physics_machine_learning: Optional[
```
nbren12, line 59: Approving since we did discuss this yesterday, but the configuration still feels a little off to me and prone to changing in the future as we add more ways of interacting with the physics:

1. still unclear to me why predicting a tendency  (fv3fit) vs a state update (emulator) cannot be done simultaneously, is there some conceptual reason for why this should be impossible? Ultimately, I expect the fv3fit and emulator models will share a common serialized format. Another example is that we may want to use the emulator's masking code for the "fv3fit" models. Can we structure the config in such a way that these future changes could be backward compatible? 
2. should be grouped with `tendency_prescriber` since they essentially act in similar ways. 

What do you think about the following:

```
physics:
  tendencies:
    prescribed: Optional[TendencyPrescriberConfig]
    predicted: Optional[runtime.transformers.fv3fit.Config]
  state_updates:
    predicted: Optional[runtime.emulator.Config]
```

workflows/prognostic_c48_run/tests/test_tendency_prescriber.py
```diff
@@ -85,5 +84,8 @@ def add_one():
         derived_state["air_temperature"],
         (derived_state_copy["air_temperature"] + 2).assign_attrs(units="degK"),
     )
-    for variable in sorted(diags):
-        print(variable, joblib.hash(diags[variable]), file=regtest)
```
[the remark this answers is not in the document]
nbren12: I would leave it if possible since these provide a lot of coverage. Are you running these tests in a recent version of the prognostic run image? It is easy to get out of sync by using the `latest` tag.


Also using `.values` helps (see a7c278998462819d2ff36910d2cc19b6663827ed)

```
    for variable in sorted(diags):
        print(variable, joblib.hash(diags[variable].values), file=regtest)
```

workflows/prognostic_c48_run/examples/fine_res_ml.yml
oliverwm1: Done

PR #1401, opened 2021-10-12
Add precomputed fine-res sources to catalog and fine-res and hybrid mappers for precomputed data

brianhenn:
Adds pre-computed coarsened fine-res apparent sources of heat and moisture (Q1 and Q2) from the May 2020 X-SHiELD run, both at original 15-minute and 3-hourly averaged time-resolution and with appropriate time chunking, to the vcm catalog.  Also adds new hybrid and fine-res loaders that can load a precomputed fine-res dataset.

Added public API:
- Adds `open_precomputed_fine_resolution_nudging_hybrid` and `open_precomputed_fine_resolution` to `loaders.mappers`


oliverwm1 (approved):
LGTM, I think this is the best approach of the ones we considered.

PR #1402, opened 2021-10-13
Allow monitored tendencies for ML diags in prog report

oliverwm1:
Allow use of monitored tendencies/storages for machine learning diagnostics in the prognostic run report. This will be useful for the fine-res ML prognostic runs. We may want to refactor our "standard" prognostic runs to use these names as well, but that is left for a possible future PR.


AnnaKwa (approved):
LGTM

PR #1404, opened 2021-10-13
fv3fit: sklearn RF serializing metadata as dict

### LINKED ISSUE #1403: fv3fit RF loading not backward compatible
Due to a quirk in the way that fv3fit RF models are saved, #1392 wasn't backward compatible for these models (it looks like it probably was for the other fv3fit models). Specifically this model class dumps a yaml list of metadata instead of a dict, which then requires tracking the number/sequence of items in the list, thus the breaking change in that PR. Probably should switch the metadata serialization to a dict and add backward compatible loading.

brianhenn:
The fv3fit `SklearnWrapper` model class dumped a yaml list of metadata instead of a dict, which then required tracking the number and sequence of the items in the list, creating overhead and making breaking changes likely. This PR switches the metadata serialization to a dict, and adds backward compatible loading. Also added a missing base class init call, which allows seeing the input and output variables of `SklearnWrapper` as public attributes.

Significant internal changes:

Resolves #1403

### DIFF
--- a/external/fv3fit/fv3fit/sklearn/_random_forest.py
+++ b/external/fv3fit/fv3fit/sklearn/_random_forest.py
@@ -213,11 +213,11 @@ class SklearnWrapper(Predictor):
             output_variables: list of output variables
             model: a scikit learn regression model
         """
+        super().__init__(input_variables, output_variables)
         self.model = model
         self.scaler_type = scaler_type
         self.scaler_kwargs = scaler_kwargs or {}
         self.target_scaler: Optional[scaler.NormalizeTransform] = None
-        # TODO: remove internal sample dim name once sample dim is hardcoded everywhere
         self._input_variables = input_variables
         self._output_variables = output_variables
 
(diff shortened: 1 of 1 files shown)


external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -310,35 +310,21 @@ def load(cls, path: str) -> "SklearnWrapper":
...
+            output_features_dict_ = metadata["output_features"]
         output_features_ = _tuple_to_multiindex(output_features_dict_)
 
         obj = cls(input_variables, output_variables, model)
         obj.target_scaler = scaler_obj
         obj.output_features_ = output_features_
 
         return obj
-
-    # these are here for backward compatibility with pre-unified API attribute names
```
brianhenn, line 325: This was from over a year ago.

oliverwm1:
It might be nice to generalize this a bit more to dump/load the full random forest hyperparameters. That way if someone added a new item to the hyperparameters (like config for vertical level slicing ;) ) they would automatically be dumped/loaded as necessary instead of needing to edit the serialization methods. 

See the dense keras model for an example:

https://github.com/ai2cm/fv3net/blob/a841fe501b449050ccc13ac9aa0e3acede1c030a/external/fv3fit/fv3fit/keras/_models/models.py#L320-L329

and

https://github.com/ai2cm/fv3net/blob/a841fe501b449050ccc13ac9aa0e3acede1c030a/external/fv3fit/fv3fit/keras/_models/models.py#L366-L372

It is a bit complicated here because of the RandomForest vs SklearnWrapper classes.

PR #1406, opened 2021-10-13
xfail the fv3run_python_mass_conserving test

oliverwm1:
For some reason (dependency change?) the `test_fv3run_python_mass_conserving` test has started failing even on PRs that are completed unrelated. This PR splits that test into two: one that does the metrics validation and one that does the mass check. It xfails the mass check one, although we should probably understand better why this is failing and remove the xfail when possible.


oliverwm1:
Also, the checksums for the 2D storages are not changing, suggesting it's either an issue with the global mean reduction or with the saving/loading of the globally averaged statistics to the log file.

PR #1410, opened 2021-10-14
Emulation layer  and microphsyics model loading/saving

frodre:
This PR is split off from the now closed #1407 to make easier on the review side.  It does a few things: 

1. Adds `get_config` methods to the architecture layers as a step in the direction of having all the layers be serializable without the call-trace.  
    - I had implementations  for the `fields.py` layers, but those ended up being much more complicated because they are a container for other layers.  Not essential, but I worked on this because I thought weights and biases needed it for incremental saving.  Turns out that's not true.
1. Added tests that at least confirm the more complicated field i/o layers are saveable in models and reproduce results after loading
1. Slight adjustments to the layer call signatures to either simplify arguments or set defaults
1.  Added `asdict` and `from_dict` methods (and tests) to the `MicrophysicsConfig` that handles conversion of lists to slices for feature subselection.  Slices are not serializable but the sequences are.
2. Add the ability to turn off/on limiting for the Microphysics models

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/config.py
+++ b/external/fv3fit/fv3fit/emulation/data/config.py
@@ -2,7 +2,7 @@ import dacite
 import dataclasses
 import logging
 from toolz.functoolz import compose_left
-from typing import Any, Dict, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, Mapping, Optional, Sequence
 
 from . import transforms
 
@@ -72,7 +72,9 @@ class TransformConfig:
 
     def __post_init__(self):
         if self.vertical_subselections is not None:
-            self.vert_sel_as_slices = {k: v.slice for k, v in self.vertical_subselections.items()}
+            self.vert_sel_as_slices = {
+                k: v.slice for k, v in self.vertical_subselections.items()
+            }
         else:
             self.vert_sel_as_slices = None
 
--- a/external/fv3fit/fv3fit/emulation/models/microphysics.py
+++ b/external/fv3fit/fv3fit/emulation/models/microphysics.py
@@ -1,6 +1,6 @@
 import dacite
 import dataclasses
-from typing import List, Mapping, Sequence, Optional
+from typing import List, Mapping, Sequence
 import tensorflow as tf
 
 from ._core import ArchitectureConfig, get_combine_from_arch_key
--- a/external/fv3fit/tests/emulation/models/test_microphysics_config.py
+++ b/external/fv3fit/tests/emulation/models/test_microphysics_config.py
@@ -44,8 +44,7 @@ def test_Config_asdict():
     sl1_kwargs = dict(start=0, stop=10, step=2)
     sl2_kwargs = dict(start=None, stop=25, step=None)
     sel_map = dict(
-        dummy_in=SliceConfig(**sl1_kwargs),
-        dummy_out=SliceConfig(**sl2_kwargs)
+        dummy_in=SliceConfig(**sl1_kwargs), dummy_out=SliceConfig(**sl2_kwargs)
     )
 
     original = MicrophysicsConfig(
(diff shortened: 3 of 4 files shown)


external/fv3fit/fv3fit/emulation/layers/fields.py
```diff
@@ -42,31 +45,65 @@ def call(self, tensor: tf.Tensor):
...
+            serializable_selection = [
+                self._selection.start,
+                self._selection.stop,
+                self._selection.step,
+            ]
+        else:
+            serializable_selection = None
+
+        config.update(
+            {"norm_layer": self.normalize, "selection": serializable_selection}
```
frodre, line 62: Normalize layers are already serializable so passing into the config dictionary prevents from having to re-fit the layers with a sample tensor.

external/fv3fit/fv3fit/emulation/data/config.py
```diff
@@ -25,9 +25,18 @@ def _sequence_to_slice(seq: Sequence[Union[None, int]]):
     return slice_
 
 
-def _convert_map_sequences_to_slices(map_: Mapping[str, Sequence[int]]):
+def convert_map_sequences_to_slices(map_: Mapping[str, Union[Sequence[int], slice]]):
```
nbren12, line 28: return type?

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -17,17 +17,25 @@ def __init__(
...
+        if self._expand_axis is not None:
             inputs = [
-                tf.expand_dims(tensor, axis=self.expand_axis) for tensor in inputs
+                tf.expand_dims(tensor, axis=self._expand_axis) for tensor in inputs
             ]
 
-        return tf.concat(inputs, axis=self.combine_axis)
+        return tf.concat(inputs, axis=self._combine_axis)
+
+    def get_config(self):
```
nbren12, line 32: not sure why these `get_config` methods are needed. Sounds like the call tracing approach is superior in terms of backwards compatibility since these classes won't need to be stable to load a saved model. did the ser/deser not work before?

The tensorflow [docs](https://www.tensorflow.org/tutorials/keras/save_and_load#saving_custom_objects) claim:
> If you are using the SavedModel format, you can skip this section. The key difference between HDF5 and SavedModel is that HDF5 uses object configs to save the model architecture, while SavedModel saves the execution graph. Thus, SavedModels are able to save custom objects like subclassed models and custom layers without requiring the original code.

external/fv3fit/fv3fit/emulation/models/microphysics.py
```diff
@@ -47,10 +49,26 @@ class MicrophysicsConfig:
     selection_map: Mapping[str, slice] = dataclasses.field(default_factory=dict)
     tendency_outputs: Mapping[str, str] = dataclasses.field(default_factory=dict)
     timestep_increment_sec: int = 900
+    enforce_positive: bool = True
 
     @classmethod
     def from_dict(cls, dict_) -> "MicrophysicsConfig":
-        return dacite.from_dict(cls, dict_, dacite.Config(strict=True))
```
nbren12, line 53: This special handling of the slice data types suggests that the user config is perhaps not lining up well with what the code implements. Can you modify the config to avoid the need for code like this?

external/fv3fit/fv3fit/emulation/data/config.py
```diff
@@ -25,9 +25,18 @@ def _sequence_to_slice(seq: Sequence[Union[None, int]]):
...
 
-    return {key: _sequence_to_slice(seq) for key, seq in map_.items()}
```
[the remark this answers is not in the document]
frodre: Switched it toward's Oli's representation

nbren12 (approved):
LGTM

PR #1411, opened 2021-10-14
Skip prognostic run if no config provided to e2e workflow

oliverwm1:
This is useful when wanting to take advantage of the training+offline part of the e2e workflow without doing a prognostic run.

If user does not provide a `prognostic-run-config` parameter to the `train-diags-prog.yaml` workflow then the prognostic run will be skipped.

### DIFF
--- a/workflows/argo/train-diags-prog.yaml
+++ b/workflows/argo/train-diags-prog.yaml
@@ -24,11 +24,11 @@ spec:
       - name: training-data-config
       - name: validation-data-config
       - name: initial-condition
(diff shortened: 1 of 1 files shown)


nbren12:
@oliverwm1  Umm...I'm a bit rusty on argo, but this logic seems to do the opposite of what you want. It appears to run the prognostic run when `skip-prognostic-run` is true.
```
      - name: prognostic-run
        when: "{{inputs.parameters.skip-prognostic-run}} == true"
```

PR #1413, opened 2021-10-14
refactor appendSegment run to python

nbren12:
The runfv3 bash script is hard to maintain and extend, we have been slowly refactoring it to python. This PR finishes the job.


workflows/post_process_run/fv3post/post_process.py
```diff
@@ -158,6 +158,10 @@ def post_process(rundir: str, destination: str, chunks: str, skip: str):
     This script rechunks the python zarr output and converts the netCDF
     outputs to zarr.
     """
+    _post_process(rundir, destination, chunks, skip)
+
+
+def _post_process(rundir: str, destination: str, chunks: str, skip: str):
```
AnnaKwa, line 164: It looks kind of strange to have the function `_post_process` be imported outside of here in the prognostic run workflow, can this function be `post_process` and the version that is used in the `post_process_run` entrypoint be called something else? `post_process_entrypoint`?

PR #1415, opened 2021-10-15
Refactor ArrayPacker to use xr.to_stacked_array

oliverwm1:
Refactor `ArrayPacker` to use the `pack` and `unpack` functions which use the xarray `to_stacked_array` and `to_unstacked_dataset` functions. This is mostly deleting a bunch of code and modifying the serialization of the `ArrayPacker`.

Refactored public API:
- The serialization format of `ArrayPacker` has changed in a non backwards compatible way

Significant internal changes:
- use `pack` and `unpack` instead of custom `to_array` and `to_dataset` functions

- [x] Tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,10 +211,26 @@ workflows:
             branches:
               ignore:
                 - master
+      # manually trigger on PRs
       - pytest:
+          name: "Test Dataflow (held)"
           kind: test_dataflow
           requires:
             - hold
+          filters:
+            tags:
+              ignore: /^v.*/
+            branches:
+              ignore: master
+      # automatically trigger on master
+      - pytest:
+          name: "Test Dataflow (not held)"
+          kind: test_dataflow
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              only: master
       - build_default:
           matrix:
             parameters:
@@ -223,36 +239,17 @@ workflows:
                 - post_process_run
                 - fv3net
                 - artifacts
+      # manually trigger on PRs
       - integration_tests:
           requires:
             - build_default
             - hold
-  master_tests:
-    jobs:
-      - lint
-      - mypy
-      - argo
-      - pytest:
-          kind: test_dataflow
-          name: "Test Dataflow"
           filters:
             tags:
-              only: /^v.*/
+              ignore: /^v.*/
             branches:
-              only: master
-      - build_default:
-          matrix:
-            parameters:
-              image:
-                - prognostic_run
-                - post_process_run
-                - fv3net
-                - artifacts
-          filters:
-            tags:
-              only: /^v.*/
-            branches:
-              only: master
+              ignore: master
+      # automatically trigger on master
       - integration_tests:
           requires:
             - build_default
--- a/external/diagnostics_utils/diagnostics_utils/utils.py
+++ b/external/diagnostics_utils/diagnostics_utils/utils.py
@@ -4,7 +4,7 @@ from .config import (
     DOMAINS,
     PRIMARY_VARS,
 )
-from vcm import thermo, safe, mass_integrate
+from vcm import thermo, safe, mass_integrate, weighted_average
 import xarray as xr
 import numpy as np
 import logging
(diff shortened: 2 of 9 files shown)


external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -163,108 +156,18 @@ def dump(self, f: TextIO):
...
         data = yaml.safe_load(f.read())
         packer = cls(data["sample_dim_name"], data["pack_names"])
-        packer._n_features = data["n_features"]
-        packer._dims = data["dims"]
+        packer._feature_index = tuple_to_multiindex(data["feature_index"])
+        packer._n_features = _count_features(packer._feature_index)
         return packer
 
 
-def _to_array_2d(
```
mcgibbon, line 171: Yay deleted code!

external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -163,108 +156,18 @@ def dump(self, f: TextIO):
...
-            raise ValueError(
-                f"cannot pack value for {name} whose first dimension is not the "
-                f"sample dimension ({sample_dim_name}), has dims {value.dims}"
-            )
-        else:
-            return_dict[name] = value.shape[1]
-    return return_dict
+def _count_features(index: pd.MultiIndex, variable_dim="variable") -> Mapping[str, int]:
+    variable_idx = index.names.index(variable_dim)
+    count = {}
```
mcgibbon, line 166: You could use `collections.defaultdict` for this.

external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -39,7 +37,7 @@ def pack(data: xr.Dataset, sample_dim: str) -> Tuple[np.ndarray, pd.MultiIndex]:
     feature_dim_name = _unique_dim_name(data, sample_dim)
     stacked = data.to_stacked_array(feature_dim_name, sample_dims=[sample_dim])
     return (
-        stacked.transpose(sample_dim, feature_dim_name).data,
+        stacked.transpose(sample_dim, feature_dim_name).values,
```
oliverwm1, line 40: without this change, `pack` might return a `dask` array for example. We want `numpy.ndarray` in all cases.

external/fv3fit/fv3fit/_shared/io.py
```diff
@@ -30,6 +32,12 @@ def _register_class(self, cls: Type[Predictor], name: str) -> Type[Predictor]:
         return cls
 
     def _load_by_name(self, name: str, path: str) -> Predictor:
+        if name in DEPCRECATED_NAMES:
```
mcgibbon, line 35: This is really nice, I'd say keep this.

external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -72,7 +80,7 @@ def __init__(self, sample_dim_name, pack_names: Iterable[Hashable]):
         self._pack_names: List[str] = list(str(s) for s in pack_names)
         self._n_features: Dict[str, int] = {}
         self._sample_dim_name = sample_dim_name
-        self._dims: Dict[str, Sequence[str]] = {}
+        self._feature_index: pd.MultiIndex = pd.MultiIndex(levels=[[]], codes=[[]])
```
mcgibbon, line 83: Is this default value useful? If not, can you please make this Optional and add logic to raise an exception if it isn't initialized? Otherwise this could be used and cause silent errors. If you expect this to cause loud errors when used, that can be OK as well.

If you want to avoid putting if checks, you can make a `feature_index` property method which raises an exception when accessed if this hasn't been initialized.

external/fv3fit/fv3fit/train.py
```diff
@@ -53,9 +53,9 @@ def dump_dataclass(obj, yaml_filename):
 
 def main(args):
     with open(args.training_config, "r") as f:
-        training_config = fv3fit.TrainingConfig.from_dict(yaml.load(f))
+        training_config = fv3fit.TrainingConfig.from_dict(yaml.safe_load(f))
```
mcgibbon, line 56: Nice catch, this was only needed before because of the xarray serialization involved with the previous packer.

external/fv3fit/tests/test_packer.py
```diff
@@ -98,8 +98,7 @@ def test_to_dataset(names, dims_list, array: np.ndarray):
     packer = ArrayPacker(SAMPLE_DIM, names)
     packer.to_array(dataset)  # must pack first to know dimension lengths
     result = packer.to_dataset(array)
-    # to_dataset does not preserve coordinates
-    xr.testing.assert_equal(result, dataset.drop(dataset.coords.keys()))
+    xr.testing.assert_equal(result, dataset)
```
mcgibbon, line 101: Nice!

mcgibbon (commented):
Looks great to me! Thanks!

external/fv3fit/fv3fit/_shared/packer.py
oliverwm1: This isn't really a new issue... the same could be said for the other attrs like `self._dims` and `self._n_features` which only really get 'initialized' on the first `to_array` call. Anyway, I'll raise an exception if someone tries to unpack before packing, but I would like to refactor the API of ArrayPacker as we discussed over slack to remove the possibility of this issue anyway.

PR #1419, opened 2021-10-15
Improve cache use

nbren12:
f3a49e6 appears to have broken caching on the fv3net and prognostic run images. The other images (e.g. artifacts) have not had any code updated since f3a49e6, so their caches still work.

Because the issue affects the fv3net image too it is probably not the changes to the prognostic run dockerfile. The commit also changed the makefile and circle builds. This suggests the order of tagging/building/pushing docker images is somehow important for cache performance.

After a few hours, I still don't understand what the issue is so I am shooting from the hip. This PR reverts most of the makefile and circle ci changes back to feadec2, and then makes a more minimal change to call the prog run tests that should be less likely to break things. I suggest reviewing commit-by-commit.

We won't know if this works until this is merged to master.

### DIFF
--- a/.circleci/build_and_push_image.sh
+++ b/.circleci/build_and_push_image.sh
@@ -13,6 +13,7 @@ fi
 
 apt-get install -y make jq
 make build_image_$IMAGE
+make image_test_$IMAGE
 
 
 if [[ "$CIRCLE_BRANCH" == "master" ]]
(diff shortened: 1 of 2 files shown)


oliverwm1 (approved):
Agreed it's not obvious why the caching stopped working or why this would help, but worth a shot.

PR #1425, opened 2021-10-18
fix integration test

nbren12:
vcm.to_url uses the gcs:// prefix for GCSfs paths, but fv3config requires "gs://".


AnnaKwa (approved):
LGTM

PR #1428, opened 2021-10-19
prog run: inject monitored diagnostics into state

nbren12:
The dynamics tendency could be a valuable ML input variable, but currently we cannot pass it to an ML model online.

Significant internal changes:
- added a class `runtime.derived_state.MergedState` that provides a uniform interface to python and fortran state. This class can be used to store arbitrary data arrays for future use.
- monitored output tendencies are available in `loop._state`


oliverwm1 (approved):
LGTM. From a user-perspective, it could be helpful to have methods to know which keys correspond to Fortran state vs python state. Although I suppose users could directly query `MergedState.left` or `MergedState.right` as necessary.

PR #1429, opened 2021-10-19
Add "convolutional" training to fv3fit

mcgibbon:
This PR takes the ConvolutionalNetwork component and defines a "convolutional" training function to use it.

Added public API:
- Added "convolutional" training type to fv3fit
- Added `stack` function similar to `stack_non_vertical` which allows you to specify which dimensions to leave unstacked, and supports multiple unstacked dimensions
- Added ConvolutionalHyperparameters to public API

Refactored public API:
- Convolutional network now uses "same" padding, so outputs are all same size as inputs
- Made transpose invariance optional when configuring convolutional network
- Fixed bug in transpose-invariant constraint where it assumed an incorrect kernel shape
- Increased defaults for convolutional network based on training result
- Refactored "pack" and "unpack" to take sample_dims (sequence of str) instead of sample_dim (str), supporting outputs with multiple sample dimensions
- Refactored StackedBatches and PureKerasModel to take an additional optional argument `unstacked_dims` which can be set to stack data with multiple sample dimensions, for example to be output as [stacked_dim, x, y, z] instead of only as [stacked_dim, z]
- Updated emulation normalization layers to support arbitrarily dimensioned data, so long as the last dimension is the sample dimension.

- [x] Tests added


external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -115,8 +116,96 @@ def predict(self, X: xr.Dataset) -> xr.Dataset:
...
+    _MODEL_FILENAME = "model.tf"
+    _CONFIG_FILENAME = "config.yaml"
+
+    def __init__(
+        self,
+        input_variables: Sequence[Hashable],
+        output_variables: Sequence[Hashable],
+        output_metadata: Iterable[Dict[str, Any]],
+        model: tf.keras.Model,
+        loss=None,
```
AnnaKwa, line 152: Does the Predictor need the loss? It doesn't look like it's dumped or loaded.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -115,8 +116,96 @@ def predict(self, X: xr.Dataset) -> xr.Dataset:
...
+                            "input_variables": self.input_variables,
+                            "output_variables": self.output_variables,
+                            "output_metadata": self._output_metadata,
+                        }
+                    )
+                )
+
+
+@io.register("all-keras-no-stack")
+class PureKerasNoStackModel(Predictor):
```
AnnaKwa, line 140: Is there a reason to have a separate class for the 'no stacking' predictor rather than a single `PureKerasModel` class with an optional boolean arg to not stack and use a different `*_prediction_to_dataset` function depending on the flag? Is it just considered cleaner?

external/fv3fit/fv3fit/keras/_models/shared/convolutional_network.py
```diff
@@ -21,8 +21,9 @@ class TransposeInvariant(tf.keras.constraints.Constraint):
     """Constrains `Conv2D` kernel weights to be unchanged when transposed."""
 
     def __call__(self, w: tf.Tensor):
-        # Conv2D kernels are of shape (kernel_column, kernel_row, channels)
-        return tf.scalar_mul(0.5, w + tf.transpose(w, perm=(1, 0, 2)))
+        # Conv2D kernels are of shape
+        # (kernel_column, kernel_row, input_channel, output_channel)
+        return tf.scalar_mul(0.5, w + tf.transpose(w, perm=(1, 0, 2, 3)))
```
AnnaKwa, line 26: Is this tested anywhere?

external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -0,0 +1,194 @@
...
+        model=predict_model,
+    )
+    return predictor
+
+
+def batch_to_array_tuple(
+    batch: xr.Dataset, input_variables: Sequence[str], output_variables: Sequence[str]
+) -> Tuple[Tuple[np.ndarray, ...], Tuple[np.ndarray, ...]]:
+    return (
+        tuple(batch[name].values for name in input_variables),
```
AnnaKwa, line 109: Maybe I missed this somewhere else, is there an assumption here about the dimension ordering in the dataset such that the last dimension is z? Is the dimension order guaranteed to be (..., z)?

AnnaKwa: I think the no-stacking limitation is fine as long as the assumptions are clearly documented. I was mainly wondering if it was possible for a dataset to be provided as input with some other ordering than x, y, z and have this result in some silent bug in the results or fail with a confusing error.

mcgibbon: I had to add the stacking anyways to get the tests to pass after changing some other things, so this is no longer an issue. it still assumes the non-stacked dimension ordering.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
mcgibbon: I ended up deleting this class since I had to add stacking to this model.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
mcgibbon: This class is now deleted.

external/fv3fit/fv3fit/keras/_models/shared/convolutional_network.py
mcgibbon: Test added, passes.

mcgibbon (commented):
Should be ready for re-review!

PR #1432, opened 2021-10-19
Add mypy to pre-commit hooks

mcgibbon:
I frequently find my PRs fail type checking only after I push them, since mypy checks are currently run outside of pre-commit. We planned a while back to enable this in pre-commit. This PR accomplishes that, with minimal changes to mypy configuration.

Significant internal changes:
- Mypy checks now run as part of pre-commit hooks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,28 @@ repos:
       files: "__init__.py"
       # ignore unused import error in __init__.py files
       args: ["--ignore=F401,E203", --config, setup.cfg]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.770
+    hooks:
+    - id: mypy
+      args: [
+          --no-strict-optional,
+          --ignore-missing-imports,
+          --follow-imports, silent  # needed so we only test enabled files
+        ]
+      files: |
+        (?x)^(
+        external/vcm/vcm/.+ |
+        workflows/dataflow/fv3net/pipelines/restarts_to_zarr/.+ |
+        workflows/prognostic_c48_run/.+ |
+        workflows/prognostic_c48_run/tests/.+ |
+        external/fv3fit/fv3fit/.+ |
+        external/loaders/loaders/.+ |
+        workflows/diagnostics/fv3net/diagnostics/offline/.+
+        )$
+      exclude: |
+        (?x)^(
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -185,8 +185,7 @@ setup-hooks:
 	pip install -c constraints.txt pre-commit
 	pre-commit install
 
-typecheck:
-	./check_types.sh
+typecheck: lint
```
nbren12, line 188: could you restore the previous behavior (e.g. just type-checking but not linting). It's nice to be able to check types on unformatted code.

.pre-commit-config.yaml
```diff
@@ -25,3 +25,26 @@ repos:
...
       args: ["--ignore=F401,E203", --config, setup.cfg]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.770
+    hooks:
+    - id: mypy
+      args: [
+          --follow-imports, silent
+        ]
+      files: |
+        (?x)^(
```
nbren12, line 36: I'm not too familiar with this syntax, but I have verified that `make typecheck` fails when I introduce the following fault:
```
diff --git a/workflows/prognostic_c48_run/runtime/loop.py b/workflows/prognostic_c48_run/runtime/loop.py
index 368d42ba..8222022b 100644
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -50,6 +50,8 @@ from .names import AREA, DELP, TOTAL_PRECIP
 
+a: int = "a"
+
 
 class Stepper(Protocol):
     """Stepper interface
```

mcgibbon: Yeah the syntax is confusing, basically it takes any regex. This regex in particular matches the regex patterns on each line below, which generally mean "all files under these directories".

That should cause a failure, shouldn't it? `workflows/prognostic_c48_run` was passed as an argument to mypy before.

Makefile
mcgibbon: Done.

.pre-commit-config.yaml
nbren12: I was just verifying the code worked yah. More of an FYI.

nbren12:
I'll disable the mypy requirement so that this can merge.

PR #1435, opened 2021-10-20
Allow monitoring derived quantities in step transformers

### LINKED ISSUE #1433: Derived quantities not properly monitored in step transformers
Tendencies monitored within a step transformer are manually computed as:

https://github.com/ai2cm/fv3net/blob/09c11a8d405e9fb00c94b9ca14ffb81d0e98eaae/workflows/prognostic_c48_run/runtime/transformers/core.py#L81-L83

instead of taking a `checkpoint` before and after. This means derived quantities such as `internal_energy` will have tendencies equal to zero (even when air temperature is being predicted) because they are not in the `prediction` dict.

oliverwm1:
It is useful to be able to monitor derived quantities (such as `internal_energy`). Currently the monitored tendencies of derived quantities due to a "transformed" step (see [here](https://github.com/ai2cm/fv3net/blob/09c11a8d405e9fb00c94b9ca14ffb81d0e98eaae/workflows/prognostic_c48_run/runtime/transformers/core.py#L81-L83)) will be zero because the "after" state provided to the `compute_change` function is a static dict.

Significant internal changes:
- wrap the "after" state provided to `compute_change` with `vcm.DerivedMapping`
- to test, added the possibility to include non-zero dQ1 tendencies to the ML mock in the prognostic run tests

- [x] Tests added

Resolves #1433

### DIFF
--- a/workflows/prognostic_c48_run/docs/conf.py
+++ b/workflows/prognostic_c48_run/docs/conf.py
@@ -27,7 +27,12 @@ author = "Noah D. Brenowitz"
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ["sphinx.ext.autodoc", "sphinx.ext.napoleon", "sphinx.ext.graphviz"]
+extensions = [
+    "sphinx.ext.autodoc",
+    "sphinx.ext.napoleon",
+    "sphinx.ext.graphviz",
+    "sphinx.ext.doctest",
+]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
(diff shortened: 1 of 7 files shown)


workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -1,26 +1,49 @@
 import joblib
 import fv3fit
+import vcm
+import xarray as xr
 from runtime.transformers.fv3fit import Config, Adapter
 from runtime.transformers.core import StepTransformer
 from machine_learning_mocks import get_mock_predictor
 
 
+class MockDerivedState:
```
nbren12, line 10: Probably worth adding setters for the derived mapping at some point.

nbren12 (approved):
Cool. Thanks for putting this together.

PR #1437, opened 2021-10-20
offline diags: use first timestep from test dataset, not full mapper, in snapshots

AnnaKwa:
Recent changes inadvertently made the snapshot diagnostics use the first timestep in the available data, instead of the first timestep in the test dataset as specified by the `config.kwargs["timesteps"]`. This PR fixes the snapshots to use the latter.

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
@@ -359,7 +359,8 @@ def main(args):
     mapper = _get_data_mapper_if_exists(config)
     if mapper is not None:
         snapshot_time = (
-            args.snapshot_time or config.kwargs.get("timesteps", list(mapper.keys()))[0]
+            args.snapshot_time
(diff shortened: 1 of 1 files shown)


workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -358,7 +358,9 @@ def main(args):
 
     mapper = _get_data_mapper_if_exists(config)
     if mapper is not None:
-        snapshot_time = args.snapshot_time or sorted(list(mapper.keys()))[0]
+        snapshot_time = (
+            args.snapshot_time or config.kwargs.get("timesteps", list(mapper.keys()))[0]
```
oliverwm1, line 362: Should `config.kwargs.get("timesteps", list(mapper.keys()))` be sorted?

oliverwm1 (approved):
I'm approving—but should there be a `sorted` call?

PR #1438, opened 2021-10-20
Emulation run from spun up ic

nbren12:
This is a convenience runscript for running experiments with microphysics. This is needed for two reasons
1. `prepare_config` does not support initializing the model from restarts.
2. Microphysics emulation is configured using environment variables rather than reading data from `fv3config.yml`.

In principle this script would not be required if we can address those two issues, but I placed it in `projects/microphysics` to clarify its experimental nature.

Other new public APIS:
- python equivalents of `runfv3` commands now in `runtime.segmented_run.api`

### DIFF
--- a/projects/microphysics/README.md
+++ b/projects/microphysics/README.md
@@ -11,7 +11,7 @@ WANDB_API_KEY=<wandb api key>
 EOF
 
 # login to google cloud
-gcloud auth login application-default login
+gcloud auth application-default login
 ```
 
 Pull docker image


projects/microphysics/scripts/fv3config.yml
```diff
@@ -0,0 +1,342 @@
+base_version: v0.5
```
nbren12, line 1: I downloaded this from one of andre's runs.

PR #1440, opened 2021-10-21
Fix-exit-code-handling

### LINKED ISSUE #1420: FV3 exit codes not properly handled and FV3 exits 1
This is actually two related bugs:

- The prognostic runs in the tests all actually exit with status 1 because python objects are not cleaned up before `wrapper.cleanup` is called.  (cherry-pick 9352fe1535a71e7dacd330871042d4986eb99f39 to fix this)
- `runfv3` does not properly handle the exit code

Fixing these bugs will change the regression data because the final timestep of the run is not being currently being saved.

This was found in #1413

nbren12:
Resolves #1420. see commits for more details. This bug affected the regression data since it caused the final timestep to not be saved.


oliverwm1:
The changes in this PR confuse me. The regression test is configured to run for 30 minutes, with a 15-minute timestep. Since we save data at the end of each timestep, there should be 2 timesteps in the output data. @nbren12

PR #1443, opened 2021-10-21
prog_run_diags: Merge 'physics' and 'dycore' to a single 2d dataset

oliverwm1:
The distinction between 'physics' and 'dycore' diagnostic sin the prog run report code is only marginally useful. (It is somewhat useful if we output physics diagnostics more often than dycore diagnostics, which is true for the may2020 shield data). The distinction prevents computing diagnostics that depend on both 'physics' and 'dycore' data (like 2D joint histogram of PWAT vs Q2). Therefore, this PR merges both categories into a single "2d" category.

Refactored public API:
- delete `.physics` and `.dycore` from `SegmentedRun` and `CatalogSimulation` and replace with `.data_2d`

Significant internal changes:
- replace 'physics' and 'dycore' registries in `compute.py` with a single 2D registry

An example report generated with these changes is [here](https://storage.googleapis.com/vcm-ml-public/oliwm/test-merged-report/index.html).


AnnaKwa (approved):
Thanks, LGTM! It's nice how this cuts down on a lot of repeats of functions getting registered in the compute module.

PR #1457, opened 2021-10-27
Allow clipping of input variables in random forest models

oliverwm1:
It is useful to be able to ignore some vertical levels as inputs. This implements this feature for random forest models. With a bit more work, this could also work for outputs and for dense keras models.

Added public API:
- `PackerConfig` which can be supplied as part of `RandomForestHyperparameters`

The config would be specified as:
```
hyperparameters:
  ...
  packer_config:
    clip:
      air_temperature:
        z:
          start: 10
          step: null
          stop: null
```

This is a bit verbose, so open to suggestions. I struggled to implement via dataclasses in a more compact but still general way.

Significant internal changes:
- if `PackerConfig.clip` is not empty, do appropriate selection of slices of given variables/dimensions

- [x] Tests added


mcgibbon:
The only way I can think to make the API more compact is to remove the "dimension" as an argument here and assume it's always the z-dimension (whichever of "z" or "z_interface" is present in the data).

external/fv3fit/tests/test_packer.py
```diff
@@ -188,6 +189,49 @@ def test_sklearn_unpack(dataset: xr.Dataset):
...
+
+
+def test_clip(dataset: xr.Dataset):
+    name = list(dataset.data_vars)[0]
+    if FEATURE_DIM in dataset[name].dims:
+        indices = {name: {FEATURE_DIM: SliceConfig(4, 8)}}
+        clipped_data = clip(dataset, indices)
+        expected_da = dataset[name]
+        for dim, slice_config in indices[name].items():
+            expected_da = expected_da.isel({dim: slice_config.slice})
```
mcgibbon, line 215: This change would let the test pick up errors in `slice_config.slice` implementation.
```suggestion
            expected_da = expected_da.isel({dim: slice(4, 8)})
```

external/fv3fit/tests/test_packer.py
```diff
@@ -188,6 +189,49 @@ def test_sklearn_unpack(dataset: xr.Dataset):
...
+    )
+    clip_config = {
+        "var1": {FEATURE_DIM: SliceConfig(4, 8)},
+        "var2": {FEATURE_DIM: SliceConfig(None, 6, 2)},
+    }
+    clipped_data = clip(ds, clip_config)
+    for name in ds:
+        expected_da = ds[name]
+        for dim, slice_config in clip_config[name].items():
+            expected_da = expected_da.isel({dim: slice_config.slice})
```
mcgibbon, line 231:
```suggestion
    clip_config = {
        "var1": {FEATURE_DIM: SliceConfig(4, 8)},
        "var2": {FEATURE_DIM: SliceConfig(None, 6, 2)},
    }
    slices = {"var1": slice(4, 8), "var2": slice(0, 6, 2)}
    clipped_data = clip(ds, clip_config)
    for name in ds:
        expected_da = ds[name]
        for dim, slice_config in clip_config[name].items():
            expected_da = expected_da.isel({dim: slices[name]})
```

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -210,3 +211,55 @@ def test_predict_returns_unstacked_dims():
...
+    arr = np.arange(np.prod(shape)).reshape(shape)
+    input_data = xr.Dataset(
+        {"a": (dims, arr), "b": (dims[:-1], arr[:, :, 0]), "c": (dims, arr + 1)}
+    )
+    wrapper.fit([input_data])
+    wrapper.predict(input_data)
+
+
+# clipping for ML outputs is not yet implemented
+def test_SklearnWrapper_fit_with_clipped_output_data():
```
mcgibbon, line 240: What does this comment mean? It looks like it's implemented based on this test existing?

Should the test be named `fit_predict` for consistency with the test above?

mcgibbon (approved):
Just some minor suggestions e.g. on default handling. LGTM.

external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -39,7 +41,7 @@ def _unique_dim_name(
 
 
 def pack(
-    data: xr.Dataset, sample_dims: Sequence[str]
+    data: xr.Dataset, sample_dims: Sequence[str], config: Optional[PackerConfig] = None,
```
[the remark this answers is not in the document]
oliverwm1: I think I already merged in your changes.

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -42,7 +45,8 @@ def _parse_metadata_backward_compatible(metadata: dict) -> tuple:
         input_variables = metadata["input_variables"]
         output_variables = metadata["output_variables"]
         output_features_tuple = metadata["output_features"]
-    return input_variables, output_variables, output_features_tuple
+        packer_config = metadata.get("packer_config", {})
```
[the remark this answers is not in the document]
oliverwm1: The casting to the `PackerConfig` occurs at the higher level func—here it would just be a dict if it is in `metadata`.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -314,6 +334,7 @@ class RandomForestHyperparameters(Hyperparameters):
     max_features: Union[str, int, float] = "auto"
     max_samples: Optional[Union[int, float]] = None
     bootstrap: bool = True
+    packer_config: Optional[PackerConfig] = None
```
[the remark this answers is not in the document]
oliverwm1: Okay, yeah I had ensured backwards compatibility by checking `if packer_config is None` during init, but this is cleaner.

external/fv3fit/tests/test_packer.py
oliverwm1: True. Although that is tested elsewhere already.

external/fv3fit/tests/test_sklearn_wrapper.py
oliverwm1: Okay I slimmed down the test to just check for raising NotImplementedError. There was some crud in there from earlier work where I was trying to implement for outputs too.

external/fv3fit/tests/test_packer.py
oliverwm1: I'm going to skip this, since the `.slice` property of `SliceConfig` is already tested.

oliverwm1:
> The only way I can think to make the API more compact is to remove the "dimension" as an argument here and assume it's always the z-dimension (whichever of "z" or "z_interface" is present in the data).

I think I will just leave it verbose, since currently the packing does not make any assumptions about feature dimension name, so I don't want to introduce one.

mcgibbon:
OK, though keep in mind if you ever do work with non-z dimensions that the stacking does make a default assumption that the feature dimension name is a z-dim. This came up for me in my last PR (where there wasn't even an option to override).

PR #1461, opened 2021-10-29
Offline diags remove diagnostics utils

AnnaKwa:
This refactors the time mean, conditional averaging, and diurnal cycle calculations to use the registry instead of functions from `diagnostics_utils`. Plotting functions for these diagnostics are relocated out of `diagnostics_utils` (no changes, just moving the code) into `diagnostics.offline.views.plot`. 

This also adds total water path as a derived mapping variable, which will be computed using the specific humidity if it is not available in the dataset. If dQ2 or Q2 is a predicted variable, water vapor path will be loaded so that a scatter plot of Q2 vs water vapor path can be added to the report.

There is a minor change to `fv3fit.DerivedModel` to facilitate wrapping an existing `DerivedModel` with another one while keeping the underlying base (non-derived) model the same.

I'm pretty sure these changes delete all current usages of the `diagnostics_utils`, but I'll leave it to a follow up PR to delete it because it's still referenced in a `training_data_diags` workflow. I think that can also be deleted but that would make this PR too big.

- [x] Tests added


workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
```diff
@@ -0,0 +1,160 @@
+import matplotlib.pyplot as plt
```
AnnaKwa, line 1: These are all just copied from `diagnostics_utils.plot`

PR #1462, opened 2021-10-29
Offline diags: refactor to use registry for remaining computations 

AnnaKwa:
It's easiest to review this PR commit by commit; I tried to break it up into comprehensible pieces and make the commit messages descriptive.

- This refactors the time mean, conditional averaging, and diurnal cycle calculations to use the registry instead of functions from diagnostics_utils. Plotting functions for these diagnostics are relocated out of diagnostics_utils (no changes, just moving the code) into diagnostics.offline.views.plot.

- The PR also adds total water path as a derived mapping variable, which will be computed using the specific humidity if it is not available in the dataset. If dQ2 or Q2 is a predicted variable, water vapor path will be loaded so that a scatter plot of Q2 vs water vapor path can be added to the report.

- There is a minor change to fv3fit.DerivedModel to facilitate wrapping an existing DerivedModel with another one while keeping the underlying base (non-derived) model the same. Without it, it was possible to have the base model of a `DerivedModel` be another `DerivedModel`, which causes issues.

I'm pretty sure these changes delete all current usages of the diagnostics_utils, but I'll leave it to a follow up PR to delete it because it's still referenced in a training_data_diags workflow. I think that can also be deleted but that would make this PR too big.

example checking that the 3D and 2D outputs come out ok: https://storage.googleapis.com/vcm-ml-public/annak/2021-10-29/offline-diags-0/index.html


workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -203,3 +187,27 @@ def _shorten_coordinate_label(coord: str):
         .replace("positive", "> 0")
         .replace("negative", "< 0")
     )
+
+
+def insert_column_integrated_vars(
```
AnnaKwa, line 192: Copied from `diagnostics_utils.utils`, no new code here

workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
```diff
@@ -0,0 +1,185 @@
...
+                existing_cols.append(col_tag)
+
+        if len(grouped) > 0:
+            grouped_ds[var] = xr.concat(
+                grouped, dim=pd.Index(existing_cols, name="domain")
+            ).squeeze()
+    return grouped_ds
+
+
+def plot_transect(
```
AnnaKwa, line 34: Everything from here down is copied from diagnostics_utils.plot, no new code here

workflows/diagnostics/tests/offline/test_compute_diagnostics.py
```diff
@@ -0,0 +1,161 @@
...
+)
+
+
+da = xr.DataArray(np.arange(1.0, 5.0), dims=["z"])
+da_nans = xr.DataArray(np.full((4,), np.nan), dims=["z"])
+ds = xr.Dataset({"a": da})
+weights = xr.DataArray([0.5, 0.5, 1, 1], dims=["z"])
+weights_nans = xr.DataArray(np.full((4,), np.nan), dims=["z"])
+
+
```
AnnaKwa, line 20: Tests copied from `diagnostics_utils`, no new code. The tests from `test_compute_metrics` have also been moved here.

workflows/diagnostics/tests/offline/test_helpers.py
```diff
@@ -1,12 +1,31 @@
...
-
+from vcm import thermo
 from fv3net.diagnostics.offline._helpers import (
     get_variable_indices,
     _count_features_2d,
+    insert_column_integrated_vars,
 )
 
 
+def test_insert_column_integrated_vars():
```
AnnaKwa, line 11: Copied from diagnostics_utils

workflows/diagnostics/fv3net/diagnostics/_shared/transform.py
```diff
@@ -278,10 +278,17 @@ def _mask_array(
 @add_to_input_transform_fns
 def subset_variables(variables: Sequence, arg: DiagArg) -> DiagArg:
     """Subset the variables, without failing if a variable doesn't exist"""
-    prognostic, verification, grid = arg.prediction, arg.verification, arg.grid
+    prognostic, verification, grid, delp = (
+        arg.prediction,
+        arg.verification,
+        arg.grid,
+        arg.delp,
```
oliverwm1, line 285: just to confirm—we don't use `arg.delp` in any of the online diags, right? I find it a bit confusing to have it separate from the prediction/verification data, since the pressure thicknesses might be different for those datasets.

workflows/diagnostics/fv3net/diagnostics/_shared/transform.py
```diff
@@ -278,10 +278,17 @@ def _mask_array(
...
+    )
     prognostic_vars = [var for var in variables if var in prognostic]
     verification_vars = [var for var in variables if var in verification]
-    return DiagArg(prognostic[prognostic_vars], verification[verification_vars], grid)
+    return DiagArg(
+        prognostic[prognostic_vars], verification[verification_vars], grid, delp
+    )
```
oliverwm1, line 291:
```suggestion
    return DiagArg(
        prognostic[prognostic_vars], verification[verification_vars], grid, arg.delp
    )
```
and this way you could revert the change on line 281 above.

oliverwm1 (approved):
I'm approving. However I noticed the feature sensitivity plot is missing from the report you linked. Not sure if that is due to this PR. https://storage.googleapis.com/vcm-ml-public/annak/2021-10-29/offline-diags-0/model_sensitivity.html

AnnaKwa:
> I'm approving. However I noticed the feature sensitivity plot is missing from the report you linked. Not sure if that is due to this PR. https://storage.googleapis.com/vcm-ml-public/annak/2021-10-29/offline-diags-0/model_sensitivity.html

I believe that's because the test report used a `PrecipitativeModel`, and the Jacobian method used in the offline report is only implemented for DenseModels. I'll make a to do issue to implement that for PureKerasModels. In the meantime I can add something to say on that page that it's empty for models where the feature sensitivity is not implemented.

workflows/diagnostics/fv3net/diagnostics/_shared/transform.py
AnnaKwa: Correct

PR #1463, opened 2021-10-29
fv3fit: allow input clipping in DenseModel

AnnaKwa:
This adds the `PackerConfig` as an optional hyperparameter for `DenseModel`, which will allow it to only use a subset of features along the given dimensions.

- [x] Tests added


external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -91,25 +93,37 @@ def clip(
...
 
     Used for ML training/prediction.
     """
 
-    def __init__(self, sample_dim_name, pack_names: Iterable[Hashable]):
+    def __init__(
+        self,
+        sample_dim_name,
+        pack_names: Iterable[Hashable],
+        config: Optional[PackerConfig] = None,
```
mcgibbon, line 125: New argument to add to docstring.

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -91,6 +94,9 @@ class DenseHyperparameters(Hyperparameters):
     loss: str = "mse"
     save_model_checkpoints: bool = False
     nonnegative_outputs: bool = False
+    packer_config: PackerConfig = dataclasses.field(
```
mcgibbon, line 97: Not an issue with/for your PR, but it continues to feel weird to me that we have a configuration container with one element which is another configuration container.

external/fv3fit/tests/test_models.py
```diff
@@ -76,3 +77,35 @@ def test_nonnegative_model_outputs():
...
+
+    nz = 5
+    dims = ["x", "y", "z"]
+    shape = (2, 2, nz)
+    arr = np.arange(np.prod(shape)).reshape(shape).astype(float)
+    input_data = xr.Dataset({"a": (dims, arr), "b": (dims, arr), "c": (dims, arr + 1)})
+    model.fit([input_data])
+    model.predict(input_data)
+    assert model.X_packer._n_features["a"] == 3
+    assert model.X_packer._n_features["b"] == 5
```
mcgibbon, line 98: This assertion makes a lot of assumptions about the internal workings of the model, and I'm worried it could continue to pass despite the model being broken. Can you add a test that asserts that a model's predictions are unchanged if the values it is supposedly clipping are set to NaN in the inputs?

mcgibbon (approved):
There's one test I'd like you to add and a docstring arg, but after that it should be good to merge without re-review.

PR #1464, opened 2021-10-30
Simplify-prepare-config

nbren12:
Improvements to the prepare config CLI and API. These are breaking changes

(Delete unused sections)
Added public API:
- config.duration, a more convenient way of specifying run durations.

Refactored public API:
- Removed initial condtions related arguments from `prepare-config` and the argo workflows.

### DIFF
--- a/docs/quickstarts.rst
+++ b/docs/quickstarts.rst
@@ -89,6 +89,9 @@ This example submits a "baseline" run of our fv3gfs model run to the Kubernetes 
 
   cat <<"EOF" > test_fv3config.yaml
   base_version: v0.5
+  initial_conditions:
+    base_url: gs://vcm-ml-experiments/2020-06-02-fine-res/coarsen_restarts
+    timestep: "20160805.000000"
   namelist:
     coupler_nml:
       days: 0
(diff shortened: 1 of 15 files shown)


nbren12:
Fixed bug and running integration tests manually here: integration-test-45ca143-0d5f

nbren12:
Fixed one more bug...new tests here: integration-test-7e4a3bd-3f8a

oliverwm1:
I just restarted the last integration test, which looks like it got terminated for some reason.

nbren12:
i'm confused about why the integration tests are failing. Seems unrelated.

> How would you configure the start time for a run that does not use coarsened restarts for initial conditions? I think it would be necessary to directly set the namelist.coupler_nml.current_date?

I tested this with the microphysics prognostic run script. It pulls the date from the restart files as long as  `force_date_from_namelist` is not true.

oliverwm1:
> It pulls the date from the restart files as long as force_date_from_namelist is not true.

Sure, but what about GFS analysis initial conditions? They do not provide any date information, you need to configure it.

nbren12:
I see. We could make an initial date timestamp an option for GFS as well.

PR #1468, opened 2021-11-02
fv3fit: dump and load packer config with ArrayPacker

AnnaKwa:
The previous PR #1463  to add the optional PackerConfig hyperparameter to the DenseModel did not test whether the input clipping worked with loaded models, so it missed a bug where loaded ArrayPackers did not clip inputs because they did not dump or load their configs. This PR addresses that issue and adds tests.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -279,16 +279,12 @@ class SliceConfig:
         return slice(self.start, self.stop, self.step)
 
 
-ClipConfig = Mapping[Hashable, Mapping[str, SliceConfig]]
-
-
-def _clip_config_factory() -> ClipConfig:
-    return {}
+ClipDims = Mapping[Hashable, Mapping[str, SliceConfig]]
 
 
 @dataclasses.dataclass(frozen=True)
 class PackerConfig:
-    clip: ClipConfig = dataclasses.field(default_factory=_clip_config_factory)
+    clip: ClipDims = dataclasses.field(default_factory=dict)
 
 
 @dataclasses.dataclass
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -282,9 +282,13 @@ class SliceConfig:
 ClipConfig = Mapping[Hashable, Mapping[str, SliceConfig]]
 
 
+def _clip_config_factory() -> ClipConfig:
+    return {}
+
+
 @dataclasses.dataclass(frozen=True)
 class PackerConfig:
-    clip: ClipConfig = dataclasses.field(default_factory=dict)
+    clip: ClipConfig = dataclasses.field(default_factory=_clip_config_factory)
 
 
 @dataclasses.dataclass
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -284,7 +284,7 @@ ClipConfig = Mapping[Hashable, Mapping[str, SliceConfig]]
 
 @dataclasses.dataclass(frozen=True)
 class PackerConfig:
-    clip: ClipConfig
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -284,7 +284,7 @@ def slice(self):
 
 @dataclasses.dataclass(frozen=True)
 class PackerConfig:
-    clip: ClipConfig
+    clip: ClipConfig = dataclasses.field(default_factory=dict)
```
mcgibbon, line 287: Shouldn't the default factory be `ClipConfig`?

external/fv3fit/tests/test_models.py
```diff
@@ -107,6 +108,26 @@ def test_DenseModel_clipped_inputs():
     xr.testing.assert_allclose(prediction_nan_filled, prediction_clipped, rtol=1e-3)
 
 
+def test_loaded_DenseModel_predicts_with_clipped_inputs(tmpdir):
```
mcgibbon, line 111: No need to change, but FYI in general there are a bunch of tests like this in `test_training.py`, you could get something like this in the future by parameterizing the dump-load-predict test there to use these special hyperparameters.

external/fv3fit/tests/test_models.py
```diff
@@ -107,6 +108,26 @@ def test_DenseModel_clipped_inputs():
...
+    nz = 5
+    dims = ["x", "y", "z"]
+    shape = (2, 2, nz)
+    arr = np.arange(np.prod(shape)).reshape(shape).astype(float)
+    input_data = xr.Dataset({"a": (dims, arr), "b": (dims, arr), "c": (dims, arr + 1)})
+    model.fit([input_data])
+    output_path = str(tmpdir.join("trained_model"))
+    fv3fit.dump(model, output_path)
+    model_loaded = fv3fit.load(output_path)
+    model_loaded.predict(input_data)
```
mcgibbon, line 128: Please perform prediction before and after dump/load-ing, and assert that the answer is unchanged.

external/fv3fit/fv3fit/_shared/config.py
AnnaKwa: `ClipConfig` is just defined as `Mapping[Hashable, Mapping[str, SliceConfig]]`, is it possible to use that as a default factory?

AnnaKwa: I added a factory function for ClipConfig, LMK if that's not what you meant.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -282,9 +282,13 @@ def slice(self):
...
 
 
+def _clip_config_factory() -> ClipConfig:
+    return {}
+
+
 @dataclasses.dataclass(frozen=True)
 class PackerConfig:
-    clip: ClipConfig
+    clip: ClipConfig = dataclasses.field(default_factory=_clip_config_factory)
```
mcgibbon, line 291: What you had before was correct, I was just thrown off by ClipConfig not being a dataclass. As we discussed in slack, please rename ClipConfig to anything without Config in the name.

mcgibbon (approved):
Sorry about the confusion, you can revert to `default_factory=dict` as before.

external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -205,14 +207,18 @@ def dump(self, f: TextIO):
...
             },
             f,
         )
 
     @classmethod
     def load(cls, f: TextIO):
         data = yaml.safe_load(f.read())
-        packer = cls(data["sample_dim_name"], data["pack_names"])
+        packer_config = dacite.from_dict(PackerConfig, data["packer_config"])
+        packer = cls(data["sample_dim_name"], data["pack_names"], packer_config)
```
oliverwm1, line 221: can this be done in a backwards compatible way? for old dense models that didn't have a packer config?

PR #1471, opened 2021-11-03
Remove cluster resources from fv3net

frodre:
Removing the cluster resources that are duplicated on workflow control and removes usage if `integration-tests` service account in favor of `default`.

Companion PR to https://github.com/ai2cm/vcm-workflow-control/pull/116


oliverwm1 (approved):
Thanks for all these cleanups!

nbren12 (changes requested):
Thanks!  I didn't even realize there were so many configs scattered between here and vcm-workflow-control. I think it's cleaner to at least have all this in one place, even though one could argue about where that should be.

Looks like I added these when working on the local k8s cluster: a50443011048746db23f4e9d1b7b61d38b1e627b.
We never use that anymore, so maybe you can should delete the "Local development of argo workflows" section of `README.yml`.

frodre:
> Looks like I added these when working on the local k8s cluster: [a504430](https://github.com/ai2cm/fv3net/commit/a50443011048746db23f4e9d1b7b61d38b1e627b). We never use that anymore, so maybe you can should delete the "Local development of argo workflows" section of `README.yml`.

Yeah, I think that was a source of confusion for me because there was argo stuff in the default namespace from an errant run of the `make deploy...`

Removed the .rst doc and the makefile target.

nbren12 (approved):
Thanks for the changes.

PR #1452, opened 2021-10-27
prog_run_diags: 2D hist of water vapor path versus Q2

oliverwm1:
Joint distribution between water vapor path and precipitation is a useful process diagnostic. This PR adds a diagnostic and plot of the 2D histogram between water vapor path and column-integrated drying (`-<Q2>`). It also adds a diagnostic of the conditional-mean `-<Q2>`.

Added public API:
- `vcm.histogram2d`

Significant internal changes:
- compute 2D histogram of `WVP` vs `-<Q2>` and 1D histogram of `WVP`
- compute conditional average of `-<Q2>` on `WVP`
- add plots of both these histograms and conditional average to process diagnostics page

- [x] Tests added

PR #1475, opened 2021-11-04
offline diags: add column drying versus PW

AnnaKwa:
This adds histograms of water vapor path and column drying as well as their 2D histogram to the offline report.

I do recognize the `create_report` script is getting really long and needs a refactor to use a registry, which I'll do at a later time, but I wanted to get these figures into the report since they were added to the prognostic report recently.

The diagnostics functions and plotting functions are basically copies of the prognostic run versions that were added in PR #1452 , but with some small adjustments to account for the fact that current ML prediction data in the offline diags is often missing units that need to be filled in with the helper function.

example at bottom of index: https://storage.googleapis.com/vcm-ml-public/annak/2021-10-28/offline-diags-7/index.html

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
@@ -218,7 +218,7 @@ def plot_histogram2d(ds, xname: str, yname: str):
     y_bin_name = f"{yname}_bins"
     x_bin_widths_name = f"{xname.lower()}_bin_width_hist_2d"
     y_bin_widths_name = f"{yname.lower()}_bin_width_hist_2d"
-    fig = plt.figure(figsize=(8, 4))
+    fig = plt.figure(figsize=(12, 6))
 
     for i, source in enumerate(["target", "predict"]):
         ds_ = ds.sel({DERIVATION_DIM_NAME: source})
--- a/workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
@@ -193,7 +193,6 @@ def plot_histogram(ds, varname: str, xscale="linear", yscale="linear"):
     bin_name = varname.replace("histogram", "bins")
     v = ds[varname]
     units = units_from_name(varname.replace("_histogram", ""))
-    print(units, varname.replace("_histogram", ""))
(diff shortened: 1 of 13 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/constants.py
```diff
@@ -140,12 +139,7 @@
 
 PRECIP_RATE = "total_precip_to_surface"
 MASS_STREAMFUNCTION_MID_TROPOSPHERE = "mass_streamfunction_300_700_zonal_and_time_mean"
-WVP = "water_vapor_path"
```
AnnaKwa, line 143: Moved to `_shared.constants` since they are used by offline diags as well

oliverwm1 (commented):
Generally looks good to me—but can we include the Q2 histogram and WVP vs Q2 joint histogram for the training data as well as for the predictions?

AnnaKwa:
Sure, I added both distributions to the histograms.

workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
```diff
@@ -183,3 +185,66 @@ def plot_generic_data_array(
...
+def plot_histogram2d(ds, xname: str, yname: str):
+    """Plot 2D histogram of xname versus yname."""
+
+    count_name = f"{xname.lower()}_versus_{yname.lower()}_hist_2d"
+    conditional_average_name = f"conditional_average_of_{yname}_on_{xname}"
+    x_bin_name = f"{xname}_bins"
+    y_bin_name = f"{yname}_bins"
+    x_bin_widths_name = f"{xname.lower()}_bin_width_hist_2d"
+    y_bin_widths_name = f"{yname.lower()}_bin_width_hist_2d"
+    fig = plt.figure(figsize=(8, 4))
```
oliverwm1, line 221: the current figure size makes it a bit hard to read.

oliverwm1 (approved):
LGTM. My only comment is that the 2D histogram looks pretty small now that it's plotted for the target and prediction.

workflows/diagnostics/fv3net/diagnostics/offline/views/plot.py
AnnaKwa: Fixed

PR #1476, opened 2021-11-04
fv3fit: remove backward compatible loading

brianhenn:
Removes some code for backward compatible loading of fv3fit sklearn models that no longer worked. 

Refactored public API:
- Models prior to dict metadata serialization (#1404) are not loadable.


oliverwm1:
Should we deprecate the current sklearn model name? e.g. see

https://github.com/ai2cm/fv3net/blob/22b23fe185f446a358f1c6c302a00b6112a6fed9/external/fv3fit/fv3fit/_shared/io.py#L12

This would have been clearer if we'd just done it when the metadata serialization format was changed.

external/vcm/pyproject.toml
```diff
@@ -1,4 +1,4 @@
 # credit: https://stackoverflow.com/questions/60045913/installing-numpy-before-using-numpy-distutils-core-setup
 [build-system]
 build-backend = "setuptools.build_meta"
-requires=["numpy==1.19.4", "setuptools"]
\ No newline at end of file
+requires=["numpy==1.19.4", "setuptools==49.6.0"]
```
oliverwm1, line 4: Let's keep this change separate for a clean git history.

brianhenn: Sure this is already merged, just had it to get the tests to pass

brianhenn:
Thanks didn't know we were versioning models, that's useful. As you point out if we bump it now it will prevent loading models from the last couple weeks that actually could load. Maybe wait until the next version?

oliverwm1:
> Thanks didn't know we were versioning models, that's useful. As you point out if we bump it now it will prevent loading models from the last couple weeks that actually could load. Maybe wait until the next version?

Sure that sounds fine.

oliverwm1 (approved):

PR #1478, opened 2021-11-05
Ensure proper Q on report for prescribed/fine-res runs

oliverwm1:
Currently the prognostic report sums the "fv3 physics" and "machine learning" tendencies for computing Q1 and Q2. But for prescribed/fine-res/emulation runs, the Q1/Q2 should just be the ML tendencies and not include the piggy-backed physics. This PR allows that by using the[ 'applied physics' tendency](https://github.com/ai2cm/fv3net/blob/cadd8561eb75c75ebf9fcdf6b8aa38bc07930116/workflows/prognostic_c48_run/runtime/loop.py#L474-L479) if it's available.


AnnaKwa (approved):
LGTM!

Issue #1211, opened 2021-05-14
Prognostic run "prephysics" is a bit confusing


nbren12:
It is hard to know what should go in`config.prephysics`:
https://github.com/VulcanClimateModeling/fv3net/blob/a24d15ef763a2a157e62738dbf09738cb85fe8ad/workflows/prognostic_c48_run/runtime/config.py#L27

As a user, I don't want to know if something is done pre or post physics, I just want to know that the intended effect is achieved. Would it be possible to break down this section more finely, so that it can be more easily understood. E.g.

```
config.surface:
   radiation:...
   surface_precipitation: ...
```

mcgibbon:
Closing with feedback from @nbren12

PR #1481, opened 2021-11-09
Prognostic run: allow multiple prescribe sources

AnnaKwa:
This PR allows the prognostic run to use >1 source for prescribed quantities. The immediate use for this feature would be to allow prognostic runs using both a surface radiative flux model as well as SSTs from the fine res data.

I tested this in a local prognostic run using separate prescribers for surface radiative fluxes and SST. I verified that the SST is being set correctly to the reference (figure shows the surface temperature difference between prognostic run and reference for a single timestep)
image

Added public API:
- The prognostic run config `prephysics` section can now accept a list of sources. It is still backwards compatible to allow a single non-list formatted source. ex. 
```
prephysics:
  -
    dataset_key: gs://vcm-ml-intermediate/2021-08-03-PIRE-c48-diags-post-spinup/fine-res-surface-radiative-fluxes-precip.zarr
    variables:
      - override_for_time_adjusted_total_sky_downward_shortwave_flux_at_surface
      - override_for_time_adjusted_total_sky_downward_longwave_flux_at_surface
      - override_for_time_adjusted_total_sky_net_shortwave_flux_at_surface
      - total_precipitation
  -
    dataset_key: gs://vcm-ml-intermediate/2021-08-03-PIRE-c48-diags-post-spinup/sst-time-centered.zarr
    variables:
      - ocean_surface_temperature 
 ```

Significant internal changes:
- Added a `CombinedStepper`, which takes a list of `Prescriber` and/or `PureMLStepper` steppers and combines their outputs.
- I moved the function `sst_update_from_reference` out of `runtime.nudging` and into `runtime.steppers.prescriber` since that seemed like the more natural home for it (since it prescribes SSTs). 
- Special treatment for variable `ocean_surface_temperature` if set in the `Prescriber` such that it only prescribes reference temperatures over ocean, and keeps land surface temperature from the state.

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/runtime/config.py
+++ b/workflows/prognostic_c48_run/runtime/config.py
@@ -31,8 +31,6 @@ FV3CONFIG_KEYS = {
     "gfs_analysis_data",
 }
 
-PrephysicsConfig = List[Union[PrescriberConfig, MachineLearningConfig]]
-
 
 @dataclasses.dataclass
 class UserConfig:
@@ -77,7 +75,6 @@ def get_config() -> UserConfig:
     """
     with open("fv3config.yml") as f:
         config = yaml.safe_load(f)
-    config = _standardize_prephysics_config(config)
 
     runtime_config = {key: config[key] for key in config if key not in FV3CONFIG_KEYS}
     return dacite.from_dict(UserConfig, runtime_config, dacite.Config(strict=True))
@@ -100,16 +97,3 @@ def write_chunks(config: UserConfig):
     chunks = get_chunks(diagnostic_file_configs)
     with open("chunks.yaml", "w") as f:
         yaml.safe_dump(chunks, f)
-
-
-def _standardize_prephysics_config(config) -> PrephysicsConfig:
-    # allows for backwards compatibility for old config format where
-    # only a single entry for prephysics was allowed
-    if "prephysics" not in config:
-        return config
-    elif isinstance(config["prephysics"], List):
-        return config
-    else:
-        raw_prephysics_config = config["prephysics"]
-        config["prephysics"] = [raw_prephysics_config]
-        return config
--- a/workflows/prognostic_c48_run/runtime/config.py
+++ b/workflows/prognostic_c48_run/runtime/config.py
@@ -41,7 +41,10 @@ class UserConfig:
         fortran_diagnostics: list of Fortran diagnostic file configurations
         prephysics: optional configuration of computations prior to physics,
             specified by either a machine learning configuation or a prescriber
-            configuration
+            configuration. If quantity is in runtime.names.PREPHYSICS_OVERRIDES,
+            it will be applied during prephysics step, all others will be updated
+            postphysics (yes, this contradicts the name of the config class,
+            there is a github issue open to address this).
         scikit_learn: a machine learning configuration
         nudging: nudge2fine configuration. Cannot be used if any scikit_learn model
             urls are specified.
--- a/workflows/prognostic_c48_run/runtime/config.py
+++ b/workflows/prognostic_c48_run/runtime/config.py
@@ -31,6 +31,8 @@ FV3CONFIG_KEYS = {
     "gfs_analysis_data",
 }
 
+PrephysicsConfig = List[Union[PrescriberConfig, MachineLearningConfig]]
+
 
 @dataclasses.dataclass
 class UserConfig:
@@ -52,7 +54,7 @@ class UserConfig:
     fortran_diagnostics: List[FortranFileConfig] = dataclasses.field(
         default_factory=list
     )
-    prephysics: Optional[Union[PrescriberConfig, MachineLearningConfig]] = None
+    prephysics: Optional[List[Union[PrescriberConfig, MachineLearningConfig]]] = None
     scikit_learn: Optional[MachineLearningConfig] = None
     nudging: Optional[NudgingConfig] = None
     tendency_prescriber: Optional[TendencyPrescriberConfig] = None
@@ -75,6 +77,8 @@ def get_config() -> UserConfig:
     """
     with open("fv3config.yml") as f:
         config = yaml.safe_load(f)
+    config = _standardize_prephysics_config(config)
+
     runtime_config = {key: config[key] for key in config if key not in FV3CONFIG_KEYS}
     return dacite.from_dict(UserConfig, runtime_config, dacite.Config(strict=True))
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: workflows/prognostic_c48_run/runtime/steppers/prescriber.py [hunk 1]
```diff
@@ -163,3 +173,35 @@ def _open_ds(dataset_key: str, consolidated: bool) -> xr.Dataset:
     except KeyError:
         ds = intake.open_zarr(dataset_key, consolidated=consolidated).to_dask()
     return ds
+
+
+def _sst_from_reference(
```
AnnaKwa, line 178: These two functions are moved over from `runtime.nudging`.

workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -96,3 +100,16 @@ def write_chunks(config: UserConfig):
...
+    # allows for backwards compatibility for old config format where
+    # only a single entry for prephysics was allowed
+    if "prephysics" not in config:
+        return config
+    elif isinstance(config["prephysics"], List):
+        return config
+    else:
+        raw_prephysics_config = config["prephysics"]
+        config["prephysics"] = [raw_prephysics_config]
+        return config
```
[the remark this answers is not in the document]
oliverwm1: As another option, did you try type-hinting `config.prephysics` as

```
Prephysics = Union[PrescriberConfig, MachineLearningConfig]
Optional[Union[List[Prephysics], Prephysics]]
```

oliverwm1 (commented):
Can we keep the `get_config` code clean? Might still be possible while maintaining backwards compatibility.

workflows/prognostic_c48_run/runtime/config.py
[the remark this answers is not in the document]
AnnaKwa: I think I'll just make this non-backwards-compatible instead of moving the backwards compatible code to the loop.  If someone tries to use an old config this will raise an error before starting the run that is easy to fix in the config.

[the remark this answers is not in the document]
brianhenn: Agree, it's not like locking out a saved model, it's just a change in the config

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -364,6 +375,7 @@ def _step_prephysics(self) -> Diagnostics:
             "override_for_time_adjusted_total_sky_downward_shortwave_flux_at_surface",
             "override_for_time_adjusted_total_sky_net_shortwave_flux_at_surface",
             "override_for_time_adjusted_total_sky_downward_longwave_flux_at_surface",
+            "ocean_surface_temperature",
```
AnnaKwa, line 378: @brianhenn  I think prephysics is the appropriate step for the SST update so I put this here. However,  it's possible to have a case where you provide a variable in the `prephysics` through a dataset and there is no postphysics stepper, in which case the run will proceed without prescribing from the specified data. Do we want to raise some exception for those cases and have a separate configuration for prescribing state postphysics ? Rename the config section to something more general like `prescribe_state`, rely on the `prephysics_overrides` list to determine where the updates happen, and let the state update in the postphysics step regardless of whether there's a nudging or ML stepper?

brianhenn: Yeah, good point. Being able to configure where state updates get applied seems like too much flexibility though, and this is a pretty short list, so I'd be more in favor of just raising an exception if someone tries to prescribe a variable that isn't on this list and there's no postphysics stepper.

This seems like an ongoing challenge that there's not really documentation of when the different state and tendency updates will occur and under what conditions, outside of just staring at the loop for a while. Maybe it'd be helpful to at least list the allowed `prephysics_overrides` in the `prephysics` UserConfig docs and say that any other variables will only be set postphysics if there's a stepper, and that the only variable we've really tried doing this with is `total_precipitation`. I know that it'll get stale in the docstring but at least it's somewhere.

brianhenn (commented):
Thanks @AnnaKwa overall this looks pretty clean. I do think you raise a good point about what prephysics variables go where.

workflows/prognostic_c48_run/runtime/loop.py
AnnaKwa: There's an old issue that's also related to this discussion (https://github.com/ai2cm/fv3net/issues/1211). I agree that we shouldn't use the config to decide where the updates happen. As per our in person discussion just now I'm going to 

i) rename the `prephysics` part of the config to `prescribed_state`, since it also can add postphysics updates
ii) create another whitelist of `postphysics_overrides` that will get applied in the postphysics step, raise an error if there are variables in the prescribed state config that are in neither override list. This means any new override variables require a PR to update one of these lists.
iii) apply postphysics overrides regardless of whether there is a postphysics stepper

workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -52,7 +54,7 @@ class UserConfig:
     fortran_diagnostics: List[FortranFileConfig] = dataclasses.field(
         default_factory=list
     )
-    prephysics: Optional[Union[PrescriberConfig, MachineLearningConfig]] = None
+    prephysics: Optional[List[Union[PrescriberConfig, MachineLearningConfig]]] = None
```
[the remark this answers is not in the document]
AnnaKwa: I actually ended up deleting that type after responding to Oli's comment, since it was only used in the `_standardize_prephysics_config` changes.

AnnaKwa:
Thanks for the comments @brianhenn , this is ready for re-review. In the end, the changes related to the issue of how to handle prephysics/postphysics ended up being very minimal- I added some more specifics to the UserConfig docstring and made the postphysics state update independent of whether or not a postphysics stepper existed. See this comment  for my rationale for not changing the postphysics state updates to require a whitelist :https://github.com/ai2cm/fv3net/pull/1481#discussion_r746836561

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -136,7 +136,7 @@ def __call__(self, time, state):
                 # "ocean_surface_temperature". If "surface_temperature" is
```
brianhenn, line 136: Nit: Rephrase comment to only focus on special case of setting ocean surface temperature. I see what you're saying with the latter part but it might be more distracting than helpful given that prescribing surface temperature isn't a use case we've encountered yet and is not on the prephysics 'whitelist'.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -203,6 +204,22 @@ def __init__(
         self._log_info(self._fv3gfs.get_tracer_metadata())
         MPI.COMM_WORLD.barrier()  # wait for initialization to finish
 
+    def _use_diagnostic_ml_prephysics(self, prephysics_config):
```
brianhenn, line 207: So many cases to consider...good catch

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -246,7 +263,10 @@ def _get_prephysics_stepper(
                         PureMLStepper(model, self._timestep, hydrostatic)
```
brianhenn, line 263: You mentioned in your comment that 

> you don't know the overriding variables it provides until model.predict is called, so I can't put this check in the Loop init

I didn't realize that the MultimodelAdapter/RenamingAdapters don't subclass fv3fit.Predictor and so don't have output variables as attributes. Seems like they should, but outside of this PR scope, I'll make an issue.

brianhenn (approved):
Thanks @AnnaKwa , this looks good. I have a rather cosmetic optional suggestion.

PR #1484, opened 2021-11-09
add command-line training config override to fv3fit.train

mcgibbon:
This PR makes use of @frodre's code for updating configuration dictionaries using command-line arguments to expose all TrainingConfig options as modifiable from the command-line. Overriding "kwargs" arguments requires that a value is given in the provided yaml, so that the argument parser knows the names and types of modifiable values.

The `get_arg_updated_config_dict` function is also updated to work for int and float arguments being provided to dataclasses using dacite.

Refactored public API:
- All TrainingConfig parameters can now be provided as command-line arguments to `fv3fit.train`.

- [x] Tests added


external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -163,17 +163,22 @@ def _add_items_to_parser_arguments(
...
+                "is given in the configuration yaml."
             )
         elif not isinstance(value, str) and isinstance(value, Sequence):
             parser.add_argument(f"--{key}", nargs="*", default=copy.copy(value))
         elif isinstance(value, bool):
             parser.add_argument(f"--{key}", type=_bool_from_str, default=value)
+        elif isinstance(value, int):
+            parser.add_argument(f"--{key}", type=int, default=value)
+        elif isinstance(value, float):
+            parser.add_argument(f"--{key}", type=int, default=value)
```
mcgibbon, line 181:
```suggestion
            parser.add_argument(f"--{key}", type=float, default=value)
```
Typo.

external/fv3fit/fv3fit/_shared/config.py
```diff
@@ -163,17 +163,22 @@ def _add_items_to_parser_arguments(
...
+                "currently supported.  Make sure you are passing a "
+                "'flattened' dictionary to this function. If you are trying "
+                "to override a value in a 'kwargs' setting, make sure a value "
+                "is given in the configuration yaml."
             )
         elif not isinstance(value, str) and isinstance(value, Sequence):
             parser.add_argument(f"--{key}", nargs="*", default=copy.copy(value))
         elif isinstance(value, bool):
             parser.add_argument(f"--{key}", type=_bool_from_str, default=value)
+        elif isinstance(value, int):
```
frodre, line 178: Thanks for the addition here.  I was previously doing a mixture of type casting here and in my dataclass, which is confusing.  I think it makes sense to enforce standard types here.

external/fv3fit/tests/training/test_main.py
```diff
@@ -302,6 +302,36 @@ def get_config(
...
+    mock_load_batches.return_value = [config.mock_dataset for _ in range(6)]
+    assert config.hyperparameters.dense_network["width"] == 6
+    assert config.hyperparameters.optimizer_config["kwargs"]["lr"] == 0.01
+    patch_args = [
+        "--hyperparameters.dense_network.width",
+        "12",
+        "--hyperparameters.optimizer_config.kwargs.lr",
+        "0.02",
+    ]
+    fv3fit.train.main(config.args, unknown_args=patch_args)
```
frodre, line 327: Having to do the override argument testing and the `main` testing all together makes test feel too complex.  I think this would be the advantage of having some type of function which generally updates a config.  That way we wouldn't have to write this test every time?  Nothing to do now, but food for thought on something to work towards.

frodre (approved):
LGTM, thanks for the updates.

external/fv3fit/tests/training/test_main.py
mcgibbon: I could split main into "get configs" and "call main which has configs as call signature", and mock out the latter function when testing the first. This is fairly close to what the test does right now - all training is mocked out.

You do have the "function which generally updates a config" already, and it's (presumably?) already tested, so arguably this test could be deleted altogether. But I want to test that `main` actually makes proper use of that function.

mcgibbon: Another way of saying some of this is, the "main" tests are testing that command-line arguments are properly propagated to code behavior. They aren't testing anything else (training, data loading, etc.) as it's all been mocked.

frodre: >You do have the "function which generally updates a config" already, and it's (presumably?) already tested

Yep, you're right, the args -> dict is tested and general function.  

I guess the complexity here is driven by parsing args in the `__main__`, passing them into `main()` and then having to derive a `TrainConfig` from there.  I thought it was doing a larger integration test and also checking the config (missed that training is also mocked), whereas from the test name, I was just expecting something looking for a valid and updated `TrainConfig` based on arguments.  It's all coupled, so the test has to look like this.

frodre: > Another way of saying some of this is, the "main" tests are testing that command-line arguments are properly propagated to code behavior.

Thanks for clarifying this.  Makes sense to me now.

PR #1486, opened 2021-11-10
Computes skill scores for the surface precipitation

nbren12:
Add summary skill scores and time series for the surface precipitation prediction.

Example report: https://wandb.ai/ai2cm/microphysics-emulation/runs/2ab33um4/overview?workspace=user-nbren12


frodre (approved):
Skill plot looks good.  I think it would be nice to add in a spatial precip plot alongside the tendencies as well.

PR #1488, opened 2021-11-10
fv3fit convolutional model: Allow 2D inputs

AnnaKwa:
I ran into a couple issues while trying out the convolutional model with 2D input fields (ex. surface geopotential). This PR modifies some of the dimension checking and input layer concatenation to allow these types of inputs to be used.

Significant internal changes:
- change `_ensure_5d` to `ensure_4d` and allow 3d shape in the `_ensure_4d` check
- add a z dim of size 1 to the input layer shape for 2D inputs
- minor fix for the check on whether validation batches are present; the training script passes an empty list, not `None` as assumed in the convolutional training function

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/convolutional.py
+++ b/external/fv3fit/fv3fit/keras/_models/convolutional.py
@@ -136,15 +136,14 @@ def _count_array_features(arr: np.ndarray) -> int:
         return arr.shape[3]
 
 
-def _ensure_5d(array: np.ndarray) -> np.ndarray:
-    if len(array.shape) == 5:
+def _ensure_4d(array: np.ndarray) -> np.ndarray:
+    # ensures [sample, x, y, z] dimensionality
+    if len(array.shape) == 4:
         return array
-    elif len(array.shape) == 4:
-        return array[:, :, :, :, None]
     elif len(array.shape) == 3:
-        return array[:, :, :, None, None]
+        return array[:, :, :, None]
     else:
-        raise ValueError(f"expected 3d, 4d or 5d array, got shape {array.shape}")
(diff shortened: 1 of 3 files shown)


external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -156,7 +172,11 @@ def build_model(
         X: example input for keras fitting, used to determine shape and normalization
         y: example output for keras fitting, used to determine shape and normalization
     """
-    input_layers = [tf.keras.layers.Input(shape=array.shape[1:]) for array in X]
+    input_layer_shapes = _get_input_layer_shapes(X)
+    input_layers = [
+        tf.keras.layers.Input(shape=input_shape) for input_shape in input_layer_shapes
+    ]
```
mcgibbon, line 178: Yes.

mcgibbon (commented):
It may also be a good idea to modify the `test_train.py` tests so one of the input variables is 1D for generic models (not for Precipitative, that one has special data handling). This would involve modifying `get_dataset` in that function to add another input variable.

external/fv3fit/tests/training/test_train.py
```diff
@@ -123,6 +123,10 @@ def get_dataset(model_type, sample_func):
         )
     else:
         output_values = input_values
+        # add an input with feature dim size 1
+        input_variables.append("var_in_2d")
```
mcgibbon, line 127: nit: instead of modifying input_variables here, can you update the hard-code above to be `input_variables=["var_in", "var_in_2d"]`? Then you can clip input_values[-1] below. Alternatively, add a comment above that says `# "var_in_2d" is appended below`

mcgibbon (approved):
Thanks!

PR #1492, opened 2021-11-12
Argo workflow for microphysics project prognostic runs

frodre:
Add an Argo workflow and Makefile with examples of running a prognostic microphysics run followed by the prognostic and piggy-backed evaluation scripts.


projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,163 @@
...
+          cpu: "7500m"
+        limits:
+          memory: "8Gi"
+          cpu: "7500m"
+      envFrom:
+      - secretRef:
+          name: wandb-andrep
+      env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS 
+        value: /secret/gcp-credentials/key.json
```
nbren12, line 81: Move `wandb_run_group` here.

```suggestion
        value: /secret/gcp-credentials/key.json
      - name: WANDB_RUN_GROUP
        value: "{{inputs.parameter.tag}}"
```

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,163 @@
...
+      volumeMounts:
+      - mountPath: /secret/gcp-credentials
+        name: gcp-key-secret
+      command: ["bash", "-c", "-x", "-e"]
+      args: 
+      - |
+        cd projects/microphysics
```
nbren12, line 126: You can better leverage k8s features here. Our use of `bash  -c` is kind of an unsafe workaround. e.g. using `{{inputs.paramters.tag}}`  below could be used to run an arbitrary command/expose a secret.  

In some cases this is necessary, but not here since this command requires no config yaml.

Use `workingDir` to change the working directory and `env` to setup environment variables, and `args` to specify the command to run. I'm not sure why the `LS` variable is required.
```
workingDir: /fv3net/projects/microphysics
env:
  - name: WANDB_RUN_GROUP
    value: "{{inputs.parameter.tag}}"
args:
  - python3
  - scripts/piggy-back.py
  - "{{input.parameters.tag}}"
```


Note that I also specify the WANDB_RUN_GROUP here. This is important for grouping these outputs in the console.

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,163 @@
...
```
nbren12, line 161: See my note above about adding `WAND_RUN_GROUP` and better leveraging k8s features here too.

nbren12 (commented):
just saw this pr. noticed a couple of things. Let me know when it's ready and i can give a full review.

projects/microphysics/argo/prog-run-and-eval.yaml
frodre: Very good point.  `LS` is a forgotten leftover from doing some debugging so I'll go ahead and remove that.

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,169 @@
...
+      resources:
+        requests:
+          memory: "6Gi"
+          cpu: "7500m"
+        limits:
+          memory: "8Gi"
+          cpu: "7500m"
+      envFrom:
+      - secretRef:
+          name: wandb-andrep
```
nbren12, line 78: Maybe document how you created this secret?

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,169 @@
...
+        value: /secret/gcp-credentials/key.json
+      - name: WANDB_RUN_GROUP
+        value: "{{inputs.parameters.tag}}"
+      volumeMounts:
+      - mountPath: /secret/gcp-credentials
+        name: gcp-key-secret
+      command: ["bash", "-c", "-x", "-e"]
+      args: 
+      - |
+        WANDB_RUN_GROUP="{{inputs.parameters.tag}}"
```
nbren12, line 90: Forgot to remove this after adding it to `env`.

```suggestion
```

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,169 @@
...
+  arguments:
+    parameters:
+      - name: tag
+      - name: config
+      - name: tf_model
+      - name: baseline_tag
+        value: "baseline"
+      - name: on_off_flag
+        value: "--online"
+      - name: do_piggy
```
nbren12, line 16: When would we not want to do this? Maybe some differences in the configs above.

projects/microphysics/argo/prog-run-and-eval.yaml
```diff
@@ -0,0 +1,169 @@
...
+        value: "true"
+  volumes:
+  - name: gcp-key-secret
+    secret:
+      secretName: gcp-key
+  templates:
+  - name: prognostic-run
+    steps:
+    - - name: run-model
```
nbren12, line 25: Maybe for later, but I would suggest running a "baseline" run for every model. This will allow us to compare the offline and online performance of a given ML model over the same time period. This workflow will also be more reproducible since the only input data will then be the tf model and fv3config inputs. It should also remove the need for separate baseline and non-baseline configs, since the  `--online/--offline` flag can control this.

nbren12 (changes requested):
Seems like there is more copy paste in the configs than necessary. Unsure why we need separate configs for baseline and non-baseline? Would be nice to collect them all in one place as well rather than scattered in `argo` and `scripts`.

Do you have time today to work on a solution together?

nbren12:
Also, what's up with all the numbers in the fortran outputs: https://wandb.ai/ai2cm/microphysics-emulation/runs/12rsoxoo/logs?workspace=user-nbren12

Has some print statement been added to the model? or is this from python.

frodre:
> Also, what's up with all the numbers in the fortran outputs

Haha, wow that is a lot of numbers.  I don't think I've added anything intentionally to do this.  Maybe I missed a configuration parameter.

nbren12:
> > Also, what's up with all the numbers in the fortran outputs
> 
> Haha, wow that is a lot of numbers. I don't think I've added anything intentionally to do this. Maybe I missed a configuration parameter.

I reran with one of the default.yml's locally and the problem didn't happen. Maybe related to the k8s environment?

nbren12 (approved):

PR #1494, opened 2021-11-13
Update fv3gfs-fortran to version with post condensation outputs

nbren12:
fv3gfs-fortran added some new outputs that we would like to use for ML but at the same time, modified several names used by call py fort. Specifically the `_output` suffix was replaced with `_after_precpd` in the variables used to override the fv3 state. 

This PR adds a backwards compatibility layer that allows us to use previously trained models with the new fortran version.

I will re-create the training set and train new models in a follow-up PR.

Here is a simulation using an old model that proves the backwards compatibility works: https://wandb.ai/ai2cm/microphysics-emulation/runs/1m0zs93y?workspace=user-nbren12.


external/emulation/emulation/_emulate/microphysics.py
```diff
@@ -104,7 +130,7 @@ def microphysics(self, state: FortranState) -> None:
 
         # grab model-required variables and
         # switch state to model-expected [sample, feature]
-        inputs = [state[name].T for name in self.model.input_names]
+        inputs = {name: state[name].T for name in self.model.input_names}
```
frodre, line 133: Is this just to make the I/O more congruent?

frodre (approved):
Nice and clean update.  Thanks for the backwards compatibility check.  LGTM

external/emulation/emulation/_emulate/microphysics.py
nbren12: I don't think was necessary, but I hopefully it makes it a bit more robust to order or argument errors.

PR #1495, opened 2021-11-13
relax coupling between diag_table and prognostic run diags

nbren12:
The prognostic run diagnostics do not work for arbitrary diag tables and harcodes the names of specific files like `atmos_dt_atmos.zarr` deeply within a stack of helper functions. In practice this tightly couples the diagnostic suite to the configurations for runs and makes it hard to generate reports for different diag tables (e.g. microphysics). For instance, the diagnostics will fail if the input data in `sfc_dt_atmos.zarr`, etc is missing.

This PR relaxes this last assumption and generally cleans-up the code in `load_run_data.py` by removing layers of nearly identical helpers and reformulating the coarsening tools.


workflows/diagnostics/fv3net/diagnostics/prognostic_run/load_run_data.py
```diff
@@ -250,7 +202,22 @@ class SegmentedRun:
...
+        url = self.url
+        catalog = self.catalog
+        path = os.path.join(url, "atmos_dt_atmos.zarr")
+        diags_url = os.path.join(url, "diags.zarr")
+        sfc_dt_atmos_url = os.path.join(url, "sfc_dt_atmos.zarr")
+
+        return xr.merge(
+            [
+                load_coarse_data(path, catalog),
+                # TODO fillna required because diags.zarr may be saved with an
```
nbren12, line 194: Is `fillna` also appropriate (if not necesary) for the fortran data? If so, I could push it into `load_coarse_data` and cleanup/generalize this logic.

workflows/diagnostics/tests/prognostic/_regtest_outputs/test_load_diag_data.test_Simulations[simulation0].out
```diff
@@ -43,191 +43,31 @@ variables:
 // global attributes:
```
Comment (nbren12) [inline workflows/diagnostics/tests/prognostic/_regtest_outputs/test_load_diag_data.test_Simulations[simulation0].out:43]: This regression changed order because I switched to merging all three datasets at once. rather than merging the "physics"  and "dycore" files separately.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/load_run_data.py
oliverwm1: I think we could get rid of the `fillna`... it was there because of an issue with the `ZarrMonitor` that was fixed a long time ago IIRC.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/load_run_data.py
```diff
@@ -138,52 +104,18 @@ def load_grid(catalog):
...
-    """
-    logger.info(f"Processing physics data from run directory at {url}")
-    prognostic_output = _load_prognostic_run_physics_output(url)
-    input_grid, coarsening_factor = _get_coarsening_args(prognostic_output, 48)
-    area = catalog[input_grid].to_dask()["area"]
-    return _coarsen(prognostic_output, area, coarsening_factor)
+    try:
+        ds = _load_standardized(path)
+    except (FileNotFoundError, KeyError):
+        ds = xr.Dataset()
```
oliverwm1, line 113: Could you add a warning here? Would be useful for debugging.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -523,16 +538,22 @@ def main(args):
...
-    diags["pwat_run_final"] = input_data["2d"][0].PWAT.isel(time=-2)
-    diags["pwat_verification_final"] = input_data["2d"][0].PWAT.isel(time=-2)
+    diags["pwat_run_initial"] = (
+        input_data["2d"][0].PWAT.isel(time=0).rename({"time": "initial_time"})
+    )
+    diags["pwat_run_final"] = (
+        input_data["2d"][0].PWAT.isel(time=-2).rename({"time": "final_time"})
+    )
+    diags["pwat_verification_final"] = (
+        input_data["2d"][0].PWAT.isel(time=-2).rename({"time": "final_time"})
```
oliverwm1, line 548: These names (`initial_time` and `final_time`) are not quite accurate. Maybe call the former `first_output_time` and select the `time=-1` index for the latter two so that they are accurate?

oliverwm1 (approved):
Thanks for generalizing the loading code! Hopefully this will allow computing the diags for more runs.

PR #1500, opened 2021-11-17
prog_run: specify input data with mapper config

oliverwm1:
By forcing the user to specify the input data with a "mapper config" it is easier to ensure consistent data is being used for ML training and prescribed runs. It also allows easier iterating on prescribed runs with modified definitions of the apparent sources.

Refactored public API:
- Instead of taking a `url`, the `TendencyPrescriberConfig` now takes a `mapper_config`

Significant internal changes:
- open zarrs with an LRU cache for the `open_zarr` mapper and all fine-res mappers

- [x] Tests added


external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -87,7 +88,7 @@ class FineResBudget(Protocol):
 
 
 def open_zarr(url, consolidated=False):
-    mapper = fsspec.get_mapper(url)
+    mapper = zarr.LRUStoreCache(fsspec.get_mapper(data_path), 128 * 2 ** 20)
```
oliverwm1, line 91: this cache size (128MB) worked well for prescribed runs with Q1 and Q2 precomputed. If we are computing on-the-fly, we might need to make cache bigger, since we need about 7-8x as much input data. (7-8 terms are needed each to compute Q1 and Q2).

brianhenn: Is the idea here that the MPI will call the same chunks multiple times?

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -87,7 +88,7 @@ class FineResBudget(Protocol):
 
 
 def open_zarr(url, consolidated=False):
-    mapper = fsspec.get_mapper(url)
+    mapper = zarr.LRUStoreCache(fsspec.get_mapper(url), 128 * 2 ** 20)
```
brianhenn, line 91: Is the idea with caching that MPI will call the same chunks multiple times since there's not chunking across tiles?

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -44,16 +44,17 @@ class TendencyPrescriber:
...
+        self._tendency_names = list(self.config.variables.values())
+        self._tile = self.communicator.partitioner.tile_index(self.communicator.rank)
 
     def _open_tendencies_dataset(self, time: cftime.DatetimeJulian) -> xr.Dataset:
+        timestamp = vcm.encode_time(time)
+        tile = self.communicator.partitioner.tile_index(self.communicator.rank)
         ds = xr.Dataset()
         if self.communicator.tile.rank == 0:
-            ds = self._tendency_ds.sel(time=time).drop_vars("time").load()
+            ds = self._mapper[timestamp].isel(tile=tile)[self._tendency_names].load()
```
brianhenn, line 57: Nit: use get_variables

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -44,16 +44,17 @@ class TendencyPrescriber:
...
         if self.communicator.rank == 0:
-            logger.debug(f"Opening tendency overriding dataset from: {self.config.url}")
-        tile = self.communicator.partitioner.tile_index(self.communicator.rank)
-        mapper = zarr.LRUStoreCache(fsspec.get_mapper(self.config.url), 128 * 2 ** 20)
-        ds = xr.open_zarr(mapper, consolidated=True)
-        self._tendency_ds = ds[list(self.config.variables.values())].isel(tile=tile)
+            logger.debug(f"Opening tendency override from: {self.config.mapper_config}")
+        self._mapper = self.config.mapper_config.load_mapper()
+        self._tendency_names = list(self.config.variables.values())
+        self._tile = self.communicator.partitioner.tile_index(self.communicator.rank)
```
brianhenn, line 50: Duplicated below and unused

brianhenn (commented):
Looks good, couple very minor things

external/loaders/loaders/mappers/_hybrid.py
oliverwm1: The idea with caching is to take advantage of the fact that in prescribed runs our access pattern is to want subsequent steps in time (which are usually in the same chunk assuming the chunk size is greater than 1 in time). So the first time accessing data from a particular chunk, it gets loaded from GCS and then stored in the cache for subsequent timesteps.

w.r.t. the tiles... in the prog run, we separately load and scatter data from each tile using each tile's master rank. (for the case of 1 rank per tile, no scattering is actually required). So it's better to have a separate chunk for each tile, so we are not loading all the tiles on each rank.

brianhenn: Thanks makes sense

brianhenn (approved):
LGTM

PR #1504, opened 2021-11-17
prog_run: make state fixture have module scope

### LINKED ISSUE #1501: `test_tendency_prescriber` depends on other tests
Within the prognostic run image, if I run

`pytest tests/test_tendency_prescriber.py`

I get a failure but if I run all tests via

`pytest tests`

then the `test_tendency_prescriber` test passes. Clearly this test depends on other tests, possibly through the `state` fixture which is used elsewhere.

oliverwm1:
Some tests (`test_emulator`, `test_fv3fit_transformer` etc) modify the `state` fixture. This PR makes the `state` fixture have a module-level scope (instead of session-level) so that the same input data is used in these separate modules. This ensures tests run as individual modules will gave same results as running many modules worth of tests at once.

Resolves #1501.


nbren12 (approved):
Thanks for tracking down this mutation. Oh...fixtures.

PR #1505, opened 2021-11-17
structured open fine resolution approaches

brianhenn:
Adds a set of approaches to opening the fine-res data based on our current priorities.

Refactored public API:
- `loaders.open_fine_resolution` now takes in an enumerated `approach` argument, a fine-res url, and an optional sequence of additional urls that will be merged into the fine-res data. Available approaches are the pre-existing `apparent_sources_only`, the previous plus adding dynamics nudging tendencies or dynamics differences, and the apparent sources with the bottom layer overwritten with the layer above it.

Significant internal changes:
- Separated the i/o and budget computations behind `open_fine_resolution`
- Refactors the hybrid and fine-resolution code into separate modules for the budget computations and the hybrid and fine res opener functions 

- [X] Tests added


external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
-    Returns:
-        a mapper
-    """
-    return XarrayMapper(
-        open_fine_resolution_dataset(
-            fine_url=fine_url,
-            input_feature_url=input_feature_url,
-            include_temperature_nudging=include_temperature_nudging,
-        )
+    fine["Q1"] = fine.Q1 + additional_source.Q1
```
brianhenn, line 334: Not sure if the "additional source" will always have Q1, Q2 (obviously some renaming of the nudging tendencies is needed), could split this into separate options for the nudging tendency and dynamics difference.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
+    "apparent_sources_extend_lower",
+]
+
+
+@dataclass
+class FineResolutionOpenerConfig:
+    approach: str
+    fine_url: str
+    input_feature_url: Optional[str] = None
+    additional_source_url: Optional[str] = None
```
nbren12, line 374: Suggest removing the url information from this config since these are not necessarily relevant to the re-chunked data.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
+    include_temperature_nudging: bool = False
+
+
+@mapper_functions.register
+def open_fine_resolution(approach: Optional[str] = None, **kwargs: dict):
```
nbren12, line 379: I suggest using this interface

```suggestion
def open_fine_resolution(fine_url: str, input_feature_url: str, additional_source_url: str, config: BudgetConfig):
```

Only the `BudgetConfig` is shared with the prog run...not the urls.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
         a mapper
     """
-    return XarrayMapper(
-        open_precomputed_fine_resolution_dataset(
-            fine_url=fine_url, input_feature_url=input_feature_url
-        )
+    if approach is None:
+        approach = "apparent_sources_only"
+
+    fine_res_config = dacite.from_dict(
```
nbren12, line 402: Better to pass in config directly and leave `dacite` to the calling routine.

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
+    return _add_fine_res_inputs(fine, input_feature_url)
+
+
+FINE_RESOLUTION_APPROACHES = [
+    "apparent_sources_only",
+    "precomputed_apparent_sources",
+    "apparent_sources_plus_nudging",
+    "apparent_sources_plus_dynamics_differences",
+    "apparent_sources_extend_lower",
+]
```
nbren12, line 366: Use `Enum` instead

```
from enum import Enum

class BudgetKinds(Enum):
    apparent_sources_only = 1
    ...

File: external/loaders/loaders/mappers/_hybrid.py [hunk 6]
```diff
@@ -313,52 +321,120 @@ def _add_fine_res_inputs(fine: xarray.Dataset, input_feature_url: Optional[str]
...
+
+
+@dataclass
+class FineResolutionOpenerConfig:
+    approach: str
```
oliverwm1, line 371:
```suggestion
    approach: Literal[*FINE_RESOLUTION_APPROACHES] = "apparent_sources_only"
```
could simplify your default handling and error checking. I think we should be able to use `Literal` now that we're on py3.8

external/loaders/loaders/mappers/_hybrid.py
oliverwm1: Maybe this function should take an argument for the names of the additional sources? Then these could be hard-coded in the function one level up. E.g. for nudging tendencies it would be

`{"Q1": "air_temperature_tendency_due_to_nudging", "Q2": "specific_humidity_tendency_due_to_nudging"}`

nbren12 (commented):
Thanks Brian for putting this together. Cool to see the comprehensive options.

The configuration isn't exactly what in mind. Most of these functions do two things 1) open  some urls and merge the outputs and 2) define from a merged dataset budget.

(1) will not be relevant to prognostic runs since we presumably plan to merge and then save the input data in a single zarr with rechunked time. The config should only be oriented at (2). 


This is what I have in mind:

```

def compute_budget(ds: xr.Dataset, approach: Approaches):
    """
    
    Args:
        ds: a dataset with .... variables.
    """
    if approach  == Approaches.some_approach:
        return compute_budge_with_some_approach(ds)
    ...
    
def open_fine_resolution(fine_url, input_feature_url, additional_source_url, approach):
    merged = # open all the urls, merge them, and do other messy metadata things
    return XarrayMapper(compute_budget(merged: xr.Dataset, approach))
```

`compute_budget` will be shared with the prog run...the merging I/O logic will not.

external/loaders/loaders/mappers/_hybrid.py
oliverwm1, line 402:
```suggestion
    config = dacite.from_dict(
```
the `fine_res` seems a bit redundant here.

nbren12:
Might be easier to pair program on this since I'm asking for some pretty extensive changes.

oliverwm1 (commented):
Generally looks good although it is a bit confusing that some of the kwargs (e.g. `additional_source_url`) are only used for certain values of `approach`. I'm not sure of an easy/simple solution to this. 

This module is getting to be a bit of a monster. Maybe we want to split the "hybrid" and pure "fine-res" funcs? Or split the code that actually computes the budget versus I/O etc?

nbren12:
> Or split the code that actually computes the budget versus I/O etc?

I think this is the key. Let's define a consistent interface for what the output of I/O and merging should be (e.g. what variable names are expected).  You can use a protocol to do this as I have done elsewhere in this module. 


```
class MergedData(Protocol):
    tendency_of_yada_due_to_yada: xr.DataArray
    temperature: xr.DataArray
    ...

class Budget(Protocol):
    Q1: xr.DataArray
    Q2: xr.DataArray

def compute_budget(data: MergedData) -> Budget:
     pass

def _open_blah(blah, blah, blah) -> MergedData:
     ....

def open_blah(blah, blah, blah, approach):
    merged = _open_blah(blah, blah, blah)
    return XarrayMapping(compute_budget_with_approach(merged, approch))
```

external/loaders/loaders/mappers/_hybrid.py
nbren12: Do we need an "additional source" configuration? I could see this getting pretty confusing and makes the config less declarative because it starts to describe behavior (i.e. addition).

brianhenn: OK it sounds like the preference then is to just split this into two declarative approaches that are self-evident (eg `apparent_sources_with_nudging` and `apparent_sources_with_dynamics_differences`)

external/loaders/loaders/mappers/_hybrid.py
brianhenn: Since this is a mapper function we don't get access to the config here, because these functions are enforced to be called as `mapper_func(**kwargs)`. The calling routine only knows about the MapperConfig. I don't think we want to have configuration objects at that level that depend on which mapper function is being called, though maybe there's some clean way to do that. 

I was on the fence about specifying a configuration object that's used within a mapper function, but it works pretty well to help ensure some structure to the arguments (for this function specifically across the different approaches).

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,236 @@
...
+    return fine_source_extended_lower
+
+
+def _ml_standard_names(merged: xr.Dataset):
+
+    # since ML target is Q1/Q2, dQ1=Q1 and pQ1=0 and same for moistening
+    merged["dQ1"] = merged["Q1"]
+    merged["dQ2"] = merged["Q2"]
+    merged["pQ1"] = xr.zeros_like(merged.Q1)
+    merged["pQ1"].attrs = {"units": "K/s", "long_name": "coarse-res physics heating"}
```
brianhenn, line 165: This is copied over but wondering if we can drop the pQs now that they're in the derived mapping?

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,236 @@
...
+        additional_datasets = []
+        for url in additional_dataset_urls:
+            additional_datasets.append(open_zarr_maybe_consolidated(url))
+        merged = xr.merge([fine_shifted, *additional_datasets], join="inner")
+        if "latitude" in merged:
+            merged["latitude"] = merged.latitude.isel(time=0)
+        if "longitude" in merged:
+            merged["longitude"] = merged.longitude.isel(time=0)
+    else:
+        merged = fine_shifted
```
brianhenn, line 57: In the case where no additional datasets are provided, the previous renaming of the fine-res dataset input features (e.g., `T` -> `air_temperature`) got dropped. I could put it back in but it seems like of klunky to have here and not having it doesn't actually prevent training a model with those features.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,236 @@
...
+            "description": merged.Q2.attrs.get("description")
+            + " + dynamics nudging tendencies",
+        }
+    )
+    return Q1, Q2
+
+
+def _add_dynamics_differences(merged: xr.Dataset):
+    with xr.set_options(keep_attrs=True):
+        Q1 = merged.Q1 + merged.fine_minus_coarse_tendency_of_air_temperature_due_to_dynamics
```
brianhenn: This is what Jeremy and I came up with about what this will likely be called. It's a mouthful but pretty unambiguous and vaguely CF-compliant.

external/loaders/loaders/mappers/_fine_res_budget.py
```diff
@@ -0,0 +1,132 @@
+import xarray
```
brianhenn, line 1: This is just a refactor

external/loaders/tests/_regtest_outputs/test__nudged.test_open_fine_resolution.out
```diff
@@ -129,18 +129,6 @@ variables:
 		Q2:units = kg/kg/s ;
 		Q2:long_name = apparent moistening from high resolution data ;
 		Q2:description = Apparent moistening due to physics and sub-grid-scale advection. Given by sat adjustment (dycore) + physics tendency + eddy-flux-convergence ;
-	float32 air_temperature(tile, z, y, x) ;
```
brianhenn, line 132: This is due to the renaming being dropped, see above

external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -213,7 +70,7 @@ def _open_nudged_hybrid_portion(fine_shifted: xarray.Dataset, nudge_url: str):
     merged["latitude"] = merged.latitude.isel(time=0)
     merged["longitude"] = merged.longitude.isel(time=0)
 
-    return merged.astype(numpy.float32)
+    return merged.astype(np.float32)
 
 
 @mapper_functions.register
```
brianhenn, line 76: To avoid too much scope creep I did not also add `approach` to the hybrid opener, so effectively the `apparent_sources_only` option is the only one for the hybrid, which seems fine for now.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,241 @@
...
+
+@mapper_functions.register
+def open_fine_resolution(
+    approach: str,
+    fine_url: str,
+    include_temperature_nudging: bool = False,
+    additional_dataset_urls: Sequence[str] = None,
+) -> GeoMapper:
+    """
+    Open the fine-res mapper using several configuration options
```
brianhenn, line 183: Other than the enumerated `approach` the args are not enforced but hopefully it's simple and documented enough to avoid too much user error.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,241 @@
...
+        }
+    )
+    return Q1, Q2
+
+
+def _extend_lower(fine_source: xr.DataArray, vertical_dim: str = "z") -> xr.Dataset:
+    if fine_source.sizes[vertical_dim] < 2:
+        raise ValueError("vertical_dim must be greater than 1.")
+    fine_source_new_bottom = fine_source.isel({vertical_dim: -2})
+    fine_source_without_bottom = fine_source.isel({vertical_dim: slice(None, -1)})
```
oliverwm1, line 144: my gut would be to use the 3rd level to overwrite the bottom two. Even the 2nd level will be contaminated by the surface level coarsening effects because we do a vertical interpolation during eddy flux convergence calculation.

external/loaders/loaders/mappers/_fine_res.py
brianhenn: Putting them back in to enforce that temp and sphum come from the fine-res data, as discussed

external/loaders/loaders/mappers/_fine_res.py
oliverwm1: These renames were done for the offline diags, but I don't have a great sense of whether they are actually necessary. Is just leaving ML target as Q1/Q2 and not having a dQ1/dQ2 okay? @AnnaKwa

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def standardize_coords(
+    ds: xr.Dataset, time_shift=-timedelta(minutes=7, seconds=30)
+) -> xr.Dataset:
+    ds_shifted = ds.assign(time=ds.time + time_shift)
+    return gfdl_to_standard(ds_shifted).drop("tile")
+
+
+def _open_fine_resolution_dataset(
```
oliverwm1, line 40:
```suggestion
def _open_merged_dataset(
```

oliverwm1 (commented):
Will take another look once changes we discussed on Slack are implemented.

external/loaders/loaders/mappers/_fine_res.py
brianhenn: Sure, happy to make that change.

external/loaders/loaders/mappers/_fine_res.py
brianhenn: My reading of the offline diags is that it's hardcoded that Q2 is derived from dQ2 + pQ2 so this would not work currently.

oliverwm1 (approved):

PR #1510, opened 2021-11-22
Zonal mean nudging tendencies in prog run report

oliverwm1:
Open nudging tendency zarr and plot zonal mean tendencies for
temperature and specific humidity nudging. Useful for evaluating
nudged runs with the prognostic run report.


AnnaKwa (approved):
LGTM

PR #1513, opened 2021-11-23
emulator: diagnose precipitation

### LINKED ISSUE #1485: Diagnose precipitation from microphysics tendencies
For the zhao-carr scheme, the column-integrated total moisture sink and the surface precipitation are identical. See the paper and this [notebook](
https://github.com/ai2cm/explore/blob/master/noahb/emulation/microphysics/2021-11-00-surface%20precip-vs-column%20tendencies.ipynb).

It is not necessary to train a separate ML model for the precip because of this.

nbren12:
For the zhao-carr microphysics, the surface precipitation equals the column
integrated total water sink. Using this relationship, should better enforce
conservation principles.

New Configurations:
- `config.conservative_model` makes a conservative precipitation model if
`config.model==null`.

Resolves #1485

Refactored public API:
- `config.model` is now an `Optional`


external/fv3fit/fv3fit/emulation/models/microphysics.py
```diff
@@ -67,11 +71,7 @@ def name(self):
 
     @property
     def output_variables(self):
-        return (
-            self.direct_out_variables
-            + list(self.residual_out_variables.keys())
-            + list(self.tendency_outputs.values())
-        )
+        return self.direct_out_variables + list(self.residual_out_variables.keys())
```
nbren12, line 74: I changed this property to only include the target variables that must be in the dataset. `.output_variables` is only used in `train_microphysics.py` anyway.

external/fv3fit/fv3fit/emulation/models/microphysics.py
```diff
@@ -155,3 +155,143 @@ def build(self, data: Mapping[str, tf.Tensor],) -> tf.keras.Model:
                 **self._get_residual_outputs(inputs, data, hidden),
             },
         )
+
+
+@dataclasses.dataclass(frozen=True)
+class Field:
```
nbren12, line 161: This way of configuring avoids the coupling between `Microphysics.{input_variables,residual_out_variables, etc}`. The keys, values of those configurations have various constraints that make it easy to make an invalid config.

We could probably modify `MicrophysicsConfig` to use `List[Field]` instead of those configurations.

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -84,6 +81,20 @@ class TrainConfig:
     verbose: int = 2
     shuffle_buffer_size: Optional[int] = 100_000
 
+    @property
+    def _model(
+        self,
+    ) -> Union[models.MicrophysicsConfig, models.ConservativeWaterConfig]:
+        if self.model:
```
frodre, line 88: Defaults are no models so it would be harder to do, but I think adding an exclusivity check to error if both models are set would be good.

external/fv3fit/fv3fit/emulation/models/microphysics.py
```diff
@@ -155,3 +155,143 @@ def build(self, data: Mapping[str, tf.Tensor],) -> tf.keras.Model:
...
+    input_name: str = ""
+    residual: bool = True
+    # only used if residual is True
+    tendency_name: str = ""
+    selection: SliceConfig = dataclasses.field(default_factory=SliceConfig)
+
+
+@dataclasses.dataclass(frozen=True)
+class ZhaoCarrFields:
+    cloud_water: Field = Field(
```
frodre, line 172: It wouldn't work for the current training data, but consider switching these to the new names so the default works going forward.

external/fv3fit/fv3fit/emulation/models/microphysics.py
frodre: Yeah, this is a nice field definition that I think would work well for simplifying `MicrophysicsConfig`.

external/fv3fit/fv3fit/emulation/models/microphysics.py
```diff
@@ -1,11 +1,14 @@
...
+import dacite
 import tensorflow as tf
+from fv3fit._shared import SliceConfig
 
-from ._core import ArchitectureConfig, get_combine_from_arch_key
 from ..layers import FieldInput, FieldOutput, IncrementedFieldOutput
-from fv3fit._shared import SliceConfig
+from ._core import ArchitectureConfig, get_combine_from_arch_key
+
+from fv3fit.emulation import thermo
```
frodre, line 11: Add `__all__` definition since we import from "*" in the module?

frodre: Also, please add some docstrings to at least `Field` and `ConservativeWaterConfig`

frodre (approved):
LGTM, minor comments and suggestions. I'm excited to try this out!

external/fv3fit/fv3fit/train_microphysics.py
nbren12: added a warning for this case.

external/fv3fit/fv3fit/emulation/models/microphysics.py
nbren12: > Add __all__ definition since we import from "*" in the module?

The only private member of this module is prefixed with `_` so `__all__` is unneeded in this case.

external/fv3fit/fv3fit/emulation/models/microphysics.py
nbren12: Done.

PR #1514, opened 2021-11-24
Dynamics difference method for estimating eddy term

nbren12:
Implement the dynamics difference method for estimating the eddy flux term. This method computes the apparent source like this:

    Q = (high_res dyn - coarse dyn) + high_res physics 
        = high res storage - coarse dyn tendency 

This is different from the `apparent_sources_plus_dynamics_difference` method, which is not what I had in mind when originally proposing it. This approach doesn't directly use the eddy flux, so should be less susceptible to numerical discretization errors. 

I deleted the `apparent_sources_plus_dynamics_difference` method. @brianhenn I can revert the last commit if you were independently interested in that method.

### DIFF
--- a/external/loaders/loaders/mappers/_fine_res.py
+++ b/external/loaders/loaders/mappers/_fine_res.py
@@ -70,7 +70,6 @@ def _open_merged_dataset(
 class Approach(Enum):
     apparent_sources_only = 1
     apparent_sources_plus_nudging_tendencies = 2
-    apparent_sources_plus_dynamics_differences = 3
     apparent_sources_extend_lower = 4
     dynamics_difference = 5
 
@@ -109,8 +108,6 @@ def compute_budget(
 
     if approach == Approach.apparent_sources_plus_nudging_tendencies:
         merged["Q1"], merged["Q2"] = _add_nudging_tendencies(merged)
-    elif approach == Approach.apparent_sources_plus_dynamics_differences:
-        merged["Q1"], merged["Q2"] = _add_dynamics_differences(merged)
     elif approach == Approach.apparent_sources_extend_lower:
         merged["Q1"] = _extend_lower(merged["Q1"])
         merged["Q2"] = _extend_lower(merged["Q2"])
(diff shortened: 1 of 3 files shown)


external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -19,6 +20,15 @@ class MLTendencies(Protocol):
     dQ2: xr.DataArray
 
 
+class NudgedRun(Protocol):
+    tendency_of_air_temperature_due_to_dynamics: xr.Dataset
```
nbren12, line 24: @oliverwm1 Can you confirm that these are the names in your prescribed run outputs? Thanks.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -19,6 +20,15 @@ class MLTendencies(Protocol):
...
+    tendency_of_specific_humidity_due_to_dynamics: xr.Dataset
```
oliverwm1, line 25:
```suggestion
    tendency_of_air_temperature_due_to_dynamics: xr.DataArray
    tendency_of_specific_humidity_due_to_dynamics: xr.DataArray
```

external/loaders/loaders/mappers/_fine_res.py
oliverwm1: Yes these are the names.

oliverwm1 (approved):
LGTM, very interested to see the results.

> This approach doesn't use the eddy flux at all, so should be less susceptible to numerical discretization errors.

It does indirectly use it, in that the eddy flux is used to drive the nudged+prescribed run which the coarse dynamics tendencies are coming from. I wonder how different the coarse dynamics are in a 'normal' nudged run.

nbren12:
> It does indirectly use it, in that the eddy flux is used to drive the nudged+prescribed run which the coarse dynamics tendencies are coming from.

Good point. Would be nice to have another way to bootstrap the prescribed run. Maybe running without any dynamics difference estimate for the first prescribed run?

nbren12:
@oliverwm1 Would you have time to run this configuration or point me to the configs you have been using?

PR #1515, opened 2021-11-24
Fix dynamics difference temperature nudging

nbren12:
I mixed up the clause of the boolean `include_temperature_nudging` mixed up.

The returned temperature source should be larger when including the temperature nudging.


oliverwm1 (approved):
Good catch.

PR #1521, opened 2021-11-30
Fix microphysics preformance regression

### LINKED ISSUE #1520: Fix skill regression in microphysics emulation
#1511 significantly damaged the initial skill of an ML model (see 5e7ae958d8de8989d51032f15d00f09b67b1a3c2) . Now I need to revert and figure out why.

Working on this branch: fix-microphysics-regression (

nbren12:
Resolves #1520.

Significant internal changes:
- Initialize `IncrementedFieldOutput` correctly.


nbren12:
> The update didn't have this so denormalizing was to the full field value not the tendency?

That's correct. I could have been more careful in the original PR. Didn't realize the data used to "train" the normalization was the "tendency_of_...." field.

Issue #1519, opened 2021-11-25
Train existing models from new dataset


nbren12:
See dataset in #1491.

nbren12:
Here is the path to the dataset:

> A new dataset has been generated at gs://vcm-ml-experiments/microphysics-emulation/2021-11-24/microphysics-training-data-v3-training_netcdfs

> Here is a link to the artifact: https://wandb.ai/ai2cm/microphysics-emulation/artifacts/training_netcdfs/microphysics-training-data/fcb06e6685228051532b

nbren12:
The new models have all been trained. See this for example: https://wandb.ai/ai2cm/microphysics-emulation/runs/39i42dhe?workspace=user-nbren12

PR #1523, opened 2021-11-30
emulation: train with new data

nbren12:
Related to #1519. I need to update the configs and modify several hardcodes to train against the new data.

I also collapse the scripting layers in the argo experiment launching framework. Previously, this had two bash scripts and a makefile. now it is a single bash script.

Once this is merged I can bump the argo configs and launch the experiments.

Refactored public API:
- refactored microphysics specific hardcodes from `vcm.derived_mapping` into `fv3fit.emultion.data.transforms`.


external/vcm/tests/test_derived_mapping.py
```diff
@@ -194,29 +194,3 @@ def var(self):
...
-        }
-    )
-
-    return ds
-
-
-@pytest.mark.parametrize(
-    "varname", ["air_temperature", "specific_humidity", "cloud_water_mixing_ratio"]
-)
-def test_microphysics_tendencies(varname):
```
nbren12, line 215: Added this coverage to an existing test:
```
tests/emulation/test_data_transforms.py::test_derived[air_temperature] PASSED                                                                                                                                                                                                                                                                                                                                            [ 86%]
tests/emulation/test_data_transforms.py::test_derived[specific_humidity] PASSED                                                                                                                                                                                                                                                                                                                                          [ 93%]
tests/emulation/test_data_transforms.py::test_derived[cloud_water_mixing_ratio] PASSED  ```

File: external/fv3fit/fv3fit/wandb.py [hunk 2]
```diff
@@ -70,6 +60,19 @@ def log_to_table(log_key: str, data: Dict[str, Any], index: Optional[List[Any]]
 def _plot_profiles(target, prediction, name):
     """Plot vertical profile comparisons"""
 
+    units = defaultdict(lambda: "unknown")
+    units.update(
+        {
+            "total_precipitation": "mm / day",
+            "specific_humidity_output": "g/kg",
```
frodre, line 67: `_after_precpd`

projects/microphysics/train/argo.yaml
```diff
@@ -27,6 +30,8 @@ spec:
             value: /secret/gcp-credentials/key.json
           - name: CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE
             value: /secret/gcp-credentials/key.json
+          - name: WANDB_NAME
```
frodre, line 33: It might also be useful to group trainings under something like WANDB_RUN_GROUP={RUNDATE}-{GROUP_NUMBER}.  I've found the sorting by job-type / group has been helpful so far, but prognostic runs are much more numerous, so it might not add much I guess.

projects/microphysics/train/run.sh
```diff
@@ -1,8 +1,27 @@
...
+else
+    bucket="vcm-ml-experiments"
+fi
 
-argo submit argo.yaml \
-    -p training-config="$(< $config_file)"
\ No newline at end of file
+group="$(openssl rand -hex 3)"
+
+for config in all-tendency-limited direct-cloud-limited; do
```
frodre, line 15: Good consolidation.

frodre (approved):
Thanks for running this from end-to-end again!  It's good that both of us have been through it   and hopefully the previous models are reproduced.  

Only necessary changes are that I think the derived variables / units in the profile logging should be tied to the new `_after_precipd` suffix. Other than that LGTM.

projects/microphysics/train/argo.yaml
nbren12: I found searching on the unique "group" suffix makes sense. Currently, we use groups to mean groups of the jobs involved in a single "trial" of an experiment (e.g. all the prog runs for a single model). Perhaps using tags is the way to go since we can change them after the fact, nest tags within other tags, etc.

nbren12:
@frodre I tried updating the hardcodes but ran into more errors in the integration tests. The "default config" used for testing also has hardcodes. I think updating the entire test suite/removing all the hardcodes from the code is a bit out of scope. For the most part, these hardcodes are involved in non mission critical features (e.g. the profile plots). In principle all of these harcodes could probably be inferred from the configuration.

Issue #1522, opened 2021-11-30
Verify that RNN is not actually "downward dependent" using jacobian plots/tests


frodre:
The design of the microphysics model RNNBLock and FieldOutput "undoes" the vertical dependence structure we tried to enforce.

frodre:
Plots show that d(output[i])/d(input[j]) != 0 for i > j

![image](https://user-images.githubusercontent.com/1440875/144109638-856a0378-ad3c-46d6-a881-74880ed2496b.png)

https://github.com/ai2cm/explore/blob/master/andrep/microphysics/2021-11-29-top-down-gradient-constraint.ipynb

nbren12:
Cool plots! I'm also surprised by how noisy the jacobian is for some fields. Maybe some regularization could help.

PR #1525, opened 2021-12-02
Add RNN (x)failing downards dependence test

frodre:
Add a test to verify issue #1522

### DIFF
--- a/external/fv3fit/tests/emulation/models/test_microphysics_config.py
+++ b/external/fv3fit/tests/emulation/models/test_microphysics_config.py
@@ -1,7 +1,8 @@
-from fv3fit._shared import SliceConfig
 import numpy as np
+import pytest
 import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
Looks awesome. For now, I think we'll want to preserve the `rnn` architecture as it currently is since it is the best configuration. Maybe we can call the fix `rnn-v1`?

frodre:
I have it in a "fixed" branch right now as `HybridRNN` and was going to add an arch key like references that model so we can still use it.  Could also go with `RNNv1`

frodre:
I like the `rnn-v1` and `rnn-v2` since that clearly delineates vs updated names.

nbren12:
The advantage of using `rnn` and `rnn-v1` is that it makes it easier to know which architectures are the same in W and B.

frodre:
Ahh, okay, so new rnn is `rnn-v1`.  That makes sense.

PR #1527, opened 2021-12-03
Correct downwards constraint in RNN

### LINKED ISSUE #1522: Verify that RNN is not actually "downward dependent" using jacobian plots/tests

frodre:
This PR reworks output layer connections to preserve downwards dependence of an RNN. 

Resolves #1522 

API changes
- "rnn-v1" and "rnn-v1-shared-weights" for new `RNN` class available via the `ArchitectureConfig`

Internal Chagnes
- Added `RNN` as corrected version of `RNNBlock`, old `RNNBlock` is now `HybridRNN`
- Initial output connections to network are now handled within the architecture (like the input stacking) via new layers
    - `StandardOutput` uses the old style moved from `FieldOutput` of connecting a dense layer with linear activations
    - `RNNOutput` uses local vertical connections to preserve the downwards dependence
- Downwards dependence test checked and now passes in `test_microphysics_config.py`


external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -112,6 +115,83 @@ def get_config(self):
         return config
 
 
+class RNN(tf.keras.layers.Layer):
```
nbren12, line 118: Not essential, but could this be turned into a function returning `tf.keras.Sequential`. e.g. 

```
tf.keras.Sequential([SimpleRNN(depth, go_backwards, channels) for i in range(depth)])
```

`Sequential` is a instance of `Layer` too.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +258,88 @@ def NoWeightSharingSLP(
...
+        """
+        super().__init__(*args, **kwargs)
+        self._feature_lengths = feature_lengths
+
+        self.output_layers = {
+            name: tf.keras.layers.LocallyConnected1D(1, 1, name=f"rnn_output_{name}")
+            for name in self._feature_lengths.keys()
+        }
+
+    def call(self, rnn_outputs) -> Mapping[str, tf.Tensor]:
```
nbren12, line 324: Can you type hint the arg. To do any typechecking, mypy needs all the arguments (except self) to be type hinted (even if just `Any`)

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +258,88 @@ def NoWeightSharingSLP(
...
+        """
+        Args:
+            feature_lengths: Map of output variable names to expected
+                feature dimension length
+        """
+        super().__init__(*args, **kwargs)
+        self._feature_lengths = feature_lengths
+
+        self.output_layers = {
+            name: tf.keras.layers.LocallyConnected1D(1, 1, name=f"rnn_output_{name}")
```
nbren12, line 320: Would also be interesting to try a single dense model for all the output levels. Can you add an option for that?

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +258,88 @@ def NoWeightSharingSLP(
...
+        """
+        super().__init__(*args, **kwargs)
+        self._feature_lengths = feature_lengths
+
+        self.output_layers = {
+            name: tf.keras.layers.Dense(feature_length, name=f"standard_output_{name}")
+            for name, feature_length in self._feature_lengths.items()
+        }
+
+    def call(self, hidden_output) -> Mapping[str, tf.Tensor]:
```
nbren12, line 280: type hint the arg pls.

external/fv3fit/fv3fit/emulation/models/_core.py
```diff
@@ -56,7 +61,25 @@ def get_combine_from_arch_key(key: str):
...
     """
 
-    if key == "rnn":
+    if key == "rnn" or key == "rnn-v1":
         return CombineInputs(-1, expand_axis=-1)
     else:
         return CombineInputs(-1, expand_axis=None)
+
+
+def get_outputs_from_arch_key(key: str, output_features: Mapping[str, int]):
```
nbren12, line 70: Type hint for the output type pls
```suggestion
def get_outputs_from_arch_key(key: str, output_features: Mapping[str, int]) -> tf.keras.Layer:
```

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +258,88 @@ def NoWeightSharingSLP(
             tf.keras.layers.Flatten(),
         ]
     )
+
+
+class StandardOutputConnector(tf.keras.layers.Layer):
```
nbren12, line 263: Can you rename `Connector` to `Layer`...these seem like synonyms in this context. I'd also make these `Connector` classes private APIs. They don't seem optimally composable...a better possible API would be an object that could be used like this: 
```
DictLayer({a: Dense(10), b: Dense(10)}(x)
```
but I am not suggesting you actually do that.

However to support future refactors, you should re-target any coverage provided by the tests of `StandardOutputConnector` and `RNNOutputConnector` to focus on the top-level and `HybridRNN` and  `RNN` classes, respectively.

external/fv3fit/tests/emulation/layers/test_architectures.py
```diff
@@ -91,11 +102,33 @@ def test_no_weight_sharing_num_weights():
...
+
+
+@pytest.mark.parametrize(
+    "connector_cls, hidden_out",
+    [
+        (StandardOutputConnector, _get_tensor((10, 64))),
+        (RNNOutputConnector, _get_tensor((10, 79, 64))),
+    ],
+)
+def test_OutputConnectors(connector_cls, hidden_out):
```
nbren12, line 122: Retarget to tests of `RNN` and `HybridRNN` please if these coverage isn't already there.

nbren12 (approved):
Great to see that we have a passing test here! 

I had some suggestions below. I think you should make the `Connector` classes private API and probably delete their tests---since they will block future refactors and probably don't provide any coverage beyond the tests of `RNN` and `HybridRNN`.

nbren12 (changes requested):
Changing to `Request changes`.

external/fv3fit/fv3fit/emulation/layers/architecture.py
frodre: I am thinking that `go_backwards` and also depending upon the state output would end up with alternating directional dependence, and the output vertical level orientation would depend on the depth in this example.  That's why I opted to use the `tf.reverse` instead.

nbren12: ok...not a big deal. I wouldn't expect SimpleRNN to flip the direction of the output though.

nbren12:
@frodre  It looks like this comment was dismissed: https://github.com/ai2cm/fv3net/pull/1527#discussion_r762153132

I'd prioritize that topic to remove the coupling between `get_outputs_from_arch_key` and the `RNN` architectures.

external/fv3fit/tests/emulation/layers/test_architectures.py
frodre: `test_ArchitectureConfig` adds coverage to whole "architecture".

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +261,165 @@ def NoWeightSharingSLP(
...
+
+    def call(self, tensors: Sequence[tf.Tensor]):
+
+        combined = self.input_combiner(tensors)
+        net_output = self.arch(combined)
+        return self.outputs(net_output)
+
+
+@dataclasses.dataclass
+class ArchitectureConfig:
```
frodre, line 429: I removed all the special handling of input combining, the ML hidden network type, and output dispensing from `emulation.model._core` and the `MicrophysicsConfig`  and hid all of our architecture layers behind `ArcitectureConfig`.

frodre: The `_HiddenArchitecture` with helper functions was simpler than creating a class for each model key, but that's an option if we wanted to make some of them public at some point in the future.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +261,193 @@ def NoWeightSharingSLP(
...
+        Args:
+            feature_lengths: Map of output variable names to expected
+                feature dimension length
+        """
+        super().__init__(*args, **kwargs)
+        self._feature_lengths = feature_lengths
+        self._single_dense = share_conv_weights
+
+        if share_conv_weights:
+            layer_cls = tf.keras.layers.Conv1D
```
frodre, line 342: I believe this would work for the "single dense model" suggestion.  Convolution shares weights for each vertical level instead of individual weights for each level.   Added to `ArchitectureConfig` under the name `rnn-v1-shared-weights`.

frodre (commented):
Thanks @nbren12 for pointing out the issues with the weird state of the API.  I managed to clean it up under the `ArchitectureConfig` so now nothing outside has to worries about the details of how to gather inputs, grab the correct network layer, and output separation.

I tried to add the suggested RNN output adjustment, but let me know if you were thinking of something different.

external/fv3fit/fv3fit/emulation/layers/architecture.py
frodre: `rnn-v1-shared-weights`, LMK if that's what you were thinking.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -178,3 +261,193 @@ def NoWeightSharingSLP(
...
+        feature_lengths: Mapping[str, int],
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        self.arch = _get_arch_layer(key, net_kwargs)
+        self.input_combiner = _get_combine_layer(key)
+        self.outputs = _get_output_layer(key, feature_lengths)
+
+    def call(self, tensors: Sequence[tf.Tensor]):
```
nbren12, line 421: Output type? I think it is
```suggestion
    def call(self, tensors: Sequence[tf.Tensor]) -> Mapping[str, tf.Tensor]:
```

external/fv3fit/fv3fit/emulation/layers/architecture.py
nbren12: I like the approach of hiding this logic behind `_HiddenArchitecture`. I think we are close enough for this PR, but probably need some further iteration/removing of layers.

Do you think a per-output-variable architecture configuration could be compatible with this API (e.g. RNN for temperature and Dense for humidity)? This potential use might break some of our assumptions.

frodre: Hmm, yeah I didn't really design this for per-variable architectures and that would break the assumptions currently.  Not sure if the easiest way to try this would be to just have multiple `MicrophysicsConfigs` and some multi-model emulator, or push down I/O information to the architecture config.  I think as long as the per-variable architectures didn't need to share any information the former would be easy.

nbren12: > sure if the easiest way to try this would be to just have multiple MicrophysicsConfigs and some multi-model emulator

That probable makes the most sense...although maybe it's just cleaner to define new architectures if we want that kind of control.

nbren12 (approved):

PR #1531, opened 2021-12-04
Loaders: fix precomputed fine-res opener

brianhenn:
#1505 made the precomputed fine-res mapper hard to use because it coupled that opener function's metadata formatting to that of `open_fine_resolution`. This PR adds an arg specifying whether to do metadata formatting, or not if it's already been done in the pre-computation.


external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -41,24 +41,31 @@ def standardize_coords(
     return gfdl_to_standard(ds_shifted).drop("tile")
 
 
-def _open_merged_dataset(
-    fine_url: str, additional_dataset_urls: Optional[Sequence[str]]
-) -> FineResBudget:
-
-    fine = open_zarr(fine_url)
-    fine_shifted = standardize_coords(fine)
+def _merge(fine, additional_dataset_urls):
```
oliverwm1, line 44: Add type hints for this func.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -41,24 +41,31 @@ def standardize_coords(
...
+    fine_url: str, additional_dataset_urls: Optional[Sequence[str]]
+) -> FineResBudget:
+
+    fine = open_zarr(fine_url)
+    fine_shifted = standardize_coords(fine)
+
+    merged = _merge(fine_shifted, additional_dataset_urls)
 
     # enforce that these ML inputs come from fine dataset
     merged["air_temperature"] = fine_shifted.T
```
oliverwm1, line 71: should this not be part of the `_merge` func so it happens for both cases?

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -41,24 +41,31 @@ def standardize_coords(
...
     else:
-        merged = fine_shifted
+        merged = fine
+
+    return merged
+
+
+def _open_merged_dataset(
+    fine_url: str, additional_dataset_urls: Optional[Sequence[str]]
+) -> FineResBudget:
```
oliverwm1, line 63: instead of splitting this function (and using a different one in each mapper), could we just have `standardize_fine_res_metadata` be a boolean kwarg for this func?

external/loaders/loaders/mappers/_fine_res.py
brianhenn: Was actually trying to exclude this since the offline datasets I've made don't tend to have T and sphum (just Q1 and Q2).

oliverwm1: Okay, maybe add some if statements to check for existence of `T` and `sphum`? I think code will be easier to follow if both mappers use the same function.

external/loaders/loaders/mappers/_fine_res.py
brianhenn: Depends on answer to above I think.

brianhenn: Done as discussed

oliverwm1 (approved):
LGTM, thanks

PR #1537, opened 2021-12-07
Fix incorrect timestep bug when computing fine res budgets

mcgibbon:
In the fine_res_budget workflow, we were passing `factor` (8) as the timestep in seconds, giving storage terms that were 112.5 times too large. This PR fixes that bug.

Refactored public API:

Significant internal changes:
- Fixed bug where 8 seconds was used in place of the timestep when computing fine-res storage terms

- [ ] Tests added

### DIFF
--- a/workflows/fine_res_budget/budget/config.py
+++ b/workflows/fine_res_budget/budget/config.py
@@ -43,3 +43,4 @@ VARIABLES_TO_AVERAGE = set(
 ) - {"area_coarse", "delp"}
 # coarsening factor. C384 to C48 is a factor of 8
 factor = 8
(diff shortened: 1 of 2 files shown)


workflows/fine_res_budget/budget/config.py
```diff
@@ -43,3 +43,4 @@
 ) - {"area_coarse", "delp"}
 # coarsening factor. C384 to C48 is a factor of 8
 factor = 8
+dt = 15 * 60  # timestep in seconds
```
nbren12, line 46: Thanks for adding the "config" option. should be clearer now!

nbren12 (approved):
Thanks!

PR #1538, opened 2021-12-08
Re-submit fine res coarsening

nbren12:
@mcgibbon fixed some bugs in this workflow so it needs to be re-run.

I was unable to run as is without first fixing some version bugs in the setup.py. This workflow requires the `region` argument of `to_zarr` which was not in `xarray==0.15.0`. I just hacked the new versions into the `setup.py` since it will take much effort to hook this into our version locking sytem.

I submitted the job locally for similar reasons and it ran succesfully:
https://console.cloud.google.com/dataflow/jobs/us-central1/2021-12-07_21_06_53-2096919572283850840?project=vcm-ml

See `submit.sh` for the new output path: gs://vcm-ml-experiments/default/2021-12-07/2020-05-27-40-day-X-SHiELD-simulation-v2/fine-res-budget.zarr


mcgibbon (approved):
LGTM. Good readme notes.

PR #1541, opened 2021-12-09
emulation: save data to tfrecord

nbren12:
To train microphysics emulators we currently use folders of netCDF. Since netCDF is not a native format for tensorflow...this requires using external dependencies like xarray/gcsfs and a custom data loading pipeline to create a `tf.data.Dataset` object. The custom python code can be slow...and potentially hard to distribute externally.

Tfrecords are a performant and native format for tensorlfow. They can be readly quickly in parallel without being blocked by the python GIL. They only need tensorflow builtin functions.

### DIFF
--- a/external/emulation/README.md
+++ b/external/emulation/README.md
@@ -54,3 +54,21 @@ The microphysics emulation loads a keras model specified by the `TF_MODEL_PATH` 
 The model input/output names are used to update the state, so they should match the variables pushed into global state by call_py_fort.  See `GFS_physics_driver.F90` to see the list of available state variables.  Any state field that intersects with a field produced by the emulator will be adjusted to {state_name}_physics_diag so that piggy-backed information is present.
 
 Running with `save_zc_microphysics = True` will save the emulator outputs and diagnostic physics info to file.
+
+
+## Loading TFRecord data
+
+TFRecord is a binary file format that it native to tensorflow. This means that
+loading it requires no pure-python code and should be portable and performant.
+If `SAVE_TFRECORD` is set, then the model will save each rank to a separate
+`.tfrecord` file. Because tfrecords contain raw serialized data, we also save a
+tf module to `parser.tf` to parse it. To open the data, use the
+following boilerplate:
+
+```python
+url = "gs://vcm-ml-experiments/microphysics-emulation/2021-12-13/pzp6alfw/artifacts/20160611.000000/tfrecords"
+parser = tf.saved_model.load(f"{url}/parser.tf")
+tf_ds = tf.data.TFRecordDataset(
+    tf.data.Dataset.list_files(f"{url}/*.tfrecord")
+).map(parser.parse_single_example)
+```
--- a/external/emulation/README.md
+++ b/external/emulation/README.md
@@ -72,3 +72,6 @@ tf_ds = tf.data.TFRecordDataset(
     tf.data.Dataset.list_files(f"{url}/*.tfrecord")
 ).map(parser.parse_single_example)
 ```
+
+Note that tensorflow I/O routines support `GCS` links beginning with
+`gs://some-bucket/some-path`.
--- a/external/emulation/emulation/_filesystem.py
+++ b/external/emulation/emulation/_filesystem.py
@@ -10,6 +10,5 @@ import fsspec
 def get_dir(path: str):
     with tempfile.TemporaryDirectory() as tmpdir:
         fs, _, _ = fsspec.get_fs_token_paths(path)
-        # fsspec places the directory inside the tmpdir, as a subdirectory
         fs.get(path, tmpdir, recursive=True)
         yield tmpdir
--- a/external/emulation/emulation/_filesystem.py
+++ b/external/emulation/emulation/_filesystem.py
@@ -10,5 +10,6 @@ import fsspec
 def get_dir(path: str):
     with tempfile.TemporaryDirectory() as tmpdir:
         fs, _, _ = fsspec.get_fs_token_paths(path)
+        # fsspec places the directory inside the tmpdir, as a subdirectory
         fs.get(path, tmpdir, recursive=True)
         yield tmpdir
--- a/external/emulation/emulation/serialize.py
+++ b/external/emulation/emulation/serialize.py
@@ -23,7 +23,54 @@ def serialize_tensor_dict(data: Mapping[str, Tensor]) -> bytes:
     return example.SerializeToString()
 
 
-def get_parser(data: Mapping[str, Tensor]):
+def get_parser(data: Mapping[str, Tensor]) -> tf.Module:
+    """Returns a ``tensorflow`` module object for parsing byte records saved by
+    ``serialize_tensor_dict``.
+
+
+    Args:
+        data: a dataset object to use as a template
+
+    Returns:
+        a parser object with the following methods:
+
+        - ``parse_single_example``: parser a single bytes record into a
+                dictionary of tensors. Analogous to ``tf.io.parse_single_example``_.
+        - ``parse_example``: goes from a vector of bytes records to a dictionary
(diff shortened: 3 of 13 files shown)


external/emulation/README.md
```diff
@@ -54,3 +54,21 @@ The microphysics emulation loads a keras model specified by the `TF_MODEL_PATH`
...
+`.tfrecord` file. Because tfrecords contain raw serialized data, we also save a
+tf module to `parser.tf` to parse it. To open the data, use the
+following boilerplate:
+
+```python
+url = "gs://vcm-ml-experiments/microphysics-emulation/2021-12-13/pzp6alfw/artifacts/20160611.000000/tfrecords"
+parser = tf.saved_model.load(f"{url}/parser.tf")
+tf_ds = tf.data.TFRecordDataset(
+    tf.data.Dataset.list_files(f"{url}/*.tfrecord")
+).map(parser.parse_single_example)
```
mcgibbon, line 73: Is it possible to load remote data with fsspec?

Should we have a helper function in `vcm` or in `loaders` to simplify this? Possibly one that enables remote data through a local tempdir. I would still keep this explicit tensorflow example if you add a helper.

external/emulation/emulation/_filesystem.py
```diff
@@ -10,6 +10,5 @@
 def get_dir(path: str):
     with tempfile.TemporaryDirectory() as tmpdir:
         fs, _, _ = fsspec.get_fs_token_paths(path)
-        # fsspec places the directory inside the tmpdir, as a subdirectory
```
mcgibbon, line 13: This information isn't obvious from fsspec's function name and isn't present in their documentation, so it may be helpful to keep it.

external/emulation/emulation/serialize.py
```diff
@@ -0,0 +1,68 @@
...
+                parsed,
+                fn_output_signature={
+                    key: tf.TensorSpec(sizes[key], dtypes[key]) for key in sizes
+                },
+            )
+            ds = tf.data.Dataset.from_tensor_slices(parsed)
+            return ds.map(self._parse_dict_of_bytes)
+
+        @tf.function(input_signature=[tf.TensorSpec(shape=[], dtype=tf.string)])
+        def parse_single_example(self, record: tf.Tensor):
```
mcgibbon, line 64: Some docstrings would help me understand when these should get used. What's the reason behind the `_example` naming?

projects/microphysics/scripts/prognostic_run.py
```diff
@@ -40,6 +42,9 @@ def get_env(args):
     default="",
     help="A unique tag. Can be used to look-up these outputs in subsequent timesteps.",
 )
+parser.add_argument("--save-zarr", action="store_true", default=False)
```
mcgibbon, line 45: You've modified the default behavior from storing zarrs to not storing zarrs, is that intended?

mcgibbon (commented):
Really cool! Definitely makes sense to save in this format. My main suggestion/ask is to formalize the API around the parser a bit more. On the Parser, it doesn't make sense to me yet why the methods are named what they are, or what they're doing/why they're needed. A docstring may be enough there. It could be helpful to have a helper function to perform the 3 operations needed to load data at a path, particularly if this would allow the data to be at a remote path.

external/emulation/README.md
nbren12: tf supports remote gs links natively. This code actually loads the data at that URL. There are performance implications to this...which is why i exposed the raw parsing funtions.

external/emulation/emulation/_filesystem.py
nbren12: sorry...i didn't intended to commit this...was related to another PR.

nbren12: reverted.

external/emulation/tests/test_zc_emu_monitor.py
```diff
@@ -1,23 +1,23 @@
...
-    _get_attrs,
     _convert_to_quantities,
     _convert_to_xr_dataset,
-    _translate_time,
     _create_nc_path,
+    _get_attrs,
+    _get_timestep,
+    _load_nml,
+    _remove_io_suffix,
+    _translate_time,
```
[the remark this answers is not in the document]
nbren12: looks like alphabetical order. I added some imports...so decided to run vscode's sort imports feature.

external/emulation/emulation/serialize.py
nbren12: it mirrors tensorflow's functions with the same names. I don't like the names either...but probably best to stick with the established upstream convention. I'll add some info to the docstrings.

projects/microphysics/scripts/prognostic_run.py
nbren12: It was...we should be fine. Better not to save loads of redundant data by default,

external/emulation/tests/test_zc_emu_monitor.py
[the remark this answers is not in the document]
mcgibbon: Oh interesting, I thought we run `isort` as part of pre-commit. On the DSL side, it makes this a lot less ambiguous (and more automatic).

external/emulation/README.md
mcgibbon: I'll say it's OK if you want to merge without one, but a helper function would make it a lot easier for me to work with this data format. This is a lot of minor name and operation details to remember. I feel the same way every time I need to open a zarr (it's always a google search), and that's a simpler set of operations. I don't currently need to work with this data format, so it's not important for now.

nbren12: Sure. we can add a higher level api to `fv3fit` in the future...would be nice for notebook use. Only reason i didn't is that this "higher level" API would tend to grow many options...e.g. shuffle the file names before loading...number of processors for the `.map`...batch size...etc. At some point it's easier to just have the 4 lines of boiler plate rather than a helper with 10 arguments.

external/emulation/emulation/serialize.py
mcgibbon: `tf.io.parse_example` takes as input anything that can be passed to `ops.convert_to_tensor`, and calls it "serialized" if you'd like to keep the same API.

mcgibbon (approved):
LGTM. Some small doc updates to bring them more in line with the tensorflow parser's docs, and food for thought on the call signature.

external/emulation/emulation/serialize.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+    Args:
+        data: a dataset object to use as a template
+
+    Returns:
+        a parser object with the following methods:
+
+        - ``parse_single_example``: parser a single bytes record into a
+                dictionary of tensors. Analogous to ``tf.io.parse_single_example``_.
```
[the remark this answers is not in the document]
nbren12: Thanks. good idea.

external/emulation/emulation/serialize.py
```diff
@@ -0,0 +1,125 @@
...
+            return {
+                key: tf.ensure_shape(
+                    tf.io.parse_tensor(x[key], dtypes[key]), sizes[key]
+                )
+                for key in sizes
+            }
+
+        @tf.function(input_signature=[tf.TensorSpec(shape=[None], dtype=tf.string)])
+        def parse_example(self, records: tf.Tensor) -> tf.data.Dataset:
+            parsed = tf.io.parse_example(records, features)
```
mcgibbon, line 100: If you're updating the other function you may want to update this one to show these are the same concept.
```suggestion
        def parse_example(self, serialized: tf.Tensor) -> tf.data.Dataset:
            parsed = tf.io.parse_example(serialized, features)
```

nbren12: True...but tests passing right now, so I think i'll just merge it to avoid being distracted by this PR this PM.

PR #1543, opened 2021-12-10
Add vcm functions to project field onto a vertical flux

mcgibbon:
This PR adds functions to represent DataArrays as a vertical flux field.

Added public API:
- `vcm.project_vertical_flux` which determines a flux field that approximates a tendency field under the given boundary conditions
- `vcm.flux_convergence` which converts a flux field into a tendency field
- `vcm.cell_center_convergence` is moved from a helper function in the fine res budget mapper, to live closer to similar code

Refactored public API:
- Deleted unused module `vcm.constants`

- [x] Tests added


external/vcm/vcm/projection.py
```diff
@@ -0,0 +1,143 @@
...
+        eddy,
+        delp,
+        input_core_dims=[[dim], [dim]],
+        output_core_dims=[[dim]],
+        dask="parallelized",
+        output_dtypes=[eddy.dtype],
+    )
+
+
+def project_vertical_flux(
```
nbren12, line 38: `flux` is the direct object of this name...implying that this function should take `flux` as an argument...but it actually takes a source. To me... "project" also implies that the output has the same type as the input...e.g. P(source) is a also source so that things like `source - P(source) = error`  can be done.

I'm having trouble thinking of an pithy name. `fit_approximate_flux`?

nbren12 (approved):
Very nice and clean...great to rescue the convergence routines from `loaders`.

Just some quibbles about naming below...

external/vcm/vcm/projection.py
mcgibbon: I'll use `fit_field_as_flux`. Fits are generally understood to be approximate, and `fit_approximate_flux` still has `flux` as its direct object. I do think it's important to be specifying "vertical" here, but none of the other functions use that naming convention at present.

external/vcm/vcm/projection.py
```diff
@@ -0,0 +1,143 @@
...
+        output_dtypes=field.dtype,
+    )
+
+
+def _project_vertical_flux(
+    field: xr.DataArray,
+    pressure_thickness: xr.DataArray,
+    first_level_flux: xr.DataArray,
+    last_level_flux: xr.DataArray,
+) -> xr.DataArray:
```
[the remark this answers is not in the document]
mcgibbon: Are these files not under mypy? We should consider expanding coverage if not.

external/vcm/vcm/projection.py
```diff
@@ -0,0 +1,143 @@
+import xarray as xr
```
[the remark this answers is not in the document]
mcgibbon: I'll use vcm/calc/vertical_flux.py. The module name should help document what we aren't documenting in the function names.

external/vcm/vcm/projection.py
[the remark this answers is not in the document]
nbren12: agreed

external/vcm/vcm/calc/vertical_flux.py
```diff
@@ -0,0 +1,149 @@
...
+    # pad interfaces assuming eddy = 0 at edges
+    return -np.diff(padded, axis=-1) / delp
+
+
+def convergence_cell_center(
+    eddy: xr.DataArray, delp: xr.DataArray, dim: str = "p"
+) -> xr.DataArray:
+    """Compute vertical convergence of a cell-centered flux.
+
+    This flux is assumed to vanish at the vertical boundaries
```
oliverwm1, line 25: I know you just moved these functions here—but could you remove the incorrect comments on lines 16 and 25? The `convergence_cell_center` function does not assume zero flux at the vertical boundaries. It linearly extrapolates from the bottom/top two levels to get the boundary value. See `_center_to_interface` func.

mcgibbon: Yeah no problem.

PR #1545, opened 2021-12-13
bump fsspec/gcsfs to 2021.11.1

nbren12:
Prognostic runs were unable to download some models with `fs.get(...recursive=True)`.


frodre (approved):
LGTM

PR #1546, opened 2021-12-14
fv3fit: set deadline=None for flaky test

AnnaKwa:
I think this might actually fix the flaky test.

Resolves #<1529>

### DIFF
--- a/external/fv3fit/tests/emulation/thermobasis/test_emulator.py
+++ b/external/fv3fit/tests/emulation/thermobasis/test_emulator.py
@@ -17,7 +17,7 @@ from fv3fit.emulation.thermobasis.loss import (
 )
 from utils import _get_argsin
 import pytest
-from hypothesis import given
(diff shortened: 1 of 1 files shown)


external/fv3fit/tests/emulation/thermobasis/test_emulator.py
```diff
@@ -152,6 +152,7 @@ def test_Config_register_parser(args, loss_cls):
 
 
 @given(lists(integers(min_value=0, max_value=100)))
+@settings(deadline=None)
 def test_Config_multi_output_levels(levels):
     str_levels = ",".join(str(s) for s in levels)
```
nbren12, line 157: I think this test may also fail when `levels` has a duplicate entry.  Can you also add this?
```suggestion
    str_levels = ",".join(str(s) for s in set(levels))
```

nbren12 (approved):
Thanks!

PR #1547, opened 2021-12-14
Add water vapor path to timeseries and time-mean diags

oliverwm1:
The prog run report only shows total water path, which includes water condensate species. It's helpful to be able to see timeseries/maps of the water vapor path without the condensates.


brianhenn (approved):
Thanks! I don't think I realized that PWAT included condensate, since the typical definition of precipitable water doesn't.

PR #1551, opened 2021-12-15
emulation: add lat vs pressure skill plots

nbren12:
Previously we did not know the skill as a function of lat/lon.  Turns out the
model has negative skill in antarctica. 

* plot  the lat vs pressure skill to the piggy backed wandb reports.

Example report: https://wandb.ai/ai2cm/microphysics-emulation/runs/1r2y7j3b?workspace=user-nbren12

### DIFF
--- a/projects/microphysics/scripts/piggy-back.py
+++ b/projects/microphysics/scripts/piggy-back.py
@@ -6,6 +6,7 @@ import vcm.catalog
 import vcm.fv3.metadata
 import wandb
 import fv3viz
+import plotly.express as px
 
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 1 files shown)


projects/microphysics/scripts/piggy-back.py
```diff
@@ -144,9 +167,39 @@ def time_dependent_dataset(skills):
...
+        return {
+            f"lat_p_skill/{field}": wandb.Plotly(
+                plot_cloud_skill_zonal(ds, field, "mean")
+            )
+        }
+
+    return func
+
+
+log_functions = [
```
nbren12, line 183: I refactored this script to build up a list of functions with the same interface rather than calling them directly...this moves towards a "registry" style approach.

frodre (approved):
Looks great

PR #1556, opened 2021-12-16
fv3fit: generalize interface of emulator loss functions

nbren12:
Keras' .compile function is poorly documented for multiple output problems.
Also, for more complex loss functions (e.g. temperature scaled loss), this
interface is not suitable.

* refactor loss functions to have the interface `loss, info = loss_fn(data, prediction)`
* remove the unused StandardLoss object
* move loss functions to `fv3fit.emulation.losses`
* `fv3fit.emulation.train`: separate loss and optimzer into separate args

### DIFF
--- a/external/fv3fit/fv3fit/emulation/losses.py
+++ b/external/fv3fit/fv3fit/emulation/losses.py
@@ -1,0 +1,82 @@
+import dataclasses
+from typing import List, Mapping
+
+import tensorflow as tf
+from fv3fit.emulation.layers.normalization import NormalizeConfig
+
+from .._shared.config import OptimizerConfig
+
+
+class NormalizedMSE(tf.keras.losses.MeanSquaredError):
+    """
+    Keras MSE that uses an emulation normalization class before
+    scoring
+    """
+
+    def __init__(self, norm_cls_name, sample_data, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._normalize = NormalizeConfig(norm_cls_name, sample_data).initialize_layer()
+
+    def call(self, y_true, y_pred):
+        return super().call(self._normalize(y_true), self._normalize(y_pred))
+
+
+@dataclasses.dataclass
+class CustomLoss:
+    """
+    Use custom custom normalized MSE-based losses for specified
+    variables
+
+    Args:
+        optimizer: configuration for the optimizer to
+            compile with the model
+        normalization: the normalization type (see normalization.py) to
+            use for the MSE
+        loss_variables: variable names to include in the MSE loss dict
+        metric_variables: variable names to include in the metrics dict
+        weights: custom scaling for the loss variables applied in the
+            overall keras "loss" term
+    """
+
+    optimizer: OptimizerConfig = dataclasses.field(
+        default_factory=lambda: OptimizerConfig("Adam")
+    )
+    normalization: str = "mean_std"
+    loss_variables: List[str] = dataclasses.field(default_factory=list)
+    metric_variables: List[str] = dataclasses.field(default_factory=list)
+    weights: Mapping[str, float] = dataclasses.field(default_factory=dict)
+
+    def prepare(self, output_samples: Mapping[str, tf.Tensor]):
+        """
+        Prepare the normalized losses for each variable by creating a
+        fitted NormalizedMSE object and place them into the respective
+        loss (+ weights) or metrics group
+
+        Args:
+            output_names: names of each output the model produces
+            output_samples: sample tensors for each output to fit
+                the normalizing layer
+
+        """
+        self.loss_funcs = {}
+        for out_varname, sample in output_samples.items():
+            self.loss_funcs[out_varname] = NormalizedMSE(self.normalization, sample)
+
+    def __call__(self, x, y):
+        try:
+            self.loss_funcs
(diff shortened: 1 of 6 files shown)


external/fv3fit/fv3fit/emulation/losses.py
```diff
@@ -0,0 +1,82 @@
...
+            use for the MSE
+        loss_variables: variable names to include in the MSE loss dict
+        metric_variables: variable names to include in the metrics dict
+        weights: custom scaling for the loss variables applied in the
+            overall keras "loss" term
+    """
+
+    optimizer: OptimizerConfig = dataclasses.field(
+        default_factory=lambda: OptimizerConfig("Adam")
+    )
```
frodre, line 43: Remove optimizer from the dataclass

external/fv3fit/fv3fit/emulation/losses.py
```diff
@@ -0,0 +1,82 @@
...
+        loss = 0.0
+        for out_varname in x:
+            if out_varname in self.loss_variables + self.metric_variables:
+                loss_value = self.loss_funcs[out_varname](
+                    x[out_varname], y[out_varname]
+                )
+                weight = self.weights.get(out_varname, 1.0)
+                if out_varname in self.loss_variables:
+                    loss += loss_value * weight
+                metrics[out_varname] = loss_value
```
frodre, line 81: Nice! So analagously, everything that's in metrics *or* loss will still have an individual value tracked.

external/fv3fit/fv3fit/emulation/trainer.py
```diff
@@ -12,10 +15,31 @@ def call(self, x):
...
 
 
+class _LayerTrainer(tf.keras.Model):
+    """A class for training multiple output layers
+    
+    This uses the more explicit `add_loss` API for clarity
+    """
+
+    def __init__(self, model: tf.keras.Model, loss: LossFunction):
+        # TODO doesn't work for "StandardLoss"...can we delete that one?
```
frodre, line 25:
```suggestion
```

external/fv3fit/fv3fit/emulation/trainer.py
```diff
@@ -63,25 +88,15 @@ def train(
...
     # This minimizes our contact points to keras APIs...which are often
     # poorly documented
     return wrapped_model.fit(
-        dataset.map(split_in_out),
+        dataset,
         epochs=epochs,
-        validation_data=validation_data.map(split_in_out)
-        if validation_data is not None
-        else None,
+        validation_data=validation_data if validation_data is not None else None,
```
frodre, line 99:
```suggestion
        validation_data=validation_data,
```

frodre (commented):
Looks good and +1 to making it use a more clearly defined keras API since we're on dictionaries now. 

Would be nice to see an example wandb training just to verify that we still get the output metrics tracked as expected.  Also, not sure if the hypothesis failure is something that is actually broken with the unit tests.

external/fv3fit/fv3fit/emulation/losses.py
nbren12: I wanted to preserve backwards compatibility for now...so didn't change this interface. I'll probably separate the config from the `__call__`  implementation in a future PR.

external/fv3fit/fv3fit/emulation/losses.py
nbren12: Yes...I believe so. I tried to leave this behavior unchanged.

nbren12: Here's a run with the new model: https://wandb.ai/ai2cm/microphysics-emulation/runs/30u6vfcn

nbren12: Actually, I needed to append `_loss` to make the names unchanged: https://wandb.ai/ai2cm/microphysics-emulation/runs/3iijf23y/logs?workspace=user-nbren12.

frodre: Thanks for checking it.

external/fv3fit/fv3fit/emulation/losses.py
```diff
@@ -0,0 +1,83 @@
...
+        for out_varname in x:
+            if out_varname in self.loss_variables + self.metric_variables:
+                loss_value = self.loss_funcs[out_varname](
+                    x[out_varname], y[out_varname]
+                )
+                weight = self.weights.get(out_varname, 1.0)
+                if out_varname in self.loss_variables:
+                    loss += loss_value * weight
+                # append "_loss" for backwards compatibility
+                metrics[out_varname + "_loss"] = loss_value
```
frodre, line 82: It's probably okay to have everything suffixed for simplicity, but it might be confusing down the line if we have metric variables that are not included in the loss labeled as such.

frodre (approved):
LGTM and thanks for the clarifications.  Just fix up the test function w/ the suffixes and should be good to go.

external/fv3fit/fv3fit/emulation/losses.py
nbren12: Agreed. I think we may want to remove the "metric_variables" config and replace it with separate "metrics" functions with a similar interface.

PR #1557, opened 2021-12-16
rename argument of parse_example from 'records' to 'serialized'

mcgibbon:
Follow on PR to https://github.com/ai2cm/fv3net/pull/1541 renaming an argument so the names match between methods.


nbren12 (approved):
LGTM. sorry for being lazy about this change.

mcgibbon:
No problem, it was reasonable.

PR #1559, opened 2021-12-17
Add emulator jacobian plots

### LINKED ISSUE #1560: Jacobian scoring into train_microphysics.
#1559

frodre:
In creating the fixed RNN, it was useful to use output sensitivity plots to investigate model behavior.  This PR adds jacobian plots to the WandB training process.

[Example](https://wandb.ai/ai2cm/crapcrap/runs/1s3wu2tq?workspace=user-aperkins)

Resolves #1560 

The jacobians are added to the end of the training script because reloading saved models had some missing metadata that made the gradient calculation much more difficult (probably custom layer data not saved in the trace).  Everything important is stored before attempting the gradient calculation.

Breaking Changes:
- Updates `TrainConfig` defaults to new datasets and associated derived mapping hardcodes for tendencies,  (`_output` suffix -> `_after_precpd` suffix) 

Minor Adjustments:
- fixed the ability to specify tendency output names via config
- fixed some inconsistencies in the keras model outputs vs. the `MicrophysicsConfig.output_variables`

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/transforms.py
+++ b/external/fv3fit/fv3fit/emulation/data/transforms.py
@@ -27,19 +27,19 @@ def _register_derived_variables(mapping: DerivedMapping, timestep):
     # the data or configuration.
     @mapping.register(
         "tendency_of_air_temperature_due_to_microphysics",
-        required_inputs=["air_temperature_input", "air_temperature_output"],
+        required_inputs=["air_temperature_input", "air_temperature_after_precpd"],
     )
     def _(self):
-        end = self._mapper["air_temperature_output"]
+        end = self._mapper["air_temperature_after_precpd"]
         begin = self._mapper["air_temperature_input"]
         return (end - begin) / timestep
 
     @mapping.register(
         "tendency_of_specific_humidity_due_to_microphysics",
-        required_inputs=["specific_humidity_input", "specific_humidity_output"],
+        required_inputs=["specific_humidity_input", "specific_humidity_after_precpd"],
     )
     def tendency_of_specific_humidity_due_to_microphysics(self):
-        end = self._mapper["specific_humidity_output"]
+        end = self._mapper["specific_humidity_after_precpd"]
         begin = self._mapper["specific_humidity_input"]
         return (end - begin) / timestep
 
@@ -47,11 +47,11 @@ def _register_derived_variables(mapping: DerivedMapping, timestep):
         "tendency_of_cloud_water_mixing_ratio_due_to_microphysics",
         required_inputs=[
             "cloud_water_mixing_ratio_input",
-            "cloud_water_mixing_ratio_output",
+            "cloud_water_mixing_ratio_after_precpd",
         ],
     )
     def tendency_of_cloud_water_mixing_ratio_due_to_microphysics(self):
-        end = self._mapper["cloud_water_mixing_ratio_output"]
+        end = self._mapper["cloud_water_mixing_ratio_after_precpd"]
         begin = self._mapper["cloud_water_mixing_ratio_input"]
         return (end - begin) / timestep
 
--- a/external/fv3fit/fv3fit/emulation/keras.py
+++ b/external/fv3fit/fv3fit/emulation/keras.py
@@ -182,7 +182,8 @@ def get_jacobians(model: ModelType, inputs: Mapping[str, tf.Tensor]) -> Mapping[
 
     Args:
         model: model to calculate sensitivity matrices with
-        inputs: inputs to calculate sensitivity against
+        inputs: inputs to calculate sensitivity against, expects
+            tensors with dimensions of [1, nfeatures]
     """
 
     with tf.GradientTape(persistent=True) as g:
@@ -198,10 +199,9 @@ def get_jacobians(model: ModelType, inputs: Mapping[str, tf.Tensor]) -> Mapping[
     return all_jacobians
 
 
-def get_model_output_sensitivities(
-    model: ModelType,
+def normalize_jacobians(
+    all_jacobians: Mapping[str, OutputSensitivity],
     sample: Mapping[str, tf.Tensor],
-    input_names: Sequence[str],
 ) -> Mapping[str, OutputSensitivity]:
     """
     Generate sensitivity jacobions for each output of a model and
@@ -213,21 +213,12 @@ def get_model_output_sensitivities(
     by [ std_input / std_output ].
     """
 
-    avg_profiles = {
-        name: tf.reduce_mean(data, axis=0, keepdims=True)
-        for name, data in sample.items()
-    }
-
     # normalize factors so sensitivities are comparable but still
     # preserve level-relative magnitudes
     normalize_factors = {
         name: float(standard_deviation_all_features(data))
         for name, data in sample.items()
     }
-
-    input_data = {name: avg_profiles[name] for name in input_names}
-
-    all_jacobians = get_jacobians(model, input_data)
 
     normalized_jacobians = {}
     for out_name, per_input_jacobians in all_jacobians.items():
@@ -236,4 +227,4 @@ def get_model_output_sensitivities(
             factor = normalize_factors[in_name] / normalize_factors[out_name]
             normalized_jacobians.setdefault(out_name, {})[in_name] = j * factor
 
-    return all_jacobians
+    return normalized_jacobians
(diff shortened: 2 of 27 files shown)


external/fv3fit/fv3fit/emulation/data/transforms.py
```diff
@@ -27,31 +27,31 @@ def _register_derived_variables(mapping: DerivedMapping, timestep):
     # the data or configuration.
     @mapping.register(
         "tendency_of_air_temperature_due_to_microphysics",
-        required_inputs=["air_temperature_input", "air_temperature_output"],
+        required_inputs=["air_temperature_input", "air_temperature_after_precpd"],
```
frodre, line 30: These are very brittle hardcodes.  Should probably switch to `Field` definition used by the conservative model and we can build up a derived mapping necessary for sample processing based on that.

external/fv3fit/fv3fit/emulation/layers/fields.py
```diff
@@ -136,6 +136,7 @@ def __init__(
         sample_out: Optional[tf.Tensor] = None,
         denormalize: Optional[str] = None,
         enforce_positive: bool = False,
+        tendency_name: str = None,
```
nbren12, line 139: This is not a valid type hint. Type hints are pretty useful...I'm wondering if this part of the code is not being checked.
```suggestion
        tendency_name: Optional[str] = None,
```

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
             self.cache = False
 
 
+def get_model_output_sensitivities(
+    model: tf.keras.Model, sample: Mapping[str, tf.Tensor],
+):
```
nbren12, line 213: Output type hint please. mypy does nothing if there is no output type and it is nice to know at a glance what this function returns since the docstring isn't clear and the code is long.

This function does a lot...I'd consider separating the "jacobian of a dict" part from the "compute mean and normalize output part".

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
             self.cache = False
 
 
+def get_model_output_sensitivities(
+    model: tf.keras.Model, sample: Mapping[str, tf.Tensor],
```
nbren12, line 212: this function should work for any tensorflow code...so best to pose broadly.
```suggestion
    model: Callable[[Mapping[str, tf.Tensor]], Mapping[str, tf.Tensor]], sample: Mapping[str, tf.Tensor],
```

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
...
+        normalize_factors[name] = factor
+
+    input_data = {name: avg_profiles[name] for name in model.input_names}
+
+    with tf.GradientTape(persistent=True) as g:
+        g.watch(input_data)
+        outputs = model(input_data)
+
+    all_jacobians = {}
+    for out_name in model.output_names:
```
nbren12, line 239: this will remove the dependence on the `model.output_names` attribute and allow this to work with any tf function code and be more robust.

```suggestion
    for out_name in outputs:
```

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
             self.cache = False
 
 
+def get_model_output_sensitivities(
+    model: tf.keras.Model, sample: Mapping[str, tf.Tensor],
+):
+    """
+    Generate sensitivity jacabion for each output relative to a mean
+    value for each input
```
nbren12, line 216: Please describe outputs and the normalization strategy (e.g. what the units of the outputs)?

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
...
+
+        all_jacobians[out_name] = normalized
+
+    return all_jacobians
+
+
+OutputSensitivity = Mapping[str, np.ndarray]
+
+
+def plot_single_output_sensitivities(name: str, jacobians: OutputSensitivity):
```
nbren12, line 256: output type please.

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -204,6 +208,93 @@ def __post_init__(self) -> None:
             self.cache = False
 
 
+def get_model_output_sensitivities(
```
nbren12, line 211: this file is getting rather long. Can you move the jacobian code (no plots) to a seperate module...and the plotting code elsewhere too...maybe there is an appropriate place already.

nbren12 (changes requested):
Thanks for adding this...it would be nice if the plots had labels and units...some other nits below.

> The jacobians are added to the end of the training script because reloading saved models had some missing metadata that made the gradient calculation much more difficult 

Maybe add an xfail test demonstrating the problem so we can fix it later?

external/fv3fit/fv3fit/train_microphysics.py
frodre: Separated the jacobian  calculation from the jacobian normalizing.   Added the specifics for the functions to `emulation/keras.py` and the plotting to `wandb.py`.

external/fv3fit/tests/emulation/models/test_microphysics_config.py
```diff
@@ -185,3 +186,37 @@ def test_RNN_downward_dependence():
             sensitivity = jacobian[output_level, input_level]
             if output_level > input_level and sensitivity != 0:
                 raise ValueError("Downwards dependence violated")
+
+
+@pytest.mark.xfail
+def test_saved_model_jacobian():
```
frodre, line 192: Failing test for jacobian calculation on saved RNNs

external/fv3fit/fv3fit/train_microphysics.py
frodre: Moved jacobian specifics into `emulation/keras.py` and the plotting into `wandb.py`.

frodre:
New example with some axes labels: https://wandb.ai/ai2cm/crapcrap/runs/44jadbkr?workspace=user-aperkins

external/fv3fit/fv3fit/emulation/keras.py
```diff
@@ -1,10 +1,11 @@
 import dataclasses
```
nbren12: Maybe move to a new module named `jacobian.py` instead. `keras.py` is a pretty random assortment of things.

external/fv3fit/tests/emulation/test_keras.py
```diff
@@ -171,3 +172,22 @@ def test_StandardLoss_prepare_arbitrary_kwargs():
...
+def test_get_model_output_sensitivities():
+    sample = {
+        "a": tf.random.normal((100, 5)),
+        "b": tf.random.normal((100, 5)),
+    }
+    sample["field"] = sample["a"] + sample["b"]
+
+    inputs = {"a": tf.keras.Input(5), "b": tf.keras.Input(5)}
+    out = tf.keras.layers.Lambda(lambda x: {"field": x["a"] + x["b"]})(inputs)
+    model = tf.keras.Model(inputs=inputs, outputs=out)
```
nbren12, line 132: Optional...but does it need to be a model? I thought this would work with a plain old function. This would also ensure `get_model_output_sensitivies` is not using any of the more arcane APIs of `tf.keras.Model`.
```
def model(x):
    return {"field": x["a"] + x["b"]}
```

external/fv3fit/tests/emulation/test_keras.py
```diff
@@ -112,3 +114,26 @@ def test_model_score_no_outputs():
...
+
+    jacobians = standardize_jacobians(get_jacobians(model, profiles), sample,)
+
+    assert set(jacobians) == {"field"}
+    assert set(jacobians["field"]) == {"a", "b"}
+    for j in jacobians["field"].values():
+        assert j.shape == (5, 5)
```
nbren12, line 139: Required. You should check the actual values:
```suggestion
        n = 5  # maybe move to the top of the test
        np.testing.assert_array_almost_equal(jacobians["field"]["a"], np.eye(n))            
        np.testing.assert_array_almost_equal(jacobians["field"]["b"], np.eye(n))

```

nbren12 (approved):
Thanks for the improvements. I have a couple final suggestions below.

external/fv3fit/tests/emulation/test_keras.py
frodre: It did not need a model to work, so I switched to use this

PR #1563, opened 2021-12-20
Prognostic run: interpolate prescribed tendencies

brianhenn:
To do runs with physics prescribed from the SHiELD run, but at physics timesteps other than 15 minutes, we need to interpolate the prescribed Q values. This PR refactors the function time interpolation decorator we've been using in the nudging routine to be used in the tendency prescriber as well, since the context is similar (data is accessed via a timestep key and caching is advantageous). It also refactors some of the setup routines of the tendency prescriber to make testing easier.

Added public API:
- Added `reference_initial_time` and `reference_frequency_seconds` optional args to the `TendencyPrescriberConfig` (the same as are found in the nudging config). 

Refactored public API:
- Refactored function time interpolation routines into a new `runtime.interpolate` module.

Significant internal changes:
- Moved setup of `time_lookup_function` and the quantile limiter for the tendency prescriber into its own function, and pass the `time_lookup_function` function as an argument to the `TendencyPrescriber`.

- [X] Tests added


nbren12:
>Changed the API of the TendencyPrescriber (see below) but in practice we don't access this object directly, and the compatibility of its configuration class is maintained.

Perhaps reword this to clarify that you only change the __init__ API. I would go even further argue this isn't a public API, so maybe just delete the comment altogether.

workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -51,10 +59,48 @@ def get_tendency_prescriber(
...
             communicator,
             timestep,
+            prescriber_config.variables,
+            mapper_function,
+            limit_quantiles=prescriber_config.limit_quantiles,
             diagnostic_variables=set(config.diagnostic_variables),
         )
+
+
+def _get_mapper_function(
```
nbren12, line 84: Can we use a more specific name e.g. `_get_time_lookup_func` or something?

workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -51,10 +59,48 @@ def get_tendency_prescriber(
...
+        ds = mapper[timestamp]
+        return {var: ds[var] for var in tendency_variables}
+
+    if initial_time is not None:
+        initial_time = label_to_time(initial_time)
+        return_func = time_interpolate_func(
+            mapper_function, timedelta(seconds=frequency_seconds), initial_time
+        )
+    else:
+        return_func = mapper_function
```
nbren12, line 104: Nit...but I personally prefer to eagerly return things in these kinds of statements rather than introducing a new symbol that could be modified later on.
```suggestion
        return time_interpolate_func(
            mapper_function, timedelta(seconds=frequency_seconds), initial_time
        )
    else:
        return mapper_function
```

workflows/prognostic_c48_run/runtime/interpolate.py
```diff
@@ -0,0 +1,72 @@
...
+import xarray as xr
+import cftime
+from datetime import timedelta
+import functools
+from typing import Callable
+
+
+# list of variables that will use nearest neighbor interpolation
+# between times instead of linear interpolation
+INTERPOLATE_NEAREST = [
```
nbren12, line 12: Not new...but the indirection with the `MASK` constant threw me off. Can you add a comment like this:
```suggestion
INTERPOLATE_NEAREST = [
    # mask is integer-valued so interpolate using nearest time point
```

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -27,55 +24,56 @@ class TendencyPrescriberConfig:
...
 
 @dataclasses.dataclass
 class TendencyPrescriber:
     """Wrap a Step function and prescribe certain tendencies."""
 
-    config: TendencyPrescriberConfig
     state: DerivedFV3State
     communicator: pace.util.CubedSphereCommunicator
     timestep: float
+    variables: Mapping[str, str]
```
nbren12, line 47: Can you document this attribute with an example?

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -27,55 +24,56 @@ class TendencyPrescriberConfig:
...
+    mapper_func: Callable[[cftime.DatetimeJulian], State]
+    limit_quantiles: Optional[Mapping[str, float]] = None
```
nbren12, line 49: While you are extracting things from this class you could  remove the limiter logic too. This could be bundled into the `mapper_func` using a easily tested function like this:
```
def limit_mapper_func(limiter: DatasetQuantileLimiter, mapper_func):
    def limited_func(time):
        return limiter.transform(mapper_func(time)) 
    return limited_func
```

This will allow you to delete two methods, the optional `self._limiter` attribute + corresponding if statements, and remove one argument from the constructor. You can initialize the `limiter` close to where you initialize the mapper_func.

I suppose this optional since the limiting logic isn't new, but I think it would make this class align much better with the Single Responsibility Principle. The whole purpose of the class is just to execute mapper_func on the root rank and scatter it to the others.

brianhenn: Agree that separation is desirable, will do

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
brianhenn: That variable is documented with an example in the config class, seems like that's probably enough

brianhenn: I added a docstring for the `TendencyPrescriber` itself, mainly so that the `time_lookup_function` has some description

nbren12 (approved):
LGTM. Thanks for the last round of refactors.

PR #1565, opened 2021-12-20
RNN training optimizations

frodre:
This PR includes a new image (`prognostic_run_gpu`) built and pushed by CircleCI capable of training using our GPU nodes (`gpu-sim-pool`) on the cluster.   I upgraded the tensorflow requirement to [2.5.2](https://github.com/tensorflow/tensorflow/releases/tag/v2.5.2) to be congruent with the CUDA libraries in the prognostic run image.  Instead of forcing this update across the whole repo w/ a merge to master, I chose our paper branch for now.

Experiment configs separated into #1570 

resolves #1564

PR #1572, opened 2021-12-24
emulation:  train RNNs on GPUs

nbren12:
These commits are all from #1565. I thought #1565  would be going into master rather than the `emulation-paper` branch. There are several changes that should be in master.

I already approved that PR...so I can I get someone's rubber stamp?


frodre:
I didn't know if the tensorflow update to 2.5.2 would be okay for master so I opted to keep it on the branch.  I did see that this broke the unit test for something in the report diags.  Maybe everything but the TF update could work?

nbren12:
This looks like the failing test:

    workflows/diagnostics/tests/offline/test_compute.py::test_offline_diags_integration

nbren12:
As long as the tests pass...I don't see any issue with using a newer tensorflow version.

AnnaKwa (approved):
rubber stamp of approval

AnnaKwa:
@spencerkclark FYI this is the dependency issue Noah was referring to:

The weird offline diags test failure happens when the report generation tries to reference a key in a dimension created using `pd.Index`, caused by updating the version for the dependency`h5py== 3.1.0` (the tensorflow version required in this PR pins h5py to this version).

```
E                   KeyError: "not all values found in index 'derivation'"

/opt/conda/envs/fv3net/lib/python3.7/site-packages/xarray/core/indexes.py:232: KeyError
```

This doesn't happen in the master image but can be replicated by updating just the `h5py` version.

spencerkclark:
Thanks @AnnaKwa -- I looked into this a little more and was reminded of some discussion in xarray (https://github.com/pydata/xarray/issues/4570, https://github.com/pydata/xarray/pull/4893).  The h5netcdf-backend-related fix in https://github.com/pydata/xarray/pull/4893 should be in the version of xarray that we are currently pinning in fv3net, but I think we need to update our version of the h5netcdf backend to at least v0.10.0 ([latest is v0.12.0](https://github.com/h5netcdf/h5netcdf/tags)) to take in account the changes in https://github.com/h5netcdf/h5netcdf/pull/82, and https://github.com/h5netcdf/h5netcdf/pull/85.

I'm not 100% sure that's the issue, but this seems string-related and has to do with the upgrade to h5py, so I think it is worth a try.

AnnaKwa:
> Thanks @AnnaKwa -- I looked into this a little more and was reminded of some discussion in xarray ([pydata/xarray#4570](https://github.com/pydata/xarray/issues/4570), [pydata/xarray#4893](https://github.com/pydata/xarray/pull/4893)). The h5netcdf-backend-related fix in [pydata/xarray#4893](https://github.com/pydata/xarray/pull/4893) should be in the version of xarray that we are currently pinning in fv3net, but I think we need to update our version of the h5netcdf backend to at least v0.10.0 ([latest is v0.12.0](https://github.com/h5netcdf/h5netcdf/tags)) to take in account the changes in [h5netcdf/h5netcdf#82](https://github.com/h5netcdf/h5netcdf/pull/82), and [h5netcdf/h5netcdf#85](https://github.com/h5netcdf/h5netcdf/pull/85).
> 
> I'm not 100% sure that's the issue, but this seems string-related and has to do with the upgrade to h5py, so I think it is worth a try.

Awesome, thanks, that works.

PR #1575, opened 2022-01-06
emulation: allow rnn to train with 2d inputs

nbren12:
When passed scalar inputs like surface pressure, the RNN architecture complains about incompatible shapes. It does not handle the broadcasting of inputs appropriately.

I suggest reviewing commit-by-commit.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/layers/architecture.py
+++ b/external/fv3fit/fv3fit/emulation/layers/architecture.py
@@ -31,6 +31,7 @@ def combine_inputs(
     """
         Args:
             inputs: a datastructure of tensors to combine into a single tensor
+                (sorted by key)
             combine_axis: Axis to concatenate tensors along.  Note that if expand_axis
                 is specified, it is applied before concatenation.  E.g., combine_axis=1
                 and expand_axis=1 will concatenate along the newly created dimension.
--- a/external/fv3fit/fv3fit/emulation/layers/architecture.py
+++ b/external/fv3fit/fv3fit/emulation/layers/architecture.py
@@ -45,8 +45,25 @@ def combine_inputs(
 
 
 class RNNLayer(tf.keras.layers.Layer):
+    """Base class for RNN architectures
+
+    Scalar inputs (with singleton final dimensions) are passed to every element
+    of the input sequence like this::
+
+        h[i+1] = rnn(inputs[i], h[i], scalars)
+
+    """
+
     def input_layer(self, inputs: Mapping[str, tf.Tensor]) -> tf.Tensor:
-        return combine_inputs(inputs, combine_axis=-1, expand_axis=-1)
+        list_inputs = [inputs[key] for key in sorted(inputs)]
+        expanded_inputs = [tf.expand_dims(tensor, axis=-1) for tensor in list_inputs]
+        seq_lengths = set(input.shape[1] for input in expanded_inputs)
+        nz = list(seq_lengths - {1})[0]
+        broadcasted_inputs = [
+            tf.repeat(tensor, nz, 1) if tensor.shape[1] == 1 else tensor
+            for tensor in expanded_inputs
+        ]
+        return tf.concat(broadcasted_inputs, axis=-1)
 
 
 class HybridRNN(RNNLayer):
--- a/external/fv3fit/fv3fit/emulation/layers/architecture.py
+++ b/external/fv3fit/fv3fit/emulation/layers/architecture.py
@@ -52,10 +52,11 @@ def combine_inputs(
 class RNNLayer(tf.keras.layers.Layer):
     """Base class for RNN architectures
 
-    Scalar inputs (with singleton final dimensions) are passed to every element
     of the input sequence like this::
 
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -20,52 +20,53 @@
...
-        config = super().get_config()
-        config.update(
-            {"combine_axis": self._combine_axis, "expand_axis": self._combine_axis}
-        )
-        return config
+        h[i+1] = rnn(inputs[i], h[i], scalars)
+
+    """
+
+    def input_layer(self, inputs: Mapping[str, tf.Tensor]) -> tf.Tensor:
```
nbren12, line 57: key logic here...the most of the rest of the diff is refactors.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -20,52 +20,61 @@
...
-    ) -> tf.Tensor:
-        if isinstance(inputs, Mapping):
-            return self._call_with_tensors([inputs[key] for key in sorted(inputs)])
-        else:
-            return self._call_with_tensors(inputs)
+class RNNLayer(tf.keras.layers.Layer):
+    """Base class for RNN architectures
 
-    def get_config(self):
+    Scalar inputs (with singleton final dimensions) are passed to every element
```
mcgibbon, line 58: Could you put this under a `Note:`? I was confused for a while when I tried to read this as a description of the layer.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -20,52 +20,61 @@
...
-        return tf.concat(inputs, axis=self._combine_axis)
 
-    def call(
-        self, inputs: Union[Sequence[tf.Tensor], Mapping[str, tf.Tensor]]
-    ) -> tf.Tensor:
-        if isinstance(inputs, Mapping):
-            return self._call_with_tensors([inputs[key] for key in sorted(inputs)])
-        else:
-            return self._call_with_tensors(inputs)
+class RNNLayer(tf.keras.layers.Layer):
```
mcgibbon, line 55: nit: You could replace this class with a function and remove inheritance, which would mean a reader only needs to read one class to know how the layer is defined. e.g. if I read `HybridRNN(RNNLayer)` I'd probably assume the parent is `tf.keras.layers.RNN`, or at least that it inherits from it.

external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -429,11 +442,11 @@ def __init__(
...
 
         self.arch = _get_arch_layer(key, net_kwargs)
-        self.input_combiner = _get_combine_layer(key)
-        self.outputs = _get_output_layer(key, feature_lengths)
+        self.input_layer = self.arch.input_layer
+        self.outputs = self.arch.get_output_layer(feature_lengths)
 
     def call(self, tensors: Mapping[str, tf.Tensor]) -> Mapping[str, tf.Tensor]:
-        combined = self.input_combiner(tensors)
+        combined = self.input_layer(tensors)
```
mcgibbon, line 449: Not in this PR, but thought for later: the _HiddenArchitecture class could be deleted if you add a method to the layers in this file which calls their `input_layer`, current `__call__`, and `outputs` methods. This would improve useability a bit (you can inspect the return type of `ArcitectureConfig.build`), and give a better sense when reading the layer classes how they're actually used (by calling pre, compute, and post steps).

The documentation e.g. for RNNOutput describes that the class combines, RNN's, then connected outputs, so it confused me when the method doing this doesn't exist on that class. Particularly, I was trying to see how the classes use their input_layer and outputs methods. Eventually it made sense when I read more of the file.

mcgibbon (approved):
Good small improvements as you fixed this bug. Please merge the suggested changes and update the one docstring with a Note header, the other comments are suggestions or food for thought.

external/fv3fit/fv3fit/emulation/layers/architecture.py
nbren12: I ended up simplifying this class by pulling the constructor logic up the call stack. I think that `_HiddenArchitecture` is useful for composing these layers in flexible combinations. Ultimately it is pretty similar to `tf.keras.Sequential`, but it allows dictionary inputs and outputs.

external/fv3fit/fv3fit/emulation/layers/architecture.py
nbren12: Done...I named it `combine_sequence_inputs`.

external/fv3fit/fv3fit/emulation/layers/architecture.py
mcgibbon: The class would be useful for flexible combinations, if that was a need, but I see below that for a given arch, you always use the same input and output layer classes. If an RNNOutput is kind of needed for an RNNBlock arch, it can be helpful to see that codified in the RNNBlock class.

This is a fairly small difference between two good solutions.

mcgibbon (approved):

PR #1576, opened 2022-01-07
vcm: add Array type and refactor vcm.calc.thermo

oliverwm1:
The `vcm.calc.thermo` module is very large. This PR moves some functions into a new `vcm.calc.vertical_coordinate` module. It also adds a new `Array` type and improves the type hints of the functions in `vcm.calc.thermo`.

Added public API:
- new `Array` type in vcm. Intended to type-hint functions which can operate on floats, numpy arrays, xarray DataArrays or tf tensors.

Significant internal changes:
- move some functions to `vcm.calc.vertical_coordinate`


external/vcm/vcm/types.py
```diff
@@ -0,0 +1,5 @@
+from typing import TypeVar
+import numpy as np
+import xarray as xr
+
+Array = TypeVar("Array", np.ndarray, xr.DataArray, float)
```
oliverwm1, line 5: If I let `Array` be anything i.e. `Array = TypeVar("Array")` then I get mypy errors when doing arithmetic such as `arr: Array * x: float`.

However if we don't include `tf.Tensor` here (since trying to avoid VCM dep on tf) I imagine this will cause issues for fv3fit type checking...

oliverwm1: Okay based on discussion with @nbren12 I'm deleting this type.

PR #1578, opened 2022-01-10
Fix bug in prescriber

AnnaKwa:
The last commit to master (3180d4) contained a bug that would not allow for prescribed datasets to have only linearly or nearest neighbor interpolation. This fixes that bug.


workflows/prognostic_c48_run/runtime/names.py
```diff
@@ -35,6 +35,7 @@
     "override_for_time_adjusted_total_sky_net_shortwave_flux_at_surface",
     "override_for_time_adjusted_total_sky_downward_longwave_flux_at_surface",
     "ocean_surface_temperature",
+    "surface_temperature",
```
AnnaKwa, line 38: Unrelated to this PR but I noticed that surface temp was not included in the prephysics updated quantities (should be as it is basically the same update as ocean temp)

brianhenn (approved):
Approved thanks @AnnaKwa

PR #1582, opened 2022-01-14
Add DZ to RESTART_VARIABLES in fine-res budget

oliverwm1:
This will ensure that the vertical thickness is included in the coarsened fine-res budget. This variable is necessary for computing density and relative humidity.


nbren12 (approved):
Hmm. do we need to retrigger the workflow again? This will tirgger the cold line costs again....really seems like we should get the C384 data out of there for now.

PR #1584, opened 2022-01-19
emulation:  yaml configuration and some cleanups

nbren12:
The microphysics emulation uses environment variables for configuration, which is unusual compared to our typical yaml-based workflow.

* use yaml instead of env vars

I bundled this with some quality of life improvements to emulation package, by improving the directory structure and deleting a bunch of code.

Suggest reviewing as separate commits.


external/emulation/emulation/hooks/config.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@dataclasses.dataclass
+class EmulationConfig:
+    model: Optional[ModelConfig] = None
+    storage: Optional[monitor.StorageConfig] = None
+
+    def build_model_hook(self):
+        if self.model is None:
+            return error_on_call(
```
nbren12: This behavior is modified in the next commit. This breaks the `NoModel` behavior.

nbren12:
the yaml based config has been merged in another pr.

PR #1585, opened 2022-01-19
emulation: configure with yaml

nbren12:
The emulation prognostic run was configured with env vars, which was hard to extend and required idiosyncratic model execution code.


external/emulation/emulation/_monitor/monitor.py
```diff
@@ -26,18 +27,29 @@
...
-    if value.lower() in affirmatives:
-        return True
-    elif value.lower() in negatives:
-        return False
-    else:
-        raise ValueError(
-            f"Unrecognized value encountered in boolean conversion: {value}"
-        )
+    var_meta_path: str = ""
+    output_freq_sec: int = 1
```
AnnaKwa, line 49: 1 s seems pretty high frequency for a default, maybe keep the old default which saved every 10800 s.

projects/microphysics/scripts/prognostic_run.py
```diff
@@ -83,15 +67,16 @@ def get_env(args):
 
 config = dacite.from_dict(HighLevelConfig, config)
 config.namelist["gfs_physics_nml"]["emulate_zc_microphysics"] = args.online
-config = config.to_fv3config()
+if args.model:
```
AnnaKwa, line 70: Should this be allowed to overwrite an existing model if it exists in the config? If yes, can you at least warn? If no can you raise an error?

projects/microphysics/scripts/prognostic_run.py
```diff
@@ -17,26 +18,13 @@
...
-
-
 CONFIG_PATH = Path(__file__).parent.parent / "configs" / "default.yaml"
 
 parser = argparse.ArgumentParser()
 parser.add_argument(
     "--model",
     type=str,
-    default="NO_MODEL",
+    default="",
```
AnnaKwa, line 27: Is the model kept as command line arg (even though it can be provided in the config yaml) so that automated training/prognostic runs can be chained together?

AnnaKwa (approved):
This looks like a nice streamlining of the prognostic emulation run. LGTM aside from a request for a lower default save frequency! Also, I had a couple questions about the option to allow either command line or config specification of the model; I'm guessing it's supposed to allow the command line arg to be used in the same manner as the other training/prognostic pipeline but wanted to clarify.

projects/microphysics/scripts/prognostic_run.py
nbren12: This is intended, I had debated removing the flag altogether earlier. It would be useful to pipeline training/evaluation, but I'll just remove it to keep it simple.

projects/microphysics/scripts/prognostic_run.py
nbren12: in principle, but we don't actually have these pipeline right now. In practice, it's the most commonly changed argument so it's nice to have a flag for it.

PR #1589, opened 2022-01-20
Store intermediate variables in `PureMLStepper.__call__` as diagnostics

spencerkclark:
The transmissivity approach for predicting the downward and net shortwave radiative fluxes depends on first predicting a transmissivity that is neither a state update or a tendency.  Before this change, the predicted transmissivity would be incorrectly classified as a tendency in the prognostic run, leading to an error.  Now any intermediate variables are simply stored as potential diagnostics.

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
+++ b/workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
@@ -191,6 +191,8 @@ class PureMLStepper:
                 state_updates[key] = value
             elif is_tendency_variable(key):
                 tendency[key] = value
+            else:
+                diagnostics[key] = value
 
         for name in state_updates.keys():
             diagnostics[name] = state_updates[name]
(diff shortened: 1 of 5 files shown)


workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -189,8 +189,10 @@ def __call__(self, time, state):
         for key, value in prediction.items():
             if is_state_update_variable(key, state):
                 state_updates[key] = value
-            else:
+            elif is_tendency_variable(key):
```
AnnaKwa, line 192: It's good to have this check for whether the tendency corresponds to. With the new `else` statement that catches everything that isn't applied as an update, there's a new way for a ML model to silently fail by inadvertently using a real update as an intermediate variable if the output names are different or if the renaming adapter is misconfigured. Could you add a logging statement to record if model predictions are put in the diagnostics? I think the tendency and state updates are logged but I think it'll be more direct to diagnose a possible issue if the diagnostic predictions are also logged.

AnnaKwa (approved):
LGTM! Thanks. I had a request for some logging if intermediate variables are predicted but not applied.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
spencerkclark: Good idea, yeah, that could be awkward to debug otherwise.  I added logging messages within the time loop near where the other updates are logged -- let me know if those look good.

AnnaKwa: Thanks, they LGTM

PR #1590, opened 2022-01-21
offline diags: update Jacobian calculation and plot

AnnaKwa:
- Adds additional jacobian computation function `standardize_jacobians_by_vertical_level` that uses the std over each vertical level to normalize instead of a single scalar std taken over all vertical levels. Has an option to leave the result in units of outputs (only normalize by input std) or dimensionless (normalize by both input and output std)
- Adds a `get_dict_compatible_model` method to `PureKerasModel` to make the underlying keras model work for dict inputs and outputs (is compatible with the Jacobian functions). I didn't make a `PureKerasModel.dict_compatible_model` attribute because there were tests with dummy keras models that didn't have `model.inputs` etc which I didn't want to break.
- Some refactoring of the model sensitivity part of the offline diags to move more code out of `compute.py` and into `_input_sensitivity`

ex. https://storage.googleapis.com/vcm-ml-public/annak/2022-01-20-test-offline/model_sensitivity.html

(Delete unused sections)
Added public API:
- `fv3fit.compute_standardized_jacobians, fv3fit.compute_vertically_standardized_jacobians`
Refactored public API:
- Moves `jacobian.py` out of `fv3fit.emulation` into `fv3fit.shared`


external/fv3fit/tests/test_jacobian.py
```diff
@@ -0,0 +1,72 @@
...
+    normalized = standardize_jacobians(jacobians, sample)
+
+    assert set(normalized) == {"field"}
+    assert set(normalized["field"]) == {"a", "b"}
+
+    np.testing.assert_array_almost_equal(normalized["field"]["a"], np.eye(n))
+    np.testing.assert_array_almost_equal(normalized["field"]["b"], np.zeros((n, n)))
+
+
+@pytest.mark.parametrize(
```
AnnaKwa, line 40: This is the only new test, everything above was relocated out of `tests.emulation.test_jacobian.py`

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -0,0 +1,215 @@
...
+from typing import Tuple, Hashable, Iterable, Dict
+import xarray as xr
+
+import fv3fit
+from fv3fit.sklearn._random_forest import SklearnWrapper
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
```
AnnaKwa, line 14: The majority of this diff is the same as was moved out of `offline._helpers`. The only new parts are `plot_input_sensitivity` and `dataset_to_dict_input`.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -164,3 +165,23 @@ def dump(self, path: str) -> None:
                         }
                     )
                 )
+
+    def _renamed_dict_compatible_model(self) -> tf.keras.Model:
```
nbren12: This looks quite similar to some of the code here: https://github.com/ai2cm/fv3net/blob/0d539792b7ee9804e47b0cdb684e1e03682fe5d8/external/fv3fit/fv3fit/keras/adapters.py#L21

Can you re-use some of that code? The tests for `self.dict_compatible_model` are also quite long and similar to the tests here: https://github.com/ai2cm/fv3net/blob/0d539792b7ee9804e47b0cdb684e1e03682fe5d8/external/fv3fit/tests/keras/test_adapters.py#L48

If you added an `adapters.rename_inputs` function  to go along with `rename_dict_output`, then you could implement this function like this:
```
rename_dict_outputs(outputs(rename_inputs(model, self.input_variables), self.output_variables))
```

At that point, there is probably no need for any special tests...mypy is probably enough to ensure that `self.dict_compatible_model` is a keras model.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -164,3 +165,23 @@ def dump(self, path: str) -> None:
...
+        """
+        Rename keras model inputs/outputs to match input_variables and
+        output_variables, and ensure model accepts and outputs dicts.
+        """
+        renamed_inputs = [
+            # skip the first placeholder dim when specifying shape
```
nbren12, line 174: Is this necessary?

external/fv3fit/tests/keras/test_pure_keras.py
```diff
@@ -0,0 +1,45 @@
+from fv3fit.keras._models.shared.pure_keras import PureKerasModel
```
nbren12, line 1: would be nice to delete these tests...since they are pretty involved and similar to https://github.com/ai2cm/fv3net/blob/master/external/fv3fit/tests/keras/test_adapters.py#L48.

external/fv3fit/fv3fit/_shared/jacobian.py
```diff
@@ -2,7 +2,7 @@
 import tensorflow as tf
 from typing import Callable, Mapping, Dict
 
-from .layers.normalization import standard_deviation_all_features
+from ..emulation.layers.normalization import standard_deviation_all_features
```
nbren12, line 5: I'd advocate for `fv3fit.keras.jacobian` rather than the `fv3fit.keras._shared.jacobian` namespace. Would be nicer to use the former in a jupyter notebook for example (or in the offline diags).

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -0,0 +1,215 @@
...
+
+def plot_input_sensitivity(
+    model: fv3fit.Predictor, sample: xr.Dataset, output_path: str
+):
+    base_model = model.base_model if isinstance(model, fv3fit.DerivedModel) else model
+
+    try:
+        jacobians = fv3fit.compute_standardized_jacobians(
+            base_model.dict_compatible_model,
```
nbren12: I realize that this nested try-except logic is not new, but I can see it being hard to tell what attribute is missing given the amount of code it contains. Would be clearer like this:
```
try:
     model = base_model.model
except AttributeError:
     logging.warn("Model is not a keras model")
else:
     jacobians = fv3fit.compute_standardized_jacobians(...)

File: workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py [hunk 8]
```diff
@@ -0,0 +1,215 @@
...
+    )
+    start = 0
+    variable_indices = {}
+    for var, var_dim in variable_dims.items():
+        variable_indices[var] = (start, start + var_dim)
+        start += var_dim
+    return variable_indices
+
+
+def plot_input_sensitivity(
```
nbren12: I realize this function is not new, but it does induce tight coupling between `fv3fit` and these offline diags. Would be great to use polymorphism for this behavior rather than the brittle exception catching logic below. Can you add some interfaces to `fv3fit.Predictor` that help with this? e.g. `fv3fit.Predictor.get_input_sensitivity`?

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -0,0 +1,215 @@
...
+    axs[0, -1].bar(
+        range(len(scalar_features)),
+        scalar_feature_mean_importances,
+        yerr=scalar_feature_std_importances,
+        tick_label=scalar_features,
+    )
+    return axs
+
+
+def _plot_jacobians(jacobians):
```
nbren12, line 186: Hard to tell what `jacobians` is...can you add a type hint?

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -0,0 +1,215 @@
...
+    input_feature_indices: Mapping[str, Tuple[int, int]], wrapped_model: SklearnWrapper,
+):
+
+    # TODO: After the planned removal of the fv3fit.sklearn.EnsembleRegressor
+    # is complete, this for loop is not needed.
+    importances = []
+    for member in wrapped_model.model.regressors:
+        importances.append(member.feature_importances_)
+    mean_importances = np.array(importances).mean(axis=0)
+    std_importances = np.array(importances).std(axis=0)
```
nbren12, line 121: Would be nice to decouple this code from the `SklearnWrapper`...for instance it should avoid looking at 
 `SklearnWrapper.model.regressors`. Can you make `mean_importances` and `std_importances` properties of `SklearnWrapper`?

external/fv3fit/fv3fit/_shared/jacobian.py
```diff
@@ -73,3 +72,56 @@ def compute_standardized_jacobians(model, data, input_variables):
...
+    Generate sensitivity jacobians for each output of a model and
+    standardize (dimensionless or output units) for easy
+    inter-variable comparison.
+
+    The scaling uses the standard deviation of each vertical level
+    for the input (std_input) and output
+    (std_output) sample. The associated
+    jacobian result is multiplied by [ std_input / std_output ].
+    If units are "output" the associated jacobian result is multiplied
+    by [ std_input ].
```
nbren12, line 92: I worry the code in this module is being too tightly coupled to the needs of downstream plots. I think the APIs should be refactored.

As with `standardize_jacobians`, this function does two things...computes the std and scales the jacobian. I suggest a refactor below.

external/fv3fit/fv3fit/_shared/jacobian.py
```diff
@@ -73,3 +72,56 @@ def compute_standardized_jacobians(model, data, input_variables):
...
+    """
+
+    # normalize factors so sensitivities are comparable but still
+    # preserve level-relative magnitudes
+    std_factors = {name: data.std(axis=0) for name, data in sample.items()}
+    standardized_jacobians: Dict[str, OutputSensitivity] = {}
+    for out_name, per_input_jacobians in all_jacobians.items():
```
nbren12, line 102: I think this for loop would also work for `standardize_jacobian`. Can you refactor it into two shared helpers and remove the `units` flag.
-  `units='output'`. `jacobian_to_output_change(all_jacobians, delta_input)`
- `units='dimensionless'`. `nondimensionalize_jacobian(all_jacobian, scales: Mapping[str, np.ndarray])`

At this point the standard deviation code is a one liner...I'd delete `standardize_jacobians_by_vertical_level` and `standardize_jacobians` and replace any of their uses with
```
# andre's scaling
std_factors = {name: data.std() for name, data in sample.items()}
# your scaling
# std_factors = {name: data.std(axis=0) for name, data in sample.items()}
nondimensionalize_jacobian(all_jacobian, std_factors)
```
this will easily allow downstream visualizations to change the scale factors without adding another function to this module.

nbren12 (commented):
Thanks Anna. I think the jacobian plots are nice and legible now. Most of your refactors looks pretty good, but I think we need to keep an eye on the tight coupling between the offline ml diags and fv3fit. I've proposed some refactors along those lines below.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -164,3 +164,23 @@ def dump(self, path: str) -> None:
                         }
                     )
                 )
+
+    def get_dict_compatible_model(self) -> tf.keras.Model:
```
nbren12, line 168: This looks quite similar to some of the code here: https://github.com/ai2cm/fv3net/blob/0d539792b7ee9804e47b0cdb684e1e03682fe5d8/external/fv3fit/fv3fit/keras/adapters.py#L21

Can you re-use some of that code? The tests for `self.dict_compatible_model` are also quite long and similar to the tests here: https://github.com/ai2cm/fv3net/blob/0d539792b7ee9804e47b0cdb684e1e03682fe5d8/external/fv3fit/tests/keras/test_adapters.py#L48

If you added an `adapters.rename_inputs` function  to go along with `rename_dict_output`, then you could implement this function with a one-liner  like this:
```
rename_dict_outputs(rename_inputs(model, self.input_variables), self.output_variables))
```

At that point, there is probably no need for any special tests...mypy is probably enough to ensure that `self.dict_compatible_model` is a keras model.

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
nbren12: See https://en.wikipedia.org/wiki/Law_of_Demeter

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
nbren12: If you function works better, I'd be amenable to replacing `rename_dict_outputs` with it. So there could be a function like 
`rename_inputs_and_outputs(model, inputs, outputs)`. I just want to avoid the duplication (especially in the test suite).

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
AnnaKwa: Yes, the Input layer doesn't know to ignore the None dim so it adds a second placeholder dimension if you use the full shape of the original input.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
AnnaKwa: I added a `rename_dict_inputs` function, instead of moving this function into `adapters`. The function here makes an assumption that the input model has list inputs, and doesn't work for models that have already been passed to `ensure_dict_outputs`. The new function should work for either.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
nbren12: I actually wrote a func for this: https://github.com/ai2cm/fv3net/blob/0d539792b7ee9804e47b0cdb684e1e03682fe5d8/external/fv3fit/fv3fit/keras/adapters.py#L71

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
AnnaKwa: As discussed I'll punt on this comment for this PR but am tracking it in an issue https://github.com/ai2cm/fv3net/issues/1596 since the changes would probably be a large diff.

external/fv3fit/fv3fit/keras/adapters.py
```diff
@@ -108,3 +108,59 @@ def _rename_graph_outputs_to_match_output_keys(model: tf.keras.Model) -> tf.kera
...
+
+    Args:
+        model: a tensorflow model. must return dicts
+        transation: a mapping from old names (the keys returned by ``model``) to
+            the new output names.
+
+    """
+
+    # need to have list format inputs to connect renamed layers to rest of the model
+    list_input_model = _ensure_list_input(model)
```
AnnaKwa, line 150: This is kind of wonky but was the only way I could figure out to connect the renamed input layers to the rest of the model if it had dict inputs and expected the old keys as input layer names.

external/fv3fit/fv3fit/keras/adapters.py
```diff
@@ -108,3 +108,48 @@ def _rename_graph_outputs_to_match_output_keys(model: tf.keras.Model) -> tf.kera
...
+
+    renamed_inputs = [
+        # skip the first placeholder dim when specifying shape
+        tf.keras.layers.Input(
+            shape=input_layer.shape[1:],
+            name=translation.get(input_layer.name, input_layer.name),
+        )
+        for input_layer in list_input_model.inputs
+    ]
+    connect_outputs = model(renamed_inputs)
```
nbren12: Should you use `list_input_model` here instead?

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -170,17 +175,16 @@ def get_dict_compatible_model(self) -> tf.keras.Model:
...
+        renamed_inputs: Mapping[str, str] = {
+            str(layer.name): str(input_var)
+            for layer, input_var in zip(self.model.inputs, self.input_variables)
         }
-
-        dict_inputs = {input.name: input for input in renamed_inputs}
-        return tf.keras.Model(inputs=dict_inputs, outputs=renamed_outputs)
+        renamed_outputs: Mapping[str, str] = {
+            str(layer.name): str(output_var)
+            for layer, output_var in zip(self.model.outputs, self.output_variables)
```
nbren12: so this is the bug you were running into?

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -327,3 +327,18 @@ def load(cls, path: str) -> "SklearnWrapper":
         obj.output_features_ = output_features_
 
         return obj
+
+    @property
+    def feature_importances(self):
```
nbren12, line 332: Might be nice to add some type hints.

external/fv3fit/fv3fit/keras/adapters.py
```diff
@@ -111,7 +111,14 @@ def _rename_graph_outputs_to_match_output_keys(model: tf.keras.Model) -> tf.kera
 
 
 def _ensure_list_input(model: tf.keras.Model) -> tf.keras.Model:
-    # useful for connecting input layers with names different from the original model
+    """
+    Takes dict input + output model and converts it to take in
+    list inputs but keep original dict outputs. Useful for connecting
+    input layers with names different from the original model.
+    
+    Should NOT be used outside of helping rename_dict_input
```
nbren12, line 119: I would consider removing this cautionary statement...it does seem like a useful functionality.

nbren12 (approved):
Thanks! Looks great to me now. I think we've learned some useful things about tensorflow. the list  input formulation seems useful for when you want to decouple tensors from the names baked into a keras model's graph. Really nice solution IMO.

Just a couple of nits below. The merge conflicts look simple to resolve.

PR #1591, opened 2022-01-25
cleanup vcm sphinx docs

nbren12:
* resolve warnings/errors
* replace all incorrectg usage of automethod with autofunction
* add custom css to display one parameter per line
* remove redundant type information from some docstrings


spencerkclark (approved):
Looks good to me.  Thanks!

PR #1592, opened 2022-01-25
Limit prescribed extremes

brianhenn:

(Delete unused sections)
Added public API:

Refactored public API:

Significant internal changes:

Requirement changes:

- [ ] Tests added

### DIFF
--- a/external/vcm/vcm/limit.py
+++ b/external/vcm/vcm/limit.py
@@ -63,7 +63,7 @@ class LimitedDataset(xr.Dataset):
     def _fit_limits(self, ds: xr.Dataset):
         sample_ds = (
             ds.isel(**self._fit_indexers) if self._fit_indexers is not None else ds
-        )
+        ).load()
         sample_dims = (
             set(sample_ds.dims) - set(self._feature_dims)
             if self._feature_dims is not None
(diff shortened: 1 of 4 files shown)


external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,90 @@
...
+            
+    Returns: a limited data array
+    
+    """
+    upper = limits.sel(bounds="upper")
+    lower = limits.sel(bounds="lower")
+    return data.where(data < upper, upper).where(data > lower, lower)
+
+
+class LimitedDataset(xr.Dataset):
```
nbren12, line 22: Suggest turning this into a function returning an xarray object. the inheritance allows you to lazily apply the limiting, but we can do that with dask.

In general it’s not easy to subclass data structure like an xarray dataset. This goes back to the datastructure vs object stuff in the clean code book. Testing that this class is a valid dataset is nigh impossible…note how xarray itself doesn’t even have subclasses of dataset. There are simply too many methods/attributes to keep track of.

I also suggest  isolating this logic to the prognostic run? Or are you considering using it in fv3fit too for training?

nbren12 (commented):
Just poking around😉

brianhenn:
Closed as this functionality was merged under #1618

PR #1593, opened 2022-01-25
v0.7 base config:  namelist flags set to use climatology 

AnnaKwa:
This updates the fv3kube base config to set namelist flags for max snow albedo, vegetation, sea ice, and land sea mask to follow the climatology in the prognostic run, as opposed to the previous behavior of persisting the same values from the initial condition throughout the run.

It also bumps the e2e and prepare-config nudge-to-fine and prognostic tests in fv3net to use the latest base config, and deletes an obsolete reference in fv3kube to a previous default version of the base config. All other references to base configs in fv3net are left at their existing versions.


external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
```diff
@@ -0,0 +1,334 @@
...
+    interp_method: conserve_great_circle
+  nam_stochy:
+    lat_s: 96
+    lon_s: 192
+    ntrunc: 94
+  namsfc:
+    fabsl: 0
+    fabss: 0
+    faisl: 99999
+    faiss: 99999
```
spencerkclark, line 295: In reviewing things a bit more, I think these relate to the land / sea / sea-ice mask, so we should set them to zero too.  I want to double check though by doing a short test simulation.
```suggestion
    faisl: 0  # Use land / sea / sea-ice mask from forcing files instead of initial conditions
    faiss: 0  # Use land / sea / sea-ice mask from forcing files instead of initial conditions
```

spencerkclark (commented):
Thanks for putting this together -- looks good.  I suggested adding a few comments next to these namelist parameters for clarity, but I want to double check the settings for a few more before merging this.

external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
```diff
@@ -0,0 +1,334 @@
...
+    fnvmnc: grb/global_shdmin.0.144x0.144.grb
+    fnvmxc: grb/global_shdmax.0.144x0.144.grb
+    fnzorc: igbp
+    fsicl: 0
+    fsics: 0
+    fslpl: 99999
+    fsmcl:
+    - 99999
+    - 99999
+    - 99999
```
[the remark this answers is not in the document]
brianhenn: Just curious why this is set as 3 values?

external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
brianhenn: I committed this but LMK if you changed your mind here

external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
```diff
@@ -0,0 +1,334 @@
...
+    const_vg: false
+    const_vi: false
+    const_vr: false
+    const_vs: false
+    de_ice: false
+    do_qa: true
+    do_sedi_heat: false
+    dw_land: 0.16
+    dw_ocean: 0.1
+    fast_sat_adj: true
```
brianhenn, line 213: Didn't realize these were already updated, thanks @AnnaKwa

external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
spencerkclark: Thanks, yeah, I still think this change is good.

spencerkclark: (deleted my previous comment)

Actually I do want to double check how the sea / land / sea-ice mask is generated online.  In looking at the forcing files it doesn't appear that there is a climatology for that.  I see a static ocean/land mask (with no sea-ice) in the form of `seaice_newland.grb` (despite the name of the file).  I'll probably need to look carefully at the fortran code.

spencerkclark: The sea ice fraction climatology clearly comes from `CFSR.SEAICE.1982.2012.monthly.clim.grb`.

brianhenn: So setting these to 0 is OK then?

AnnaKwa:
Actually I just realized this probably needs a line added for v0.7 in fv3kube.config: https://github.com/ai2cm/fv3net/blob/86a672c6b680e01a5cb2b4a2dd4ba2b5c83453d9/external/fv3kube/fv3kube/config.py#L19

external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
brianhenn: See [notebook](https://github.com/ai2cm/explore/blob/master/brianh/2022-05-24-land-sea-mask-namelist/2022-05-24-land-sea-mask-debugging.ipynb) showing that if these parameter are set to 0, the land sea mask is dynamically calculated from the sea ice fraction, and that if they are set to 99999, it is persisted. Not sure what happens if `faiss` and `faisl` are set to one and `fsicl` and `fsics` are set to another, probably something to avoid. I think this confirms that we want all of these to be set to 0 in the PIRE and nudge-to-25km contexts, and 99999 in the nudge to 40-day SHiELD context.

spencerkclark: Thanks for doing those runs and confirming that @brianhenn!  I'm happy with this now if you'd to merge.

spencerkclark (approved):

PR #1598, opened 2022-01-27
fv3fit.emulation: add temperature scaling

nbren12:
The microphysics emulator performs poorly in cold places. Performing ML in a temperature-scaled space should help.

Added public API:
- `fv3fit.emulation.transforms.ConditionallyScaled` configuration class

Some interpolation and groupby-apply-combine functions utilties:
- `fv3fit.keras.math.piecewise`
- `fv3fit.keras.math.groupby_bins`

### DIFF
--- a/external/fv3fit/fv3fit/emulation/transforms/factories.py
+++ b/external/fv3fit/fv3fit/emulation/transforms/factories.py
@@ -1,14 +1,18 @@
 import dataclasses
-from typing import Sequence, Set
+from typing import Callable, Sequence, Set
 
-from typing_extensions import Protocol
-from fv3fit.emulation.types import TensorDict
+import tensorflow as tf
 from fv3fit.emulation.transforms.transforms import (
-    TensorTransform,
-    LogTransform,
-    UnivariateTransform,
     ComposedTransform,
+    ConditionallyScaledTransform,
+    LogTransform,
+    TensorTransform,
+    UnivariateTransform,
 )
+from fv3fit.emulation.types import TensorDict
+from fv3fit.emulation.zhao_carr_fields import Field
+from fv3fit.keras.math import groupby_bins, piecewise
+from typing_extensions import Protocol
 
 
 class TransformFactory(Protocol):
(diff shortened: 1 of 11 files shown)


nbren12:
Running an experiment with this code here: https://wandb.ai/ai2cm/microphysics-emulation/runs/1ctz7e3w/overview?workspace=user-nbrenowitz

external/fv3fit/fv3fit/emulation/transforms/factories.py
```diff
@@ -45,6 +49,83 @@ def build(self, sample: TensorDict) -> TensorTransform:
...
+    values = groupby_bins(edges, x, y, reduction)
+
+    def interp(x: tf.Tensor) -> tf.Tensor:
+        return piecewise(edges[:-1], values, x)
+
+    return interp
+
+
+@dataclasses.dataclass
+class ConditionallyScaled(TransformFactory):
```
AnnaKwa, line 72: Can you also indicate in the class name that this is for a residual?

external/fv3fit/fv3fit/emulation/transforms/transforms.py
```diff
@@ -53,6 +53,45 @@ def backward(self, y: TensorDict) -> TensorDict:
         return out
 
 
+class ConditionallyScaledTransform(TensorTransform):
```
AnnaKwa, line 56: Similar to the new factory class, can you make it clear in the name that this is for residuals?

external/fv3fit/fv3fit/keras/math.py
```diff
@@ -0,0 +1,42 @@
...
+        xg: points to interpolate at
+
+    Return:
+        yg: interpolated values. same shape as xg. Uses constant extrapolation.
+            Within the min/max of x it has the following form::
+
+                f(xg) = y[i] if x[i] <= xg < x[i+1]
+    """
+    xg_flat = tf.reshape(xg, [-1])
+    index = tf.searchsorted(x, xg_flat, side="right") - 1
```
AnnaKwa, line 20: Will this raise an error if `x` is not actually sorted, or just silently return the wrong output?

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -89,7 +90,9 @@ class TrainConfig:
     test_url: str
     out_url: str
     transform: TransformConfig = field(default_factory=TransformConfig)
-    tensor_transform: List[TransformedVariableConfig] = field(default_factory=list)
+    tensor_transform: List[
+        Union[TransformedVariableConfig, ConditionallyScaled]
```
AnnaKwa, line 94: Maybe `Union[TransformedVariableConfig, ConditionallyScaled]` can be defined outside of this class as `Transforms` or something.

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -234,7 +237,13 @@ def open_dataset(
         self, url: str, nfiles: Optional[int], required_variables: Set[str],
     ) -> tf.data.Dataset:
         nc_open_fn = self.transform.get_pipeline(required_variables)
-        return nc_dir_to_tf_dataset(url, nc_open_fn, nfiles=nfiles)
+        return nc_dir_to_tf_dataset(
+            url,
+            nc_open_fn,
+            nfiles=nfiles,
+            shuffle=True,
```
AnnaKwa, line 244: Why shuffle?

external/fv3fit/fv3fit/emulation/transforms/factories.py
```diff
@@ -45,6 +49,83 @@ def build(self, sample: TensorDict) -> TensorTransform:
...
+        min_scale: the minimium scale to normalize by. Used when the scale might
+            be 0.
+
+    """
+
+    to: str
+    condition_on: str
+    bins: int
+    field: Field
+    min_scale: float = 0.0
```
AnnaKwa, line 97: Would it be easier to just make the default some epsilon in the case of the scale being zero?

AnnaKwa (approved):
LGTM! Just some minor comments/questions.

external/fv3fit/fv3fit/keras/math.py
nbren12: It may silently fail, but I decided to let tensorflow make it's own complaints if it wants to. I did put an assertion in the `groupby` to ensure we don't do that in practice.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: If they start getting used elsewhere sure, but this Union is only used for dacite really. I used the `TransformFactory` protocol  for type-checking most places.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: We weren't shuffling before, so only Jan data was used to fit all the normalization. It's possible this could have an impact since it is southern summer, and the coldest temperatures are not present.

external/fv3fit/fv3fit/emulation/transforms/factories.py
nbren12: It would be nice if the default worked for all variables---including ones with smaller non-zero magnitude than `min_scale`. I think the most sane default is to do nothing at all.

external/fv3fit/fv3fit/emulation/transforms/transforms.py
```diff
@@ -53,6 +53,45 @@ def backward(self, y: TensorDict) -> TensorDict:
...
+        self,
+        to: str,
+        input_name: str,
+        output_name: str,
+        on: str,
+        scale: Callable[[tf.Tensor], tf.Tensor],
+        center: Callable[[tf.Tensor], tf.Tensor],
+        min_scale: float = 0.0,
+    ) -> None:
+        self.to = to
```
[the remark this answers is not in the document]
nbren12: Whoops. I failed to push the commit before merging...maybe we can just let this one slide? (or I could open another PR). Let me know.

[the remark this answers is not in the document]
AnnaKwa: That's fine, I think it's usefulness is mostly for reviewers looking at this class. If you make a future PR that touches this class you can just add it then.

PR #1599, opened 2022-01-27
Improve performance of `vcm.cos_zenith_angle`

spencerkclark:
I have noticed that `vcm.cos_zenith_angle` is a bottleneck when making offline predictions in notebooks (and presumably is a bottleneck in other settings).  It takes nearly a second to compute it for a single time across the C48 grid.  This PR makes some minor changes to improve its performance by a factor of over 800.  All vcm tests pass and plots of it look good in a [notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2022-01-26-cosz-performance/2022-01-26-validate-new-cosz.ipynb).  Obviously due to floating point arithmetic, values change slightly, but I think the performance improvement here is something to take seriously.  I'm happy to add more tests if we feel it is necessary.

Some timing results:
```
In [1]: import xarray as xr; import vcm; import vcm.catalog

In [2]: grid = vcm.catalog.catalog["grid/c48"].to_dask().load()

In [3]: times = xr.cftime_range("2020", periods=10, calendar="julian")

In [4]: times = xr.DataArray(times, dims=["time"])
```

With the existing version of `vcm.cos_zenith_angle`:
```
In [5]: %timeit vcm.cos_zenith_angle(times, grid.lon, grid.lat)
8.73 s ± 69.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

With the version of `vcm.cos_zenith_angle` in this PR:
```
In [5]: %timeit vcm.cos_zenith_angle(times, grid.lon, grid.lat)
10 ms ± 186 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### DIFF
--- a/external/fv3fit/fv3fit/_shared/stacking.py
+++ b/external/fv3fit/fv3fit/_shared/stacking.py
@@ -113,6 +113,7 @@ def shuffled(random: RandomState, datasets: Sequence[xr.Dataset]) -> xr.Dataset:
             dimensionality/coordinates if multiple datasets are given
     """
     chunks_default = (len(datasets[0][SAMPLE_DIM_NAME]),)
+    datasets = [dataset.unify_chunks() for dataset in datasets]
     chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
     chunk_indices = _get_chunk_indices(chunks)
     shuffled_inds = np.concatenate(
--- a/external/fv3fit/fv3fit/_shared/stacking.py
+++ b/external/fv3fit/fv3fit/_shared/stacking.py
@@ -112,8 +112,8 @@ def shuffled(random: RandomState, datasets: Sequence[xr.Dataset]) -> xr.Dataset:
         datasets: input data to be shuffled, must contain identical
             dimensionality/coordinates if multiple datasets are given
     """
-    chunks_default = (len(datasets[0][SAMPLE_DIM_NAME]),)
     datasets = [dataset.unify_chunks() for dataset in datasets]
+    chunks_default = (len(datasets[0][SAMPLE_DIM_NAME]),)
     chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
     chunk_indices = _get_chunk_indices(chunks)
     shuffled_inds = np.concatenate(
--- a/external/fv3fit/fv3fit/_shared/stacking.py
+++ b/external/fv3fit/fv3fit/_shared/stacking.py
@@ -112,7 +112,6 @@ def shuffled(random: RandomState, datasets: Sequence[xr.Dataset]) -> xr.Dataset:
         datasets: input data to be shuffled, must contain identical
             dimensionality/coordinates if multiple datasets are given
     """
-    datasets = [dataset.unify_chunks() for dataset in datasets]
     chunks_default = (len(datasets[0][SAMPLE_DIM_NAME]),)
     chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
     chunk_indices = _get_chunk_indices(chunks)
--- a/external/fv3fit/tests/test_stacking.py
+++ b/external/fv3fit/tests/test_stacking.py
@@ -142,9 +142,9 @@ def test__shuffled_dask_different_chunks(a_chunk_size, b_chunk_size):
     dataset = _stacked_dataset(SAMPLE_DIM_NAME)
     dataset["a"] = dataset.a.chunk({SAMPLE_DIM_NAME: a_chunk_size})
     dataset["b"] = dataset.b.chunk({SAMPLE_DIM_NAME: b_chunk_size})
-    result, = shuffled(np.random.RandomState(1), [dataset])
+    (result,) = shuffled(np.random.RandomState(1), [dataset])
     assert result.chunks[SAMPLE_DIM_NAME] == (5, 5)
-    assert result.chunks["x"] == (2, )
+    assert result.chunks["x"] == (2,)
 
 
 def test_multiple_unstacked_dims():
--- a/external/fv3fit/tests/test_stacking.py
+++ b/external/fv3fit/tests/test_stacking.py
@@ -137,6 +137,16 @@ def test__shuffled_dask():
     shuffled(np.random.RandomState(1), [dataset])
 
 
+@pytest.mark.parametrize(("a_chunk_size", "b_chunk_size"), [(10, 5), (5, 10)])
+def test__shuffled_dask_different_chunks(a_chunk_size, b_chunk_size):
+    dataset = _stacked_dataset(SAMPLE_DIM_NAME)
+    dataset["a"] = dataset.a.chunk({SAMPLE_DIM_NAME: a_chunk_size})
+    dataset["b"] = dataset.b.chunk({SAMPLE_DIM_NAME: b_chunk_size})
+    result, = shuffled(np.random.RandomState(1), [dataset])
+    assert result.chunks[SAMPLE_DIM_NAME] == (5, 5)
+    assert result.chunks["x"] == (2, )
+
+
 def test_multiple_unstacked_dims():
     na, nb, nc, nd = 2, 3, 4, 5
     ds = xr.Dataset(
--- a/external/fv3fit/tests/test_stacking.py
+++ b/external/fv3fit/tests/test_stacking.py
@@ -137,16 +137,6 @@ def test__shuffled_dask():
     shuffled(np.random.RandomState(1), [dataset])
 
 
-@pytest.mark.parametrize(("a_chunk_size", "b_chunk_size"), [(10, 5), (5, 10)])
-def test__shuffled_dask_different_chunks(a_chunk_size, b_chunk_size):
-    dataset = _stacked_dataset(SAMPLE_DIM_NAME)
-    dataset["a"] = dataset.a.chunk({SAMPLE_DIM_NAME: a_chunk_size})
-    dataset["b"] = dataset.b.chunk({SAMPLE_DIM_NAME: b_chunk_size})
-    (result,) = shuffled(np.random.RandomState(1), [dataset])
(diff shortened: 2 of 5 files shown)


external/vcm/vcm/calc/_zenith_angle.py
```diff
@@ -122,10 +118,7 @@ def _greenwich_mean_sidereal_time(model_time):
 
     theta_radians = np.deg2rad(theta / 240.0) % (2 * np.pi)
 
-    if theta_radians < 0:
```
spencerkclark, line 125: I actually don't think it is possible for `theta_radians` to be less than zero at this point since it is constructed using the modulo operator (a.k.a. the `np.remainder` function).  According to [the NumPy documentation](https://numpy.org/doc/stable/reference/generated/numpy.remainder.html):

> Computes the remainder complementary to the floor_divide function. It is equivalent to the Python modulus operator ``x1 % x2`` and has the same sign as the divisor x2.

In this case the divisor is always `2 * np.pi`, which is greater than zero, so the result should always be greater than or equal to zero.  I don't think there is much harm in keeping this check in the form I translated it to below, however, if we'd rather not change the algorithm.  I assume this check was originally here for a reason (though maybe it was that this code was ported from another language that had different behavior?).

spencerkclark:
The integration tests fail in a fv3fit step:

<details>
<p>

```
+ python3 -m fv3fit.train training_config.yaml training_data.yaml gs://vcm-ml-scratch/test-end-to-end-integration/2022-01-27/a0cab9d-8234-train-diags-prog/trained_models/sklearn_dQ1_model --validation-data-config validation_data.yaml
2022-01-27 12:46:41.380372: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: :/usr/local/esmf/lib:/FMS/libFMS/.libs/:/usr/local/lib
2022-01-27 12:46:41.380469: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
/usr/local/lib/python3.8/dist-packages/tensorflow_addons/utils/ensure_tf_install.py:54: UserWarning: Tensorflow Addons supports using Python ops for all Tensorflow versions above or equal to 2.2.0 and strictly below 2.4.0 (nightly versions are not supported).
 The versions of TensorFlow you are currently using is 2.5.2 and is not supported.
Some things might work, some things might not.
If you were to encounter a bug, do not file an issue.
If you want to make sure you're using a tested and supported configuration, either change the TensorFlow version or the TensorFlow Addons's version.
You can find the compatibility matrix in TensorFlow Addon's readme:
https://github.com/tensorflow/addons
  warnings.warn(
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/train.py", line 113, in <module>
    main(args, unknown_args)
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/train.py", line 99, in main
    model = train(
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/sklearn/_random_forest.py", line 50, in train_random_forest
    model.fit(train_batches)
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/sklearn/_random_forest.py", line 90, in fit
    return self._model_wrapper.fit(batches)
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/sklearn/_random_forest.py", line 251, in fit
    for i, batch in enumerate(stacked_batches):
  File "/usr/lib/python3.8/_collections_abc.py", line 874, in __iter__
    v = self[i]
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/_shared/stacking.py", line 27, in __getitem__
    return self._stack_batch(self._batches[idx])
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/_shared/stacking.py", line 45, in _stack_batch
    return shuffled(self._random_state, [ds])[0]
  File "/usr/local/lib/python3.8/dist-packages/fv3fit/_shared/stacking.py", line 116, in shuffled
    chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/dataset.py", line 2099, in chunks
    raise ValueError(
ValueError: Object has inconsistent chunks along dimension _fv3fit_sample. This can be fixed by calling unify_chunks().
```

</p>
</details>

This is because `vcm.cos_zenith_angle` was an eager function before and so `cos_zenith_angle` was always an unchunked NumPy array.  With these changes it is now dask-compatible.  This failure could be fixed by either keeping the computation of `cos_zenith_angle` eager or using the suggestion of calling `unify_chunks` on the dataset.

external/fv3fit/fv3fit/_shared/stacking.py
```diff
@@ -112,6 +112,7 @@ def shuffled(random: RandomState, datasets: Sequence[xr.Dataset]) -> xr.Dataset:
         datasets: input data to be shuffled, must contain identical
             dimensionality/coordinates if multiple datasets are given
     """
+    datasets = [dataset.unify_chunks() for dataset in datasets]
```
spencerkclark, line 115: Maybe I should add a logging message here if unifying chunks is necessary?

external/fv3fit/tests/test_stacking.py
```diff
@@ -137,6 +137,16 @@ def test__shuffled_dask():
     shuffled(np.random.RandomState(1), [dataset])
 
 
+@pytest.mark.parametrize(("a_chunk_size", "b_chunk_size"), [(10, 5), (5, 10)])
+def test__shuffled_dask_different_chunks(a_chunk_size, b_chunk_size):
```
mcgibbon, line 141: I can't tell for sure what this is testing, is it that `shuffled` maintains the chunking of a dataset with multiple chunk sizes? Can you make the name more verbose or add a docstring?

external/vcm/vcm/calc/_zenith_angle.py
mcgibbon: Yes, the modulo operator above should be equivalent to these two lines. For the same reason, is `+ 2 * np.pi * (theta_radians < 0)` necessary?

external/fv3fit/fv3fit/_shared/stacking.py
mcgibbon: This only makes sense to me if it's something we don't normally expect to happen. If we'd get this logging message every time we train, we'll just ignore it and add to console noise.

The docstring says we sample within chunks if chunking is present (i.e. without shuffling across chunks). Is this changed, and is there a reason for it to change?

external/vcm/vcm/calc/_zenith_angle.py
spencerkclark: Good, just wanted to make sure that my understanding was correct there.  I'll remove the `+ 2 * np.pi * (theta_radians < 0)` bit since it's a no-op.

external/fv3fit/fv3fit/_shared/stacking.py
spencerkclark: If I logged it, I would probably do it something like this:
```python
chunks_default = (len(datasets[0][SAMPLE_DIM_NAME]),)
try:
    chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
except ValueError:
    datasets = [dataset.unify_chunks() for dataset in datasets]
    chunks = datasets[0].chunks.get(SAMPLE_DIM_NAME, chunks_default)
    logging.debug(f"Dataset variables had inconsistent chunk sizes; unified the chunks to: {chunks}")
```
So it would only show up if unification was required.  If `unify_chunks` changed the chunk size to something you didn't expect I think it would be useful to know.  Imagine if in an extreme example the chunk size along the sample dimension changed to `1` -- in that case no shuffling would occur.

This mainly comes up because `cos_zenith_angle` typically will not be chunked in time, since it is generally called with a time vector that is completely within memory.  `unify_chunks` is required to chunk it more finely along the sample dimension to match the other variables in the dataset.  The other variables should generally be chunked like each other -- if they weren't we would have encountered this error already.

But upon further reflection, I'm going to make this a moot point.  I think this issue should be solved upstream of fv3fit.  If we want `cos_zenith_angle` to be lazy in our training pipeline, we should find a way to compute it with the proper chunk structure from the start.  For now I have just modified the code in the DerivedMapping to always make its computation of `cos_zenith_angle` eager.  It was eager previously, and the performance improvement here is still a positive in that context.

external/vcm/vcm/derived_mapping.py
```diff
@@ -89,7 +89,11 @@ def _recurse_find_deps(vars, deps):
 
 @DerivedMapping.register("cos_zenith_angle", required_inputs=["time", "lon", "lat"])
 def cos_zenith_angle(self):
-    return vcm.cos_zenith_angle(self["time"], self["lon"], self["lat"])
+    # TODO: eagerly loading the lon and lat inputs is required for the integration
+    # tests to pass.  It would be nice to remove this requirement now that
+    # cos_zenith_angle is dask compatible, but that would require figuring out the
+    # appropriate way to chunk the time coordinate.
+    return vcm.cos_zenith_angle(self["time"], self["lon"].load(), self["lat"].load())
```
mcgibbon, line 96: I'd normally complain about a load, but lon and lat should be small enough.

mcgibbon (approved):
Thanks for the speedup!

external/vcm/vcm/derived_mapping.py
spencerkclark: Yeah, I agree, eventually it would be nice to remove this requirement, but it was happening silently before anyway, and at least this time the values will be computed much faster.

PR #1602, opened 2022-01-28
Some fixes for offline diags

AnnaKwa:
- consistent color legends for prediction / target between line plots
-  fix units bug and use scientific notation instead of 2 decimal pt precision

ex. https://storage.googleapis.com/vcm-ml-public/annak/2022-01-28/offline_report/index.html


oliverwm1 (approved):
Thanks for the fixes.

PR #1605, opened 2022-02-04
Flexible restart categories

brianhenn:
The prognostic run initial conditions and nudging handling both assume 'standard' fv3gfs restart category names, but some sets of restarts (e.g., the C384 restarts) aren't named this way. This PR adds optional parameters to both to specify the names on disk for each type of restart file. Both routines then rename the files to their standards. 

Added public API:
- a `RestartCategoriesConfig` class added to fv3kube as an optional input to `c48_initial_conditions_overlay` to specify restart file names
- optional `restart_categories` RestartCategoriesConfig arg added to runtime's InitialCondition config class
- optional `restart_categories` RestartCategoriesConfig arg added to runtime's NudgingConfig class

- [X] Tests added


external/fv3kube/fv3kube/config.py
```diff
@@ -19,7 +19,12 @@
 }
 
 TILE_COORDS_FILENAMES = range(1, 7)  # tile numbering in model output filenames
-RESTART_CATEGORIES = ["fv_core.res", "sfc_data", "fv_tracer.res", "fv_srf_wnd.res"]
+REQUIRED_RESTART_CATEGORIES = [
```
nbren12, line 22: Overall looks good to me...maybe would be nice to use a dataclass instead of a dict to get self-documenting/mypy/typechecking goodness:
@dataclass
```
class RestartNameConfig:
    core: str = "fv_core.res"
    surface: str = "sfc_data"
    tracer: str = "fv_tracer.res"
    surface_wind: str = "fv_srf_wnd.res"
```

nbren12 (approved):

workflows/prognostic_c48_run/runtime/nudging.py
```diff
@@ -128,6 +131,8 @@ def _get_reference_state(
     if MPI.COMM_WORLD.rank == 0:
         fs = fsspec.get_fs_token_paths(dirname)[0]
         fs.get(dirname, localdir, recursive=True)
+        if restart_categories is not None:
+            _rename_local_restarts(localdir, restart_categories)
```
nbren12, line 135: I see how the renaming in needed since `fv3gfs.util.open_restart` hardcodes the name. At some point it would be nice to parameterize this hardcode upstream so that we can avoid doing additional I/O for something like this.

workflows/prognostic_c48_run/runtime/nudging.py
```diff
@@ -149,6 +154,19 @@ def _get_reference_state(
     return _to_state_dataarrays(state)
 
 
+def _rename_local_restarts(
```
nbren12, line 157: Can you test this logic?

workflows/prognostic_c48_run/runtime/nudging.py
nbren12: Can you open an issue for that?

nbren12: Actually maybe it would be more performant to use symbolic/hard links and avoid the copy.

external/fv3kube/fv3kube/config.py
```diff
@@ -76,11 +82,19 @@ def get_base_fv3config(version_key: Optional[str] = None) -> FV3Config:
     return base_yaml
 
 
+class RestartCategoriesConfig(Protocol):
```
nbren12, line 85: I can't see any need for multiple implementations of "RestartCategoriesConfig" so I don't see the need for this Protocol layer. Maybe just define the dataclass here instead of using a protocol?

nbren12 (approved):

external/fv3kube/fv3kube/config.py
brianhenn: Such a dataclass gets usage in both fv3kube and runtime, so yeah if it's imported from fv3kube into runtime then it only needs to be defined once (but is maybe less discoverable in runtime). I'll go with that.

PR #1606, opened 2022-02-04
Restore original order of shuffling and thinning of training data batches

spencerkclark:
This PR restores the original order of shuffling and thinning in producing stacked sample batches (shuffling then thinning).  When the order is flipped, it leads to sampling biases depending on the order of the stacked dimensions (see [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2021-10-01-offline-skill/2022-02-03-batching-investigation.ipynb) for more discussion).

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/stacking.py
+++ b/external/fv3fit/fv3fit/_shared/stacking.py
@@ -41,8 +41,8 @@ class StackedBatches(Sequence[xr.Dataset]):
             dim=SAMPLE_DIM_NAME
         )
         ds = check_empty(ds)
-        ds = preserve_samples_per_batch(ds)
-        return shuffled(self._random_state, [ds])[0]
+        ds = shuffled(self._random_state, [ds])[0]
+        return preserve_samples_per_batch(ds)
 
 
 def stack(ds: xr.Dataset, unstacked_dims: Sequence[str]):
--- a/external/fv3fit/tests/test_stacking.py
+++ b/external/fv3fit/tests/test_stacking.py
@@ -10,6 +10,7 @@ from fv3fit._shared.stacking import (
     stack,
     preserve_samples_per_batch,
     StackedBatches,
+    DATASET_DIM_NAME,
     SAMPLE_DIM_NAME,
 )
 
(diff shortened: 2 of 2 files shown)


external/fv3fit/tests/test_stacking.py
```diff
@@ -36,6 +37,26 @@ def test_StackedBatches_get_slice_stack():
...
+    stacked_batches = StackedBatches([batch, batch], np.random.RandomState(0))
+    stacked_batch = stacked_batches[0]
+
+    # If the order of shuffling and thinning were reversed, we would
+    # only select points from dataset zero, meaning stacked_batch.dataset
+    # would be all zeros.
+    expected_x = np.array([0, 3, 0, 3, 3, 1])
+    expected_dataset = np.array([1, 1, 2, 0, 2, 2])
+    np.testing.assert_equal(stacked_batch.x, expected_x)
+    np.testing.assert_equal(stacked_batch.dataset, expected_dataset)
```
mcgibbon, line 57: This does ensure shuffling happens before thinning, but is this really a test that it shuffles before it thins? This looks like a general regression test to me - if so you could update the function name to be more general, while keeping the comments that explain why this test is here.

No changes needed here, since regression is a pretty good test in this case. In the future, you could explicitly test that some operation happens before another by using mocks and asserting directly that e.g. the shuffle mock was called before the thinning mock. This would be more robust to other changes that could cause this test to fail.

external/fv3fit/fv3fit/_shared/stacking.py
```diff
@@ -41,8 +41,8 @@ def _stack_batch(self, ds_unstacked: xr.Dataset) -> xr.Dataset:
             dim=SAMPLE_DIM_NAME
         )
         ds = check_empty(ds)
-        ds = preserve_samples_per_batch(ds)
-        return shuffled(self._random_state, [ds])[0]
+        ds = shuffled(self._random_state, [ds])[0]
+        return preserve_samples_per_batch(ds)
```
mcgibbon, line 45: Could you rename `preserve_samples_per_batch` to something that describes what it's doing (subsampling the data by the inverse of the number of "dataset" coordinates)? From our discussion this function has nothing to do with uniformity of samples per batch.

mcgibbon (approved):
Thanks for the bug fix! Only one required change (renaming `preserve_samples_per_batch`).

We discussed over VC that preserve_samples_per_batch is kind of dangerous in what it does (silently subsamples if a "dataset" dimension is present), and should probably be removed after consulting with the team.

spencerkclark:
Subsampling is now handled differently making this PR no longer relevant.

PR #1607, opened 2022-02-04
fix Local so it loads datasets before saving to disk

mcgibbon:
Significant internal changes:
- Fixed `Local` so it will eagerly load Dataset objects before saving to disk
- Local now accepts slices in `__getitem__`

- [x] Tests added (for getitem only, only manual test for eager loading because mocking load makes the serialization fail, and dask is not a requirement of `loaders`)

### DIFF
--- a/external/loaders/loaders/batches/_sequences.py
+++ b/external/loaders/loaders/batches/_sequences.py
@@ -69,13 +69,20 @@ class Local(BaseSequence[T]):
 
     @classmethod
     def dump(cls, dataset, path):
-        joblib.dump(dataset, path)
+        try:
+            joblib.dump(dataset.load(), path)
+        except AttributeError:
+            joblib.dump(dataset, path)
 
     def __len__(self):
         return len(self.files)
 
     def __getitem__(self, i):
-        return joblib.load(self.files[i])
+        slice_value = self.files[i]
+        if isinstance(slice_value, str):
+            return joblib.load(slice_value)
+        else:
+            return [joblib.load(file) for file in slice_value]
 
 
 def to_local(sequence: Sequence[T], path: str, n_jobs: int = 4) -> Local[T]:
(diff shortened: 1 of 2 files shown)


external/loaders/loaders/batches/_sequences.py
```diff
@@ -69,13 +69,20 @@ def files(self):
 
     @classmethod
     def dump(cls, dataset, path):
-        joblib.dump(dataset, path)
+        try:
+            joblib.dump(dataset.load(), path)
+        except AttributeError:
+            joblib.dump(dataset, path)
```
nbren12, line 75: Duck-typing is fine, but I don't think you want to catch attribute errors which may occur in joblib.dump
```suggestion
        try:
            loaded_data = dataset.load()
        except AttributeError:
            pass
        else:
            loaded_data = dataset
         joblib.dump(loaded_data, path)
```

external/loaders/loaders/batches/_sequences.py
```diff
@@ -69,13 +69,20 @@ def files(self):
...
 
     def __len__(self):
         return len(self.files)
 
     def __getitem__(self, i):
-        return joblib.load(self.files[i])
+        slice_value = self.files[i]
```
nbren12, line 81: Is there a generic for containers that allow slices?

nbren12 (approved):
LGTM. but small suggestions below:

external/loaders/loaders/batches/_sequences.py
mcgibbon: I assumed it was Sequence because all the built-in sequences accept slices. The docs say "integer" but there's no type hint on the definition of Sequence.

I think we're pretty safe from this getting relied on in our package code (when would we slice a batches sequence? the idea is that batch is the largest we can process at once), but it will make things easier when I'm working in notebooks.

PR #1608, opened 2022-02-05
Allow variable_names argument in configuration for BatchesLoader

mcgibbon:
Currently BatchesLoader does not allow setting the keyword argument value for variable_names, instead taking it as an argument to `load_batches`. This PR updates that behavior to combine any value of variable_names provided in the configuration with the value given as an argument.

This will be necessary down the line for producing cached datasets directly from a configuration yaml for a BatchesLoader.

Refactored public API:
- The variables argument to BatchesLoader.load_batches is now optional, and will merge any value given in its configuration kwargs for variable_names with the argument passed to load_batches

- [x] Tests added

### DIFF
--- a/external/loaders/loaders/_config.py
+++ b/external/loaders/loaders/_config.py
@@ -1,5 +1,5 @@
 import abc
-from typing import Dict, TypeVar, Callable
+from typing import Dict, Optional, Sequence, TypeVar, Callable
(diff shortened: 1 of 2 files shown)


brianhenn (approved):
Looks good @mcgibbon thanks

PR #1613, opened 2022-02-09
add map and histogram from antarctica report to piggy-back report

nbren12:
Cloud water histograms are a useful diagnostics. I also added some time snapshots of cloud water.

* Clean-up piggy-back.py using registration pattern.

example: https://wandb.ai/ai2cm/microphysics-emulation/runs/3h2d6gwt?workspace=user-nbrenowitz


nbren12:
the maps are there, but need to hit down on the  "2 of 3" arrow in the "media" section.

The "prognostic_evaluation" report has the transposed version, so I'll just leave this one the same for now.

PR #1616, opened 2022-02-09
fv3fit: add logging callback

AnnaKwa:
- Adds a `TrainingLoopLossHistory` class with a callback that can be used in the training loop `.fit` method to save the loss history for each timestep batch in the epochs. The `log_summary` method can be used after training is complete to print the loss history to the logs. I chose this over just saving all stdout to the logs because the progress bars in the the keras training output would save in multiple lines and made parsing the logs annoying.
- Saves the logs in the model output location if using the train script. I find it easier to have it here than having to look up the name of the job name that trained the model to get the logs.

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/dense.py
+++ b/external/fv3fit/fv3fit/keras/_models/dense.py
@@ -9,7 +9,12 @@ from ..._shared.config import (
     OptimizerConfig,
     register_training_function,
 )
-from .shared import TrainingLoopConfig, XyMultiArraySequence, DenseNetworkConfig
+from .shared import (
+    TrainingLoopConfig,
+    XyMultiArraySequence,
+    DenseNetworkConfig,
+    TrainingLoopLossHistory,
+)
 from fv3fit.keras._models.shared import PureKerasModel, LossConfig, ClipConfig
 from fv3fit.keras._models.shared.utils import (
     standard_denormalize,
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/keras/_models/shared/callbacks.py
```diff
@@ -0,0 +1,37 @@
...
+
+from .training_loop import EpochResult
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class TrainingLoopLossHistory:
+    """Summarizes the history across epochs and batches
```
mcgibbon, line 11: Can you document the attributes in the Google style? i.e.
```
Attributes:
    train_loss_end_of_epoch: <description>
    ...
```

external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -117,17 +122,22 @@ def train_dense_model(
...
+        validation_data=validation_data,
+        callbacks=[loss_history.callback],
     )
     predictor = PureKerasModel(
         input_variables=hyperparameters.input_variables,
         output_variables=hyperparameters.output_variables,
         output_metadata=output_metadata,
         model=predict_model,
-        unstacked_dims=("z",),
+        unstacked_dims=("z"),
```
mcgibbon, line 137: Can you revert this? While technically this code will run, the meaning changes. previously it was a tuple containing "z" as its one element, now it is passing the string "z" as the value for unstacked_dims. This works because if you iterate over a string it will return each character, but this will break as soon as the unstacked dimension is more than one character long.
```suggestion
        unstacked_dims=("z",),
```

mcgibbon (approved):
So clean, I really like it! Nice work!

There's one minor bug to fix before merging, and a docstring suggestion.

PR #1618, opened 2022-02-10
vcm limiter class, applied to prognostic run tendency prescriber

brianhenn:
Implements a dataset limiter class in vcm to limit outliers to some specified upper and lower quantile values. The variables and dimensions along which the limits are computed are configurable. This PR also allows applying configurable upper and lower limits to the prognostic run's tendency prescriber, e.g., to help prevent crashes in the prescribed physics run. 

Added public API:
- `vcm.limit.DatasetQuantileLimiter`

Refactored public API:
- Added optional `limit_quantiles` argument to `runtime.transformers.tendency_prescriber.TendencyPrescriberConfig`

- [X] Tests added

### DIFF
--- a/external/loaders/loaders/mappers/_fine_res.py
+++ b/external/loaders/loaders/mappers/_fine_res.py
@@ -9,7 +9,6 @@ import numpy as np
 import fsspec
 
 from vcm.fv3.metadata import gfdl_to_standard
-from vcm.limit import DatasetQuantileLimiter
 from loaders._config import mapper_functions
 from loaders.mappers._base import GeoMapper
 from loaders.mappers._xarray import XarrayMapper
@@ -224,9 +223,7 @@ def open_fine_resolution(
 
 
 def _open_precomputed_fine_resolution_dataset(
-    fine_url: str,
-    additional_dataset_urls: Optional[Sequence[str]] = None,
-    limit_alpha: Optional[float] = None,
+    fine_url: str, additional_dataset_urls: Optional[Sequence[str]] = None
 ) -> MLTendencies:
 
     merged = _open_merged_dataset(
@@ -235,19 +232,12 @@ def _open_precomputed_fine_resolution_dataset(
         standardize_fine_coords=False,
     )
 
-    if limit_alpha is not None:
-        limiter = DatasetQuantileLimiter(alpha=limit_alpha, limit_only=["Q1", "Q2"])
-        limited = limiter.fit_transform(
-            merged, feature_dims=["z", "tile"], fit_indexers={"time": 0},
-        )
-        merged = limited
-
     return _ml_standard_names(merged)
 
 
 @mapper_functions.register
 def open_precomputed_fine_resolution(
-    fine_url: str, additional_dataset_urls: str = None, limit_alpha: float = None
+    fine_url: str, additional_dataset_urls: str = None
 ) -> GeoMapper:
     """
     Open a fine-res mapper from precomputed data, optionally using state
@@ -258,15 +248,11 @@ def open_precomputed_fine_resolution(
             precomputed Q1 and Q2
         additional_dataset_urls: sequence of urls which to zarrs containing additional
             data to be merged into the resulting mapper dataset
-        limit_alpha: two-tailed alpha for computing extrema quantiles of Q1 and Q2,
-            values beyond which will be reduced to the quantile
     Returns:
         a mapper
     """
     return XarrayMapper(
         _open_precomputed_fine_resolution_dataset(
-            fine_url=fine_url,
-            additional_dataset_urls=additional_dataset_urls,
-            limit_alpha=limit_alpha,
+            fine_url=fine_url, additional_dataset_urls=additional_dataset_urls
         )
     )
--- a/external/loaders/loaders/mappers/_fine_res.py
+++ b/external/loaders/loaders/mappers/_fine_res.py
@@ -9,6 +9,7 @@ import numpy as np
 import fsspec
 
 from vcm.fv3.metadata import gfdl_to_standard
+from vcm.limit import DatasetQuantileLimiter
 from loaders._config import mapper_functions
 from loaders.mappers._base import GeoMapper
 from loaders.mappers._xarray import XarrayMapper
@@ -223,7 +224,9 @@ def open_fine_resolution(
 
 
 def _open_precomputed_fine_resolution_dataset(
-    fine_url: str, additional_dataset_urls: Optional[Sequence[str]] = None
+    fine_url: str,
+    additional_dataset_urls: Optional[Sequence[str]] = None,
+    limit_alpha: Optional[float] = None,
 ) -> MLTendencies:
 
     merged = _open_merged_dataset(
@@ -232,12 +235,19 @@ def _open_precomputed_fine_resolution_dataset(
         standardize_fine_coords=False,
     )
 
+    if limit_alpha is not None:
+        limiter = DatasetQuantileLimiter(alpha=limit_alpha, limit_only=["Q1", "Q2"])
+        limited = limiter.fit_transform(
+            merged, feature_dims=["z", "tile"], fit_indexers={"time": 0},
+        )
+        merged = limited
+
     return _ml_standard_names(merged)
 
 
 @mapper_functions.register
 def open_precomputed_fine_resolution(
-    fine_url: str, additional_dataset_urls: str = None
+    fine_url: str, additional_dataset_urls: str = None, limit_alpha: float = None
 ) -> GeoMapper:
     """
     Open a fine-res mapper from precomputed data, optionally using state
@@ -248,11 +258,15 @@ def open_precomputed_fine_resolution(
             precomputed Q1 and Q2
         additional_dataset_urls: sequence of urls which to zarrs containing additional
             data to be merged into the resulting mapper dataset
+        limit_alpha: two-tailed alpha for computing extrema quantiles of Q1 and Q2,
+            values beyond which will be reduced to the quantile
     Returns:
         a mapper
     """
     return XarrayMapper(
         _open_precomputed_fine_resolution_dataset(
-            fine_url=fine_url, additional_dataset_urls=additional_dataset_urls
+            fine_url=fine_url,
+            additional_dataset_urls=additional_dataset_urls,
+            limit_alpha=limit_alpha,
         )
     )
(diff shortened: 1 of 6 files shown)


external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,113 @@
...
+            limited = data.copy()
+            vars_ = self._limit_only if self._limit_only is not None else data.data_vars
+            for var in vars_:
+                limited[var] = _limit_extremes(data[var], self._limits[var])
+        elif isinstance(data, xr.DataArray):
+            if (self._limit_only is None) or (data.name in self._limit_only):
+                limited = _limit_extremes(data, self._limits[data.name])
+            else:
+                limited = data
+        return limited
```
nbren12, line 108: Does this API need to be polymorphic or is it just a small convenience for e.g. interactive use? In addition to making this function more complicated, returning union types can make life difficult since any downstream code won't know if `.transform` returned  a dataset or dataarray. You can type hint this more effectively using `TypeVar`, but I think you should just delete the DataArray support. This will fail anyway if `array.name` is not defined.

Users can manually cast the array to dataset like this if they want: `arr.to_dataset(name="the name")`.

```suggestion
    def transform(
        self, data: xr.Dataset) ->xr.Dataset:
        """Limit data.
        
        Args:
            data: Dataset to be limited
            
        Returns: Limited dataset
        
        """
        if self._limits is None:
            raise ValueError("Limiter method .fit must be called before .transform")
     
        limited = data.copy()
        vars_ = self._limit_only if self._limit_only is not None else data.data_vars
        for var in vars_:
           limited[var] = _limit_extremes(data[var], self._limits[var])
        return limited

File: external/vcm/tests/test_limit.py [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+            expected_arr[0 : (int(scale) + 1), :] = lower"""
+
+
+def get_dataset(alpha):
+    arr = _get_arr(alpha)
+    da = xr.DataArray(arr, dims=["x", "z"])
+    return xr.Dataset({"Q1": da, "Q2": da})
+
+
+def get_expected_arr(alpha, feature_dims, fit_indexers):
```
nbren12, line 37: This mocking code is pretty complex...two levels deep of if-statements and fairly complex random (?) data. I find this test harder to read then the code it is testing.

I think part of the issue is you are trying to unit-test the integration of the fitting and limiting, rather than each on their own. Something like this would be simpler IMO:

```
def test_fit_limits_correct():
    # ~ two lines for the data mock
    simple_data = ...
    limiter = DatasetQuantileLimiter.fit(simple_data)
    assert limiter._limits == ...


def test_transform():
    
    limiter = DatasetQuantileLimiter
    limiter._limits = simple mock
    assert limiter.transform(simple_data) == ...
```

```

Comment (nbren12) [reply external/vcm/tests/test_limit.py:37]: Those tests would be easier to refactor if we e.g. develop new ways of computing the limits.

File: external/vcm/vcm/limit.py [hunk 3]
```diff
@@ -0,0 +1,113 @@
...
+            set(sample_ds.dims) - set(feature_dims)
+            if feature_dims is not None
+            else sample_ds.dims
+        )
+        lower = sample_ds.quantile(self._alpha / 2.0, dim=sample_dims)
+        upper = sample_ds.quantile(1.0 - self._alpha / 2.0, dim=sample_dims)
+        self._limits = xr.concat(
+            [lower, upper],
+            dim=xr.DataArray(["lower", "upper"], dims=["bounds"], name="bounds"),
+        )
```
nbren12, line 82: Maybe simpler to just store the upper and lower bounds in different attrs?
```suggestion
        self._lower_limit = sample_ds.quantile(self._alpha / 2.0, dim=sample_dims)
        self._upper_limit = sample_ds.quantile(1.0 - self._alpha / 2.0, dim=sample_dims)
```

external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,113 @@
...
+        Returns: Limited dataset or dataarray
+        
+        """
+        if self._limits is None:
+            raise ValueError("Limiter method .fit must be called before .transform")
+        if isinstance(data, xr.Dataset):
+            limited = data.copy()
+            vars_ = self._limit_only if self._limit_only is not None else data.data_vars
+            for var in vars_:
+                limited[var] = _limit_extremes(data[var], self._limits[var])
```
nbren12, line 102: I suggest deleting `_limit_extremes`. My suggestion above removes the need for the "above" and "below" hardcodes. The only remaining logic is a simple one-liner that can be copied here.



```suggestion
                limited[var] = data.where(data[var] < self._upper[var], self._upper[var]).where(data[var] > self._lower[var], self._lower[var])
```

external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,113 @@
...
+        elif isinstance(data, xr.DataArray):
+            if (self._limit_only is None) or (data.name in self._limit_only):
+                limited = _limit_extremes(data, self._limits[data.name])
+            else:
+                limited = data
+        return limited
+
+    @property
+    def limits(self) -> xr.Dataset:
```
nbren12, line 111: This feels a bit Java-y. I'd rename `self._limits` to `self.limits` and delete this property.

external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -9,6 +9,7 @@
 import fsspec
 
 from vcm.fv3.metadata import gfdl_to_standard
+from vcm.limit import DatasetQuantileLimiter
```
nbren12, line 12: Limiting is a post-processing feature that could be applied to any mapper, so it's odd to couple it to just these few mappers. Ideally mappers would just be simple data objects...and any downstream tools (e.g. prog run or fv3fit) can use `vcm.DatasetQuantileLimiter` in their own pre/post processing pipelines.

For now, you should move this logic to the prognostic run...which I think is the only place you currently plan to use it now.

nbren12 (changes requested):
Thanks @brianhenn. I think this will be a pretty useful feature. But I do have some requested changes below:

external/vcm/vcm/limit.py
brianhenn: Switched to dataset input as suggested

external/vcm/vcm/limit.py
brianhenn: Since we're storing the upper and lower limits separately, I like keeping this as a way of accessing both via a single attribute.

external/loaders/loaders/mappers/_fine_res.py
brianhenn: Good suggestion, I moved it to the tendency prescriber

external/vcm/vcm/limit.py
nbren12: sure.

external/vcm/tests/test_limit.py
```diff
@@ -0,0 +1,84 @@
...
+    "limit_only", [pytest.param(None, id="all"), pytest.param(["Q1"], id="Q1_only")]
+)
+def test_limiter_transform_limit_only(limit_only):
+    alpha = 0.25
+    ds = get_dataset(alpha)
+    limiter = DatasetQuantileLimiter(alpha, limit_only=limit_only)
+    limiter._upper, limiter._lower = get_limits(3.5, 0.5, alpha, ["z"])
+    limited = limiter.transform(ds, deepcopy=True)
+    expected_ds = ds.copy(deep=True)
+    for var in limit_only or ds.data_vars:
```
nbren12, line 81: I got a little confused by the syntax here. I think a new variable name would help:

```suggestion
    vars_to_limit = limit_only or ds.data_vars
    for var in vars_to_limit:
```

external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,99 @@
+import xarray as xr
+from sklearn.base import BaseEstimator, TransformerMixin
```
nbren12, line 2: I don't think it's worth adding the direct scikit-learn dep to `runtime` just for these base classes. Also, sklearn classes usually take in and return numpy arrays. 

Duck typing should be sufficient here, and we can define our own protocol if we extend the method.

external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,99 @@
...
+    Args:
+        alpha: two-tailed alpha for computing extrema quantiles, values beyond which
+            will be reduced to the quantile
+        limit_only: variables in the dataset to be limited; if not specified all
+            variables are limited
+
+    """
+
+    def __init__(
+        self, alpha: float, limit_only: Optional[Sequence[str]] = None,
```
nbren12, line 22: I suggest using two knobs for upper and lower bounds:

```suggestion
        self, lower_quantile_limit: float, upper_quantile_limit: float, limit_only: Optional[Sequence[str]] = None,
```

I could see wanting to use different quantiles for the upper and lower threshold. Also, its hard to tell from the docstring how `alpha` is turned into upper/lower bounds. Finally, the "alpha to upper/lower quantile" formula is in two places...the tests and this class, so increases the implicit coupling.

external/vcm/vcm/limit.py
```diff
@@ -0,0 +1,99 @@
...
+        
+        """
+        if fit_indexers is not None and feature_dims is not None:
+            for index_dim in fit_indexers.keys():
+                if index_dim in feature_dims:
+                    raise ValueError(
+                        f"Indexer dim {index_dim} may not be in feature_dims."
+                    )
+
+        sample_ds = (ds.isel(**fit_indexers) if fit_indexers is not None else ds).load()
```
nbren12, line 55: Does this function need to know about `fit_indexers`? Seems like you should extract the subselection from this method, and call it like this:

    limiter.fit(ds.isel(fit_indexers), dims=["lon"])

this is personally clearer to me since I know what `.isel` does, but not what a fit_indexer is.

nbren12 (approved):
Thanks for addressing my larger scale comments! I'm hitting approve since the remaining stuff is pretty small. The confusing/unneeded concepts of "alpha" and "fit_indexers" are the two remaining issues that should be addressed before merging.

external/vcm/vcm/limit.py
brianhenn: As above the use for this is in `fit_transform` which takes a single dataset and where you want to fit on a portion of it while transforming the whole thing.  If `fit` and `transform` are always separate calls then it isn't needed. I'll remove both for simplicity then.

PR #1619, opened 2022-02-10
Fix "antarctic-only" selection for emulation training

### LINKED ISSUE #1617: Antarctic selection transform broken
Using the Antarctic selection in the emulation load transform pipeline now fails on tiles with no Antarctic points.  Indexing zero length slices on netCDF doesn't work, so a `microphysics_traning.py` with `antarctica_only=True` in the config fails.

```
  File "/usr/local/lib/python3.8/dist-packages/xarray/backends/netCDF4_.py", line 104, in _getitem
    array = getitem(original_array, key)
  File "netCDF4/_netCDF4.pyx", line 4428, in netCDF4._netCDF4.Variable.__getitem__
  File "/usr/local/lib/python3.8/dist-packages/netCDF4/utils.py", line 467, in _out_array_shape
    c = count[..., i].ravel()[0] # All elements should be identical.
IndexError: index 0 is out of bounds for axis 0 with size 0

During handling of the above exception, another exception occurred:

  File "test_select.py", line 9, in <module>
    data.isel({"sample": mask}).load()
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/dataset.py", line 873, in load
    v.load()
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/variable.py", line 451, in load
    self._data = np.asarray(self._data)
  File "/usr/local/lib/python3.8/dist-packages/numpy/core/_asarray.py", line 83, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/indexing.py", line 548, in __array__
    self._ensure_cached()
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/indexing.py", line 545, in _ensure_cached
    self.array = NumpyIndexingAdapter(np.asarray(self.array))
  File "/usr/local/lib/python3.8/dist-packages/numpy/core/_asarray.py", line 83, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/indexing.py", line 518, in __array__
    return np.asarray(self.array, dtype=dtype)
  File "/usr/local/lib/python3.8/dist-packages/numpy/core/_asarray.py", line 83, in asarray
    return array(a, dtype, copy=False, order=order)
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/indexing.py", line 419, in __array__
    return np.asarray(array[self.key], dtype=None)
  File "/usr/local/lib/python3.8/dist-packages/xarray/backends/netCDF4_.py", line 91, in __getitem__
    return indexing.explicit_indexing_adapter(
  File "/usr/local/lib/python3.8/dist-packages/xarray/core/indexing.py", line 710, in explicit_indexing_adapter
    result = raw_indexing_method(raw_key.tuple)
  File "/usr/local/lib/python3.8/dist-packages/xarray/backends/netCDF4_.py", line 114, in _getitem
    raise IndexError(msg)
IndexError: The indexing operation you are attempting to perform is not valid on netCDF4.Variable object. Try loading your data into memory first by calling .load().
```

`xarray` and `netcdf4` haven't been changed in the constraints in almost half a year, so I'm not sure what triggered the difference.

frodre:
Somewhere along the line the antarctic-only selection pipeline broke when encountering tile data outside of the Antarctic.  

resolves #1617 

It seems to be a change in behavior when using `.isel` with an empty mask (i.e., all false values) on an unloaded xarray dataset opened from a netCDF.  Added a workaround and test for this case, but we'll probably just scrub this in place of a more general selection transform down the line.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/transforms.py
+++ b/external/fv3fit/fv3fit/emulation/data/transforms.py
@@ -97,7 +97,16 @@ def select_antarctic(dataset: xr.Dataset, sample_dim_name="sample") -> xr.Datase
 
     logger.debug("Reducing samples to antarctic points (<60S) only")
     mask = dataset["latitude"] < -np.deg2rad(60)
-    dataset = dataset.isel({sample_dim_name: mask})
+
+    if not np.any(mask):
+        # https://github.com/ai2cm/fv3net/issues/1617
+        # Workaround without having to load entire dataset
+        # just to then toss that data out, while still returning an
+        # empty dataset with correct metadata
+        dataset = dataset.isel({sample_dim_name: slice(0, 1)}).load()
+        dataset = dataset.isel({sample_dim_name: [False]})
+    else:
+        dataset = dataset.isel({sample_dim_name: mask})
 
     return dataset
 
--- a/external/fv3fit/tests/emulation/test_data_transforms.py
+++ b/external/fv3fit/tests/emulation/test_data_transforms.py
@@ -74,7 +74,6 @@ def test_select_antarctic(lats, data, expected):
     xr.testing.assert_equal(expected_da, result["field"])
 
 
-@pytest.mark.xfail
 def test_select_antarctic_xarray_netCDF():
     """
     xarray can't use an empty index mask (i.e., all False) along
--- a/external/fv3fit/tests/emulation/test_data_transforms.py
+++ b/external/fv3fit/tests/emulation/test_data_transforms.py
@@ -74,6 +74,28 @@ def test_select_antarctic(lats, data, expected):
     xr.testing.assert_equal(expected_da, result["field"])
 
 
+@pytest.mark.xfail
+def test_select_antarctic_xarray_netCDF():
+    """
+    xarray can't use an empty index mask (i.e., all False) along
+    a selection dimension for multi-dimensional data with an unloaded
+    netCDF backend
+    """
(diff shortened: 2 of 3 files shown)


nbren12 (approved):
Preemptive approval. Pretty weird corner case if you ask me...maybe worth raising with xarray?

FWIW, I think the antarctica subselection/(and many of the other `transforms`) would be slightly cleaner if deferred until after the `tf.data.Dataset` object is loaded. Implementing this logic in tensorflow would make it easier to e.g. build a merged model which uses one model in antarctica and another elsewhere.

frodre:
> Preemptive approval. Pretty weird corner case if you ask me...maybe worth raising with xarray?

Yeah, the behavior appears to already be the result of an edge case, so I was kind of thinking this was working as intended? (I.e., we don't have a way to support this behavior).  See https://github.com/pydata/xarray/issues/593  Could open an issue just to make sure.

> FWIW, I think the antarctica subselection/(and many of the other transforms) would be slightly cleaner if deferred until after the tf.data.Dataset object is loaded. Implementing this logic in tensorflow would make it easier to e.g. build a merged model which uses one model in antarctica and another elsewhere.

I agree.  I was thinking of some potential refactors to consolidate behaviors into the tensor transforms.

PR #1622, opened 2022-02-11
Add functionality from StackedBatches to batches_from_mapper

mcgibbon:
We are working to consolidate preprocessing functionality into data configuration, to allow saving preprocessed datasets based only on data configuration. This PR moves existing functionality in StackedBatches to batches_from_mapper for that purpose.

Refactored public API:
- Added argument `drop_nans` to `batches_from_mapper`, which if set to True (default False) will drop data with NaN values and raise an exception if no data remains
- added argument `unstacked_dims` to `batches_from_mapper`, which if given will stack and shuffle data retaining these unstacked dimensions

Significant internal changes:
- `stack` is duplicated from `fv3fit` to `loaders`, with plans to delete `stack` from `fv3fit` once our training routines no longer stack directly

- [x] Tests added

### DIFF
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -122,7 +122,7 @@ def batches_from_mapper(
         in_memory: if True, load data eagerly and keep it in memory
         unstacked_dims: if given, produce stacked and shuffled batches retaining
             these dimensions as unstacked (non-sample) dimensions
-        drop_nans: if True, drop NaN values from the data, and raise an
+        drop_nans: if True, drop samples with NaN values from the data, and raise an
             exception if all values in a batch are NaN. requires unstacked_dims
             argument is given, raises a ValueError otherwise.
     Raises:
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -122,8 +122,8 @@ def batches_from_mapper(
         in_memory: if True, load data eagerly and keep it in memory
         drop_nans: if True, drop NaN values from the data, and raise an
             exception if all values in a batch are NaN
-        unstacked_dims: if given, produce stacked batches retaining these
-            dimensions as unstacked (non-sample) dimensions
+        unstacked_dims: if given, produce stacked and shuffled batches retaining
+            these dimensions as unstacked (non-sample) dimensions
     Raises:
         TypeError: If no variable_names are provided to select the final datasets
 
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -104,8 +104,8 @@ def batches_from_mapper(
     res: str = "c48",
     needs_grid: bool = True,
     in_memory: bool = False,
-    drop_nans: bool = False,
     unstacked_dims: Optional[Sequence[str]] = None,
+    drop_nans: bool = False,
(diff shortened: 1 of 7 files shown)


external/loaders/loaders/batches/_batch.py
```diff
@@ -147,12 +157,19 @@ def batches_from_mapper(
...
 
-    transforms += [add_derived_data(variable_names)]
+    transforms.append(add_derived_data(variable_names))
+
+    if unstacked_dims is not None:
+        transforms.append(curry(stack)(unstacked_dims))
+        transforms.append(shuffle)
+
+    if drop_nans:
+        transforms.append(dropna)
```
AnnaKwa, line 167: `dropna` assumes the input is stacked with a `sample` dimension, can you put this check within the block above? 

Although that creates some ambiguity in the config since dropna needs a `dim` arg and only makes sense if `unstacked_dims` is provided. Maybe we should go with your original plan of removing `dropna` from the data loading, and assume that our training data should have non nan values. I'd then worry about the case where a bunch of training data gets downloaded only for the training to start and choke on a NaN.

AnnaKwa: As discussed, let's go with raising an exception if data is unstacked but `drop_nans` is True.

external/loaders/loaders/batches/_batch.py
```diff
@@ -115,6 +120,10 @@ def batches_from_mapper(
         needs_grid: Add grid information into batched datasets. [Warning] requires
             remote GCS access
         in_memory: if True, load data eagerly and keep it in memory
+        drop_nans: if True, drop NaN values from the data, and raise an
+            exception if all values in a batch are NaN
```
AnnaKwa, line 124: Can you add to this to say it only operates if data is stacked?

AnnaKwa (changes requested):
LGTM! Only request is to clarify the behavior of `drop_nans` in the docstring and if statement since it only works if `unstacked_dims` is provided.

external/loaders/loaders/batches/_batch.py
mcgibbon: Done.

external/loaders/loaders/batches/_batch.py
mcgibbon: Done.

mcgibbon (commented):
PR should be ready for re-review.

AnnaKwa (approved):
Thanks! The changes LGTM.

external/loaders/loaders/batches/_batch.py
```diff
@@ -115,6 +120,11 @@ def batches_from_mapper(
         needs_grid: Add grid information into batched datasets. [Warning] requires
             remote GCS access
         in_memory: if True, load data eagerly and keep it in memory
+        unstacked_dims: if given, produce stacked and shuffled batches retaining
+            these dimensions as unstacked (non-sample) dimensions
+        drop_nans: if True, drop NaN values from the data, and raise an
+            exception if all values in a batch are NaN. requires unstacked_dims
+            argument is given, raises a ValueError otherwise.
```
brianhenn, line 127: Since this requires stacking maybe specify that what drop_nans does is drop samples with nans, or rename argument to reflect that.

brianhenn (commented):
LGTM, thanks @mcgibbon .

Issue #1596, opened 2022-01-26
Reduce coupling between offline diags and fv3fit in input sensitivity plots


AnnaKwa:
The offline diags `_input_sensivitiy` module is currently very much coupled to models in fv3fit, see this comment: https://github.com/ai2cm/fv3net/pull/1590#discussion_r792226035

It would be nice to clean up the try/except logic that produces the input sensitivity plots by having a common interface for different predictors. 

Noah suggests the following could be used to return a common object and subsequent plotting code could look into the container and figure out which type of plot to make.

```
@dataclass
class InputSensitivity:
    feature_importance: Optional[<some numpy arrays + indices>]
    jacobian: Optional[dict of numpy arrays...]

fv3fit.Predictor.get_input_sensitivity -> InputSensitivity  
```

PR #1623, opened 2022-02-11
fv3fit: Add `input_sensitivity` method to `Predictor`

AnnaKwa:
This adds a method `Predictor.input_sensitivity` which takes in a sample dataset and returns a `InputSensitivity` container which uses different formats to describe RF and NN input sensitivity. This helps to decouple some of the input sensitivity calculations from the offline diags (https://github.com/ai2cm/fv3net/issues/1596). 

I also cleaned up some of the dataset handling in the input sensitivity calculation of the offline diags so that the `input_sensitivity` methods take an already-stacked dataset of sample data.

The actual calculation and plotting of the input sensitivities in the offline diags is unchanged in this PR. A follow on PR will change that code to use the `input_sensitivity` method and adjust the plotting code to use the data from the `InputSensitivity` container.

Added public API:
-  `Predictor.input_sensitivity()`

Resolves #<[github issues](https://github.com/ai2cm/fv3net/issues/1596)>

### DIFF
--- a/external/fv3fit/fv3fit/_shared/input_sensitivity.py
+++ b/external/fv3fit/fv3fit/_shared/input_sensitivity.py
@@ -9,5 +9,5 @@ JacobianInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
 
 @dataclass
 class InputSensitivity:
-    rf_feature_importances: Optional[RandomForestInputSensitivity]
-    jacobians: Optional[JacobianInputSensitivity]
+    rf_feature_importances: Optional[RandomForestInputSensitivity] = None
+    jacobians: Optional[JacobianInputSensitivity] = None
--- a/external/fv3fit/fv3fit/_shared/input_sensitivity.py
+++ b/external/fv3fit/fv3fit/_shared/input_sensitivity.py
@@ -1,13 +1,34 @@
 import numpy as np
-from typing import Optional, Mapping
+from typing import Optional, Mapping, Sequence
 from dataclasses import dataclass
 
 
-RandomForestInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
+# Jacobian format: {y_name: {x_name: dy/dx}}
 JacobianInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
 
 
 @dataclass
+class RandomForestInputSensitivity:
+    """
+    Attributes:
+        name: input feature name
+        mean_importances: average value of feature importance across ensemble
+        std_importances: standard deviation of feature importance across ensemble
+        indices: for input features of length >1 along feature dimension, the indices
+            corresponding to position along feature dimension. Has length of 1 with
+            nan index for scalar features.
+    """
+
+    name: str
+    mean_importances: Sequence[float]
+    std_importances: Sequence[float]
+    indices: Sequence[int]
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/_shared/input_sensitivity.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+from typing import Optional, Mapping
+from dataclasses import dataclass
+
+
+RandomForestInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
```
AnnaKwa, line 6: Ideally I wanted to make these dataclasses that would be constructed in the RF/NN `input_sensitivity` methods but couldn't find a way to define these in the individual model's modules without running into circular imports when trying to import here for type hinting. If you have a better suggestion let me know.

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -78,11 +71,20 @@ def _get_variable_indices(
     return variable_indices
 
 
+def _stack_sample_data(ds: xr.Dataset) -> xr.Dataset:
```
AnnaKwa, line 74: Moved the stacking out of the input sensitivity calculations so the input here is the same as what the `Predictor.input_sensitivity` method expects. I guess I should have saved this for the follow on PR but I'd already commited this change.

external/fv3fit/fv3fit/_shared/input_sensitivity.py
brianhenn: I don't have any solutions to the circular import issue, could you still use dataclasses to better specify the `{method}InputSensitivity` data structures? Right now the jacobian and RF importance ones look identical from the typing alone, but of course they have pretty different data meanings. It would help make it more obvious what the data are/how they should be plotted.

brianhenn (commented):
This looks useful to me =, thanks. I can follow up but also happy to let someone with more familiarity with the Jacobians look as well.

external/fv3fit/fv3fit/_shared/input_sensitivity.py
AnnaKwa: Sure, I added one for the RF input sensitivity. I couldn't think of a good way to use a data struct for the Jacobian data as it's a dict of dicts, which gets kind of unwieldy and harder to traverse later if I try to make a formal data structure out of it.

external/fv3fit/fv3fit/_shared/input_sensitivity.py
```diff
@@ -1,13 +1,34 @@
...
-RandomForestInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
+# Jacobian format: {y_name: {x_name: dy/dx}}
 JacobianInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
 
 
+@dataclass
+class RandomForestInputSensitivity:
+    """
+    Attributes:
+        name: input feature name
```
brianhenn, line 14: Nit: Make`RandomForestInputSensitivities` a mapping with the `name` as the key?

brianhenn (approved):
LGTM @AnnaKwa

PR #1621, opened 2022-02-10
Move piggy-back.py to diagnostics tree

nbren12:
Prepare piggy-back.py for further development. Currently there is no framework for testing this code. Once this refactor is merged, I can add some new plots comparing the condensation and precip skill.

Move this script into the `diagnostics` package and test it.

Added Public API:
- vcm.cdl_to_dataset. This is makes it easy to generate netCDF-like datasets. Common Data Language (CDL) is much more concise than the yaml format used by `synth`.
- CLI: `projects/microphysics/scripts/piggy-back.py` renamed to `prognostic_run_diagnostics piggy`

PR #1625, opened 2022-02-11
Implement CDL parser in python

nbren12:
Depending on a command line tool like `ncgen` within python scripts
makes dependency management harder, so I implemented a pure python parser. This should also allow building e.g. a dask dataset from CDL directly without first writing a large netCDF file to disk using `ncgen`.

I tried to document this code well, but realize that parsers are outside of our typical skill set. That said, I was able to figure out how to use Lark in an hour or two.

* implement cdl_to_dataset in pure python using the Lark parser

Requirement changes:
- Revert the lock files back to the state before #1621 was merged.
- Add lark


external/vcm/vcm/cdl/generate.py
```diff
@@ -0,0 +1,113 @@
+from typing import Mapping
+import lark
+import numpy as np
+import xarray
+
+from .parser import grammar
+
```
AnnaKwa, line 7: Can you add back the info comment that was in the removed cdl.py pointing to the UCAR site for people to reference how to write CDL strings?

AnnaKwa (approved):
The parser is cool! LGTM. Can you add back the url for referencing the UCAR page?

external/vcm/vcm/cdl/generate.py
nbren12: I added these docs to the cdl_to_dataset docstring.

Issue #1430, opened 2021-10-19
`batches_from_geodata` redundancy


AnnaKwa:
The batch opening functions `batches_from_geodata` and `batches_from_mapper` fulfill the same need now, but with different APIs and config loading classes (`BatchesLoader` and `BatchesFromMapper`). Since `batches_from_geodata` itself requires a `mapping_function` arg, it doesn't really make sense that its corresponding config class is `BatchesLoader` and not `BatchesFromMapper`.  Any configuration using `batches_from_geodata` can be written to also be usable by `BatchesFromMapper`, and allowing multiple options for specifying data configs of mapper-like input data necessitates additional logic in the offline report where the snapshot specific diagnostics check whether the test data is a mapper with timestep keys. A common case is when the data config uses `batches_from_geodata` and the mapping function `open_zarr`.

I propose to remove the `batches_from_geodata` function, which forces people to use `batches_from_mapper` for any data config that has timestep keys. This is a breaking change for a lot of existing configurations that use `open_zarr` though. I could also add code for backwards compatibility that allows `batches_from_geodata` to still be used in data configs but initializes a `BatchesFromMapper` config instead of `BatchesLoader`.

PR #1626, opened 2022-02-11
add subsample_ratio argument to batches_from_mapper

mcgibbon:
This PR adds an argument to batches_from_mapper which allows subsampling data along the sample dimension.

Refactored public API:
- Added `subsample_ratio` argument to `batches_from_mapper` which will keep only the requested fraction of the input data, selected randomly
- removed random_seed argument from batches_from_mapper and batches_from_geodata, instead using `np.random` directly
- removed unused argument subsample_size from batches_from_geodata

- [x] Tests added


external/loaders/loaders/batches/_batch.py
```diff
@@ -163,8 +158,13 @@ def batches_from_mapper(
     if unstacked_dims is not None:
         transforms.append(curry(stack)(unstacked_dims))
         transforms.append(shuffle)
```
brianhenn, line 160: I think in the future we'd want an option to sample over a dimension like time, x/y/tile, or dataset (for Spencer) evenly eg using `xr.Dataset.thin` and then stack+shuffle _after_ sampling. I'm thinking about this since at C384 the subsampling is probably going to have to be severe and the odds of getting batches that are unbalanced in some way seem higher. Is this API suitable to adding that option later? `subsample_ratio` might be so specific that it kind of precludes that approach. I guess we could add a `subsample_every_n` arg that's mutually exclusive with `subsample_ratio`. Something to keep in mind.

external/loaders/loaders/_utils.py
```diff
@@ -87,6 +87,17 @@ def dropna(ds: xr.Dataset) -> xr.Dataset:
...
 
 
+@curry
+def select_first_samples(fraction: float, ds: xr.Dataset) -> xr.Dataset:
+    """
+    Return a dataset with only the given fraction of samples from the
+    input dataset, taking the first samples.
+    """
+    n_samples = len(ds[SAMPLE_DIM_NAME])
+    n_samples_keep = int(fraction * n_samples)
```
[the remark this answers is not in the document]
mcgibbon: Updated.

external/loaders/loaders/batches/_batch.py
mcgibbon: Following our standup discussion, I'll leave this as-is for now. It should be fairly easy to update this later to swap out the shuffling transform based on an argument, or just replace it completely.

external/loaders/loaders/batches/_batch.py
```diff
@@ -74,11 +75,10 @@ def batches_from_geodata(
...
         data_mapping,
-        variable_names,
-        timesteps_per_batch,
-        random_seed,
-        timesteps,
-        res,
+        variable_names=variable_names,
+        timesteps_per_batch=timesteps_per_batch,
+        timesteps=timesteps,
+        res=res,
```
AnnaKwa, line 81: There are some cases (surface flux training data) where people use `batches_from_geodata` with `open_zarr` as the mapping function. Can you also pass the new `batches_from_mapper` args for `unstacked_dims, subsample_ratio, drop_nans` here and add them as optional args to `batches_from_geodata`?

Not relevant to this PR, but I was reminded of this issue I created noting that `batches_from_geodata` is somewhat redundant now that `batches_from_mapper` can return unstacked data, and could be removed (https://github.com/ai2cm/fv3net/issues/1430).

AnnaKwa (changes requested):
I just had one request for the new optional args to also be available in `batches_from_geodata`.

external/loaders/loaders/_utils.py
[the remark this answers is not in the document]
brianhenn: Oh good to know, still like having it not subsample when no subsampling is requested.

external/loaders/loaders/batches/_batch.py
mcgibbon: Updated. As you noted, it should be possible to write any config that uses `batches_from_geodata` to use `batches_from_mapper` instead. If it is, I would suggest doing that instead of adding these arguments to a config using `batches_from_geodata`.

AnnaKwa (approved):
Thanks for updating, LGTM.

external/loaders/loaders/batches/_batch.py
brianhenn: Sounds good

mcgibbon:
Got the go-ahead from @brianhenn to merge.

PR #1627, opened 2022-02-11
emulation: add dense-local architecture

nbren12:
Condensation is local to each (x,y,z) grid point.

Added public API:
- add name=`dense-local` option to ArchitectureConfig for emulation


frodre (approved):
LGTM

PR #1628, opened 2022-02-14
Prognostic run: Fill nans in added tendencies

brianhenn:
This PR fills nans in tendencies from the postphysics step (i.e., from nudging and ML) with zeros in the loops's function to add them to the state. Very occasionally coarsened restart files at C384 contain nans in the nudging reference state, which then crash the prognostic run when they propagate into the nudging tendencies and state, which this PR fixes.

Significant internal changes:
- Postphysics tendency nans are now filled with zeros before being added to state, and global fraction of filled tendencies for each variable is logged. 

- [X] Tests added


workflows/prognostic_c48_run/tests/test_nudging.py
```diff
@@ -190,6 +190,24 @@ def test_get_nudging_tendency(
         assert result[name].attrs["units"] == tendency.attrs["units"]
 
 
+def test_nudging_tendency_fillna(
```
nbren12, line 193: This is a lot of test code for the small change. It's a tricky integration to test, and the test is quite coupled to the implementation (e.g. sharing STATE_NAME_TO_TENDENCY). The test is also 100x longer than the added code. All this suggests the API/behavior being tested could be problematic. 

Maybe it would be easier test if we  always fillna when *applying* any kind of tendency? This would also have the benefit of working in other settings than the nudging. I'm not sure if there is a convenient `apply_tendency` function. If so, you should refactor the fillna there. Otherwise, I suggest just removing  this test.

nbren12 (changes requested):

workflows/prognostic_c48_run/tests/test_nudging.py
brianhenn: Moving to the `add_tendency` function as discussed.

nbren12 (approved):
Thanks! It might be nice to add a logging/warning saying which tendency name has NaNs.

brianhenn:
Thanks. That's a good idea, I originally had it logging the number of cells with nans at each level for each variable (like for the sphum limiter) but decided that was too much logging.

nbren12:
I agree per-level is a bit much. Maybe an overall fraction of NaN would be a better balance. It would be nice to know if a tendency is 100% NaN or 1%.

PR #1633, opened 2022-02-14
Fix fv3fit.keras.adapters.get_inputs

frodre:
The `_convert_dict_output` uses the function `get_inputs` to retrieve the input dictionary for the model, but this was untested.  It did not always preserve names for dictionary inputs.

 The bug is related to the fact that `tf.keras.Model.inputs` seems to return a list despite the input format. The previous implementation worked in the case where the layer names already correspond to the dict keys, but failed if input layers just used defaults (e.g., "layer_2").  The fix uses the `.input` model property instead of `.inputs`, but I can't seem to find any documentation on these model properties/attributes. 🙃

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/adapters.py
+++ b/external/fv3fit/fv3fit/keras/adapters.py
@@ -69,8 +69,14 @@ def ensure_dict_output(model: tf.keras.Model) -> tf.keras.Model:
 
 
 def get_inputs(model: tf.keras.Model) -> Mapping[str, tf.Tensor]:
-    if isinstance(model.inputs, Mapping):
-        return model.inputs
+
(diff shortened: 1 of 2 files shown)


nbren12 (approved):
Thanks for adding the test/finding the bug. I'm not if `.inputs` is documented either. If it is than maybe returning `{input.name: input for input in model.inputs}` would be more robust than using the undocumented `.input` attribute.

frodre:
After doing some more digging, I see `input` is a `tf.keras.layers.Layer` attribute

https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#attributes
> input:  Retrieves the input tensor(s) of a layer.
Only applicable if the layer has exactly one input, i.e. if it is connected to one incoming layer.

PR #1636, opened 2022-02-15
Use Predictor.input_sensitivity in offline diags and redo RF plot

AnnaKwa:
This follows up on PR https://github.com/ai2cm/fv3net/pull/1623 to use the new Predictor.input_sensitivity API in the offline diags. It cleans up a lot of the code in offline_diags._input_sensitivity and revamps the RF plotting function to work with the new format that is returned in the InputSensitivity for random forests.


workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -2,73 +2,18 @@
...
-) -> Dict[Hashable, Tuple[int, int]]:
 
-    variable_dims = _count_features_2d(variables, stacked, "sample")
-    start = 0
-    variable_indices = {}
-    for var, var_dim in variable_dims.items():
-        variable_indices[var] = (start, start + var_dim)
-        start += var_dim
-    return variable_indices
+RandomForestInputSensitivity = Mapping[str, Mapping[str, np.ndarray]]
```
brianhenn, line 15: Missing usage of this ... does it make sense to import `fv3fit.RandomForestInputSensitivity` for use here instead of having to redefine it again here and as `info` below, or does that overly couple this to fv3fit? Seems pretty reasonable I think.

AnnaKwa: Yeah, I think so- since this is already super coupled to fv3fit from having to extract the relevant structured information out of the `InputSensitivity`, if that coupling breaks it'd be easier to debug if the types were imported.

brianhenn (approved):
Thanks @AnnaKwa

PR #1643, opened 2022-02-17
Composable model blending

frodre:
This PR ads the ability to create a new model with blended outputs from two keras models given a masking function.  The blending is currently used in our experiments of using a separate for the Antarctic region.


frodre:
Example of how I use this to create a blended antarctic model.

https://github.com/ai2cm/fv3net/blob/andrep/antarctic-feb22/projects/microphysics/antarctic/create_blended.py

external/fv3fit/fv3fit/emulation/compositions/blended.py
```diff
@@ -0,0 +1,54 @@
...
+) -> tf.keras.Model:
+    """
+    Take two models and blend their outputs given a mask function.
+
+    The masking operation is broadcastable so take care to ensure
+    that large tensors aren't created by omitting a feature
+    dimension from the mask, when singleton feature fields exist.
+
+    Args:
+        model_a: model to use at points where mask == True
```
AnnaKwa, line 23: For more clarity could name these `model_mask_true/ model_mask_false`

external/fv3fit/fv3fit/emulation/compositions/blended.py
```diff
@@ -0,0 +1,54 @@
...
+        model_b: model to use at points where mask == False
+        mask_func: masking function determining model output selection.  the
+            function uses the model inputs and should return a mask tensor
```
AnnaKwa, line 26: For clarity can you instead say "the function uses the blended model's `inputs`"? At first I read this as the `mask_func` inputs had to be in the individual models' inputs, which I think is not the case e.g. the example script adds the mask func arg `latitude` to the inputs returned by `get_model_inputs`.

external/fv3fit/fv3fit/emulation/compositions/blended.py
```diff
@@ -0,0 +1,54 @@
...
+            with [sample, feature] or [sample, 1] dimensions
+        inputs: input tensors for the model, assumed to be
+            cover all necessary inputs for model_a and model_b
+    """
+
+    # avoid name conflicts in blended model
+    model_a._name = "ModelA"
```
AnnaKwa, line 33: Ditto about labeling with "mask_true/false" over "A/B"- it might make debugging easier.

AnnaKwa:
Looks good to me! Just a couple small comments about naming. I approved.

AnnaKwa (approved):
Looks good to me! Just a couple small comments about naming. I approved.

PR #1645, opened 2022-02-17
fv3fit: add range_config for limiting output values

AnnaKwa:
This PR adds a `RangeConfig` class which allows the training config to specify upper and/or lower bounds for individual  output variables. The immediate use case is for training models to predict both downward longwave surface flux and column shortwave transmissivity, where the former should >=0 and the latter should be in the range [0, 1]. 

Currently it is only implemented for the dense model (is new a hyperparameter option `DenseHyperparameters.range_config`).

ex.
```
hyperparameters:
  range_config:
    ranges:
        upper_bounded_variable:
          max: 0.
        lower_bounded_variable:
          min: 0.
        both_sides_bounded_variable:
          min: 0.
          max: 1.
```

I ended up implementing this with a custom activation rather than using the existing keras ReLU, because setting the `threshold` arg to negative values using the keras ReLU results in a discontinuity at threshold where the floor jumps back to zero (no error is raised if you set `threshold < 0` but this seems to be discouraged).

Added public API:
- `OutputLimtConfig`, `DenseHyperparameters.output_limit_config`

Refactored public API:
- breaking change: the hyperparameter `DenseHyperparameters.nonnegative_outputs` is removed. To achieve the same effect, use `DenseHyperparameters.output_limit_config` with variables' `min` field set to 0.

Internal changes:
- deleted the old deprecated `DenseModel` and related methods and tests. I was in the process of removing the `nonnegative_outputs` reference and decided to just take out everything that was not used anymore to be consistent.

- [x ] Tests added


external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -174,6 +177,8 @@ def build_model(
         config.input_variables,
     )
 
+    # note that n_features_out=1 is not used, as we connect subsequent layers
+    # to the last hidden output layer
```
mcgibbon, line 181: Thanks for adding a comment, suggesting here because it's documenting code I wrote.
```suggestion
    # note that n_features_out=1 is not used, as we want a separate output
    # layer for each variable (norm_output_layers)
```

external/fv3fit/fv3fit/keras/_models/models.py
```diff
@@ -136,98 +130,6 @@ def model(self) -> tf.keras.Model:
             raise RuntimeError("must call fit() for keras model to be available")
         return self._model
 
-    def _fit_normalization(self, X: np.ndarray, y: np.ndarray):
```
mcgibbon, line 139: You were able to delete so much of this! Is DenseModel used at all anymore? Or could it be deleted? If we haven't used it in a while I wouldn't be worried about backwards compatibility.

external/fv3fit/fv3fit/keras/_models/shared/output_range.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy as np
+import tensorflow as tf
+from typing import Optional, Union, Mapping, Hashable, Sequence
+
+
+Output = Union[np.ndarray, tf.Tensor]
+
+
+@dataclasses.dataclass
+class Range:
```
mcgibbon, line 11: Can you call this `OutputRange`, `Limits`, or something else to distinguish it from the `range` keyword?

external/fv3fit/fv3fit/keras/_models/shared/output_range.py
```diff
@@ -0,0 +1,71 @@
...
+                    output,
+                    x,
+                )
+        if self.max is not None:
+            x = tf.where(tf.math.greater_equal(output, self.max), zeros + self.max, x)
+        return x
+
+
+@dataclasses.dataclass
+class RangeConfig:
```
mcgibbon, line 51: You could decide to rename all "range" to "limit", to be consistent also with the names you chose in the method on this class.

mcgibbon (approved):
LGTM! Could consider renaming everything from "range" to "limit", but it's up to you.

external/fv3fit/fv3fit/keras/_models/models.py
AnnaKwa: It's not used in current experiments, so I think it's safe to delete.

external/fv3fit/fv3fit/keras/_models/shared/output_range.py
AnnaKwa: Updated all the names to `limit`, `OutputLimit`, etc.

PR #1648, opened 2022-02-18
loaders: fix extra kwarg

AnnaKwa:
This adds a `variable_names` arg to the  `batches_from_netcdf` function, as all calls of `load_batches` pass this arg. Only these variables will be loaded.

Adds if statement to skip calling `append_halos` (which requires unstacked data) in `XyMultiSequenceArray` if n_halos is 0  (which it is for stacked data).


mcgibbon (approved):
LGTM, thanks for the fix!

PR #1651, opened 2022-02-22
Recursive dependency check for ordered transforms

frodre:
A different design for transform dependency tracking that allows for cyclical naming, e.g., A -> B -> A or A -> A just requires A.  Tradeoff here is that it assumes transforms are correctly ordered, which is true since they're currently defined by the user.

Added public API:

Refactored public API:

Significant internal changes:

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/emulation/transforms/__init__.py
+++ b/external/fv3fit/fv3fit/emulation/transforms/__init__.py
@@ -2,10 +2,11 @@ from .factories import (
     ComposedTransformFactory,
     ConditionallyScaled,
     TransformedVariableConfig,
+    Difference,
 )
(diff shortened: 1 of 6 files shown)


frodre:
Closing in favor of #1652 since it achieves the same thing with a smaller footprint.  I was not certain that the recursive backwards names would support cycles given @nbren12 reply on #1647, but it can and it was just an issue with the name gathering in the composed transform.

frodre:
Closing in favor of #1652 since it achieves the same thing with a smaller footprint.  I was not certain that the recursive backwards names would support cycles given @nbren12 reply on #1647, but it can and it was just an issue with the name gathering in the composed transform.

PR #1652, opened 2022-02-22
emulation: simplify input name resolution

nbren12:
@frodre  was right to point out that some issues with #1647.

The DAG based logic of the ComposedTransformFactory isn't necessary and imposed unneeded constraints e.g. around circular dependencies of names. Since the compositions happen in a specific order, the solution is actually much simpler.

We can restore this logic if we ever make a DAG transform without this ordering.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/transforms/factories.py
+++ b/external/fv3fit/fv3fit/emulation/transforms/factories.py
@@ -1,6 +1,5 @@
-from collections import defaultdict
 import dataclasses
-from typing import Callable, MutableMapping, Optional, Sequence, Set
+from typing import Callable, Optional, Sequence, Set
 
 import tensorflow as tf
 from fv3fit.emulation.transforms.transforms import (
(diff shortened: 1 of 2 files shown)


frodre (approved):
Thanks for this update.  I had a similar solution, but yours is a much smaller footprint than #1651 since it makes use of `backwards_names`.

nbren12:
Yah. I went down a bit of a rabbit hole with the DAG logic. That is where I see the `ComposedTransform` going, but is not where the code currently is. Certainly it's easier to think about reversing a stack of transforms rather than reversing an entire dag.

Despite its confusing name, `backward_names` contains more information than explicitly listing inputs/outputs. For instance, `backward_names({"a"})` will list only the deps of "a", not all the deps of the full transform. The information content is equivalent for the single output transforms like `Difference` but not for the `ComposedTransform`. If we want to move to a DAG approach later then we'll need something equivalent to `backward_names`.

PR #1653, opened 2022-02-22
Fix end of epoch loss logging bug 

AnnaKwa:
Fixes bug where the training loop callback logged the losses at each timesteps batch step instead of just the end of the epoch.


mcgibbon (approved):
Thanks for the unit test!

PR #1654, opened 2022-02-23
Add cpu parameter to training argo workflow

oliverwm1:
This allows training ML models with more than one cpu using kubernetes. In some cases this can greatly speed training.

Added public API:
- `cpu` parameter to `training` argo workflow
- `training-cpu` parameter to end-to-end argo workflow


AnnaKwa (approved):
LGTM

PR #1655, opened 2022-02-23
Remove StackedBatches from random forest training

mcgibbon:
This PR removes the use of StackedBatches from random forest training, requiring the user configure stacking in the input data configuration.

Refactored public API:
- When using "sklearn_random_forest" training, the user must supply a data configuration that provides pre-stacked batches. If the first input has more than 2 dimensions, they will get a ValueError explaining this limitation.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/sklearn/_random_forest.py
+++ b/external/fv3fit/fv3fit/sklearn/_random_forest.py
@@ -42,6 +42,13 @@ def train_random_forest(
     train_batches: Sequence[xr.Dataset],
     validation_batches: Sequence[xr.Dataset],
 ):
+    """
+    Args:
+        hyperparameters: configuration for training
+        train_batches: batched data for training, must be stacked
+            with at most one non-sample dimension
+        validation_batches: ignored in this function
+    """
     model = RandomForest(
         hyperparameters.input_variables,
         hyperparameters.output_variables,
--- a/external/fv3fit/fv3fit/sklearn/_random_forest.py
+++ b/external/fv3fit/fv3fit/sklearn/_random_forest.py
@@ -24,7 +24,6 @@ from .._shared.config import RandomForestHyperparameters
 from .. import _shared
 from .._shared import (
     scaler,
-    StackedBatches,
     stack_non_vertical,
     match_prediction_to_input_coords,
     SAMPLE_DIM_NAME,
@@ -228,29 +227,34 @@ class SklearnWrapper(Predictor):
         return "SklearnWrapper(\n%s)" % repr(self.model)
 
     def _fit_batch(self, data: xr.Dataset):
-        x, _ = pack(data[self.input_variables], [SAMPLE_DIM_NAME], self.packer_config)
-        y, self.output_features_ = pack(data[self.output_variables], [SAMPLE_DIM_NAME])
+        first_input_dims = data[next(iter(self.input_variables))].dims
+        if len(first_input_dims) > 2:
+            raise ValueError(
+                "was given input data with more than 2 dimensions, "
+                "data should all be [sample] or [sample, feature]"
+            )
+        sample_dim_name = first_input_dims[0]
+        x, _ = pack(data[self.input_variables], [sample_dim_name], self.packer_config)
+        y, self.output_features_ = pack(data[self.output_variables], [sample_dim_name])
 
         if self.target_scaler is None:
-            self.target_scaler = self._init_target_scaler(data)
+            self.target_scaler = self._init_target_scaler(data, sample_dim_name)
 
         y = self.target_scaler.normalize(y)
         self.model.fit(x, y)
 
-    def _init_target_scaler(self, batch):
+    def _init_target_scaler(self, batch, sample_dim_name):
         return get_scaler(
             self.scaler_type,
             self.scaler_kwargs,
             batch,
             self._output_variables,
-            SAMPLE_DIM_NAME,
+            sample_dim_name,
         )
 
     def fit(self, batches: Sequence[xr.Dataset]):
         logger = logging.getLogger("SklearnWrapper")
-        random_state = np.random.RandomState(np.random.get_state()[1][0])
-        stacked_batches = StackedBatches(batches, random_state)
-        for i, batch in enumerate(stacked_batches):
+        for i, batch in enumerate(batches):
             logger.info(f"Fitting batch {i+1}/{len(batches)}")
             self._fit_batch(batch)
             logger.info(f"Batch {i+1} done fitting.")
--- a/external/fv3fit/fv3fit/sklearn/_random_forest.py
+++ b/external/fv3fit/fv3fit/sklearn/_random_forest.py
@@ -216,24 +216,29 @@ class SklearnWrapper(Predictor):
         self.scaler_type = scaler_type
         self.scaler_kwargs = scaler_kwargs or {}
         self.target_scaler: Optional[scaler.NormalizeTransform] = None
-        self._input_variables = input_variables
-        self._output_variables = output_variables
         self.packer_config = packer_config
         for name in self.packer_config.clip:
-            if name in self._output_variables:
+            if name in self.output_variables:
                 raise NotImplementedError("Clipping for ML outputs is not implemented.")
 
     def __repr__(self):
         return "SklearnWrapper(\n%s)" % repr(self.model)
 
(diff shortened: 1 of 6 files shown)


external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -228,29 +227,34 @@ def __repr__(self):
...
             batch,
             self._output_variables,
-            SAMPLE_DIM_NAME,
+            sample_dim_name,
         )
 
     def fit(self, batches: Sequence[xr.Dataset]):
         logger = logging.getLogger("SklearnWrapper")
-        random_state = np.random.RandomState(np.random.get_state()[1][0])
-        stacked_batches = StackedBatches(batches, random_state)
```
mcgibbon, line 252: Deleting this call to StackedBatches here is the one "main" code change in this PR. The changes to sample dimension name handling are a consequence of this - since stacking is done externally, we can't assume a particular sample dimension name, and instead want to assume the first dimension is the sample dimension.

external/fv3fit/tests/_regtest_outputs/test_sklearn_wrapper.test_predict_is_deterministic.out
```diff
@@ -1 +1 @@
-fe1edd06a4cdb3bbca2652fc980a4d65
```
mcgibbon, line 1: Had to update the shape of the data used for this test to be pre-stacked in training.

external/fv3fit/tests/training/test_main.py
```diff
@@ -387,6 +392,8 @@ def cli_main(args: MainArgs):
         + validation_args
         + local_download_args
     )
+    # if you need pdb support, temporarily replace the check_call above with this:
+    # fv3fit.train.main(args)
```
mcgibbon, line 396: Could delete this, but I found it handy, and it took a little while to figure out what to replace it with here.

external/fv3fit/tests/training/test_main.py
```diff
@@ -424,9 +431,18 @@ def test_cli(
...
     )
-    mock_load_batches.return_value = [config.mock_dataset for _ in range(6)]
-    cli_main(config.args)
-    fv3fit.load(config.args.output_path)
-    if use_local_download_path:
-        assert len(os.listdir(config.args.local_download_path)) > 0
+    if not use_local_download_path and model_type == "dense":
+        # dense requires caching if unstacked_dims is set as double-stacking
+        # a MultiIndex causes an error, can remove this block when dense training
+        # no longer stacks
```
mcgibbon, line 439: Note this exception only happens if no local download path is set, so we won't see this with argo in practice.

external/fv3fit/tests/training/test_train.py
```diff
@@ -323,24 +328,6 @@ def test_dump_and_load_default_maintains_prediction(model_type):
     xr.testing.assert_equal(loaded_result, original_result)
 
 
-@pytest.mark.parametrize("model_type", ["sklearn_random_forest"])
-def test_train_predict_multiple_stacked_dims(model_type):
```
mcgibbon, line 327: Another test already handles predicting with multiple dimensions, and we don't support training in this mode any more.

external/fv3fit/tests/test_pack.py
```diff
@@ -57,6 +57,19 @@ def test_roundtrip(
         np.testing.assert_array_equal(unpacked[name].values, ds[name].values)
 
 
+def test_pack_2d_and_3d():
```
mcgibbon, line 60: I thought this might be a bug reproducer but then it passed, I figure might as well keep it.

external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -228,29 +227,34 @@ def __repr__(self):
         return "SklearnWrapper(\n%s)" % repr(self.model)
 
     def _fit_batch(self, data: xr.Dataset):
-        x, _ = pack(data[self.input_variables], [SAMPLE_DIM_NAME], self.packer_config)
-        y, self.output_features_ = pack(data[self.output_variables], [SAMPLE_DIM_NAME])
+        first_input_dims = data[next(iter(self.input_variables))].dims
```
brianhenn, line 230: What if the latter input variables have different and erroneous dims?

external/fv3fit/fv3fit/sklearn/_random_forest.py
brianhenn, line 252: @mcgibbon I inadvertently deleted your comment here trying to delete my own :0, but I think you were basically just saying that this is the main point of the PR

brianhenn (commented):
@mcgibbon thanks. Is it in the scope of this PR (or another PR you're planning) to also remove the dependence on StackedBatches in the precipitative NN? Then we could delete StackedBatches entirely, and would (effectively) be down to one way of stacking data in all of our training workflows, which would be nice.

external/fv3fit/fv3fit/sklearn/_random_forest.py
mcgibbon: Yes, that is correct.

mcgibbon:
> @mcgibbon thanks. Is it in the scope of this PR (or another PR you're planning) to also remove the dependence on StackedBatches in the precipitative NN? Then we could delete StackedBatches entirely, and would (effectively) be down to one way of stacking data in all of our training workflows, which would be nice.

Yes, this is part of my next task on this line. I'll need to do this before updating everything to use tf datasets (whether it's in the same or different PR than that conversion.

external/fv3fit/tests/training/test_main.py
AnnaKwa: Actually, I think by default no local download path is passed in the template, so it's possible.

external/fv3fit/fv3fit/sklearn/_random_forest.py
mcgibbon: Then this won't catch that. Some possible consequences:
- If only some variables are stacked, the training will fail or give very strange results. I think it's unlikely we'll run into this situation.
- If variables are stacked using different stacked dimension names, the training will fail or give very strange results. Also unlikely.
- If all variables are unstacked, in practice all inputs will have at least 2 dimensions - time and tile, plus at least one spatial dimension, probably two more is the minimum in the case of surface variables. This will always be caught by checking only one variable.

I could update the logic to safeguard against the first two failure modes, if you think that would be helpful. It shouldn't be too expensive to do so, I just added the exception checks here as a better-than-nothing solution.

If you don't think I need to update this let me know, otherwise I will have it check the dimensions of every input and output variable.

AnnaKwa (approved):
This LGTM. Could you add to the docstring to indicate that the batches need to be stacked before being passed to .fit?

external/fv3fit/fv3fit/sklearn/_random_forest.py
brianhenn: OK makes sense. I think my comment had more to do with the readability of the code, which might be improved by adding those checks that do clarify that we care about all variables, but I'll leave that up to you.

brianhenn:
Once `StackedBatches` is gone is there any use for batches that _aren't_ stacked? ie should `unstacked_dims` become a required argument at that point?

external/fv3fit/fv3fit/sklearn/_random_forest.py
mcgibbon: That's a great point. I'll update it.

mcgibbon: Updated.

external/fv3fit/tests/training/test_main.py
mcgibbon: OK, in practice I haven't seen any `run.sh` examples where people haven't set this - should I post in slack about it? This issue should go away within a month or so.

mcgibbon:
> Once StackedBatches is gone is there any use for batches that aren't stacked? ie should unstacked_dims become a required argument at that point?

Good point, I should make that change when updating the rest of the models to require stacked batches.

external/fv3fit/tests/training/test_main.py
AnnaKwa: Yeah, that sounds fine.

PR #1656, opened 2022-02-23
Add positive value limiting transform to emulation

frodre:
This PR adds a new transform, `LimitValueTransform` that clips values outside of a given range to 0 on the call to `.backwards`.

Other API Changes:
* `UnivariateTransform` arguments switched to have an optional `to`.  If not provided, defaults to `source`
* Removed `enforce_positive` from `MicrophysicsConfig` and `fields.py`


external/fv3fit/fv3fit/emulation/transforms/transforms.py
```diff
@@ -71,8 +71,37 @@ def backward(self, x: tf.Tensor) -> tf.Tensor:
...
+    y := x
+    x := ReLU(y)
+
+    Attributes:
+        enforce_positive: used only for identifying this transform in the
+            configuration
+    """
+
+    # TODO: Is there a better way to do config instance via dacite w/o uniqueness?
+    enforce_positive: str = "yes_please"
```
frodre, line 88: Don't use `enforce_positive`, but I need something to differentiate it for instantiation via `dacite`.  Know a better way to handle it while maintaining the nice type-based dispatch?

external/fv3fit/fv3fit/emulation/transforms/transforms.py
```diff
@@ -71,8 +71,37 @@ def backward(self, x: tf.Tensor) -> tf.Tensor:
...
+
+    def __post_init__(self):
+        self.relu = tf.keras.layers.ReLU()
+
+    def forward(self, x: tf.Tensor) -> tf.Tensor:
+        return x
+
+    def backward(self, x: tf.Tensor) -> tf.Tensor:
+        return self.relu(x)
```
nbren12, line 97: Can use a function instead of the RelU layer since it is has no tunable parameters, best to avoid `__post_init__` if we can IMO.
 
```suggestion
    def forward(self, x: tf.Tensor) -> tf.Tensor:
        return x

    def backward(self, x: tf.Tensor) -> tf.Tensor:
        return tf.keras.activations.relu((x)
```

external/fv3fit/fv3fit/emulation/transforms/factories.py
```diff
@@ -148,3 +149,23 @@ def build(self, sample: TensorDict) -> ComposedTransform:
...
+    A convenience factory to apply PositiveTransform to
+    multiple fields with a single configuration entrypoint.
+
+    Attributes:
+        enforce_positive_on: List of variable names to apply limiter to
+    """
+
+    enforce_positive_on: Sequence[str]
+
+    def __init__(self, enforce_positive_on: Sequence[str]):
```
nbren12, line 166: I think the inheritance makes this code a bit magical, and subject to breaking if the `.factories` attribute changes for some reason. It's also odd/not correct for a `dataclass` to have a __init__, which is why I suspect you needed to manually patch in the redundant `enforce_positive_on` attribute below. This really should just be a function returning an `TransformedVariableConfig`, but unfortunately that probably won't work with dacite.

Can you delete the class? The convenience it provides seems small relative to the maintenance burden, and we can already configure the positive enforcement explicitly using the `UnivariateTransform`.

external/fv3fit/fv3fit/emulation/transforms/transforms.py
```diff
@@ -71,8 +71,37 @@ def backward(self, x: tf.Tensor) -> tf.Tensor:
...
+
+
+UnivariateCompatible = Union[LogTransform, PositiveTransform]
+
+
 class UnivariateTransform(TensorTransform):
-    def __init__(self, source: str, to: str, transform: LogTransform):
+    def __init__(self, source: str, to: str, transform: UnivariateCompatible):
```
nbren12, line 104: After deleting `EnforcePositiveValues` you can make this interface a bit more convenient by letting `to` default to `source`:
```suggestion
    def __init__(self, source: str, to: Optional[str] = None, transform: UnivariateCompatible):
        to = to or source
```

Then you can configure the positivity enforcment like this:
```
tensor_transforms:
- {source: air_temperature_after_gscond, transform: {enforce_positive: True}}
```

external/fv3fit/fv3fit/emulation/transforms/transforms.py
nbren12: Hmm. if we keep going down this road we'll need name based dispatch...but that starts to feel like even more overengineering. Maybe would be more attractive to make this a boolean or generalize the class to limit the outputs between `upper` and `lower` limits.

nbren12 (commented):
It gets the job done, but I think this PR highlights the downsides of exposing too much math to the yaml config and working around the fact that `yaml+dacite` is not a fully fledged programming language, so things like loops over variable names are not cleanly expressed. Perhaps a better alternative would be to rewrite our yaml configs in code so we can do things like

```
for name in positive_variables:
    transforms.append(FunctionalTransform(backward=tf.keras.activations.relu, source=name, to=name))
```

external/fv3fit/fv3fit/emulation/transforms/factories.py
frodre: > It's also odd/not correct for a dataclass to have a init, which is why I suspect you needed to manually patch in the redundant enforce_positive_on attribute below.

Yeh, 😅, I after starting to write the init, I changed it to be just a class, but then I didn't think I could still use `dacite` to initialize it.   My main goal was to have something easy to configure here, but yeah if it's confusing/hard to maintain I can see why other suggestions would be preferable.

external/fv3fit/fv3fit/emulation/transforms/transforms.py
frodre: > generalize the class to limit the outputs between upper and lower limits.

I like this. I'll give it a shot.

nbren12 (approved):
LGTM. Only possible suggestion is to add the limiter tranforms to `rnn.yaml`.

nbren12:
I would just merge this PR though to get it off your plate.

PR #1659, opened 2022-02-24
prognostic run: load prescriber data at each timestep

brianhenn:
This PR allows lazy loading of `Prescriber` data at each timestep during the prognostic run. Previously all timesteps' prescribed data were loaded at the outset of the run, which was fine for C48 but exceeded memory limits when using all but a few timesteps at C384. For the `Prescriber`, the`time_interpolate_func` decorator is now used instead of `xr.interp`, which is better for loading less data, more in line with other time interpolation (eg in the `TendencyPrescriber` and nudging stepper), and allowed for deleting some code. 

Refactored public API:
- `PrescriberConfig` now has optional `reference_initial_time` and `reference_frequency_seconds` arguments needed for time interpolation. This is a breaking change if time interpolation of the reference dataset is needed, because previously it happened automatically. 

Significant internal changes:
- `Prescriber` now loads data at each timestep and scatters within each tile, and is constructed with a new factory function (`get_prescriber`). A common function (`scatter_within_tile`) is now used to accomplish this shared task between the `Prescriber` and the `TendencyPrescriber`.


workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -62,11 +64,14 @@ def get_tendency_prescriber(
     else:
         prescriber_config = config.tendency_prescriber
         tendency_variables = list(prescriber_config.variables.values())
-        if communicator.rank == 0:
```
brianhenn, line 65: This was mainly to clean up from the last PR that the non-zero tile ranks don't need to all be opening the target dataset.

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -109,3 +104,33 @@ def step() -> Diagnostics:
 
         step.__name__ = func.__name__
         return step
+
+
+def scatter_within_tile(
```
AnnaKwa, line 109: Is there some module where this function could live that both this module and `prescriber.py` import from?

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -162,45 +98,6 @@ def get_momentum_diagnostics(self, state, tendency) -> Diagnostics:
...
-    ds = _open_ds(dataset_key, consolidated)
-    ds = get_variables(ds, variables)
-    if timesteps is not None:
-        ds = _time_interpolate_data(ds, timesteps, variables)
-    time_coord = ds.coords["time"]
-    return ds.drop_vars(names="time").load(), time_coord
-
-
-def _time_interpolate_data(ds, timesteps, variables):
-    vars_interp_nearest = [var for var in variables if var in INTERPOLATE_NEAREST]
```
AnnaKwa, line 181: Just to confirm, this functionality for discrete values is covered in the `time_interpolation_func`?

AnnaKwa (approved):
LGTM, I had a couple minor comments. Mainly I want to confirm that the nearest neighbor interpolation will still happen for discrete variables i.e. land sea mask.

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
```diff
@@ -109,3 +104,33 @@ def step() -> Diagnostics:
...
+    """
+    if communicator.tile.rank == 0:
+        state: State = time_lookup_function(time)
+    else:
+        state = {}
+
+    tile = communicator.partitioner.tile_index(communicator.rank)
+    if communicator.tile.rank == 0:
+        ds = xr.Dataset(state).isel(tile=tile)
+    else:
```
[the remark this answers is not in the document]
brianhenn: True, looks like it can be dropped since the `Communicator` handles generating the dict for the receiving ranks.

workflows/prognostic_c48_run/runtime/transformers/tendency_prescriber.py
brianhenn: Yeah, we could make a new `scatter` module for this, don't think there's an existing one that makes sense. But probably a good idea in order to keep the import dag clean.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
brianhenn: Correct, this refactor means it no longer has to be implemented twice.

PR #1660, opened 2022-02-24
Update integration test config to stack training data

AnnaKwa:
The previous PR https://github.com/ai2cm/fv3net/pull/1655 changed the RF training to take in stacked data. The integration test config needs to be updated to stack the training data. It also needs to use a cache directory when training to avoid errors that occur when trying to stack data that is already stacked with a multi index. The latter issue will be resolved in the future when tf datasets are used in training.

Breaking change:
Since our current train-diags-prog argo workflow currently uses the same configuration file for validation during training and offline diags, there'd have to be a new distinction between the validation-config.yaml passed to the training workflow vs the test data config passed to the offline diags. 

This isn't that big of a deal as the using validation data in training can be optional, but it'd be a breaking change in the sense the parameter formerly known as validation-data-config would now be optional and only be used for training time validation, and a new parameter test-data-config would have to be provided for the offline diags.

If this is approved, I'll make those changes to vcm-workflow-control as well.

### DIFF
--- a/tests/end_to_end_integration/run_test.sh
+++ b/tests/end_to_end_integration/run_test.sh
@@ -92,10 +92,16 @@ yq -y --arg data_path $n2fDataPath '.mapper_config.kwargs.data_path|=$data_path'
     training-data-config.yaml > \
     training-data-config-compiled.yaml
 
+yq -y --arg data_path $n2fDataPath '.mapper_config.kwargs.data_path|=$data_path' \
+    test-data-config.yaml > \
+    test-data-config-compiled.yaml
+
+
 deployWorkflows "$registry" "$commit"
 argo submit argo.yaml -p bucket="${bucket}" -p project="${project}" \
     -p training-data-config="$(< training-data-config-compiled.yaml)" \
     -p validation-data-config="$(< training-data-config-compiled.yaml)" \
+    -p test-data-config="$(< test-data-config-compiled.yaml)" \
     -p tag="${tag}" --name "$name"
 
 trap "argo logs \"$name\" | tail -n 100" EXIT
--- a/workflows/argo/training.yaml
+++ b/workflows/argo/training.yaml
@@ -18,7 +18,7 @@ spec:
         parameters:
           - name: training_config
           - name: training_data_config
-          - name: validation_data_config
+          - {name: validation_data_config, value: " "}
           - name: output
           - {name: cpu, value: 1000m}
           - {name: memory, value: 6Gi}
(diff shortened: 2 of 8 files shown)


workflows/argo/training.yaml
```diff
@@ -18,7 +18,7 @@ spec:
         parameters:
           - name: training_config
           - name: training_data_config
-          - name: validation_data_config
+          - {name: validation_data_config, value: " "}
```
AnnaKwa, line 21: I made this optional to reflect that the training script does not need this as an input, and often times users may not provide a separate validation config to the `train-diags-prog` workflow in addition to the test data config. Previously the same data config was used both as train time validation data and offline evaluation, but that is no longer possible.

oliverwm1: Thanks this will be useful. The RF training script wastes time preprocessing the validation data which it doesn't end up using during training.

workflows/argo/training.yaml
```diff
@@ -56,12 +56,18 @@ spec:
             echo "Validation Data Configuration:"
             cat validation_data.yaml
 
+            if grep -q '[^[:space:]]' "validation_data.yaml";
```
oliverwm1, line 59: Is this tested? What would happen if user passes some string other than `" "` for this argument?

tests/end_to_end_integration/run_test.sh
```diff
@@ -92,10 +92,16 @@ yq -y --arg data_path $n2fDataPath '.mapper_config.kwargs.data_path|=$data_path'
     training-data-config.yaml > \
     training-data-config-compiled.yaml
 
+yq -y --arg data_path $n2fDataPath '.mapper_config.kwargs.data_path|=$data_path' \
+    test-data-config.yaml > \
+    test-data-config-compiled.yaml
```
oliverwm1, line 97: `test-data-config-compiled.yaml` should be added to `.gitignore`

oliverwm1 (commented):
It is unwieldy to have three data config yamls, but at the same time we might want this flexibility if doing a proper train/validation/test split, especially for NN training. So I think I agree this is the right approach, as long as we have good working examples.

Is there a way we can reduce the copy-and-paste between the three configs? In particular, do we expect the `mapper_config` part to ever change? I guess it's possible we'd want to test on a different dataset than what was trained on.

workflows/argo/training.yaml
AnnaKwa: The same would happen as for if a bad train or test data config was passed- the `validation-data-config` arg would be passed to the python call and then some error would be raised when converting it to a DataConfig object. I tested the if statement in a simple test case of passing a "config" through the command line and verifying that it either did not add a validation arg for an empty string passed, or added the validation arg if a string was passed.
```
#!/bin/bash
validation_config=$1


cat <<EOF >validation.yaml
$validation_config
EOF

if grep -q '[^[:space:]]' "validation.yaml";
then
     validation_arg="--validation-data-config validation.yaml"
else
     validation_arg=" "
fi

echo "validation arg: $validation_arg"
```

AnnaKwa:
> Is there a way we can reduce the copy-and-paste between the three configs? In particular, do we expect the mapper_config part to ever change?
Yes, sometimes- if the preprocessing and saving is done beforehand, and the training data points to some preexisting subsampled netcdfs, then the train config will use `batches_from_netcdf` while the test data config will probably still use `batches_from_mapper` (since the offline diags wants unstacked data).

oliverwm1 (approved):
LGTM. Happy to review a vcm-workflow-control PR also.

PR #1661, opened 2022-02-24
Add LRUCache option to n2f mapper

AnnaKwa:
This adds the option to use LRU caches when using the nudge to fine mapper. This may or may not speed up your usage of the mapper, depending on your chunking and the sequence in which you are accessing the various chunks. Testing this in a not-cache-size-limited case for opening 3 batches of 2 timesteps each in the same chunk, using the cache was about 40% faster than without and avoided repeated downloads.

Added public API:
- Optional arg `cache_size_mb` for `open_nudge_to_fine`, if provided will create a `zarr.storage.LRUCache` of this size for each zarr dataset that the mapper has to open. By default this is `None`, which reproduces the current behavior.

### DIFF
--- a/external/loaders/loaders/mappers/_nudged/_nudged.py
+++ b/external/loaders/loaders/mappers/_nudged/_nudged.py
@@ -125,7 +125,7 @@ def open_nudge_to_fine(
         "nudging_tendencies.zarr",
         "state_after_timestep.zarr",
     ),
-    cache_size_mb: float = None,
+    cache_size_mb: Optional[float] = None,
 ) -> XarrayMapper:
     """
     Load nudge-to-fine data mapper for use with training. Merges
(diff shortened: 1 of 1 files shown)


external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -124,6 +126,7 @@ def open_nudge_to_fine(
         "nudging_tendencies.zarr",
         "state_after_timestep.zarr",
     ),
+    cache_size_mb: float = None,
```
mcgibbon, line 129: Should be Optional if default is None (how does this pass mypy checks?), can also be integer since this gets cast to an integer anyways when it's used, and I doubt we'd want to specify fractional MB.
```suggestion
    cache_size_mb: Optional[int] = None,
```

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -207,12 +216,22 @@ def open_nudge_to_fine_multiple_datasets(
 
 
 def _get_datasets(
-    url: str, sources: Sequence[str], consolidated: bool = True
+    url: str,
+    sources: Sequence[str],
+    consolidated: bool = True,
+    cache_size_mb: float = None,
```
mcgibbon, line 222: Should have an `Optional` type hint, or you can make this required as it's a local function and you're always passing this argument.

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -207,12 +216,22 @@ def open_nudge_to_fine_multiple_datasets(
...
+        if cache_size_mb is not None:
+            mapper = zarr.LRUStoreCache(
+                fsspec.get_mapper(os.path.join(url, f"{source}")),
+                max_size=int(cache_size_mb * 1e6),
+            )
+            ds = xr.open_zarr(mapper, consolidated=consolidated)
+        else:
+            ds = intake.open_zarr(
+                os.path.join(url, f"{source}"), consolidated=consolidated
+            ).to_dask()
```
mcgibbon, line 235: Can you reduce the number of differences between these two pathways? e.g. you should be able to use the mapper returned by get_mapper directly when not using a cache, with the only change being that you wrap that mapper in a cache layer if caching is enabled. Currently the code pathways are very different, using (xarray+fsspec)/intake and using dask/maybe-non-dask(?) datasets.

mcgibbon (approved):
Really impressive speed up number! Thanks!

PR #1664, opened 2022-02-28
Ordered subsample option for batches

AnnaKwa:
Adds a `keep_ordered_samples` option to `batches_from_mapper` to not shuffle samples within stacked batches. This can speed up loading by a lot, depending on the number of timesteps per batch and the data chunking. 
ex. preprocessing about half the timesteps in the PIRE dataset took >3-4 hrs, with this option enabled and using the LRUCache option for `batches_from_mapper` it takes about 20-25 min.

Adds the `in_memory` option to `batches_from_netcdf`. Existing copies of each batch aren't getting garbage collected after use, so memory usage builds up and eventually crashes during training. The future usage of tf dataset may fix this. In the meantime since I believe the preprocessed data should fit in memory, we can use this to speed up training and prevent OOM'ing.

(Delete unused sections)
Added public API:
- `keep_ordered_samples` option in `batches_from_mapper`. Defaults to False.
- `in_memory` option for `batches_from_netcdf`. Defaults to False.

Internal change:
- Deleted the `select_first_samples` function and replaced it with `select_fraction` so it could be used for either case of samples being shuffled/ not shuffled. `select_fraction` returns a random subset that can be drawn from the entire dataset, but keeps its original order.

- [x] Tests added

### DIFF
--- a/external/loaders/loaders/_utils.py
+++ b/external/loaders/loaders/_utils.py
@@ -71,7 +71,6 @@ def stack(unstacked_dims: Sequence[str], ds: xr.Dataset) -> xr.Dataset:
     return ds_stacked.transpose(SAMPLE_DIM_NAME, *unstacked_dims)
 
 
-@curry
 def sort_by_time(ds: xr.Dataset) -> xr.Dataset:
     return ds.sortby("time")
 
--- a/external/loaders/loaders/_utils.py
+++ b/external/loaders/loaders/_utils.py
@@ -71,6 +71,18 @@ def stack(unstacked_dims: Sequence[str], ds: xr.Dataset) -> xr.Dataset:
     return ds_stacked.transpose(SAMPLE_DIM_NAME, *unstacked_dims)
 
 
+@curry
+def select_random_ordered_subset(fraction: float, ds: xr.Dataset) -> xr.Dataset:
+    if fraction == 1.0:
+        out = ds
+    else:
+        n_samples = len(ds[SAMPLE_DIM_NAME])
+        n_samples_keep = int(fraction * n_samples)
+        sample_indices = sorted(np.random.choice(n_samples, n_samples_keep))
+        out = ds.isel({SAMPLE_DIM_NAME: sample_indices})
+    return out
+
+
 def shuffle(ds: xr.Dataset) -> xr.Dataset:
     shuffled_indices = np.arange(len(ds[SAMPLE_DIM_NAME]), dtype=int)
     np.random.shuffle(shuffled_indices)
--- a/external/loaders/loaders/_utils.py
+++ b/external/loaders/loaders/_utils.py
@@ -72,13 +72,21 @@ def stack(unstacked_dims: Sequence[str], ds: xr.Dataset) -> xr.Dataset:
 
 
 @curry
-def select_random_ordered_subset(fraction: float, ds: xr.Dataset) -> xr.Dataset:
+def sort_by_time(ds: xr.Dataset) -> xr.Dataset:
+    return ds.sortby("time")
+
+
+@curry
+def select_fraction(fraction: float, ds: xr.Dataset) -> xr.Dataset:
+    # subselects a random fraction of samples, preserving original order.
     if fraction == 1.0:
         out = ds
     else:
         n_samples = len(ds[SAMPLE_DIM_NAME])
         n_samples_keep = int(fraction * n_samples)
-        sample_indices = sorted(np.random.choice(n_samples, n_samples_keep))
+        sample_indices = sorted(
+            np.random.choice(n_samples, n_samples_keep, replace=False)
+        )
         out = ds.isel({SAMPLE_DIM_NAME: sample_indices})
     return out
 
@@ -97,21 +105,6 @@ def dropna(ds: xr.Dataset) -> xr.Dataset:
     if len(ds[SAMPLE_DIM_NAME]) == 0:
         raise ValueError("Check for NaN fields in the training data.")
     return ds
-
-
-@curry
-def select_first_samples(fraction: float, ds: xr.Dataset) -> xr.Dataset:
-    """
-    Return a dataset with only the given fraction of samples from the
-    input dataset, taking the first samples.
-    """
-    if fraction == 1.0:
-        out = ds
-    else:
-        n_samples = len(ds[SAMPLE_DIM_NAME])
-        n_samples_keep = int(fraction * n_samples)
-        out = ds.isel({SAMPLE_DIM_NAME: slice(0, n_samples_keep)})
-    return out
 
 
 @curry
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -199,7 +199,7 @@ def batches_from_mapper(
     batch_func = compose_left(*transforms)
 
     seq = Map(batch_func, batched_timesteps)
-    seq.attrs["times"] = timesteps
+    seq.attrs["times"] = shuffled_times
 
     if in_memory:
         out_seq: Batches = tuple(ds.load() for ds in seq)
--- a/external/loaders/loaders/batches/_batch.py
+++ b/external/loaders/loaders/batches/_batch.py
@@ -1,5 +1,6 @@
 import logging
 from loaders.typing import Batches
+import numpy as np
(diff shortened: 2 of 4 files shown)


external/loaders/loaders/_utils.py
```diff
@@ -71,6 +71,18 @@ def stack(unstacked_dims: Sequence[str], ds: xr.Dataset) -> xr.Dataset:
     return ds_stacked.transpose(SAMPLE_DIM_NAME, *unstacked_dims)
 
 
+@curry
+def select_random_ordered_subset(fraction: float, ds: xr.Dataset) -> xr.Dataset:
```
nbren12, line 75: This function is pretty short, but does 2-3 things and could be hard to test: computes `n_samples`, . shuffle, orders, selects, and it has a hard coded dependence on `SAMPLE_DIM_NAME` that will make future refactors more difficult. I would try to make these more composable. Would something like this be equally performant?
```
batches.map(shuffle).map(select_fraction(fraction)).map(sort_by_time)
```

nbren12 (commented):
Amazing speedup! I'm glad this strategy worked.

I realize the diff is small, but I'm a bit concerned about the lack of composability of the new interfaces. My heuristic is that it would be hard to write a test for the `select_random_ordered_subset` because it does many things. Note the name has 3-4 verbs in it.

Also some of these changes affect the default behavior and could be answer changing.

external/loaders/loaders/batches/_batch.py
```diff
@@ -158,8 +161,7 @@ def batches_from_mapper(
 
     if timesteps is None:
         timesteps = list(data_mapping.keys())
-    shuffled_times = np.random.choice(timesteps, len(timesteps), replace=False).tolist()
-    batched_timesteps = list(partition_all(timesteps_per_batch, shuffled_times))
+    batched_timesteps = list(partition_all(timesteps_per_batch, timesteps))
```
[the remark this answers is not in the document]
AnnaKwa: Reverted this to the original shuffling behavior, if `shuffle_samples=False` then the times are sorted in the list of transforms.

external/loaders/tests/test__batch.py
```diff
@@ -286,3 +286,22 @@ def test_batches_from_netcdf(tmpdir):
     )
     for ds_saved, ds_loaded in zip(saved_batches, loaded_batches):
         xr.testing.assert_equal(ds_saved, ds_loaded)
+
+
+def test_batches_from_mapper_stacked_data_is_not_shuffled():
```
mcgibbon, line 291: Does the test `test_batches_from_mapper_stacked_data_is_shuffled` need to be modified/deleted?

external/loaders/loaders/batches/_batch.py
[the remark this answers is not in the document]
mcgibbon: Originally I suggested not having an option for this to keep the logic simple, since we have no use case that needs timestamp shuffling in the batches pipeline (always need to shuffle at train time, anyways).

I mostly think this should be consistent, so I'm fine with either of these:
- Add an option for shuffle_timesteps similar to shuffle_samples, because we want to be able to toggle shuffling options.
- Remove the keep_ordered_samples/shuffle_samples argument, because we don't need to be able to toggle shuffling options.

external/loaders/loaders/_utils.py
mcgibbon: @nbren12 we can't `shuffle` at the start because the random subset must have its indices in-order for dask access. Shuffle-then-order would break performance. tfdataset might do something smarter here, but our sequences code won't.

This could use a test. I'd suggest selecting from an artificially-constructed dataset where each sample has a specific value, then asserting that the result has 1) the correct number of samples, 2) no repeated sample values, and 3) monotonically-increasing sample values.

nbren12: > @nbren12 we can't shuffle at the start because the random subset must have its indices in-order for dask access.

I see, but can we just always use the `in_memory` behavior and put a `.load` as the very first transform? Seems unnecessary to have a lazy object (Batches) which contain lazy objects (dask arrays).

nbren12: isel with random indices is only an issue with dask...so maybe just stop using dask?

external/loaders/loaders/batches/_batch.py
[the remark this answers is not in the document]
nbren12: > Add an option for shuffle_timesteps similar to shuffle_samples, because we want to be able to toggle shuffling options.

This was my suggestion above, so I agree that we should give a separate flag. 

With tf.data.Dataset we will still want to shuffle timesteps, since `.shuffle` only mixes things within a given buffer size. Mixing 3PM and 12PM of the same day together is not as much shuffling as mixing different days.

external/loaders/loaders/_utils.py
mcgibbon: I think what you're proposing (prepending a `load` transform, instead of removing shuffling) would work.

If you go this route, we don't need the "sort by time" transform at the end, this was solely to improve performance.

external/loaders/tests/test__batch.py
AnnaKwa: Since I kept `shuffle_samples` and `shuffle_timesteps` as options, I didn't change that test.

external/loaders/loaders/_utils.py
AnnaKwa: If `shuffle_samples` is True, the shuffle occurs after sorting by time.

`select_random_ordered_subset` is deleted and broken into a short `sort_by_time` function and `select_fraction`. The latter replaces the `select_first_samples` function and has the suggested tests added.

AnnaKwa:
Now this has both shuffle_timesteps and shuffle_samples args but the default behavior should be the same as before (timesteps are still all shuffled, samples are shuffled).

There is a small change in behavior for subsampling stacked data since I replaced select_first_samples (or whatever it was called) with select_fraction which selects out of the entire dataset but keeps the order. But since I think the only experiments that would change are Oli's current RF training, and this effect is not nearly the same as changing the timestep shuffling before batching, I thought that was ok.

external/loaders/loaders/batches/_batch.py
```diff
@@ -158,8 +165,14 @@ def batches_from_mapper(
...
         timesteps = list(data_mapping.keys())
-    shuffled_times = np.random.choice(timesteps, len(timesteps), replace=False).tolist()
-    batched_timesteps = list(partition_all(timesteps_per_batch, shuffled_times))
+
+    if shuffle_timesteps:
+        final_timesteps = np.random.choice(
+            timesteps, len(timesteps), replace=False
+        ).tolist()
+    else:
+        final_timesteps = timesteps
```
mcgibbon, line 174: We realized in slack that there's no scenario where we'd want to leave the timesteps un-shuffled, so this can be reverted to the previous version.

external/loaders/loaders/batches/_batch.py
```diff
@@ -121,6 +122,8 @@ def batches_from_mapper(
     unstacked_dims: Optional[Sequence[str]] = None,
     subsample_ratio: float = 1.0,
     drop_nans: bool = False,
+    shuffle_timesteps: bool = True,
+    shuffle_samples: bool = True,
```
mcgibbon, line 126: If you're changing the numbers output anyways by going from shuffle-select_first to select_fraction, I'd suggest setting shuffle_samples to False by default since this is what we expect to want to use in essentially every case.

As noted in other comment, shuffle_timesteps can be True always and removed as an option.

mcgibbon (commented):
From my side looks good to merge as-is, gave a suggestion to change a default boolean option and remove an option we're unlikely to change.

This doesn't implement the solution @nbren12 asked for though, so we should maybe check with him before merging.

external/loaders/loaders/batches/_batch.py
nbren12: I'm confused. Isn't `shuffle_timesteps=False` required to iterate through the data sequentially...which in combination with the LRUCache accounts for like 90% of the speedup in the PR?

mcgibbon: To get this merged let's say keep the shuffle_timesteps (default True) option in. If there isn't a use case for it, it just won't get used.

nbren12 (approved):

PR #1665, opened 2022-02-28
Fix backwards names to remove intermediate variable

### LINKED ISSUE #1658: Emulation: intermediate variables are present in `config.model_variables`
I am running into an issue when configuring a basic emulation model w/ a difference-transformed variable on master.  

```
Exception has occurred: KeyError
'temperature_gscond_difference'

During handling of the above exception, another exception occurred:
```

I was wondering if there a was a reason we stopped removing the "to" variable name from the `backwards_names` return value (e.g., for [Difference](https://github.com/ai2cm/fv3net/blob/dfaa394707dde2e0b18b6cf70f8a803a7308660a/external/fv3fit/fv3fit/emulation/transforms/transforms.py#L35))?  That would be consistent with converting an intermediate variable to dependencies in the `requested_names`, I think. 

Right now it provides all the variables back in `requested_names` (e.g., `before`, `after`, `to`) and the xarray lookup fails in the initial data loading when trying to find the `to` variable name.

@nbren12 Let me know when you get back.  In the interim I'll adjust so it works how I think it should.

frodre:
Some updates to the transform dependency tracking included intermediate variables in the output.  This PR reverts that behavior so we are correctly requesting the basic set of variables to fulfill all the specified transforms.

resolves #1658

### DIFF
--- a/external/fv3fit/fv3fit/emulation/transforms/factories.py
+++ b/external/fv3fit/fv3fit/emulation/transforms/factories.py
@@ -99,10 +99,12 @@ class ConditionallyScaled(TransformFactory):
 
     def backward_names(self, requested_names: Set[str]) -> Set[str]:
         """List the names needed to compute ``self.to``"""
-        new_names = (
-            {self.condition_on, self.source} if self.to in requested_names else set()
-        )
-        return requested_names.union(new_names)
+
+        if self.to in requested_names:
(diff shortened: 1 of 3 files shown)


nbren12 (approved):
LGTM. Thanks for the quick fix.

PR #1666, opened 2022-03-01
Loaders: C96/C384 data batches + add C384 wind rotation to catalog

brianhenn:
This PR allows C96/C384 data to be loaded as batches for training with derived grid variables. Batches have a grid associated that allows cos_zenith_angle, wind rotation etc., to be computed, but this only worked for C48 resolution because it assumed C48 grid metadata, and the C384 wind rotation matrices were missing from the catalog.

Significant internal changes:
- Adds C384 wind rotation data to catalog.


AnnaKwa (approved):
LGTM. I know the fortran got updated but it may be worthwhile to check the distribution of wind coefficients just to be extra sure they're between [-1, 1] and not [-0.5, 0.5].

brianhenn:
Thanks I knew there was a gotcha there but I couldn't remember what it was. [Looks like we're good](https://github.com/ai2cm/explore/blob/master/brianh/2022-02-28-c384-wind-rotation-check/2022-02-28-c384-wind-rotation-check.ipynb).

PR #1667, opened 2022-03-01
Best RNN configuration so far

frodre:
We've had a series of tuning experiments including input changes and conditional loss scaling adjustments to improve our offline RNN performance.  This PR commits our current best config for the RNN to the training directory.


nbren12 (approved):
LGTM

PR #1668, opened 2022-03-01
fv3fit training: use data config's `in_memory` for cached data

AnnaKwa:
~The last PR that added an `in_memory` arg to the batches from netcdf loader to address a memory usage issue in training, in which garbage collection does not delete existing batches when reopening the netcdfs.~ Edit- this was not the cause of the memory issue, see below comment about garbage collection.

However, the value of this arg in the config didn't actually get passed to the call to `batches_from_netcdf` within `get_cached_data`. This PR adds a config field `cache` to the `TrainingConfig`, ex.
```
cache:
    in_memory: True
    local_download_path: some_dir
```
**Breaking change**: The command line arg `--local-download-path` is removed, and the training config field `cache.local_download_path` is used instead to provide this information. It can still be provided via the command line, e.g. `--cache.local_download_path your_download_dir`. I will update the vcm-workflow-config repo to reflect this.

It turned out that even with the data loaded into memory, jobs would crash with OOM errors multiple epochs into the training with memory usage accumulating steadily. (ignore the "used" value, the cursor is hovering over a time range outside the training job)
image

I don't know why the garbage collection was not freeing up memory in the training loop, but when I added a call to `gc.collect()` at the end of each epoch in the training loop the memory usage remains constant over epochs.
image

### DIFF
--- a/external/fv3fit/fv3fit/train.py
+++ b/external/fv3fit/fv3fit/train.py
@@ -57,6 +57,19 @@ def get_data(
     variable_names: Sequence[str],
     in_memory: bool = False,
 ) -> Tuple[loaders.typing.Batches, loaders.typing.Batches]:
+    """
+    Args:
+        training_data_config: configuration of training data
+        validation_data_config:  if provided, configuration of validation data. If None,
+            an empty list will be returned for validation data.
+        local_download_path: if provided, cache data locally at this path
+        variable_names: names of variables to include when loading data
+        in_memory: if True and local_download_path is also set, batches will be
+            returned as a tuple of eagerly-loaded Datasets. Has no effect if
+            local_download_path is not set.
+    Returns:
+        Training and validation data batches
+    """
     if local_download_path is None:
         return get_uncached_data(
             training_data_config=training_data_config,
--- a/external/fv3fit/fv3fit/train.py
+++ b/external/fv3fit/fv3fit/train.py
@@ -61,6 +61,7 @@ def get_data(
     validation_data_config: Optional[str],
     local_download_path: Optional[str],
     variable_names: Sequence[str],
+    in_memory: bool = False,
 ) -> Tuple[loaders.typing.Batches, loaders.typing.Batches]:
     if local_download_path is None:
         return get_uncached_data(
@@ -74,6 +75,7 @@ def get_data(
             validation_data_config=validation_data_config,
             local_download_path=local_download_path,
             variable_names=variable_names,
+            in_memory=in_memory,
         )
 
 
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/train.py
```diff
@@ -61,6 +61,7 @@ def get_data(
     validation_data_config: Optional[str],
     local_download_path: Optional[str],
     variable_names: Sequence[str],
+    in_memory: bool = False,
 ) -> Tuple[loaders.typing.Batches, loaders.typing.Batches]:
```
mcgibbon, line 65:
```suggestion
    in_memory: bool = False,
) -> Tuple[loaders.typing.Batches, loaders.typing.Batches]:
    """
    Args:
        training_data_config: configuration of training data
        validation_data_config: if given, configuration of validation data. If not given,
            an empty list will be returned for validation data.
        local_download_path: if given, cache the data locally at this path
        variable_names: names of variables to include int he batches
        in_memory: if True and local_download_path is set, batches will be
            returned as a tuple of eagerly-loaded Datasets. Has no effect if
            local_download_path is not set.
```

external/fv3fit/fv3fit/train.py
```diff
@@ -156,6 +162,7 @@ def main(args, unknown_args=None):
         args.validation_data_config,
         args.local_download_path,
         variable_names=training_config.variables,
+        in_memory=training_data_config.kwargs.get("in_memory", False),
```
mcgibbon, line 165: I think you might need to move this to a top-level configuration option. Whether the BatchesLoader should use in-memory is a separate question from whether the cached batches_from_netcdf data should be kept in-memory. When the full dataset doesn't fit in memory, we'd need the batches config to have `in_memory=False` even if we want the cached (probably subsampled) data files to load with `in_memory=True`.

There are a couple ways you can present this in the config. You could group local_download_path and in_memory into their own `CacheConfig` dataclass. If you keep them as separate flat options, I'd suggest renaming `in_memory` to something that mentions the cache, so it's different from the BatchesLoader kwarg option.

AnnaKwa: Good point. I removed the command line arg `--local-download-path` and there is now a `CacheConfig` class with attributes `in_memory` and `local_download_path`.

external/fv3fit/fv3fit/train.py
```diff
@@ -42,12 +42,6 @@ def get_parser():
             "by default an empty sequence is used"
         ),
     )
-    parser.add_argument(
-        "--local-download-path",
-        type=str,
-        default=None,
-        help=("path for downloading data before training"),
-    )
```
mcgibbon, line 50: Make sure people are aware of the change they need to make to `run.sh` or similar because of this.

mcgibbon (approved):
Thanks for the update! Perplexed that `gc.collect()` is the solution, I wish we knew why.

PR #1670, opened 2022-03-02
add duration_seconds wandb metric for prognostic runs

nbren12:
This is the most important scalar metric we evaluate, but we didn't have it in wandb yet.

See this example: https://wandb.ai/ai2cm/microphysics-emulation/runs/3ar6kqen/overview?workspace=user-nbrenowitz


oliverwm1 (approved):
LGTM.

PR #1671, opened 2022-03-02
Make TransformConfig more testable

nbren12:
Creating a Pipeline object allows for a "seam" in the code that we can test the TransformConfig object.

Note: not strictly needed for any feature...I did the refactor while working on a classifier + regression model approach for microphysics ML, but realized that I didn't actually need to modify the Transform code. Still I think it's a helpful refactor that gives us some more coverage and makes the design easier to extend in the future.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/config.py
+++ b/external/fv3fit/fv3fit/emulation/data/config.py
@@ -1,14 +1,32 @@
 import dataclasses
 import logging
-from typing import Dict, Mapping, Optional, Set
+from typing import Any, Callable, Dict, List, Mapping, Optional, Set
 
 import dacite
+import xarray
 from fv3fit._shared import SliceConfig
-from toolz.functoolz import compose_left
+from toolz.functoolz import pipe
 
 from . import transforms
 
 logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class Pipeline:
+    xarray_transforms: List[Callable[[xarray.Dataset], xarray.Dataset]]
+    array_like_transforms: List[Any]
+    use_tensors: bool = True
+
+    def __call__(self, dataset: xarray.Dataset) -> Mapping[str, Any]:
+        transformed_dataset = pipe(dataset, *self.xarray_transforms)
+
+        if self.use_tensors:
+            array_like = transforms.to_tensors(transformed_dataset)
+        else:
+            array_like = transforms.to_ndarrays(transformed_dataset)
+
+        return pipe(array_like, *self.array_like_transforms)
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/emulation/data/config.py
```diff
@@ -1,16 +1,34 @@
...
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class Pipeline:
+    xarray_transforms: List[Callable[[xarray.Dataset], xarray.Dataset]]
+    array_like_transforms: List[Any]
+    use_tensors: bool = True
+
+    def __call__(self, dataset: xarray.Dataset) -> Mapping[str, Any]:
```
mcgibbon, line 21: It likely doesn't matter for the use cases here, but because this has bundled xarray and ndarray transforms, making the output and input datatypes different, you won't be able to do things like chain pipelines from different sources. If we do ever need to do that, it probably makes sense to use `pipe` directly with sequences of callables.

Not sure how this will interplay with the fv3fit.train changes, since we no longer have `xarray` representations in training functions.

mcgibbon: I'm also trying to be careful to mark routines that work on tensors versus routines that work on arrays of tensors. I think we can achieve this pretty seamlessly with compute methods operating on tensors, and wrapper methods that convert these to callables that operate on mappings. I was surprised to see `to_tensors` gives a mapping and that `array_like_transforms` operate on dicts.

e.g. I wrote this, used as `dataset.map(apply_to_mapping(ensure_feature_dim))`:
```python
@curry
def apply_to_mapping(tensor_func: Callable[[tf.Tensor], tf.Tensor], data: Mapping[str, tf.Tensor]) -> Dict[str, tf.Tensor]:


def ensure_feature_dim(tensor: tf.Tensor) -> tf.Tensor:
    """
    Given a tensor that may or may not be a scalar, return a tensor with at least
    one dimension, casting scalars to length-1 one-dimensional tensors.
    """
    if len(tensor.shape) == 0:
        return tensor[None]
    else:
        return tensor
```

mcgibbon (approved):
LGTM, no issues. I wrote some thoughts about tfdataset transformations from what I'm working on currently.

external/fv3fit/fv3fit/emulation/data/config.py
nbren12: I definitely agree with your comment. The type conversion from dataset to TensorDict makes this not composable.

`array_like_transforms` overlaps in responsibility with the composable/reversible [TensorTransform] (https://github.com/ai2cm/fv3net/blob/51eb52b5b86b9a9225383cd0b6e39245c631f280/external/fv3fit/fv3fit/emulation/transforms/transforms.py#L9) concept used for e.g. temperature scaling, so could be moved there. At this point the `Pipeline` could probably be deleted since it won't do much.

PR #1672, opened 2022-03-02
Increase default run length to 30 days

nbren12:
Some of our configurations complete the 10 day simulations, so we need an upper
bound on how long they run.

* add scripts for convenient running ./run and ./run_gscond_only
* remove tf-model argo parameter in favor of yq modification of config
* Submit argo jobs

### DIFF
--- a/projects/microphysics/argo/run
+++ b/projects/microphysics/argo/run
@@ -12,4 +12,4 @@ argo submit prog-run-and-eval.yaml \
     --name "$tag" \
     -p image_tag="$image" \
     -p tag=$tag \
-    -p config=$(base64 --wrap 0 config.yaml) | tee -a submitted-jobs.txt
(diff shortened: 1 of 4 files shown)


projects/microphysics/argo/run
```diff
@@ -0,0 +1,15 @@
...
+
+yq  --arg url "$url" '.zhao_carr_emulation.model.path |= $url' ../configs/default.yaml > config.yaml
+trap 'rm config.yaml' EXIT
+
+
+argo submit prog-run-and-eval.yaml \
+    --name "$tag" \
+    -p image_tag="$image" \
+    -p tag=$tag \
+    -p config=$(base64 --wrap 0 config.yaml) | tee -a submitted-jobs.txt
```
frodre, line 15: add the `grep -v 'config:'` to this one as well

frodre (approved):
LGTM.  Thanks for the example... looks like I should take another look at `yq`!

PR #1674, opened 2022-03-02
loaders: delete unnecessary `batches_from_geodata`, `diagnostic_batches_from_geodata`

AnnaKwa:
See issue https://github.com/ai2cm/fv3net/issues/1430

`loaders._batches` is cluttered with a lot of code for functions that we no longer use (`batches_from_geodata, diagnostic_batches_from_geodata`). This PR removes them. These were originally written because `batches_from_mapper` only returned stacked data, and the offline diagnostics used unstacked data. This is no longer the case since the `unstacked_dims` arg for `batches_from_mapper` allows the data to be either stacked or unstacked. 

Potential breaking change: There may still be some usages around of `batches_from_geodata` where the `open_zarr` function is used. However, `batches_from_mapper` can also open zarr data using `open_zarr`, so these data configurations can easily be edited to use `batches_from_mapper` instead.

### DIFF
--- a/external/loaders/tests/test__config.py
+++ b/external/loaders/tests/test__config.py
@@ -227,12 +227,30 @@ def test_batches_loader_from_dict(data, expected_class):
     assert type(result) is expected_class
 
 
-def test_safe_dump_data_config():
+def test_safe_dump_BatchesFromMapperConfig():
+    """
+    Test that dataclass.asdict and pyyaml can be used to save BatchesFromMapperConfig.
+    """
+    config = loaders.BatchesFromMapperConfig(
+        function="batches_from_mapper",
+        kwargs={"timesteps": ["1", "2", "3"]},
+        mapper_config={},
+    )
+    with tempfile.TemporaryDirectory() as tmpdir:
+        filename = os.path.join(tmpdir, "config.yaml")
+        with open(filename, "w") as f:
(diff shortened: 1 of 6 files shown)


workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -245,9 +245,6 @@ def transform(ds):
 def _get_data_mapper_if_exists(config):
     if isinstance(config, loaders.BatchesFromMapperConfig):
         return config.load_mapper()
-    elif isinstance(config, loaders.BatchesConfig):
```
AnnaKwa, line 248: `open_zarr` can still be used with `batches_from_mapper`.

external/loaders/tests/test__config.py
```diff
@@ -210,14 +208,6 @@ def test_load_batches_from_mapper_raises_if_registered_with_wrong_decorator():
 @pytest.mark.parametrize(
     "data, expected_class",
     [
-        pytest.param(
```
brianhenn, line 213: Should we replace this case with a working example of a `BatchesConfig`? Otherwise it's hard to tell what this test is trying to do

brianhenn (approved):
Thanks @AnnaKwa

PR #1675, opened 2022-03-02
`train.py`: CLI can update any TrainingConfig field

AnnaKwa:
Previously, the CLI interface for updating the training configuration could only update args that already existing in the configuration, i.e. fields that were not specified in the config could not be changed from their default values. It would be convenient though to be able to do so. This PR allows `TrainingConfig` entries to be updated through the CLI even if they are not in the original file.

This should also fix the currently failing integration test, as it is failing because it tries to set `TrainingConfig.cache.local_download_path` through the CLI but that field is not in the config file.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/config.py
+++ b/external/fv3fit/fv3fit/_shared/config.py
@@ -230,6 +230,17 @@ def to_nested_dict(d: dict):
     return new_config
 
 
+# Small modification to the arg parser so that the error raised by
+# providing invalid args is clearer.
+class ArgumentError(Exception):
+    pass
+
+
+class ArgumentParser(argparse.ArgumentParser):
+    def error(self, message):
+        raise ArgumentError(message)
+
+
 def get_arg_updated_config_dict(args: Sequence[str], config_dict: Dict[str, Any]):
     """
     Update a configuration dictionary with keyword arguments through an ArgParser.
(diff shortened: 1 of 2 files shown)


brianhenn (commented):
While this fix will help, can `[get_arg_updated_config_dict]`(https://github.com/ai2cm/fv3net/blob/df39ca8a1347d2982f4b457e1f04df6fca1165ba/external/fv3fit/fv3fit/_shared/config.py#L233) raise a more informative error in the case that someone passes an actual bad arg? [Right now it raises an error](https://app.circleci.com/pipelines/github/ai2cm/fv3net/11155/workflows/5979f2ef-80a2-423d-8729-003d8cd2ad56/jobs/98371?invite=true#step-105-8088) that suggests the usage of the `train` module is just whatever is in the config_dict file instead of the actual correct usage of `train`.

brianhenn:
Apparently argparse makes it somewhat difficult to catch exceptions, but overriding its error function is one possible solution: https://stackoverflow.com/a/16942165/10729865

We could replace the current bad arg error message with just a simple "Bad training config command line argument passed."

AnnaKwa:
Thanks, that's a useful patch for handling argparse errors.

This is the exception that is now raised if you provide a bad arg:
```
  File "/Users/annak/opt/miniconda3/envs/fv3net/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/annak/opt/miniconda3/envs/fv3net/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/annak/Code/fv3net/external/fv3fit/fv3fit/train.py", line 203, in <module>
    main(args, unknown_args)
  File "/Users/annak/Code/fv3net/external/fv3fit/fv3fit/train.py", line 158, in main
    args=unknown_args, config_dict=config_dict
  File "/Users/annak/Code/fv3net/external/fv3fit/fv3fit/_shared/config.py", line 262, in get_arg_updated_config_dict
    raise e
  File "/Users/annak/Code/fv3net/external/fv3fit/fv3fit/_shared/config.py", line 260, in get_arg_updated_config_dict
    updates = parser.parse_args(args)
  File "/Users/annak/opt/miniconda3/envs/fv3net/lib/python3.7/argparse.py", line 1758, in parse_args
    self.error(msg % ' '.join(argv))
  File "/Users/annak/Code/fv3net/external/fv3fit/fv3fit/_shared/config.py", line 241, in error
    raise ArgumentError(message)
fv3fit._shared.config.ArgumentError: unrecognized arguments: --cache.local_download_pathfdfd train-data-download-dir`

Which no longer shows the big `train.py [-h] ...` with the imcomplete list of args.

PR #1679, opened 2022-03-03
Fix prognostic run docker-compose build

oliverwm1:
Since #1572, calling `make build` within `workflows/prognostic_c48_run` results in an error because the `BASE_IMAGE` arg is not defined. This PR defines the `BASE_IMAGE` as `ubuntu:20.04` in the `docker-compose.yaml`.

### DIFF
--- a/workflows/prognostic_c48_run/docker-compose.yml
+++ b/workflows/prognostic_c48_run/docker-compose.yml
@@ -8,8 +8,7 @@ services:
       dockerfile: "docker/prognostic_run/Dockerfile"
       target: bld
       args:
-        # the tag doesn't matter since we bind-mount over these sources anyways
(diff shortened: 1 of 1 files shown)


nbren12 (approved):
no objections, but I didn't even recall that `workflows/prognostic_c48_run/Makefile` exists and my workflow is quite different. It's a bit confusing to have two different ways of building this image.

99% of the time using `:latest` works for me bc the fv3net code is all bind-mounted in.

The 1% of the time I really have to rebuild the docker image, usually do something like this `VERSION=latest make build_image_prognostic_run` from the project root.

nbren12:
>recently I was getting stuck in an image in which pace.util was not installed for the prog run dev environment.

Seems like either your ":latest" image was out of date which could be solved with `docker pull` or the docker-compose.yml does not use the latest image.

> We could remove the Makefile within workflows/prognostic_c48_run to keep everything at the top level, like the enter_emulation rule.

I think this would be helpful. I've actually written this target several times to help develop various PRs. I'd also lean towards defining a single top-level docker-compose file for the same reason.

PR #1680, opened 2022-03-04
Catalog: update PIRE 3d verification

AnnaKwa:
The dataflow job that processed the restarts into zarr format, which was subsequently used to create the pressure interpolated verification data, had about 80 failures that resulted in those timesteps not being filled with data in the final zarr. When these corrupted times are used in the verification, the resulting diagnostic plots have weird large values. Since the dataflow job is costly to rerun, and might not save 100% of the timesteps correctly, I opted to drop these timesteps from the existing verifcation as they were randomly distributed in time.


external/vcm/vcm/catalog.yaml
```diff
@@ -665,8 +665,9 @@ sources:
 
   2020_1yr_pire_pressure_interpolated:
     description: >-
-      3D dynamical core diagnostics variables from 40-day nudged simulation (May 27 2020 SHiELD), coarsened to C48 resolution and interpolated to pressure levels.
-      Generated with https://github.com/VulcanClimateModeling/explore/blob/master/annak/2021-06-17-save-c3072-verification-correct-winds.ipynb
```
spencerkclark, line 669: It might be good to leave the trail of notebooks used to create this for provenance.

spencerkclark (approved):

PR #1683, opened 2022-03-05
Replace predict_columnwise instances with predict

oliverwm1:
The `predict_columnwise` method on the `Predictor` class has been deprecated for 7 months. This PR removes this method (which is currently just a wrapper around `.predict`) and changes all of its uses to directly use the `predict` method.


AnnaKwa (approved):
LGTM, thanks!

PR #1684, opened 2022-03-07
Prognostic run: run-fv3gfs choose node pool

brianhenn:
This PR selects either `climate-sim-pool` or `ultra-sim-pool` for running fv3 segments in the `run-fv3gfs` workflow template depending on the request cpus. Any and only requests of greater than 24 CPUs (i.e., for greater than 2x2 tile layout processes) get sent to the `ultra-sim-pool`.

Significant internal changes:
- The `run-fv3gfs` workflow has an additional internal step to choose the node pool based on cpu.


oliverwm1 (approved):
LGTM

PR #1687, opened 2022-03-07
Modify ClipConfig API to remove dimension, always clip last dim

mcgibbon:
When we move to tfdatasets, we do not have access to dimension names. In practice we always want to clip the last dimension of an array. This PR modifies clipping configuration to remove the dimension name, and always assume we are clipping the last dimension.

The only training function with more than one non-sample dimension which could possibly receive data with an incorrect dimension ordering is the convolutional model. This model already assumes incoming data has the correct dimensionality, so the clipping assumption of dimension ordering doesn't add any additional danger.

We may consider adding code later to enforce that loader functions return tfdatasets with data of the correct dimensionality.

Refactored public API:
- The `clip` attribute of `PackerConfig` and `ClipConfig` is now `Mapping[Hashable, SliceConfig]`, was `Mapping[Hashable, Mapping[str, SliceConfig]]`.

Significant internal changes:
- refactored clipping functionality to use configuration as `Mapping[Hashable, SliceConfig]`, was `Mapping[Hashable, Mapping[str, SliceConfig]]`.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/dense.py
+++ b/external/fv3fit/fv3fit/keras/_models/dense.py
@@ -50,7 +50,7 @@ class DenseHyperparameters(Hyperparameters):
             each output variable.
         save_model_checkpoints: if True, save one model per epoch when
             dumping, under a 'model_checkpoints' subdirectory
-        clip_config: configuration of dataset packing.
+        clip_config: configuration of input and output clipping
         output_limit_config: configuration for limiting output values.
     """
 
--- a/external/fv3fit/fv3fit/keras/_models/dense.py
+++ b/external/fv3fit/fv3fit/keras/_models/dense.py
@@ -50,7 +50,7 @@ class DenseHyperparameters(Hyperparameters):
             each output variable.
         save_model_checkpoints: if True, save one model per epoch when
             dumping, under a 'model_checkpoints' subdirectory
-        clip_config: configuration of input and output clipping
+        clip_config: configuration of input and output clipping of last dimension
         output_limit_config: configuration for limiting output values.
     """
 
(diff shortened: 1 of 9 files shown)


external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -50,7 +50,7 @@ class DenseHyperparameters(Hyperparameters):
             each output variable.
         save_model_checkpoints: if True, save one model per epoch when
             dumping, under a 'model_checkpoints' subdirectory
-        clip_config: configuration of dataset packing.
+        clip_config: configuration of input and output clipping
```
AnnaKwa, line 53: Can you add to this to mention it is always the last/feature dim?

external/fv3fit/fv3fit/keras/_models/shared/clip.py
```diff
@@ -20,21 +19,12 @@ class ClipConfig(PackerConfig):
             dimension to clip.
     """
 
-    clip: ClipDims = dataclasses.field(default_factory=dict)
```
AnnaKwa, line 23: Can you update the docstring above as well? (it says it is 'usually the last dim')

AnnaKwa (approved):
Thanks, LGTM. Still, I feel a bit wary of having the clipping rely on an assumed dimension ordering that is out of its control since clipping along the wrong dimension could be a silent bug that is hard to catch. I do realize that we already have several other places in the training that rely on dimension ordering to be correct from the loaders though. I don't know where the best place to document the alphabetical dimension ordering behavior would be, but it would be useful to have that communicated somewhere- before our discussion I didn't know this was the case.

Could you add to the PR description to note that the training functions that might make use of this already rely on the assumption that the feature dimension is last?

external/fv3fit/fv3fit/keras/_models/dense.py
mcgibbon: Updated.

external/fv3fit/fv3fit/keras/_models/shared/clip.py
mcgibbon: Updated.

PR #1689, opened 2022-03-07
Remove output_metadata from PureKerasModel

mcgibbon:
When moving to tfdatasets in fv3fit.train, we will no longer have access to xarray metadata during training. This PR modifies PureKerasModel to use the list of unstacked dimensions instead of output_metadata.

Refactored public API:
- output_metadata is removed as an argument to PureKerasModel
- unstacked_dims is now a required argument for PureKerasModel (previously optional)
- unstacked_dims argument to PureKerasModel has changed in meaning. Previously it accepted a list of all dimensions you would accept as unstacked dimensions, in any order. Now it must be exactly the list of dimensions used in outputs, where the last dimension (commonly the vertical dimension) may be excluded for some variables.

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/convolutional.py
+++ b/external/fv3fit/fv3fit/keras/_models/convolutional.py
@@ -16,7 +16,6 @@ import logging
 from fv3fit.keras._models.shared.utils import (
     standard_denormalize,
     full_standard_normalized_input,
-    get_stacked_metadata,
 )
 
 logger = logging.getLogger(__file__)
@@ -91,11 +90,6 @@ def train_convolutional_model(
         train_data = tuple(train_data)
     X, y = train_data[0]
     train_model, predict_model = build_model(hyperparameters, X=X, y=y)
-    output_metadata = get_stacked_metadata(
-        names=hyperparameters.output_variables,
-        ds=train_batches[0],
-        unstacked_dims=UNSTACKED_DIMS,
-    )
     del train_batches
     hyperparameters.training_loop.fit_loop(
         model=train_model, Xy=train_data, validation_data=validation_data
@@ -103,9 +97,8 @@ def train_convolutional_model(
     predictor = PureKerasModel(
         input_variables=hyperparameters.input_variables,
         output_variables=hyperparameters.output_variables,
-        output_metadata=output_metadata,
         model=predict_model,
-        unstacked_dims=("x", "y", "z", "z_interface"),
+        unstacked_dims=("x", "y", "z"),
         n_halo=n_halo,
     )
     return predictor
--- a/external/fv3fit/fv3fit/keras/_models/precipitative.py
+++ b/external/fv3fit/fv3fit/keras/_models/precipitative.py
@@ -386,4 +386,5 @@ class PrecipitativeModel:
             # predictor has additional diagnostic outputs which were indirectly trained
             list(self.output_variables),
             self._predict_model,
+            unstacked_dims=("z",),
         )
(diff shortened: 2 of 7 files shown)


external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -91,21 +90,15 @@ def train_convolutional_model(
...
     hyperparameters.training_loop.fit_loop(
         model=train_model, Xy=train_data, validation_data=validation_data
     )
     predictor = PureKerasModel(
         input_variables=hyperparameters.input_variables,
         output_variables=hyperparameters.output_variables,
-        output_metadata=output_metadata,
         model=predict_model,
-        unstacked_dims=("x", "y", "z", "z_interface"),
+        unstacked_dims=("x", "y", "z"),
```
brianhenn, line 101: Were there any variables on the z_interface dimension we would want to use to begin with? I can't think of any.

external/fv3fit/fv3fit/keras/_models/precipitative.py
```diff
@@ -400,6 +385,5 @@ def predictor(self) -> PureKerasModel:
             self.input_variables,
             # predictor has additional diagnostic outputs which were indirectly trained
             list(self.output_variables),
-            get_output_metadata(self.output_packer, SAMPLE_DIM_NAME),
```
brianhenn, line 403: Don't you need to provide `unstacked_dims = ('z',)` here?

brianhenn: Oh I see, `unstacked_dims` has that as a default. 

Nit: Can the calls of `PureKerasModel` be consistent in whether they provide this argument (e.g, in the dense model it does but here it doesn't)? Would make the intended usage clearer. Maybe don't have a default for unstacked_dims so that people reading the code will see what's passed for them to help understand what's happening?

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -37,9 +36,8 @@ def __init__(
         self,
         input_variables: Iterable[Hashable],
         output_variables: Iterable[Hashable],
-        output_metadata: Iterable[Dict[str, Any]],
         model: tf.keras.Model,
-        unstacked_dims: Optional[Sequence[str]] = None,
+        unstacked_dims: Sequence[str] = ("z",),
```
brianhenn, line 40: See above about whether this should have a default.

brianhenn (approved):
Thanks @mcgibbon, approving with an optional suggestion

external/fv3fit/fv3fit/keras/_models/convolutional.py
mcgibbon: I don't think so. Even if we made a flux-based model that internally produces cell-interface variables, we'd still convert those to cell-center tendencies before making them outputs of the model. If this becomes an issue we can add that functionality when we need it.

external/fv3fit/fv3fit/keras/_models/precipitative.py
mcgibbon: Great suggestion. I made unstacked_dims a required argument.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
mcgibbon: Updated to required argument.

PR #1690, opened 2022-03-07
Small changes to improve PureKerasModel serialization

mcgibbon:
Two functions in the precipitative model are turned into layers, so that later I can look at the configuration of a save-loaded model and compare it to the configuration of starting model to ensure PureKerasModel is serializing correctly.

A get_config method is added to a normalization layer which takes an initialization argument, so it can be used within PureKerasModel. Without this method, keras complains it requires this method to serialize the layer.

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/precipitative.py
+++ b/external/fv3fit/fv3fit/keras/_models/precipitative.py
@@ -37,6 +37,15 @@ Q_TENDENCY_NAME = "dQ2"
 
 class IntegratePrecipLayer(tf.keras.layers.Layer):
     def call(self, args) -> tf.Tensor:
+        """
+        Args:
+            dQ2: rate of change in moisture in kg/kg/s, negative corresponds
+                to condensation
+            delp: pressure thickness of atmospheric layer in Pa
+
+        Returns:
+            precipitation rate in kg/m^2/s
+        """
         dQ2, delp = args  # layers seem to take in a single argument
         # output should be kg/m^2/s
         return tf.math.scalar_mul(
@@ -49,6 +58,14 @@ class IntegratePrecipLayer(tf.keras.layers.Layer):
 
 class CondensationalHeatingLayer(tf.keras.layers.Layer):
     def call(self, dQ2: tf.Tensor) -> tf.Tensor:
+        """
+        Args:
+            dQ2: rate of change in moisture in kg/kg/s, negative corresponds
+                to condensation
+
+        Returns:
+            heating rate in degK/s
+        """
         return tf.math.scalar_mul(tf.constant(-LV / CPD, dtype=dQ2.dtype), dQ2)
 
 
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/keras/_models/precipitative.py
```diff
@@ -35,27 +35,21 @@
...
-    return tf.math.scalar_mul(
-        # dQ2 * delp = s-1 * Pa = s^-1 * kg m-1 s-2 = kg m-1 s-3
-        # dQ2 * delp / g = kg m-1 s-3 / (m s-2) = kg/m^2/s
-        tf.constant(-1.0 / GRAVITY, dtype=dQ2.dtype),
-        tf.math.reduce_sum(tf.math.multiply(dQ2, delp), axis=-1),
-    )
-
+class IntegratePrecipLayer(tf.keras.layers.Layer):
+    def call(self, args) -> tf.Tensor:
+        dQ2, delp = args  # layers seem to take in a single argument
```
frodre, line 40: What was the error when using multiple args?  I've used layers before with specific arguments for each tensor and I think it worked.

https://github.com/ai2cm/fv3net/blob/d4924b40618a363a5ad27a98b6fb410f541d7a48/external/fv3fit/fv3fit/emulation/layers/fields.py#L89

external/fv3fit/fv3fit/keras/_models/precipitative.py
```diff
@@ -35,27 +35,21 @@
...
-    """
-    Args:
-        dQ2: rate of change in moisture in kg/kg/s, negative corresponds
-            to condensation
 
-    Returns:
-        heating rate in degK/s
-    """
-    return tf.math.scalar_mul(tf.constant(-LV / CPD, dtype=dQ2.dtype), dQ2)
+class CondensationalHeatingLayer(tf.keras.layers.Layer):
```
frodre, line 50: I would keep the documentation describing the specifics of the values used/output by this layer.

external/fv3fit/fv3fit/keras/_models/precipitative.py
frodre: Descriptions are embedded in comments here.  Make consistent with `CondensationalHeatingLayer`

frodre (approved):
LGTM.  Minor comments to check out the multiple argument input and make the layer documentation consistent.

external/fv3fit/fv3fit/keras/_models/precipitative.py
mcgibbon: There wasn't an error, I just (back when I wrote the original code here, the comment is pre-existing) noticed keras seems to have a "style" where most layers take in a list of inputs, instead of separate arguments.

Thanks for catching the dropped docstrings, added docstrings to both layers.

external/fv3fit/fv3fit/keras/_models/precipitative.py
mcgibbon: Restored, thanks.

PR #1693, opened 2022-03-08
Use temporary directory for emulation cache during tests

mcgibbon:
Currently running the fv3fit tests leaves a `.data` directory used for caching by the emulation code which does not get cleaned up. This PR modifies the fv3fit tests to use a temporary directory instead.

### DIFF
--- a/external/fv3fit/tests/conftest.py
+++ b/external/fv3fit/tests/conftest.py
@@ -2,6 +2,8 @@ import numpy as np
 import pytest
 import requests
 import xarray as xr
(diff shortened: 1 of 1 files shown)


mcgibbon:
I ran pytest tests/emulation and confirmed it never created a .data directory during or after the tests.

nbren12 (approved):
LGTM. Nice to not pollute the git repo with this. We should probably make this configurable with an env var like the fv3config cache...which works pretty well IMO.

PR #1695, opened 2022-03-09
fv3fit: Fix dropped derived outputs bug

AnnaKwa:
This fixes a bug in `DerivedModel` that dropped derived outputs if an existing `DerivedModel` was wrapped again in a `DerivedModel` (this can happen in the offline diags, which add Q2 as a derived output).

Added a test for this.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -1,4 +1,4 @@
-from typing import Iterable, Set, Hashable, List
+from typing import Iterable, Set, Hashable, Sequence
 import fsspec
 import yaml
 import os
@@ -16,7 +16,7 @@ class DerivedModel(Predictor):
     _BASE_MODEL_SUBDIR = "base_model_data"
 
     def __init__(
-        self, model: Predictor, derived_output_variables: List[Hashable],
+        self, model: Predictor, derived_output_variables: Sequence[Hashable],
     ):
         """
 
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -31,11 +31,16 @@ def __init__(
         """
         # if base_model is itself a DerivedModel, combine the underlying base model
         # and combine derived attributes instead of wrapping twice with DerivedModel
-        self.base_model: Predictor = model.base_model if isinstance(
-            model, DerivedModel
-        ) else model
+        if isinstance(model, DerivedModel):
+            self.base_model: Predictor = model.base_model
+            existing_derived_outputs = model._derived_output_variables  # type: ignore
```
AnnaKwa, line 36: mypy complains that it doesn't know the type of the RHS, even though this is within an `isinstance` block.

oliverwm1 (approved):
Thanks for the fix. The mypy issue feels like a bit of a hint that maybe allowing DerivedModels to wrap other DerivedModels is overly complex. It seems easy for bugs to crop up. I wonder if there are other approaches we could take—like just updating the list of derived variables instead of wrapping the model in a whole new one.

AnnaKwa:
>  if there are other approaches we could take—like just updating the list of derived variables instead of wrapping the model in a whole new one

Effectively that is what it does (well, it was supposed to, now it actually does that). My thinking behind the wrapping is that the `DerivedModel` is itself a `Predictor` and the model that it wraps is a `Predictor`, so it seemed like `DerivedModel(some_derived_model)` should be allowed. If you think the API should be clearer, we could add a method like `new_derived_model = some_derived_model.add_derived_outputs(...)` which could get used in the wrapping case as well.

PR #1699, opened 2022-03-14
Re-enable using multiple datasets in offline report

spencerkclark:
This PR makes changes to re-enable creating an offline report using a test dataset with a `"dataset"` dimension.  As before, the metrics displayed are computed in the aggregate across all datasets represented (i.e. not on a dataset by dataset basis).  

Significant internal changes:
- When a `"dataset"` dimension exists in the batches, r2 metrics are computed both in the aggregate and for each dataset.  The aggregate r2 metrics are tagged with just `"_r2"` in their names, while the per dataset metrics are tagged with `"_per_dataset_r2"`.  This allows the aggregate metrics to be easily displayed in the report.
- Something similar is done for the `"_bias"` metrics.  This is required so that scalar versions end up appropriately in the `scalar_metrics.json` file required by the report.
- Other metrics are purely saved on a dataset by dataset basis, and are averaged across datasets after the diagnostics are read in when the report is generated.

An example report can be found [here](https://storage.googleapis.com/vcm-ml-public/spencerc/2021-01-28/n2f-25km-offline-reports/tq-nn-snoalb-tapered-25-clipped-updated-pr-seed-0/index.html).  The corresponding diagnostics files can be found in this directory: `gs://vcm-ml-experiments/spencerc/2022-01-20-nudge-to-25-km-ml-models/offline_diags/tq-nn-snoalb-tapered-25-clipped-updated-pr-seed-0`.  If we read in the `offline_diagnostics.nc` from that directory and look at a per dataset r2 variable, we get something like this:

```
>>> ds.column_integrated_dq1_per_dataset_r2_2d_global
<xarray.DataArray 'column_integrated_dq1_per_dataset_r2_2d_global' (dataset: 4)>
array([0.169668, 0.181235, 0.182179, 0.184461])
Coordinates:
  * dataset  (dataset) object 'Minus 4 K' 'Unperturbed' 'Plus 4 K' 'Plus 8 K'
```

- [x] Tests added

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
@@ -11,7 +11,7 @@ from vcm.cloud import gsutil
 from vcm.catalog import catalog
 
 DELP = "pressure_thickness_of_atmospheric_layer"
-
+DATASET_DIM_NAME = "dataset"
 
 UNITS = {
     "column_integrated_dq1": "[W/m2]",
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
@@ -58,7 +58,7 @@ def is_3d(da: xr.DataArray, vertical_dim: str = "z"):
 
 def _compute_r2(ds_metrics: xr.Dataset, tag: str = "_r2") -> xr.Dataset:
     """Compute r2 values from MSE and variance metrics.
-    
+
     The names of the resulting variables use the provided tag.
     """
     mse_vars = [var for var in ds_metrics if "_mse" in var]
@@ -72,7 +72,7 @@ def _compute_r2(ds_metrics: xr.Dataset, tag: str = "_r2") -> xr.Dataset:
 
 def insert_r2(ds_metrics: xr.Dataset) -> xr.Dataset:
     """Insert r2 fields into the metrics Dataset.
-    
+
     Fields given the "_r2" tag represent r2 values over all datasets. Fields
     given the "_per_dataset_r2" tag represent the r2 computed over each
     dataset present in the testing data.
@@ -94,7 +94,7 @@ def _rename_via_replace(ds: xr.Dataset, find: str, replace: str) -> xr.Dataset:
 
 def insert_aggregate_bias(ds_metrics: xr.Dataset) -> xr.Dataset:
     """Compute the aggregate bias over all datasets from the per dataset bias.
-    
+
     Renames the per dataset bias variables using the "per_dataset_bias" tag.  Only
     meant to be called on ds_metrics Datasets with a "dataset" dimension.
     """
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
@@ -56,7 +56,23 @@ def is_3d(da: xr.DataArray, vertical_dim: str = "z"):
     return vertical_dim in da.dims
 
 
-def insert_r2(ds_metrics):
+def _insert_r2_dataset_dim(ds_metrics):
+    global_metrics = ds_metrics.mean(DATASET_DIM_NAME)
+    mse_vars = [var for var in ds_metrics if "_mse" in var]
+    global_r2 = xr.Dataset()
+    for mse_var in mse_vars:
+        variance = ds_metrics[mse_var.replace("_mse", "_variance")]
+        ds_metrics[mse_var.replace("_mse", "_r2_per_dataset")] = (
+            1.0 - ds_metrics[mse_var] / variance
+        )
+        global_variance = global_metrics[mse_var.replace("_mse", "_variance")]
+        global_r2[mse_var.replace("_mse", "_r2")] = (
+            1.0 - global_metrics[mse_var] / global_variance
+        )
+    return xr.merge([ds_metrics, global_r2])
+
+
+def _insert_r2(ds_metrics):
     mse_vars = [var for var in ds_metrics if "_mse" in var]
     for mse_var in mse_vars:
         variance = ds_metrics[mse_var.replace("_mse", "_variance")]
@@ -64,6 +80,28 @@ def insert_r2(ds_metrics):
             1.0 - ds_metrics[mse_var] / variance
         )
     return ds_metrics
+
+
+def insert_r2(ds_metrics):
+    if DATASET_DIM_NAME in ds_metrics.dims:
+        return _insert_r2_dataset_dim(ds_metrics)
+    else:
+        return _insert_r2(ds_metrics)
+
+
+def insert_global_bias_metrics(ds_metrics):
+    if DATASET_DIM_NAME in ds_metrics.dims:
+        bias_vars = [var for var in ds_metrics if "bias" in var]
+        global_bias = xr.Dataset()
+        per_dataset_bias = xr.Dataset()
+        for bias_var in bias_vars:
+            per_dataset_bias[bias_var.replace("bias", "per_dataset_bias")] = ds_metrics[
+                bias_var
+            ]
+            global_bias[bias_var] = ds_metrics[bias_var].mean(DATASET_DIM_NAME)
+        return xr.merge([ds_metrics.drop(bias_vars), per_dataset_bias, global_bias])
+    else:
+        return ds_metrics
 
 
 def insert_rmse(ds: xr.Dataset):
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
@@ -58,7 +58,7 @@ def is_3d(da: xr.DataArray, vertical_dim: str = "z"):
 
 def _compute_r2(ds_metrics: xr.Dataset, tag: str = "_r2") -> xr.Dataset:
     """Compute r2 values from MSE and variance metrics.
-
+    
     The names of the resulting variables use the provided tag.
     """
     mse_vars = [var for var in ds_metrics if "_mse" in var]
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
@@ -56,52 +56,53 @@ def is_3d(da: xr.DataArray, vertical_dim: str = "z"):
     return vertical_dim in da.dims
 
 
-def _insert_r2_dataset_dim(ds_metrics):
-    global_metrics = ds_metrics.mean(DATASET_DIM_NAME)
+def _compute_r2(ds_metrics: xr.Dataset, tag: str = "_r2") -> xr.Dataset:
+    """Compute r2 values from MSE and variance metrics.
+    
+    The names of the resulting variables use the provided tag.
+    """
     mse_vars = [var for var in ds_metrics if "_mse" in var]
-    global_r2 = xr.Dataset()
+    ds_r2 = xr.Dataset()
     for mse_var in mse_vars:
-        variance = ds_metrics[mse_var.replace("_mse", "_variance")]
-        ds_metrics[mse_var.replace("_mse", "_r2_per_dataset")] = (
-            1.0 - ds_metrics[mse_var] / variance
-        )
-        global_variance = global_metrics[mse_var.replace("_mse", "_variance")]
-        global_r2[mse_var.replace("_mse", "_r2")] = (
-            1.0 - global_metrics[mse_var] / global_variance
-        )
-    return xr.merge([ds_metrics, global_r2])
+        variance_var = mse_var.replace("_mse", "_variance")
+        r2_var = mse_var.replace("_mse", tag)
+        ds_r2[r2_var] = 1.0 - ds_metrics[mse_var] / ds_metrics[variance_var]
+    return ds_r2
 
 
-def _insert_r2(ds_metrics):
-    mse_vars = [var for var in ds_metrics if "_mse" in var]
-    for mse_var in mse_vars:
-        variance = ds_metrics[mse_var.replace("_mse", "_variance")]
-        ds_metrics[mse_var.replace("_mse", "_r2")] = (
-            1.0 - ds_metrics[mse_var] / variance
-        )
-    return ds_metrics
-
-
-def insert_r2(ds_metrics):
+def insert_r2(ds_metrics: xr.Dataset) -> xr.Dataset:
+    """Insert r2 fields into the metrics Dataset.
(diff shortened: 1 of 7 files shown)


spencerkclark:
Tests pass locally, but I'm getting a failure in something that seems unrelated (prognostic run diagnostics for emulation) in CI:
<details>
<p>

```
============================= test session starts ==============================
platform linux -- Python 3.8.12, pytest-4.6.11, py-1.11.0, pluggy-0.13.1
Matplotlib: 3.2.2
Freetype: 2.6.1
rootdir: /root/project, inifile: pytest.ini
plugins: anyio-2.2.0, hypothesis-6.17.4, mpl-0.13, regtest-1.4.4, typeguard-2.10.0
collecting ... collecting 0 items                                                             collecting 19 items                                                            collecting 95 items                                                            collecting 241 items                                                           collecting 288 items                                                           collecting 293 items                                                           collecting 374 items                                                           collecting 443 items                                                           collected 452 items / 6 deselected / 446 selected                              

external/fv3viz/tests/test_plot_helpers.py ...................           [  4%]
external/fv3viz/tests/test_segment_plot_inputs.py ........               [  6%]
external/fv3viz/tests/test_visualize.py ...........................x.... [ 13%]
.........                                                                [ 15%]
external/loaders/tests/test__base.py ...........................         [ 21%]
external/loaders/tests/test__batch.py ...............................    [ 28%]
external/loaders/tests/test__config.py .............................     [ 34%]
external/loaders/tests/test__fine_resolution.py ............             [ 37%]
external/loaders/tests/test__local.py ...                                [ 38%]
external/loaders/tests/test__merger.py ........                          [ 39%]
external/loaders/tests/test__nudged.py .............                     [ 42%]
external/loaders/tests/test__one_ahead.py ....                           [ 43%]
external/loaders/tests/test__sequences.py ........                       [ 45%]
external/loaders/tests/test__serialized_phys.py ................         [ 49%]
external/loaders/tests/test__transforms.py .....                         [ 50%]
external/loaders/tests/test__utils.py ......                             [ 51%]
external/loaders/tests/test__xarray.py ...                               [ 52%]
external/loaders/tests/test_batches_save.py ...                          [ 52%]
external/loaders/tests/test_stacking.py ...                              [ 53%]
external/report/tests/test_holoviews.py ..                               [ 54%]
external/report/tests/test_report.py .....                               [ 55%]
external/synth/tests/test__fine_res.py ..                                [ 55%]
external/synth/tests/test__nudging.py .                                  [ 55%]
external/synth/tests/test_restart.py ........                            [ 57%]
external/synth/tests/test_synth.py ..............................        [ 64%]
tests/end_to_end_integration/test_argo_training_config.py .              [ 64%]
workflows/coarsen_c384_diagnostics/test_coarsen_c384_diagnostics.py .    [ 64%]
workflows/dataflow/tests/local/test_pipeline_common.py ....              [ 65%]
workflows/dataflow/tests/local/test_pipelines_coarsen_surface_c3072.py . [ 65%]
.                                                                        [ 66%]
workflows/diagnostics/tests/offline/test__keras_loss_history.py ......   [ 67%]
workflows/diagnostics/tests/offline/test__ml_prediction.py ............. [ 70%]
                                                                         [ 70%]
workflows/diagnostics/tests/offline/test__select.py ...                  [ 71%]
workflows/diagnostics/tests/offline/test_compute.py ..                   [ 71%]
workflows/diagnostics/tests/offline/test_compute_diagnostics.py ........ [ 73%]
......                                                                   [ 74%]
workflows/diagnostics/tests/offline/test_helpers.py .                    [ 74%]
workflows/diagnostics/tests/prognostic/test_computed_diagnostics.py .... [ 75%]
...................                                                      [ 80%]
workflows/diagnostics/tests/prognostic/test_generate_movie_stills.py ..  [ 80%]
workflows/diagnostics/tests/prognostic/test_generate_report.py ......... [ 82%]
.                                                                        [ 82%]
workflows/diagnostics/tests/prognostic/test_load_diag_data.py ......     [ 84%]
workflows/diagnostics/tests/prognostic/test_logs.py .                    [ 84%]
workflows/diagnostics/tests/prognostic/test_metrics.py .......           [ 85%]
workflows/diagnostics/tests/prognostic/test_registry.py ..               [ 86%]
workflows/diagnostics/tests/prognostic/test_savediags.py sssssssssssssss [ 89%]
sssssssssssssssss...                                                     [ 94%]
workflows/diagnostics/tests/prognostic/test_transform.py ..........      [ 96%]
workflows/diagnostics/tests/prognostic/emulation/test_single_run.py F... [ 97%]
...x                                                                     [ 98%]
workflows/fine_res_budget/tests/test_data.py .                           [ 98%]
workflows/fine_res_budget/tests/test_fine_res_budget.py .......          [100%]

=================================== FAILURES ===================================
_________________ test_log_functions[plot_histogram_begin_end] _________________

func = <function plot_histogram_begin_end at 0x7f8384154160>

    @pytest.mark.parametrize(
        "func",
        [
            single_run.plot_histogram_begin_end,
            # this test fails in CI for some reason
            # single_run.plot_cloud_weighted_average,
            single_run.plot_cloud_maps,
            single_run.skill_table,
            single_run.skill_time_table,
            single_run.log_lat_vs_p_skill("cloud_water"),
        ],
    )
    def test_log_functions(func):
    
        ds = vcm.cdl_to_dataset(cdl)
        for key in ds:
            ds[key].values[:] = 0
    
        for key in set(ds.coords) - {"time"}:
            ds[key].values[:] = np.arange(len(ds[key]))
    
        nx = len(ds.x)
        ds["lat"].values[:] = np.linspace(-45, 45, nx)
>       func(ds)

workflows/diagnostics/tests/prognostic/emulation/test_single_run.py:251: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
workflows/diagnostics/fv3net/diagnostics/prognostic_run/emulation/single_run.py:86: in plot_histogram_begin_end
    return {"cloud_histogram": _get_image()}
workflows/diagnostics/fv3net/diagnostics/prognostic_run/emulation/single_run.py:60: in _get_image
    im = wandb.Image(plot_to_image(fig))
external/fv3fit/fv3fit/tensorboard.py:10: in plot_to_image
    figure.savefig(buf, format="png", bbox_inches="tight")
/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/figure.py:2203: in savefig
    self.canvas.print_figure(fname, **kwargs)
/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/backend_bases.py:2119: in print_figure
    result = print_method(
/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py:514: in print_png
    FigureCanvasAgg.draw(self)
/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py:388: in draw
    self.renderer = self.get_renderer(cleared=True)
/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py:404: in get_renderer
    self.renderer = RendererAgg(w, h, self.figure.dpi)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <matplotlib.backends.backend_agg.RendererAgg object at 0x7f82d7082af0>
width = 537.6920033094618, height = 458899.18176172086, dpi = 100

    def __init__(self, width, height, dpi):
        RendererBase.__init__(self)
    
        self.dpi = dpi
        self.width = width
        self.height = height
>       self._renderer = _RendererAgg(int(width), int(height), dpi)
E       ValueError: Image size of 537x458899 pixels is too large. It must be less than 2^16 in each direction.

/opt/conda/envs/fv3net/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py:92: ValueError
```

</p>
</details>

workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
```diff
@@ -242,6 +243,13 @@ def weighted_mean(ds, weights, dims):
         return (ds * weights).sum(dims) / weights.sum(dims)
 
 
+def _maybe_average_over_dataset(ds: xr.Dataset) -> xr.Dataset:
```
AnnaKwa, line 246: Instead of each compute function doing the 'maybe average' as a last step, could the average over datasets be done a level up by splitting the test data into the various `dataset` coordinates if that dim is present and calling `compute_diagnostics` on each of these before averaging them together at the end?

e.g. something like this at this part of the code https://github.com/ai2cm/fv3net/blob/00851a37066a1efce949dd3a2a868a4fef650d9e/workflows/diagnostics/fv3net/diagnostics/offline/compute.py#L164

```
if `dataset` in ds.dims:
    for dataset_coord in ds.dataset:
         dataset_summaries.append( compute_diagnostics(
            prediction.sel(dataset=dataset_coord ), target.sel(dataset=dataset_coord ), grid, ds[DELP], n_jobs=n_jobs
        )

... and then avg over the resulting summary datasets
```

AnnaKwa (changes requested):
Thanks Spencer, I didn't realize the multiple dataset dims were broken in the offline diags after the past changes. The additional test is helpful. 

Is it possible to do the averaging over the dataset dim one level up as opposed to within each diagnostic computation? A side benefit could be that the values calculated for individual climates could also be left in the final summary dataset after the average is calculated, in the event that you want to go back and look at those later without redoing the calculation in a notebook.

workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
spencerkclark: As discussed on Slack, we need to be careful about computing the r2 values, but it should be possible to preserve per-dataset metrics in the computed diagnostics while continuing to display purely aggregate metrics in the report.  

f6c776d is a start at implementing things this way, but needs to be cleaned up.  I will do that tomorrow.

spencerkclark:
> could the average over datasets be done a level

This is an excellent point, which allows us to avoid touching `compute_diagnostics`.  There are a couple catches with this, as we discussed on Slack, but I've tried to address them in my updated commits.  It is nice that we will now be able to have per-dataset metrics stored in the offline diagnostics.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -171,10 +172,11 @@ def _compute_diagnostics(
 
     # then average over the batches for each output
     ds_summary = xr.concat(batches_summary, dim="batch")
+    ds_summary = insert_r2(ds_summary)
+    if DATASET_DIM_NAME in ds_summary.dims:
```
AnnaKwa, line 176: I think the handling of the aggregate-over-dataset-dim quantities could be consolidated into the if statement  if  1) `insert_r2` is removed and 2) a helper `insert_aggregate_r2` similar to `insert_aggregate_bias` calls `_compute_r2` on `ds_summary.mean("dataset")` and uses `rename_via_replace` in a similar manner. Just suggesting since it was clear from reading the code that the aggregated bias was being calculated and inserted, but it was not obvious that the same was true for the R2 unless looking into the `insert_r2` code.

e.g. 
```
ds_summary.update( _compute_r2(ds_summary, tag="_r2")) )
if DATASET_DIM_NAME in ds_summary.dims:
    ds_summary = insert_aggregate_r2(ds_summary)
    ds_summary = insert_aggregate_bias(ds_summary)    
```

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -56,14 +56,53 @@ def is_3d(da: xr.DataArray, vertical_dim: str = "z"):
...
+    """Compute the aggregate bias over all datasets from the per dataset bias.
+
+    Renames the per dataset bias variables using the "per_dataset_bias" tag.  Only
+    meant to be called on ds_metrics Datasets with a "dataset" dimension.
+    """
+    bias_vars = [var for var in ds_metrics if "bias" in var]
+    aggregate_bias = safe.get_variables(ds_metrics, bias_vars).mean(DATASET_DIM_NAME)
+    per_dataset_bias = safe.get_variables(ds_metrics, bias_vars)
+    per_dataset_bias = _rename_via_replace(per_dataset_bias, "bias", "per_dataset_bias")
+    return xr.merge([ds_metrics.drop(bias_vars), aggregate_bias, per_dataset_bias])
```
AnnaKwa, line 105: Would `xr.merge([ds_metrics.update(aggregtae_bias), per_dataset_bias])` do the same thing or nah?

AnnaKwa (approved):
Thanks Spencer, especially for the added tests. I only realized after re-reviewing that the change I requested did add more places in the code where if statements had to be inserted to deal with the lingering dataset dim. I do think it will be useful to have the individual datasets' diags saved out though. Maybe a future way to deal with this would be to save the aggregated dataset as `diagnostics.nc` and split off the part of the dataset that has a `dataset` dim into a separate netcdf that can be saved with a different name, so that the report step always expects the same dimensionality of data and doesn't have to worry about the possible extra dim. That's out of scope for this PR though, I think this works fine for now. 

I had one suggestion for the part where the aggregated metrics are inserted.

Unrelated to the PR, I did not appreciate how big the change in column water vapor path distribution was in different climates until I looked at the example report!

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
spencerkclark: Indeed that's much clearer, and further simplifies the code in `_helpers.py`.

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
spencerkclark: I experimented with this.  It would work here; however, one subtle thing that `xr.Dataset.update` does is that it operates in-place (meaning it modifies `ds_metrics`, which we input to the function).  In the context in which we are using it in `compute.py` this happens to be ok, because we call `insert_aggregate_bias` like this:
```
ds_metrics = insert_aggregate_bias(ds_metrics)
```
But if we called it like this:
```
new_ds_metrics = insert_aggegrate_bias(ds_metrics)
```
`ds_metrics` would change (and would not be equal to `new_ds_metrics`), which might be unexpected, so I've opted to stick with methods that do not operate in-place.  For context, when experimenting, I was initially bit by this in the test I wrote, until I realized this was going on.

spencerkclark:
Thanks for the helpful comments!

> Unrelated to the PR, I did not appreciate how big the change in column water vapor path distribution was in different climates until I looked at the example report!

That is interesting.  The warmer climates definitely shift the distribution to higher values, and I guess that has the effect of smoothing it out when everything is aggregated together.

PR #1700, opened 2022-03-14
argo: choose node pool on restarted prognostic runs

brianhenn:
Adds the functionality of choosing node pools (eg, for c384 on the ultra pool) to the `restart-prognostic-run` template

Significant internal changes:
- Adds the `choose-node-pool` step to `restart-prognostic-run` template


AnnaKwa (approved):
LGTM

PR #1701, opened 2022-03-14
`loaders.batches_from_mapper`: remove kludge

AnnaKwa:
This removes an old kluge function `nonderived_variables` that was hard coded to add the required variables needed for `dQu, dQv, eastward_wind, northward_wind` to the list of variables to subselect from the underlying data in `_get_batch`. This PR removes the this function and its usage in `_get_batch`, which now returns the full dataset from the mapper without subsetting variables. The subselection of variables and adding derived variables is handled later on by the `add_derived_data` transform.


oliverwm1 (approved):
Thanks, this is a nice simplification. Just have one small suggestion.

PR #1703, opened 2022-03-15
Add TransformedPredictor to fv3fit, DataTransforms to vcm

oliverwm1:
The current strategy for defining 'derived' input and output variables does not easily handle cases where a variable can be computed multiple ways. As an example, for the flux-form, we want to compute `Qm` from `Q1` and `Q2` before training, while after prediction we will be computing `Qm` from `Qm_flux`. This PR adds a new `DataTransform` framework which allows configuration of a sequence of transforms that can be applied to data either 1) during 'batch' creation or 2) after ML prediction.

Added public API:
- vcm.moist_static_energy_tendency, vcm.temperature_tendency
- vcm.DataTransform, vcm.ChainedDataTransform, vcm.DataTransformConfig
- fv3fit.TransformedPredictor

Refactored public API:
- `batches_from_mapper` now takes a `data_transforms` arg, which is of type `Sequence[vcm.DataTransformConfig]`
- `fv3fit.TrainingConfig` now has a `output_transforms` arg, which is also of type `Sequence[vcm.DataTransformConfig]`

Requirement changes:
- added `dacite` to VCM requirements

I tested this PR by training two ML models, one to [predict Q1/Q2 with Qm derived](https://storage.googleapis.com/vcm-ml-public/offline_ml_diags/train-derived-qm-data-transform-v5/apparent_source/index.html), and the other to predict [Qm/Q2 with Q1 derived](https://storage.googleapis.com/vcm-ml-public/offline_ml_diags/train-derived-qm-data-transform-v5/apparent_source_Qm/index.html). Offline reports are linked.

- [x] Tests added


external/loaders/loaders/batches/_batch.py
```diff
@@ -112,6 +114,12 @@ def batches_from_mapper(
             add_wind_rotation_info(res),
         ]
 
+    if data_transforms is not None:
+        data_transform = dacite.from_dict(
+            vcm.ChainedDataTransform, {"config": data_transforms}
+        )
+        transforms.append(curry(data_transform.apply))
+
     transforms.append(add_derived_data(variable_names))
```
oliverwm1, line 123: Currently `add_derived_data` calls `vcm.DerivedMapping` on the data and selects the given variables. We could add this to the `vcm.ChainedDataTransform` to do it as part of that transformation. Need to decide whether to do it before or after all the other transforms. For now I left it after, since it does the variable selection.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,83 @@
...
+@register("q2_from_dQ2_pQ2")
+def q2_from_dQ2_pQ2(ds):
+    ds["Q2"] = ds["dQ2"] + ds["pQ2"]
+    return ds
+
+
+@dataclasses.dataclass
+class DataTransformConfig:
+    name: Literal[
+        "q1_from_qm_q2", "qm_from_q1_q2", "q1_from_dQ1_pQ1", "q2_from_dQ2_pQ2"
```
oliverwm1, line 60: Some duplication, since need to add transformation name here in addition to registering it above. But I think worth it for the validation we'll get from `dacite`.

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -177,3 +179,52 @@ def load(cls, path: str) -> "EnsembleModel":
         models = [io.load(path) for path in config["models"]]
         reduction = config["reduction"]
         return cls(models, reduction)
+
+
+@io.register("output_transformed_model")
+class TransformedPredictor(Predictor):
```
oliverwm1, line 185: This is very similar to the `DerivedModel`, the important difference being that transformation names are used for configuration, instead of variable names. We might want to switch over to using this instead of `DerivedModel`, but the `TransformedPredictor` is not currently as fully featured, since I have not implemented the concept of "required variables" for the data transforms, and so we cannot do things like adding variable names to the model inputs.

external/vcm/setup.py
```diff
@@ -31,6 +31,7 @@
     "pytest-regtest",
     "h5netcdf>=0.8",
     "intake-xarray>=0.3.1",
+    "dacite",
```
oliverwm1, line 34: not strictly needed, but it's useful to be able to use it in tests.

external/loaders/loaders/batches/_batch.py
oliverwm1: The challenge is that the inputs/outputs for the `vcm.DerivedMapping` are defined differently than the static definitions used for the registered data transforms.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,103 @@
...
+        return DATA_TRANSFORM_REGISTRY[self.config.name]["inputs"]
+
+    @property
+    def output_variables(self) -> Sequence[str]:
+        return DATA_TRANSFORM_REGISTRY[self.config.name]["outputs"]
+
+
+@dataclasses.dataclass
+class ChainedDataTransform:
+    config: Sequence[DataTransformConfig]
```
mcgibbon, line 82: suggestion: `transforms` for the name, or something else that describes the intent of the config? we already have "config" in our argo yamls to indicate the training config, as it stands this PR would put the "config" key at three levels of that yaml.

workflows/diagnostics/fv3net/diagnostics/offline/_select.py
```diff
@@ -1,16 +1,22 @@
...
 
 
-def nearest_time(select_time: str, times: Sequence[str]):
-    select_datetime = vcm.parse_datetime_from_str(select_time)
-    datetimes = np.vectorize(vcm.parse_datetime_from_str)(times)
-    closest_datetime = min(datetimes, key=lambda d: abs(d - select_datetime))
-    return vcm.encode_time(closest_datetime)
+def nearest_time_batch_index(
+    time: cftime.DatetimeJulian, time_batches: Sequence[Sequence[cftime.DatetimeJulian]]
+):
```
mcgibbon, line 12: Can you add a docstring and output type hint? Don't you need two integers - the batch index, and the time index within the batch?

Update: I see from its use in offline/compute.py that you don't, adding a description of the output here would clarify.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,103 @@
...
+@register("q2_from_dQ2_pQ2", ["dQ2", "pQ2"], ["Q2"])
+def q2_from_dQ2_pQ2(ds):
+    ds["Q2"] = ds["dQ2"] + ds["pQ2"]
+    return ds
+
+
+@dataclasses.dataclass
+class DataTransformConfig:
+    name: TransformName
+    kwargs: dict = dataclasses.field(default_factory=dict)
```
AnnaKwa, line 59: What is an example of a transform that would take kwargs?

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@dataclasses.dataclass
+class DataTransform:
+    config: DataTransformConfig
```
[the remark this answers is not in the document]
AnnaKwa: +1 on the above- I think that would make it a bit easier for interactive use to be able to just do `DataTransform(name="q1_from_...").apply()` vs having to also put a config object inside.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,103 @@
...
+
+@register("qm_from_q1_q2", ["Q1", "Q2"], ["Qm"])
+def qm_from_q1_q2(ds):
+    ds["Qm"] = vcm.moist_static_energy_tendency(ds["Q1"], ds["Q2"])
+    return ds
+
+
+@register("q1_from_qm_q2", ["Qm", "Q2"], ["Q1"])
+def q1_from_qm_q2(ds):
+    ds["Q1"] = vcm.temperature_tendency(ds["Qm"], ds["Q2"])
```
AnnaKwa, line 40: Should this use the air temperature as the input to the latent heat of vaporization? Similarly for the moist static energy tendency, is it calculated using Lv(T at z) or is it the convention to always Lv(T freezing)?

AnnaKwa (changes requested):
Thank Oli, the implementation looks straightforward. I echoed Jeremy's request that the name and kwargs just be in the `DataTransform`, unless there is a need for the config class that I am missing. I think that would help streamline its usage outside of the batches loader. I also had a question about the calculation of the MSE and temperature tendency transforms, could just be me not being familiar with the conventions though.

external/fv3fit/fv3fit/train.py
```diff
@@ -177,6 +177,8 @@ def main(args, unknown_args=None):
         in_memory=training_config.cache.in_memory,
     )
 
+    logger.info(f"Following variables are in train batches: {list(train_batches[0])}")
```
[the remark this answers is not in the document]
oliverwm1: done.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -26,7 +26,7 @@
     insert_rmse,
     insert_column_integrated_vars,
 )
-from ._select import meridional_transect, nearest_time
+from ._select import meridional_transect, nearest_time_batch_index
```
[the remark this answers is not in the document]
oliverwm1: I probably just need to merge master into this branch.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,103 @@
...
+    return ds
+
+
+@register("q1_from_dQ1_pQ1", ["dQ1", "pQ1"], ["Q1"])
```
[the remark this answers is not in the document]
oliverwm1: Well we've already hard-coded "dQ1" and "pQ1" etc in other places (e.g. prognostic run) so I don't want to change those. For the function name, I was following convention of lower-case first letter. But yah, I agree it's a bit confusing. Maybe we allow capitals in the function names?

external/vcm/vcm/data_transform.py
oliverwm1: Good question. I had some use cases but then realized they would be better suited to just defining a new transform. I still think it's possible we will have a case (e.g. thresholding at a certain value/percentile) but don't have an immediate need for the feature.

external/vcm/vcm/data_transform.py
oliverwm1: originally i had this option as a kwarg, but that doesn't really work because then the required inputs are not well-defined. So I could add a new transform that uses the temperature-dependent version. I previously found it does help a bit in closing MSE budget, but overall is a pretty small factor.

external/vcm/vcm/data_transform.py
oliverwm1: Well, we might want to have an option of whether or not to rectify the implied surface precip to be positive when I implement the flux-form. (We could just always do it, but it might be interesting to see what happens if we don't).

external/fv3fit/tests/training/test_main.py
```diff
@@ -32,8 +32,8 @@ def get_mock_dataset(n_time, unstacked_dims: Sequence[str]):
             "downward_shortwave": (dims_surface, arr_surface),
             "net_shortwave": (dims_surface, arr_surface),
             "downward_longwave": (dims_surface, arr_surface),
-            "dQ1": (dims, arr),
-            "dQ2": (dims, arr),
+            "Q1": (dims, arr),
+            "Q2": (dims, arr),
```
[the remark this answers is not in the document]
oliverwm1: it seems a bit outside the scope of this PR to add the precipitative model test (it fails even if I undo the changes I made to `test_main.py`). But I will parameterize some of the output variable/transform changes I made, so that it will be possible to run the precipitative model test without those.

external/fv3fit/tests/training/test_main.py
```diff
@@ -271,10 +318,12 @@ def get_config(
     use_validation_data: bool,
     unstacked_dims: Sequence[str],
     use_local_download_path: bool = False,
+    output_transforms: Optional[Sequence[vcm.DataTransformConfig]] = None,
 ):
+    output_transforms = [] if output_transforms is None else output_transforms
     base_dir = str(tmpdir)
     input_variables = ["air_temperature", "specific_humidity"]
-    output_variables = ["dQ1", "dQ2"]
+    output_variables = ["Qm", "Q2"]
```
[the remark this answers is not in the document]
oliverwm1: Okay, I parametrized this arg, so the default is still dQ1 and dQ2.

external/vcm/vcm/data_transform.py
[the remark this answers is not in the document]
oliverwm1: Okay, yeah I guess `DataTransform` is simple enough that a separate config object is overkill.

external/vcm/vcm/data_transform.py
oliverwm1: That sounds good. Although I actually avoided using the `ChainedDataTransform` in the top-level configs. I used `Sequence[DataTransform]` directly there, otherwise we end up with an unnecessary single-itemed dict which just confuses things IMO. It does mean I have to do some custom handling for dumping/loading (e.g. see the `TransformedPredictor` dump/load methods).

oliverwm1: Let me know if you have other ideas to handle this.

external/loaders/loaders/batches/_batch.py
```diff
@@ -52,6 +52,7 @@ def batches_from_mapper(
     drop_nans: bool = False,
     shuffle_timesteps: bool = True,
     shuffle_samples: bool = False,
+    data_transforms: Optional[Sequence[Mapping]] = None,
```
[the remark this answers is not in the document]
oliverwm1: At this point, the `batches_from_mapper` func is just getting passed the raw dict, so it would not be accurate to type-hint with the DataTransform object. If/when we define a `BatchesFromMapperConfig` we could improve this.

external/vcm/tests/test_calc.py
```diff
@@ -143,3 +144,17 @@ def test_specific_humidity(t, rh, rho):
...
+    Q2 = xr.DataArray(np.reshape(np.arange(0, 40, 2), (5, 4)), dims=["x", "y"],)
+    Qm = vcm.moist_static_energy_tendency(Q1, Q2)
+    assert Qm.shape == Q1.shape
+
+
+def test_temperature_tendency():
+    Qm = xr.DataArray(np.reshape(np.arange(0, 40, 2), (5, 4)), dims=["x", "y"],)
+    Q2 = xr.DataArray(np.reshape(np.arange(0, 40, 2), (5, 4)), dims=["x", "y"],)
+    Q1 = vcm.temperature_tendency(Qm, Q2)
+    assert Qm.shape == Q1.shape
```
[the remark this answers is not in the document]
oliverwm1: I added a test for round-tripping a temperature tendency to MSE and back.

workflows/diagnostics/fv3net/diagnostics/offline/_select.py
oliverwm1: These commits were actually already merged to master on another PR. They are no longer in the PR preview now that I've merged master in. Let me know if you feel strongly, I could make another PR to add a docstring.

external/vcm/vcm/data_transform.py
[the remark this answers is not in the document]
oliverwm1: I changed the function names to strictly follow the convention for variable names, even for cases where this means the first letter of the function is a capital. I think this improves the situation.

external/fv3fit/tests/training/test_main.py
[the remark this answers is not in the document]
mcgibbon: OK, thanks for trying. I was hoping it would pass on main.

workflows/diagnostics/fv3net/diagnostics/offline/_select.py
mcgibbon: I don't touch this code often, I just found it confusing and would have saved a minute or two of investigation if there were a docstring. I'll leave it up to you.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,117 @@
...
+def register(
+    inputs: Sequence[str], outputs: Sequence[str], func: Callable[..., xr.Dataset],
+):
+    name = func.__name__
+    if name in DATA_TRANSFORM_REGISTRY:
+        raise ValueError(f"Function {name} has already been added to registry.")
+    DATA_TRANSFORM_REGISTRY[name] = {
+        "func": func,
+        "inputs": inputs,
+        "outputs": outputs,
```
mcgibbon, line 30: optional: you could add more type-safety here with a dataclass instead of Mapping[str, Any].

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,117 @@
...
+        return ds
+
+    @property
+    def input_variables(self) -> Sequence[str]:
+        inputs: Set[str] = set()
+        for transform in self.transforms[::-1]:
+            inputs.update(transform.input_variables)
+            for output in transform.output_variables:
+                inputs.discard(output)
+        return sorted(list(inputs))
```
mcgibbon, line 110: I think this code is correct so I'll say this is optional, but this is a pretty compact and easy to test algorithm that takes a while to understand and is easy to break, e.g. by iterating in the wrong order, processing all outputs then all inputs, etc. Breakages also would cause failures in distant parts of the code.

If you think someone else will ever edit this, you should consider adding a unit test.

external/vcm/vcm/data_transform.py
```diff
@@ -0,0 +1,117 @@
...
+DATA_TRANSFORM_REGISTRY: MutableMapping[str, Mapping[str, Any]] = {}
+
+TransformName = Literal[
+    "Q1_from_Qm_Q2",
+    "Qm_from_Q1_Q2",
+    "Q1_from_Qm_Q2_temperature_dependent",
+    "Qm_from_Q1_Q2_temperature_dependent",
+    "Q1_from_dQ1_pQ1",
+    "Q2_from_dQ2_pQ2",
+]
```
mcgibbon, line 17: This isn't mypy-checkable since our inputs are in yaml files, so if it goes out of date nothing will break. Because of that I'd suggest adding a unit test to make sure it stays up-to-date. You can check the values allowed in the Literal (`TransformName.__args__`) are identical to DATA_TRANSFORM_REGISTRY.keys().

The other option is to delete this type hint so it can't go out of date, `str` would be OK. But this is pretty nice documentation.

mcgibbon (approved):
Thanks for minimizing conflicts with my PR! Please add a unit test for TransformName, the other two suggestions are optional.

external/vcm/vcm/data_transform.py
oliverwm1: This is already tested by `test_all_registered_transforms_are_added_to_data_transform_name_type`. If a user registers a new transform without adding its name to `TransformName`, that test will fail. Which is good, since otherwise they would not be able to use their transform out in the wild (dacite would complain).

external/vcm/vcm/data_transform.py
oliverwm1: This is tested by `test_chained_transform_inputs_outputs`. I'll think about adding some additional cases.

oliverwm1:
> Thanks! Could you add a transform, perhaps Q1 from dQ1 + pQ1 to the vcm workflow config example?

Yup.

external/vcm/vcm/data_transform.py
oliverwm1: good idea!

PR #1704, opened 2022-03-15
Emulation online

nbren12:
I wanted to explore training ML with a cycled online-offline setup. In this configuration, the emulator alternates between the emulator and fortran at a configurable interval.

It's a big diff, but mostly just config/version changes. The code in `projects/microphysics` doesn't need a close eye. It's probably easiest to review commit-by-commit, the earlier commits are mostly related to implementing the feature, but then the history descends into trial/error on the memory issue.

* add configurations for cycling `ModelConfig.online_schedule`
* add the SQL querying logic (about time this is in master---doesn't need careful review)
* revive `fv3fit` image, which allows for more rapid iteration on fv3fit code
* lots of memory usage tweaks. Maybe some of these will help #1710.


projects/microphysics/train/limited.yaml
```diff
@@ -0,0 +1,236 @@
+{
```
nbren12, line 1: i downloaded this from wandb since it wasn't in master yet and is an active target for development

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -271,10 +272,14 @@ def save_jacobians(std_jacobians, dir_, filename="jacobians.npz"):
...
 def main(config: TrainConfig, seed: int = 0):
     logging.basicConfig(level=getattr(logging, config.log_level))
+    logging.getLogger("fv3fit.keras.callbacks").setLevel(logging.DEBUG)
     set_random_seed(seed)
 
     # callbacks that are always active
-    callbacks = [tf.keras.callbacks.TerminateOnNaN()]
+    callbacks = [
+        tf.keras.callbacks.TerminateOnNaN(),
+        GarbageCollectionCallback(batch_frequency=100),
```
frodre, line 281: I haven't seen garbage collection affect memory usage in my testing, so I'm inclined just to remove the usage.

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -286,17 +291,17 @@ def main(config: TrainConfig, seed: int = 0):
         config.test_url, config.nfiles_valid, config.model_variables
     )
 
-    train_set = next(iter(train_ds.shuffle(1_000_000).batch(50_000)))
+    if config.shuffle_buffer_size is not None:
+        train_ds = train_ds.shuffle(config.shuffle_buffer_size)
```
frodre, line 295: Shuffle buffer over samples really seems to be detrimental to memory usage and training performance.  If using dict code, we can shuffle on the file names (gives us re-ordering every epoch) and then get a psuedo shuffle if we use the dataset `.interleave`.  I'm wondering if we can get away with it and not see a degradation in final training loss.   I'll prepare an update in a PR with some of the timing and memory usage comparisons.

nbren12:
Production configuration here: https://wandb.ai/ai2cm/microphysics-emulation/runs/30f2835o?workspace=user-nbrenowitz

frodre (approved):
LGTM.  

Just a couple minor comments and suggestions.  I'll be interested to try some profiling with the memory allocation update.

frodre:
`external/emulation/tests/test__time.py` is an empty file

external/emulation/emulation/_emulate/microphysics.py
```diff
@@ -83,11 +129,13 @@ def microphysics(self, state: FortranState) -> None:
...
         if self.orig_outputs is None:
             self.orig_outputs = set(state).intersection(model_outputs)
+            logger.debug(f"Overwriting existing state fields: {self.orig_outputs}")
 
-        logger.info(f"Overwritting existing state fields: {self.orig_outputs}")
         microphysics_diag = {
             f"{name}_physics_diag": state[name] for name in self.orig_outputs
         }
+
+        model_outputs.update(self.mask(state, model_outputs))
```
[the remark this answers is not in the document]
nbren12: Hmm. I'm not sure I follow, but I agree the naming `always_right` is a bit confusing since the user of this class doesn't care about right vs left, but rather emulator vs not. I'll change to that.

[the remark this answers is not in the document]
nbren12: Changed to 
```
def always_emulator(state: FortranState, emulator: FortranState):
    return emulator
```
if that helps.

projects/microphysics/train/rnn.yaml
```diff
@@ -3,11 +3,12 @@ wandb:
   wandb_project: microphysics-emulation
 batch_size: 512
 epochs: 25
-nfiles_valid: 100
-valid_freq: 2
+nfiles_valid: 2
```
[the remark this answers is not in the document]
nbren12: I reverted this change.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: Sure. I'll revert this.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: Seems like a good thing to try in a follow-up PR. The computational cost is probably more important for dense models. RNNs bottleneck on the actual network ops. My null hypothesis is that some degree of interleaving will be important for performance, but maybe proper shuffling doesn't matter.

nbren12:
The last experiment used over 3000 files so may have worse memory use than for fewer files. I submitted a new training with nfiles=1440:
- argo job: 2f8116a5431b927d84636715293846961791dcf5

PR #1705, opened 2022-03-16
offline diags: grab snapshot from batches

oliverwm1:
This ensures that the data that goes into the offline snapshot diagnostic has undergone the same transformations as the other "batched" data.

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/_select.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/_select.py
@@ -12,10 +12,8 @@ def nearest_time_batch_index(
 ):
     min_index = 0
     min_distance = min(abs(np.array(time_batches[0]) - time))
-    print(min_distance)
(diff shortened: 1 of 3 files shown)


AnnaKwa (approved):
Thanks, LGTM. For posterity can you add a bit more detail to the PR description- that this is related to the transforms getting added and the old way of getting the snapshot via mapper key couldn't make use of those changes in the batch loader?

oliverwm1:
> Thanks, LGTM. For posterity can you add a bit more detail to the PR description- that this is related to the transforms getting added and the old way of getting the snapshot via mapper key couldn't make use of those changes in the batch loader?

Sounds good. I do think it is a bit more general than that, since I think it could let us delete the `vcm.DerivedMapping` call in `_get_predict_function` (if the data is all loaded via `batches_from_mapper` then we know the `vcm.DerivedMapping` has already been applied to it). That said I didn't actually delete it the call.

PR #1706, opened 2022-03-17
Revert "Re-enable using multiple datasets in offline report (#1699)"

spencerkclark:
This reverts commit a7ea32002b10ad7bc196b2c273c8861f2f3271db.

I really should have realized this earlier before going down this path, but the approach taken here uses a faulty method for computing the variance -- and by extension the r2 -- across all datasets.  The variance across all datasets is in general not equal to the mean of the variance across datasets.

I'm reverting this commit for now, until I come up with a better solution.

cc: @AnnaKwa


spencerkclark:
Superseded by #1709.

PR #1707, opened 2022-03-17
Rename quantities to standard names in fine-res mapper

oliverwm1:
For the flux-form ML, we need to use additional diagnostics like the radiative fluxes and turbulent heat fluxes. In order to use these in a similar way during ML training and in the prognostic run, we need to ensure we are using the standard names for these quantities in the fine-res mapper. This PR adds a new option `use_fine_res_state` (default True) which forces the air temperature, specific humidity and radiative/turbulent fluxes to come from the fine-res data, even when we are merging with nudged-run data.

Refactored public API:
- `use_fine_res_state` option for `open_fine_resolution` mapper

Significant internal changes:
- update fine-res-zarr schema for latest fine-res budget (current regtest data is missing 2D diagnostics)
- rename air temperature, specific humidity and radiative/turbulent fluxes to standard names if `use_fine_res_state` is true
- compute column-integrated fine-res temperature nudging tendency in mapper (alternatively, we could recompute the fine-res budget with the `int_t_dt_nudge_coarse` variable included in the budget)

- [x] Tests added


external/loaders/loaders/mappers/_fine_res.py
```diff
@@ -63,11 +69,10 @@ def _open_merged_dataset(
...
 
-    # enforce that these ML inputs come from fine dataset if they exist there
-    if "T" in fine:
-        merged["air_temperature"] = fine.T
-    if "sphum" in fine:
-        merged["specific_humidity"] = fine.sphum
+    if use_fine_res_state:
+        # overwrite standard name arrays with those from fine-res budget
+        for fine_res_name, standard_name in FINE_RES_BUDGET_NAMES.items():
+            merged[standard_name] = fine[fine_res_name]
```
brianhenn, line 75: Would dropping the (duplicated) fine-res named variables when `use_fine_res_state` is true help simplify things?

brianhenn (approved):
LGTM

external/loaders/loaders/mappers/_fine_res.py
oliverwm1: Yeah, I was considering that. The benefit of not dropping is that then we can always assume that the GFDL-style names will always be available in the mapper output, no matter what the `use_fine_res_state` flag is set to. Do you have a strong preference?

brianhenn: Up to you

oliverwm1: Okay I'm going to leave as is. I don't think there'll be a performance cost, since these will just be references to the other arrays.

Issue #1710, opened 2022-03-18
Memory leaks with netcdfs + tf.data.Dataset


nbren12:
@frodre and I have both been having memory leaks while training with tf datasets. If I had to guess, this performance regression is probably caused by one of the version updates from 97d2e459d917dc7d879c5405bd830b82ac8b1b82. 

We have tried a few different things:

- decrease shuffle buffer size. tf.data.Dataset doesn't seem to deallocate these buffers. This improved initial performance but memory is still leaking.
- install libtcmalloc-minimal4. Some github issues suggested this worked. I tried it here: 201edd66a7a41eb9eb0de905544efa065b7de92e. This has had the biggest impact. RNNs complete 20/25 epochs, but they still ultimately crashed.
- Add a tensorflow callback which garbage collects every 100 batches. 067f22b32221563bd8d5b268a9c2a7fc1a644ca9. This did not solve the leak.

Some ideas we've discussed trying

- [ ] use `.map` rather than `.from_generator`. We can leverage the `dict_dataset` code here. @frodre 
- [ ] remove `.prefetch`. this may hold onto old python objects. @frodre 
- [ ] see if problem persists with tfrecord data. @nbren12 

cc @mcgibbon

AnnaKwa:
Memory usage during the nudged-to-PIRE training with the full train dataset
Left: running the  on the PR branch using tf dataset, OOMed
Right: running on master branch

<img width="587" alt="image" src="https://user-images.githubusercontent.com/16710132/159063177-e30fd9f4-4a8e-4533-be07-dd3ed2bf8a7b.png">

mcgibbon:
This problem doesn't exist in `fv3fit.train` with `in_memory=False` when using the batches_from_netcdf loader with one batch. Here are profile reports for 3 epochs of training:
[1_epoch-2022-03-18T11:32:18.199.zip](https://github.com/ai2cm/fv3net/files/8306918/1_epoch-2022-03-18T11.32.18.199.zip)
And one epoch of training:
[3_epochs-2022-03-18T11:26:19.184.zip](https://github.com/ai2cm/fv3net/files/8306920/3_epochs-2022-03-18T11.26.19.184.zip)

The memory usage is identical for 1 or 3 epochs, and in the 3 epoch training report most of the memory allocation happens before the training function is called. The memory allocated during the training function execution (208MB) is much less than the memory allocated when seq_to_tfdataset is called (609MB). This report is from a dataset which happens to have one batch.

[...]

PR #1712, opened 2022-03-18
Emulation memory optimization

frodre:
Prefetch before shuffle in the tf dataset pipeline was causing a larger amount of memory growth per epoch.  This PR adds some adjustments to reduce memory usage.

 * Remove large initial shuffle buffer for creating a sample dataset in favor of loading **more** files (pre-shuffled)
 * Change prefetch to be called after shuffling and batching
 * Lower the epoch shuffle buffer size to 6 tiles (13824 samples)

I also removed some old caching code.  `TrainConfig` no longer uses/allows the `cache` parameter.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/load.py
+++ b/external/fv3fit/fv3fit/emulation/data/load.py
@@ -5,6 +5,7 @@ import xarray as xr
 from toolz.functoolz import compose_left
 from typing import Callable, Optional, Sequence
 
+from loaders.batches import shuffle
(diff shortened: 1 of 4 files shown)


frodre:
Candidate to handle #1710. Checking with dense model run.

nbren12 (approved):
lgtm.

nbren12:
I merged this since your results seemed pretty convincing and and removing the caching seemed like a good idea regardless.

PR #1716, opened 2022-03-22
improve memory use with libtcmalloc

nbren12:
Memory performance is still poor even after merging #1704. Setting this single environment variable makes a dramatic difference! 

ld_preload_helps

This plot is produced by running the following benchmark w/ and w/o the LD_PRELOAD flag:
```
from fv3fit.train_microphysics import TrainConfig, main
import os

os.environ["WANDB_RUN_GROUP"] = "debug-memory"

config = TrainConfig.from_yaml_path("rnn.yaml")
config.use_wandb = False
config.model.architecture.name = "dense"
config.model.architecture.kwargs = {}
config.checkpoint_model = False
config.epochs = 10
config.nfiles = 10
config.nfiles_valid = 1
config.out_url = "local1"
main(config)
```


frodre (approved):
LGTM, thanks!

PR #1718, opened 2022-03-22
Tfdataset using map instead of from_generator

frodre:
The using `tf.Dataset.map` to map across netcdf filenames, open them, and convert to tensors allows for use of interleave, which can parallelize data input, and allows for reshuffling the file order with every epoch (not possible w/ generator currently).  

From our memory leak investigation, shuffling across samples incurs a severe input-pipeline performance cost, but is required to randomize our `from_generator` input pipeline.  Based on tensorflow profileing~60% of the time is spent waiting for inputs when sample shuffling is used vs. 30% when no sample shuffling is performed for i/o bound model training.  (Tested with 100 files over 10 epochs)

Using map, we might be able to remove the need for sample shuffling since interleave will mix samples from shuffled file ordering.  Using a draft PR to test that offline performance is not significantly reduced with production-like workflow.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/dict_dataset.py
+++ b/external/fv3fit/fv3fit/emulation/data/dict_dataset.py
@@ -1,25 +1,29 @@
+import logging
 from typing import Sequence
 import tensorflow as tf
+
 import vcm
 from fv3fit.emulation.data.io import get_nc_files
+from .transforms import open_netcdf_dataset
 
 __all__ = ["netcdf_url_to_dataset"]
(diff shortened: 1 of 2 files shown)


frodre:
The map operation is quite a bit slower in this context.  Perhaps something in the transformation from xarray -> tensor map takes extra time here?  Regardless, closing for now

image
[wandb](https://wandb.ai/ai2cm/microphysics-emulation/groups/sample-buffer-generator_or_map-mar-2022/workspace?workspace=user-aperkins)

PR #1719, opened 2022-03-23
emulation: implement new normalization options

nbren12:
I want to run an experiment without per-feature demeaning of the pressure. I suggest reviewing commit-by-commit as usual.

Added public API:
- added `config.model.normalize_map` this is passed a `NormFactory` dataclass object
- renamed  `config.model.normalize_key` to `config.model.normalize_default`

Significant internal changes:
- Generalize the `normalize_key` option to the `NormalizeFactory`...pass to all the appropriate places....would be nice to decouple this logic better. I'm not sure the `Field` abstractions are working that well. It would make sense to group selection/normalization/tendency logic as a user configuration, but these classes are only used as low-level helpers for the MicrophysicConfig model building. They have a lot of arguments and are hard to test for this application.  We should consider removing at least their tests. These required effort to update...I am not sure provide any additional coverage over those of `MicrophysicsConfig`. 

-  [x] Tests added


external/fv3fit/fv3fit/emulation/losses.py
```diff
@@ -60,7 +65,8 @@ def build(self, output_samples: Mapping[str, tf.Tensor],) -> Callable:
         """
         loss_funcs = {}
         for out_varname, sample in output_samples.items():
-            loss_funcs[out_varname] = NormalizedMSE(self.normalization, sample)
+            norm_layer = self.normalization.build(sample)
+            loss_funcs[out_varname] = NormalizedMSE(norm_layer)
```
nbren12, line 69: this shows the benefit of passing an instantiated object to a class constructor rather than constructing the normalization inside. Makes it possible to pass any callable to NormalizedMSE. Probably would be even better to just pass the scales.

external/fv3fit/tests/emulation/layers/test_fields.py
```diff
@@ -97,20 +106,6 @@ def test_IncrementedFieldOutput(dt_sec: float):
     assert magnitude == pytest.approx(dt_sec, rel=1.0)
 
 
-def test_IncrementedFieldOutput_tendency_layer_name():
```
nbren12, line 100: I could probably revert this change...but I think we should probably remove the IncremntedFieldOutput object anyway, since we use `tensor_transforms` for that.

external/fv3fit/tests/emulation/layers/test_normalization2.py
```diff
@@ -12,7 +12,7 @@
 
 
 def test_norm_layer():
-    norm = NormLayer(scale=2.0, center=1.0, epsilon=1.0)
+    norm = NormLayer(scale=2.0, center=1.0)
```
nbren12, line 15: I realized that none of the normalizations configurable by the microphysics emulation support `epsilon`.

external/fv3fit/tests/emulation/test_losses.py
```diff
@@ -5,20 +5,24 @@
...
 
 
 def test_NormalizeMSE():
-    sample = np.array([[25.0], [75.0]])
-    target = np.array([[50.0], [50.0]])
-
-    mse_func = NormalizedMSE("mean_std", sample)
-    mse = mse_func(target, sample)
-    np.testing.assert_approx_equal(mse, 1.0, 6)
+    sample = np.array([[0], [0]])
```
nbren12, line 17: `NormalizedMSE` is simple enough now that I don't think it's worth checking the actual values.

external/fv3fit/tests/emulation/test_train_microphysics.py
```diff
@@ -96,17 +97,12 @@ def test_TrainConfig_from_flat_dict():
...
-    with open(yaml_path, "w") as f:
-        yaml.safe_dump(asdict(default), f)
-
-        loaded = TrainConfig.from_yaml_path(yaml_path)
-
-        assert loaded == default
+def test_TrainConfig_from_yaml(tmp_path: PosixPath):
+    default = TrainConfig(".", ".", ".")
+    yaml_path = tmp_path / "train_config.yaml"
+    yaml_path.write_text(default.to_yaml())
```
nbren12, line 103: refactored to use `to_yaml`

PR #1720, opened 2022-03-23
emulation: add more normalization configs

nbren12:
I want to run an experiment without per-feature demeaning of the pressure. I suggest reviewing commit-by-commit as usual.

Added public API:

- add config.model.normalize_map this is passed a NormFactory dataclass object
- renam config.model.normalize_key to config.model.normalize_default
- `TrainConfig.to_yaml` is needed to create yaml now since Enum objects require some special handling.
- Deprecate `fv3fit.emulation.layers.normalization`

Significant internal changes:
- Generalize the normalize_key option to the NormalizeFactory...pass to all the appropriate places....would be nice to decouple this logic better. I'm not sure the Field abstractions are working that well. It would make sense to group selection/normalization/tendency logic as a user configuration, but these classes are only used as low-level helpers for the MicrophysicConfig model building. They have a lot of arguments and are hard to test for this application. We should consider removing at least their tests. These required effort to update...I am not sure they provide any additional coverage over those of MicrophysicsConfig.


external/fv3fit/tests/emulation/layers/test_fields.py
```diff
@@ -8,6 +8,7 @@
     IncrementedFieldOutput,
```
nbren12, line 8: My sense is that these tests may not be necessary since the Field objects are essentially private helpers of `MicrophysicsConfig` which is tested.

external/fv3fit/tests/emulation/layers/test_normalization.py
```diff
@@ -90,7 +89,4 @@ def test_NormLayer_Configs(config_cls):
 
     tensor = _get_tensor()
     config = config_cls("mean_std", tensor, "lol")
-    layer = config.initialize_layer()
-
-    assert isinstance(layer, NormLayer)
```
nbren12, line 95: these kinds of checks are pretty rigid esp if `NormLayer` is a concrete class rather than a protocol.

frodre (approved):
LGTM.  Nice Update

> I'm not sure the Field abstractions are working that well. It would make sense to group selection/normalization/tendency logic as a user configuration, but these classes are only used as low-level helpers for the MicrophysicConfig model building. They have a lot of arguments and are hard to test for this application.

I agree.  It's been on my mental backlog since most of the functionality is no longer needed.  
A refactor PR is probably in order to remove some unused portions of the emulation code.  The transform pipeline is another part that comes to mind since we have the nice composable tensor-transform options now.

PR #1722, opened 2022-03-23
Allow using auxiliary input variables in transformed prediction

oliverwm1:
This PR allows `TransformedPredictor` to use variables from the input dataset provided to `.predict` for its transformations. This is necessary for transformations which use an auxiliary variable that is not output by the wrapped `.predict` method.

Significant internal changes:
- merge input dataset with wrapped `.predict` output dataset
- only include the output variables in the dataset returned by `TransformedPredictor.predict`

- [ ] Tests added


AnnaKwa (approved):
LGTM

PR #1727, opened 2022-03-23
add argo workflow for end to end experiments

nbren12:
Running end to end experiments of the microphysics emulator was not so easy.

* combined the training/prog run workflows into one yaml

Can still use --entrypoint to select the workflows invidually

* add scripts/framework for launching end to end runs

See experiments/default.yaml

* launched "production" tests of a recent commit on master
```
EndToEndJob(name='limited-d6a0ed-v1', fv3fit_image_tag='d6a0ed27eb58dae25782e9c6613774562f2e64f0', image_tag='d6a0ed27eb58dae25782e9c6613774562f2e64f0', bucket='vcm-ml-experiments', project='microphysics-emulation')
EndToEndJob(name='rnn-d6a0ed-v1', fv3fit_image_tag='d6a0ed27eb58dae25782e9c6613774562f2e64f0', image_tag='d6a0ed27eb58dae25782e9c6613774562f2e64f0', bucket='vcm-ml-experiments', project='microphysics-emulation')
to monitor experiments: argo list -lexperiment=d6a0ed27eb58dae25782e9c6613774562f2e64f0-end-to-end
```


projects/microphysics/experiments/default.py
```diff
@@ -0,0 +1,21 @@
+import sys
```
nbren12, line 1: Would be nice to have this in a cron job so we could detect performance regressions faster.

projects/microphysics/argo/argo.yaml
```diff
@@ -0,0 +1,376 @@
...
+kind: Workflow
+metadata:
+  generateName: prognostic-microphysics-
+  labels:
+    project: microphysics-emulation
+    job_type: prognostic-evaluation
+spec:
+  entrypoint: prognostic-run
+  arguments:
+    parameters:
```
frodre, line 11: One missing feature I would like is a way to group multiple runs (but different `workflow.name`) under the same W&B "run group".  It seems like it's a prominent feature in the UI and it was really helpful for comparison of my recent experiments.

Perhaps just an option to override the default of the job name, which works well for single experiments.

projects/microphysics/experiments/default.py
frodre: Agree, I can cook one up after this merges.

frodre (changes requested):
Overall looks good and looking forward to giving it a spin in practice.

Just a small feature request.  I would like to see a `WANDB_RUN_GROUP` argo parameter or override mechanism that would let me group multiple experiments under a single group.  I think the current default as the workflow name makes sense.

projects/microphysics/argo/argo.yaml
nbren12: I added `wandb-group` and `wandb-tags` parameters.

nbren12:
this is ready for re-review.

projects/microphysics/argo/end_to_end.py
```diff
@@ -0,0 +1,222 @@
...
+        [
+            "argo",
+            "submit",
+            WORKFLOW_FILE.absolute().as_posix(),
+            "--entrypoint",
+            job.entrypoint,
+            "--name",
+            job.name,
+        ]
+        + _argo_parameters(job.parameters + other_parameters)
```
frodre, line 33:
```suggestion
        + _argo_parameters(job.parameters + other_parameters)
        _argo_parameters(dict(**job.parameters, **other_parameters))
```

projects/microphysics/argo/end_to_end.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+WORKFLOW_FILE = pathlib.Path(__file__).parent / "argo.yaml"
+
+
+def load_yaml(path):
+    return yaml.safe_load(pathlib.Path(path).read_bytes())
+
+
+def submit(job, labels, other_parameters=None):
```
frodre, line 21: Since this is a main public function, add a brief docstring and maybe a type for `job` so it's easy to see what goes in.  You describe it in the examples below, but this will make it abundantly clear.

projects/microphysics/argo/end_to_end.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    subprocess.check_call(args, stdout=subprocess.DEVNULL)
+
+
+def _run_job_q(job):
+    print(job)
+    return input("Y/n") != "n"
+
+
+def submit_jobs(jobs, experiment_name):
```
frodre, line 44: Brief docstring mostly to delineate that experiment name is used for grouping by tag/group.  Could add that `jobs` is a `Sequence[JobType]` or whatever you settle on for the type into `submit_job`.

projects/microphysics/argo/end_to_end.py
```diff
@@ -0,0 +1,222 @@
...
+        job = TrainingJob(
+            name="test-v5",
+            fv3fit_image_tag="d848e586db85108eb142863e600741621307502b",
+            config=load_yaml("../train/rnn.yaml"),
+        )
+
+    """
+
+    name: str
+    config: dict = dataclasses.field(repr=False)
```
frodre, line 131: Maybe add a comment describing where to find the config schema for this type of job.  Or maybe add a comment that it should be compatible with the argo training target listed under the `WORKFLOW_FILE` (which right now is `TrainConfig`? 

Not quite sure since everything is overall general but right now only use it with `TrainMicrophysics`.  Maybe we can switch the description once we're compatible with regular `fv3fit` training.  Something to think about, but no change required since I understand how to use it and this is a special case for emulation right now.

frodre (approved):
Thanks, @nbren12 for the updates.  It's exactly what I was looking for.  I was able to get a job submitted and really like how using python for submission opens things up.  There was one change that needed to be made for it to work and then I have a couple documentation requests.  Other than that, looks great!

Also, I had to upgrade my argo version to get `-l` parameter to work, but that might just be a me problem.  Don't know if argo requirements for local envs listed anywhere.

projects/microphysics/argo/end_to_end.py
nbren12: oh...thanks. sorry for not testing.

projects/microphysics/argo/end_to_end.py
nbren12: I added a small comment. Would be nice to extract the config objects from `runtime` so that they can be composed "client-side" without requiring fv3 in the environment.

PR #1728, opened 2022-03-23
Perform dataset transformations on batched data for performance, shuffling

mcgibbon:
The tfdataset refactor significantly reduced performance of data loading, due to transformations being performed on unbatched data. This PR fixes that performance by batching the data, and improves shuffling behavior by shuffing batches before shuffling samples.

Refactored public API:

Significant internal changes:
- dataset elements passed to training functions now have a sample dimension, and each contain one batch of data
- training functions are refactored to use elements with a sample dimension as input
- TrainingLoopConfig.fit_loop now shuffles batch elements before unbatching and shuffling them, improving shuffling on temporally-contiguous input data

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/convolutional.py
+++ b/external/fv3fit/fv3fit/keras/_models/convolutional.py
@@ -20,7 +20,7 @@ from fv3fit.keras._models.shared.utils import (
     full_standard_normalized_input,
 )
 from fv3fit import tfdataset
-from fv3fit.tfdataset import select_keys, ensure_nd, apply_to_mapping
+from fv3fit.tfdataset import select_keys, ensure_nd, apply_to_mapping, apply_to_tuple
 
 logger = logging.getLogger(__file__)
 
@@ -45,6 +45,8 @@ class ConvolutionalHyperparameters(Hyperparameters):
         training_loop: configuration of training loop
         loss: configuration of loss functions, will be applied separately to
             each output variable
+        normalization_fit_samples: number of samples to use when fitting normalization,
+            note that one sample is one tile
     """
 
     input_variables: List[str]
@@ -62,6 +64,7 @@ class ConvolutionalHyperparameters(Hyperparameters):
         default_factory=dict
     )
     loss: LossConfig = LossConfig(scaling="standard", loss_type="mse")
+    normalization_fit_samples: int = 30
 
     @property
     def variables(self) -> Set[str]:
@@ -81,20 +84,28 @@ def get_Xy_dataset(
     requested input variables and the second containing
     the requested output variables.
     """
-    # tile, x, y, z
-    data = data.map(apply_to_mapping(ensure_nd(4)))
+    # sample, tile, x, y, z
+    data = data.map(apply_to_mapping(ensure_nd(5)))
+
     if clip_config is not None:
-        y_source = data.map(clip_sample(clip_config))
+        clip_function = clip_sample(clip_config)
     else:
-        y_source = data
-    y = y_source.map(select_keys(output_variables))
-    X = data.map(apply_to_mapping(append_halos_tensor(n_halo))).map(
-        select_keys(input_variables)
-    )
-    # now that we have halos, we need to collapse tile back into the sample dimension
-    X = X.unbatch()
-    y = y.unbatch()
-    return tf.data.Dataset.zip((X, y))
+
+        def clip_function(data):
+            return data
+
+    def map_fn(data):
+        # clipping of inputs happens within the keras model, we don't clip at the
+        # data layer so that the model still takes full-sized inputs when used
+        # in production
+        x = apply_to_tuple(append_halos_tensor(n_halo))(
+            select_keys(input_variables, data)
+        )
+        y = select_keys(output_variables, clip_function(data))
+        return x, y
+
+    # unbatch to fold sample dimension into tile dimension as new sample dimension
+    return data.map(map_fn).unbatch()
 
 
 @register_training_function("convolutional", ConvolutionalHyperparameters)
@@ -125,7 +136,13 @@ def train_convolutional_model(
     train_Xy = get_Xy(data=train_batches, clip_config=hyperparameters.clip_config)
     # need unclipped shapes for build_model
     # have to batch data so we can take statistics for scaling transforms
-    X, y = next(iter(get_Xy(data=train_batches, clip_config=None).batch(10_000_000)))
+    X, y = next(
+        iter(
+            get_Xy(data=train_batches, clip_config=None)
+            .unbatch()
+            .batch(hyperparameters.normalization_fit_samples)
+        )
+    )
 
     train_model, predict_model = build_model(hyperparameters, X=X, y=y)
     hyperparameters.training_loop.fit_loop(
--- a/external/fv3fit/fv3fit/keras/_models/convolutional.py
+++ b/external/fv3fit/fv3fit/keras/_models/convolutional.py
@@ -77,7 +77,7 @@ def get_Xy_dataset(
     clip_config: Optional[Mapping[Hashable, SliceConfig]],
     n_halo: int,
     data: tf.data.Dataset,
-):
+) -> tf.data.Dataset:
     """
     Given a tf.data.Dataset with mappings from variable name to samples,
     return a tf.data.Dataset whose entries are two tuples, the first containing the
--- a/external/fv3fit/fv3fit/keras/_models/shared/halos.py
+++ b/external/fv3fit/fv3fit/keras/_models/shared/halos.py
@@ -193,8 +193,10 @@ def _append_halos_tensor(n_halo: int, tensor: tf.Tensor) -> tf.Tensor:
         for _ in range(6)
     ]
     for i, quantity in enumerate(quantities):
+        # split tensor into quantity list over the tile dimension
         quantity.view[:] = tensor[:, i, :].numpy()
     _halo_update(communicators=communicators, quantities=quantities, n_halo=n_halo)
+    # restore the tile dimension with None and concatenate along it
     out_tensor = tf.concat(
         [quantity.data[:, None, :] for quantity in quantities], axis=1
     )
--- a/external/fv3fit/fv3fit/keras/_models/shared/halos.py
+++ b/external/fv3fit/fv3fit/keras/_models/shared/halos.py
@@ -161,7 +161,7 @@ def append_halos(ds: xr.Dataset, n_halo: int) -> xr.Dataset:
 def append_halos_tensor(n_halo: int, tensor: tf.Tensor) -> tf.Tensor:
     """
     Given a tensor with
-    "tile", "x", "y", and "z" dimensions and no halo data,
+    "sample", "tile", "x", "y", and "z" dimensions and no halo data,
     return a tensor which has n_halo halo points of valid data
     appended to the start and end of the x and y dimensions.
     """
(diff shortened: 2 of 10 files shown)


external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -81,20 +84,28 @@ def get_Xy_dataset(
...
+
+    def map_fn(data):
+        # clipping of inputs happens within the keras model, we don't clip at the
+        # data layer so that the model still takes full-sized inputs when used
+        # in production
+        x = apply_to_tuple(append_halos_tensor(n_halo))(
+            select_keys(input_variables, data)
+        )
+        y = select_keys(output_variables, clip_function(data))
+        return x, y
```
mcgibbon, line 105: This pattern suggested by @nbren12 was already used in the dense code in the original tfdataset PR, adding it to convolutional training here.

external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -125,7 +136,13 @@ def train_convolutional_model(
     train_Xy = get_Xy(data=train_batches, clip_config=hyperparameters.clip_config)
     # need unclipped shapes for build_model
     # have to batch data so we can take statistics for scaling transforms
-    X, y = next(iter(get_Xy(data=train_batches, clip_config=None).batch(10_000_000)))
+    X, y = next(
```
mcgibbon, line 139: Copied over normalization fitting logic from the dense training. Might want to consolidate these later into some `fit_nd_model` helper (instead of only `fit_column_model`).

mcgibbon:
Training log of out-of-memory test script on master:
```
2022-03-23 17:01:36 [INFO] calling train function
Epoch 1/3
    268/Unknown - 5s 11ms/step - loss: 1.8540 - standard_normalize_Q1_loss: 0.9022 - standard_normalize_Q2_loss: 0.95172022-03-23 17:01:48 [INFO] saving end of epoch metrics for epoch 0: Logs(loss=1.851829171180725, val_loss=None)
270/270 [==============================] - 5s 13ms/step - loss: 1.8518 - standard_normalize_Q1_loss: 0.9010 - standard_normalize_Q2_loss: 0.9509
Epoch 2/3
268/270 [============================>.] - ETA: 0s - loss: 1.6183 - standard_normalize_Q1_loss: 0.7378 - standard_normalize_Q2_loss: 0.88052022-03-23 17:01:52 [INFO] saving end of epoch metrics for epoch 1: Logs(loss=1.6177815198898315, val_loss=None)
270/270 [==============================] - 4s 11ms/step - loss: 1.6178 - standard_normalize_Q1_loss: 0.7375 - standard_normalize_Q2_loss: 0.8803
Epoch 3/3
270/270 [==============================] - ETA: 0s - loss: 1.5533 - standard_normalize_Q1_loss: 0.7001 - standard_normalize_Q2_loss: 0.85322022-03-23 17:01:57 [INFO] saving end of epoch metrics for epoch 2: Logs(loss=1.5532615184783936, val_loss=None)
270/270 [==============================] - 4s 11ms/step - loss: 1.5533 - standard_normalize_Q1_loss: 0.7001 - standard_normalize_Q2_loss: 0.8532
```

Training log with current commit of this branch:
```
2022-03-23 17:02:57 [INFO] calling train function
Epoch 1/3
    266/Unknown - 2s 4ms/step - loss: 1.8594 - standard_normalize_Q1_loss: 0.9093 - standard_normalize_Q2_loss: 0.95012022-03-23 17:03:01 [INFO] saving end of epoch metrics for epoch 0: Logs(loss=1.8620867729187012, val_loss=None)
270/270 [==============================] - 2s 5ms/step - loss: 1.8621 - standard_normalize_Q1_loss: 0.9102 - standard_normalize_Q2_loss: 0.9519
Epoch 2/3
263/270 [============================>.] - ETA: 0s - loss: 1.6270 - standard_normalize_Q1_loss: 0.7425 - standard_normalize_Q2_loss: 0.88452022-03-23 17:03:02 [INFO] saving end of epoch metrics for epoch 1: Logs(loss=1.6256746053695679, val_loss=None)
270/270 [==============================] - 1s 3ms/step - loss: 1.6257 - standard_normalize_Q1_loss: 0.7422 - standard_normalize_Q2_loss: 0.8835
Epoch 3/3
270/270 [==============================] - ETA: 0s - loss: 1.5547 - standard_normalize_Q1_loss: 0.7001 - standard_normalize_Q2_loss: 0.85462022-03-23 17:03:04 [INFO] saving end of epoch metrics for epoch 2: Logs(loss=1.5547468662261963, val_loss=None)
270/270 [==============================] - 1s 3ms/step - loss: 1.5547 - standard_normalize_Q1_loss: 0.7001 - standard_normalize_Q2_loss: 0.8546
```

external/fv3fit/fv3fit/keras/_models/convolutional.py
```diff
@@ -81,20 +84,28 @@ def get_Xy_dataset(
...
+        # data layer so that the model still takes full-sized inputs when used
+        # in production
+        x = apply_to_tuple(append_halos_tensor(n_halo))(
+            select_keys(input_variables, data)
+        )
+        y = select_keys(output_variables, clip_function(data))
+        return x, y
+
+    # unbatch to fold sample dimension into tile dimension as new sample dimension
+    return data.map(map_fn).unbatch()
```
nbren12, line 108: While you are here could you add return type hint to this method?

external/fv3fit/fv3fit/keras/_models/shared/halos.py
```diff
@@ -161,7 +161,7 @@ def append_halos(ds: xr.Dataset, n_halo: int) -> xr.Dataset:
 def append_halos_tensor(n_halo: int, tensor: tf.Tensor) -> tf.Tensor:
     """
     Given a tensor with
-    "tile", "x", "y", and "z" dimensions and no halo data,
```
nbren12, line 164: Nice change. 4 dims is more general than 3.

external/fv3fit/fv3fit/keras/_models/shared/halos.py
```diff
@@ -161,7 +161,7 @@ def append_halos(ds: xr.Dataset, n_halo: int) -> xr.Dataset:
...
+    "sample", "tile", "x", "y", and "z" dimensions and no halo data,
```
nbren12, line 164: Can you add a parameter for the name "sample" or maybe infer it automatically? Would be nice to apply e.g. with `time` as  dimension too.

external/fv3fit/fv3fit/keras/_models/shared/halos.py
```diff
@@ -178,24 +178,25 @@ def append_halos_tensor(n_halo: int, tensor: tf.Tensor) -> tf.Tensor:
...
-            origin=(n_halo, n_halo, 0),
-            extent=tuple(tensor.shape[1:]),
+            origin=(0, n_halo, n_halo, 0),
+            extent=tuple([tensor.shape[0]] + list(tensor.shape[2:])),
         )
         for _ in range(6)
     ]
     for i, quantity in enumerate(quantities):
-        quantity.view[:] = tensor[i, :, :, :].numpy()
+        quantity.view[:] = tensor[:, i, :].numpy()
```
nbren12, line 196: It's tricky for me to verify if this is the expected change. How good is your test coverage here? If you introduce a fault on purpose e.g. flip the index order incorrectly does it break some test?

external/fv3fit/fv3fit/keras/_models/shared/training_loop.py
```diff
@@ -94,7 +87,13 @@ def fit_loop(
             callbacks=[EpochCallback(func) for func in callbacks], epochs=self.epochs
         )
         if self.in_memory:
-            Xy_fit, validation_fit = _tfdataset_to_tensor_sequence(Xy, validation_data)
+            Xy_fit = _tfdataset_to_tensor_sequence(Xy)
```
nbren12, line 90: Nice job moving the iffy logic up the call stack.

external/fv3fit/tests/test_tfdataset_ops.py
```diff
@@ -86,10 +86,17 @@ def test_tfdataset_from_batches_single():
         }
     )
     tfdataset = tfdataset_from_batches([xr_dataset])
-    for i, result in enumerate(iter(tfdataset)):
+    result = next(iter(tfdataset))
```
nbren12, line 89: so much better coverage compared to `isinstance(result, dict)` haha!

external/fv3fit/tests/test_training_loop.py
```diff
@@ -79,3 +80,42 @@ def test_fit_loop_calls_fit(n_epochs):
...
+    # is shuffled
+    # import matplotlib.pyplot as plt
+    # plt.figure()
+    # plt.hist2d(array_x, shuffled_x, bins=n_batches)
+    # plt.xlabel("output index")
+    # plt.ylabel("input index")
+    # plt.show()
+
+    # check that there is some inter-batch shuffling
+    hist, _, _ = np.histogram2d(array_x, shuffled_x, bins=n_batches)
```
nbren12, line 118: these tests could be a bit brittle and block future intentional changes in behavior.  I also worry about the runtime..since fv3fit already has a pretty slow test suite. I expect these assertions will survive longer through future refactors if you included them in a unit test rather than this training integration test. 

If there is no such unit to test, the perhaps one should be written since shuffling seems like a separate concern from other aspects of loop fitting.  Can you inject a shuffle implementation into `TrainingLoopConfig`? This would remove the concern of shuffling from `fit_loop`, and allow it to evolve freely/change its shuffling strategy.

external/fv3fit/fv3fit/tfdataset.py
```diff
@@ -13,6 +13,13 @@ def apply_to_mapping(
     return {name: tensor_func(tensor) for name, tensor in data.items()}
 
 
+@curry
+def apply_to_tuple(
```
nbren12, line 17: where is this used? Doesn't seem specific to to `tf.data` since it's more of a general purpose higher-order-function.

nbren12 (approved):
Thanks for following up on this! Would be interested to see a performance comparison.

As far as suggestion, I think the "concept" of shuffling should be reified into some code unit (class or function) so that fit_loop can hot-swap the kind of shuffling it does. There are many ways to shuffle and it seems we switch how we do this all the time.

external/fv3fit/tests/test_training_loop.py
mcgibbon: I'll split the shuffle and rebatching into its own helper function, and add a unit test of that function.

The test is quite fast, actually (2s).

I'd like to VC on changes past that - I intended for these assertions to be things that, at least presently, a user should really think about changing TrainingLoopConfig in a way that breaks the assertions. I think we can compartmentalize this logic into a new swappable component when we need to, but there is significant overhead and complexity when we add swappable components (like Hyperparameters), and I don't think we should do it prematurely. Moving this code to a function like you suggested will at least help keep this as an easy-to-change thing down the line.

nbren12: sounds good for now.

external/fv3fit/fv3fit/keras/_models/shared/training_loop.py
nbren12: i recently experience this too. This is why the `Maybe` class exists in haskell btw:
```
Prelude> times_2 x = 2 * x
Prelude> times_2 <$> Just 2
Just 4
Prelude> times_2 <$> Nothing
Nothing
```
Makes it so much easier to compose. This adds 1, multiples by 2, and then squares:
```
Prelude> (^2) <$> (*2) <$> (+1) <$> Nothing
Nothing
Prelude> (^2) <$> (*2) <$> (+1) <$> Just 1
Just 16
```

PR #1733, opened 2022-03-24
prepare-config: Remove flags from CLI

oliverwm1:
Remove flag options from `prepare-config` CLI. Specifically `--model_url` and `--diagnostics_ml`.

Refactored public API:
- the `prepare-config` CLI now takes only a path to a YAML, instead of the YAML and optional flags.

Significant internal changes:
- modified the e2e argo workflow to patch the prognostic config as required

- [x] Tests added


workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py
```diff
@@ -197,26 +196,6 @@ def _create_arg_parser() -> argparse.ArgumentParser:
...
-            "Remote url to a trained ML model. If a model is omitted (and not "
-            "specified in `user_config`'s `scikit-learn` `model` field either), then "
-            "no ML updating will be done. Also, if an ML model is provided, no "
-            "nudging will be done. Can be provided multiple times, "
-            "ex. --model_url model1 --model_url model2. If multiple urls are given, "
-            "they will be combined into a single model at runtime, providing the "
-            "outputs are nonoverlapping."
-        ),
-    )
-    parser.add_argument(
```
AnnaKwa, line 215: If this info is not in the docs, can you move this info there so it is still somewhere discoverable?

workflows/prognostic_c48_run/docs/config-usage.rst
```diff
@@ -89,7 +87,7 @@ reference analysis data. Here is an example ``minimal.yaml``:
 Machine learning
 ~~~~~~~~~~~~~~~~
 
-A machine learning run can be configured in two ways. The first is by
+A machine learning run is configured by
```
AnnaKwa, line 90: This is an old error, but can you update this to "one or more models"?

AnnaKwa (approved):
Thanks, LGTM- can you move the info about the `diagnostic_ml` option into the docs?

workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py
oliverwm1: The option is documented in the `MachineLearningConfig` dataclass, which is included in the API page on the docs.

PR #1734, opened 2022-03-25
fv3fit.train with emulator model

nbren12:
The model used for microphysics emulation is general purpose so can be used for loaders based training. It has several features that will be useful for projects like @oliverwm1's fine resolution ML with enforced conservation:
- multi target learning
- arbitrary pre/post ML architecture transformations. Can be extended, trained from data (or not), and all are baked into the network architecture.
- different architectures (dense, downwards rnn, etc).

This feature was quite easy to do given fv3fit's extensible configuration design, and @mcgibbon;s recent refactor! Also, both `fv3fit.train` and `fv3fit.train_emulation` had similar challenges and converged on similar solutions. I suggest reviewing commit-by-commit, since some refactoring was involved.

I left the existing entrypoint intact since we haven't figured out how to make `fv3fit.train` handle the data loading pipeline yet. There are a few hacks involved in maintaining this compatibility, but I think it's a fair trade-off for now. There are several integrations that aren't working yet:
- wandb
- Won't be able to configure the normalization since that uses `Enum` objects that require some special handling in `dacite.from_dict`
- 2D outputs. I'm not 100% sure these will work since the microphysics emulator includes a singleton dimension in its outputs.

Nonetheless, this is a good MVP.

### DIFF
--- a/external/fv3fit/fv3fit/emulation/data/load.py
+++ b/external/fv3fit/fv3fit/emulation/data/load.py
@@ -27,7 +27,7 @@ def nc_files_to_tf_dataset(
     """
 
     transform = compose_left(*[open_netcdf_dataset, convert])
-    return seq_to_tfdataset(files, transform).unbatch().prefetch(tf.data.AUTOTUNE)
+    return seq_to_tfdataset(files, transform).prefetch(tf.data.AUTOTUNE)
 
 
 def nc_dir_to_tfdataset(
--- a/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
+++ b/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
@@ -205,7 +205,7 @@ class PureKerasDictPredictor(Predictor):
         Args:
             model: a keras model. Tuple in/out models are okay, but will be cast
                 to a dict-output model before saving.
-            passthru: any data! Is NOT serialized, but can be
+            passthrough: any data! Is NOT serialized, but can be
                 used by other parts of the training workflow. Mostly exists for
                 backwards compatibility and should be avoided in new code.
         """
--- a/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
+++ b/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
@@ -199,14 +199,15 @@ class PureKerasDictPredictor(Predictor):
     _MODEL_FILENAME = "model.tf"
     _CONFIG_FILENAME = "config.yaml"
 
-    def __init__(
-        self, model: tf.keras.Model,
-    ):
+    def __init__(self, model: tf.keras.Model, passthrough: Any = None):
         """Initialize the predictor.
 
         Args:
             model: a keras model. Tuple in/out models are okay, but will be cast
                 to a dict-output model before saving.
+            passthru: any data! Is NOT serialized, but can be
+                used by other parts of the training workflow. Mostly exists for
+                backwards compatibility and should be avoided in new code.
         """
         dict_model = ensure_dict_output(model)
         tuple_model = ensure_tuple_model(dict_model)
@@ -224,6 +225,7 @@ class PureKerasDictPredictor(Predictor):
             # TODO consider this refactor
             unstacked_dims=["z"],
         )
+        self.passthrough = passthrough
 
     def predict(self, X: xr.Dataset) -> xr.Dataset:
         """Predict an output xarray dataset from an input xarray dataset."""
--- a/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
+++ b/external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
@@ -9,6 +9,7 @@ from fv3fit.keras.adapters import (
     ensure_dict_output,
     rename_dict_input,
     rename_dict_output,
+    ensure_tuple_model,
 )
 import tensorflow as tf
 from typing import Any, Dict, Hashable, Iterable, Sequence, Mapping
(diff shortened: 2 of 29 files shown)


external/fv3fit/tests/training/test_train.py
```diff
@@ -206,6 +227,9 @@ def test_train_default_model_on_identity(model_type, regtest):
     The model with default configuration options can learn the identity function,
     using gaussian-sampled data around 0 with unit variance.
     """
+    if model_type == "transformed":
+        pytest.skip()
```
mcgibbon, line 231: Do you know why your code doesn't pass this?

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -175,3 +176,80 @@ def input_sensitivity(self, stacked_sample: xr.Dataset) -> InputSensitivity:
...
+        self._pure_keras_predictor = PureKerasModel(
+            self.input_variables,
+            self.output_variables,
+            tuple_model,
+            # unstacked dims is hardcoded most places in fv3fit.keras, but it
+            # seems like useful metadata, but more of an aspect of the training
+            # data pipeline since it is the piece of code that removes the
+            # dimension names...would be nice to remove this concept...e.g. pass
+            # unstacked_dims to the .dump rather than in the constructor
+            # TODO consider this refactor
```
mcgibbon, line 225: `unstacked_dims` needs to be passed alongside the tf.keras.Model instance, since if you have a xr.Dataset and a tf.keras.Model you need the additional metadata of what are the dimensions the model is expecting (it's `["sample"] + unstacked_dims` where sample is the stacking of anything not in unstacked_dims). A PureKerasModel initialized without `unstacked_dims` cannot `predict`.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
```diff
@@ -175,3 +176,80 @@ def input_sensitivity(self, stacked_sample: xr.Dataset) -> InputSensitivity:
...
+    _MODEL_FILENAME = "model.tf"
+    _CONFIG_FILENAME = "config.yaml"
+
+    def __init__(self, model: tf.keras.Model, passthrough: Any = None):
+        """Initialize the predictor.
+
+        Args:
+            model: a keras model. Tuple in/out models are okay, but will be cast
+                to a dict-output model before saving.
+            passthru: any data! Is NOT serialized, but can be
```
mcgibbon, line 208: suggestion: snake_case? I'm also fine with saying passthrough is one word.
```suggestion
            passthrough: any data! Is NOT serialized, but can be
```

external/fv3fit/tests/emulation/test_train_microphysics.py
```diff
@@ -149,14 +148,11 @@ def test_training_entry_integration(tmp_path):
     main(config)
 
 
-def test_TrainConfig_build_model():
+def test_MicrophysicsHyperParameters_build_model():
```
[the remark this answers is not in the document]
nbren12: I struggled to find a good name since the current functionality overlaps almost 100% with the existing dense model.

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -170,6 +187,83 @@ def input_variables(self) -> Sequence:
...
+        """used for testing"""
+        return TransformedParameters(
+            model=MicrophysicsConfig(
+                input_variables=input_variables,
+                direct_out_variables=output_variables,
+                architecture=ArchitectureConfig("dense"),
+            ),
+            loss=CustomLoss(loss_variables=output_variables),
+            use_wandb=False,
+        )
```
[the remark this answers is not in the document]
nbren12: The class needs to be passed input/output variables, but those aren't top-level attributes, and the loss needs to know the loss variables separately since output_variables are decoupled from the loss. I'll rename to `init_testing`.

external/fv3fit/tests/training/test_train.py
```diff
@@ -69,29 +71,13 @@ class TrainingResult:
...
 def train_identity_model(model_type, sample_func, hyperparameters=None):
     input_variables, output_variables, train_dataset = get_dataset(
         model_type, sample_func
     )
     if hyperparameters is None:
-        hyperparameters = get_default_hyperparameters(
-            model_type, input_variables, output_variables
-        )
+        cls = get_hyperparameter_class(model_type)
+        hyperparameters = cls.get_default_model(input_variables, output_variables)
```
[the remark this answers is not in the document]
nbren12: Renamed to `init_testing`. I agree with your medium term plan, I started on a similar idea, but ended up hitting some road blocks. Somewhere in the code needs to be a table with hyperparameters, training set, and datasets.

external/fv3fit/tests/training/test_train.py
```diff
@@ -113,54 +99,89 @@ def unstack_test_dataset(test_dataset: xr.Dataset):
...
-    # originally we trained on unstacked data, which is why we don't
-    # produce unstacked data directly above and have this conversion here
-    if model_type == "convolutional":
-        train_dataset = stack(train_dataset, unstacked_dims=["tile", "x", "y", "z"])
-    else:
-        train_dataset = stack_non_vertical(train_dataset)
+    return xr.Dataset(data_vars=data_vars)
+
+
+def _get_dataset_precipitative(sample_func):
```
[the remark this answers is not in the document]
nbren12: Yah. I'll punt on it for now. I was planning on it, but the indirection felt a little unnecessary in that case since the function was smaller.

external/fv3fit/tests/training/test_train.py
nbren12: No, but I think it's an intermittent error. To be solved another day I think. This PR added many tests so I think is still a step in the right direction.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
nbren12: I understand. I just chose to put the "z" hardcode here rather than in the training func as you did here: https://github.com/ai2cm/fv3net/blob/04c4d4c358f6f00f635c6803da14046200084ec8/external/fv3fit/fv3fit/keras/_models/dense.py#L230

I think unstacked dims should be passed to dump. The hardcodes exist because the source data has been stripped of this metadata by the time the model is built. The thing that dumps the model knows what the dimensions are though, so I think this would work nicely:
```
predictor.dump(path, unstacked_dims)
```

nbren12: Or maybe we'd have to make a PredictorFactory...ugh:
```
predictor_factory = PredictorFactory(model, inputs, outputs)
# ...in the higher scope
predictor = predictor_factor.build(unstacked_dims)
predictor.dump(path)
```

nbren12: would be easy to do with `functools.partial` or curry.

external/fv3fit/tests/training/test_train.py
mcgibbon: I'd like to pull down your PR and dig into it a little bit, if that's OK. Alternatively, have you already or can you test the refactored code on a production case?

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
mcgibbon: Does the thing that dumps the code know the dimensions? I thought `.dump` gets called in `fv3fit.train`, but the stacked dimensions are only known inside the context of a training function (where they are assumed). We'd have to extract the unstacked dimensions from the training function into the top level to pass this directly to `dump`.

Specializing the predictor on column prediction by passing "z" explicitly here is fine. We can always refactor it if we use this for more than columnwise prediction (e.g. the convolutional model).

mcgibbon: We should talk more if we want to later implement a change like this - as I understand it, the scope in `fv3fit.train` where `dump` is called does not have access to / any dependence on `unstacked_dims`.

external/fv3fit/fv3fit/emulation/data/load.py
```diff
@@ -27,7 +27,7 @@ def nc_files_to_tf_dataset(
     """
 
     transform = compose_left(*[open_netcdf_dataset, convert])
-    return seq_to_tfdataset(files, transform).unbatch().prefetch(tf.data.AUTOTUNE)
+    return seq_to_tfdataset(files, transform).prefetch(tf.data.AUTOTUNE)
```
mcgibbon, line 30: Double-checking this: I don't see any changes to calls in the non-testing code of `nc_files_to_tf_dataset`. Is this an issue / have I overlooked one?

nbren12: There are changes. easy to see in commit view: https://github.com/ai2cm/fv3net/pull/1734/commits/124c7b3e4813dcabd4e9aab11b33440715ae52e7

nbren12: and we have good coverage here.

external/fv3fit/fv3fit/keras/_models/shared/pure_keras.py
nbren12: I usually find the pulling info that needs to be shared to the mini-max scope and injecting it where needed is a good option. Lot's of ways to do it...maybe the train funcs get passed it as an arg if they should know about the dimension metadata.

mcgibbon: It feels like hardcoding it in the training function is fine because this information is already contextually hardcoded everywhere. e.g. the dense training function assumes its arrays are [sample, z] when it does clipping or when it does model building, and the convolutional training assumes the dimension order when it applies halo updates. In the training function, there isn't a need to make the dimensions arbitrary and it keeps things much simpler if they're not. Kind of like when you decorate the required dims for xarray apply_ufuncs.

mcgibbon: Another way to say this is, the information flow would be in the opposite direction. The training functions are all hard-coded to a specific unstacked_dims, and would need to pass this upwards to `fv3fit.train` if it needed it (e.g. to assert the data being loaded has the correct unstacked_dims).

nbren12: That does make sense. Thanks for the explanation.

PR #1736, opened 2022-03-25
use more batches in shuffling test

### LINKED ISSUE #1735: fv3fit: shuffling test is flaky
@mcgibbon I just got a test failure on the new training loop randomness test:
```
FAILED tests/test_training_loop.py::test__shuffle_batched_tfdataset - assert (-0.3750280852770701 ** 2) < 0.1
```

Should we set the random seed or relax the 0.1 threshold or both?

mcgibbon:
This PR increases the number of batches shuffled in the shuffling test from 20 to 200 by making each of the batches smaller, which should be large enough to give consistent shuffling results. In manual testing, 5 tests did not produce a correlation over 0.002, and the threshold checked is 0.01.

Fixes #1735


AnnaKwa (approved):
LGTM

PR #1740, opened 2022-03-29
add report.upload

nbren12:
Add a tool for uploading an html document with the appropriate metadata to GCS. If run in interactive mode it will ask for permission before uploading. Otherwise (if stdin is not a TTY) it will just upload.


external/report/report/_io.py
```diff
@@ -0,0 +1,28 @@
+import pathlib
+import hashlib
+import sys
+from google.cloud import storage
```
oliverwm1, line 4: I'm a bit skeptical about adding `google.cloud.storage` as a dependency of this package. What about having this code in `vcm.cloud`?

nbren12: Why?  It's pretty convenient to have it here...and the logic is html specific rather than general i/o. Uploading a report to a shareable link is a pretty important part of the report workflow, so makes sense to bundle it IMO. I was inspired by this feature of datapane: https://docs.datapane.com/reports/publishing-and-sharing.

GCS is a pure python package and is a google GA product, so quite stable. Adding vcm as a dependency to report would have a much larger impact on the installed footprint of report...GCS is a dep of vcm after all.

oliverwm1 (approved):
Looks like a useful tool. Approving, but could you add the new dep to the `report` requirements?

external/report/report/_io.py
```diff
@@ -0,0 +1,28 @@
...
+
+def upload_html(path: str, contents: str):
+    client = storage.Client(project=PROJECT)
+    bucket = client.get_bucket(BUCKET)
+    blob = bucket.blob(path)
+    blob.content_type = "text/html"
+    blob.upload_from_string(contents.encode(), content_type=blob.content_type)
+
+
+def upload(html: str):
```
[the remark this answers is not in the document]
nbren12: maybe extending this CLI to work on dicts of `path: html`?

PR #1741, opened 2022-03-29
fv3fit: Fix packer multi-index bug

### LINKED ISSUE #1742: Random Forest model predicts dataset with a too-large z dim
See failing test on this PR: #1741

The RF model 'predict' method sometimes returns a dataset with the wrong z-dim length. This does not happen if the RF only has one output. For two outputs, it predicts a dataset with a z-dim length that is twice as long as it should be. This bug does not exist prior to #1685.

cc @mcgibbon

oliverwm1:
Resolves #1742


mcgibbon (approved):
May or may want to have someone else review since I wrote this code. Or if you think it's good you can be the reviewer on your own PR by merging.

Issue #1744, opened 2022-03-30
Running `make lock_pip` currently fails


spencerkclark:
Running `make lock_pip` currently fails with the development version of fv3net in a fresh environment created by `make create_environment`.  The truncated output containing the relevant error looks like this:

```
pip-compile --no-annotate \
	--output-file docker/fv3fit/requirements.txt \
	external/fv3fit/setup.py \
	external/loaders/setup.py \
	external/vcm/setup.py
Could not find a version that matches vcm (from loaders (external/loaders/setup.py))
No versions found
Was https://pypi.org/simple reachable?
Makefile:218: recipe for target 'docker/fv3fit/requirements.txt' failed
make: *** [docker/fv3fit/requirements.txt] Error 2
```

A solution suggested by @nbren12 is to remove `"vcm"` from the `setup.py` file in the loaders packages.  This indeed fixes the issue.

PR #1745, opened 2022-03-30
fix `make lock_pip`

### LINKED ISSUE #1721: make constraints.txt fails on mac
`make constraints.txt` fails with
```
pip._internal.exceptions.DistributionNotFound: No matching distribution found for google-python-cloud-debugger==2.18 (from -r .dataflow-versions.txt (line 48))
```
due to this package not being available on mac.

There are a couple different ways to fix it, like using docker to compile the constraints, or filtering this requirement out.

nbren12:
Resolves #1721 and #1744. Also, ensures that this cannot happen again.

@frodre The fv3fit image was using pretty out of date dependencies due to a bug with how lock_pip worked. Is it possible this was causing some of our issues with performance regressions?


constraints.txt
```diff
@@ -11,6 +11,7 @@ altair==4.1.0
 anyio==2.2.0
 apache-beam==2.37.0
 appdirs==1.4.4
+appnope==0.1.2
```
nbren12, line 14: BSD.

nbren12 (commented):
new deps ok.

spencerkclark (approved):
Looks good to me.  Hopefully the outage impacting the image builds will be fixed soon.

PR #1746, opened 2022-03-30
Add env vars for wandb to argo templates

AnnaKwa:
This adds environment variables `WANDB_PROJECT` and `WANDB_ENTITY` to the argo training workflow so that wandb enabled jobs will be logged. The default project is `argo-default` and can be set via the parameter `wandb-project`. The entity is always `ai2cm`. This is fixed since if you try to set this to your personal account it probably won't work with permissions that weren't generated using your account's API key.


nbren12 (approved):
LGTM. maybe want to also add WANDB_TAGS and WANDB_RUN_GROUP. We've found those quite handy.

PR #1748, opened 2022-03-30
Add progrun url to artifact metdata for prognostic evaluation fix

frodre:


frodre:
https://forums.developer.nvidia.com/t/error-apt-get-updating-from-nvidia-cuda11-2-1-base-ubuntu20-04/209836/8

projects/microphysics/scripts/prognostic_run.py
```diff
@@ -41,7 +41,11 @@ def run_from_config_dict(config: dict):
         logging.info(f"Running segment {i+1} of {segments}")
         api.append(url)
 
-    artifact = wandb.Artifact(tag, type="prognostic-run")
+    # Might be a bug in wandb that doesn't let us grab references to files below
+    # the added reference directory level when checksum=False, adding to metadata
+    # allows us to have a ref without downloading whole directory, keeping ref in
+    # case downloading is important later
+    artifact = wandb.Artifact(tag, type="prognostic-run", metadata={"url": url})
```
nbren12, line 48: Feels a bit redundant, and won't be backwards compatible. Just did some poking around and it looks like this works:
```
In [24]: entry = next(iter(artifact.manifest.entries.values()))

In [25]: entry.ref
Out[25]: 'gs://vcm-ml-experiments/microphysics-emulation/2022-03-30/dqc-precpd-limiter-rnn-limited-qc-87e803-6hr-v3-offline'

In [26]: artifact = run.use_artifact('ai2cm/microphysics-emulation/dqc-precpd-limiter-rnn-limited-qc-87e803-6hr-v3-offline:v0'
    ...: , type='prognostic-run')
    ...:

In [27]: entry = next(iter(artifact.manifest.entries.values()))

In [28]: entry.ref
Out[28]: 'gs://vcm-ml-experiments/microphysics-emulation/2022-03-30/dqc-precpd-limiter-rnn-limited-qc-87e803-6hr-v3-offline'
```
however this does recurse pretty far into the artifact data structure and might break if we ever add more artifacts in `prognostic_run.py`.

nbren12: Another approach to consider is the one used by the piggy-backed diags: https://github.com/ai2cm/fv3net/blob/cca7bc1f85422f36de0a7a5cfb7ae4e8eb247b22/workflows/diagnostics/fv3net/diagnostics/prognostic_run/emulation/single_run.py#L281-L283.

docker/prognostic_run/Dockerfile
```diff
@@ -2,6 +2,8 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE} as bld
 
+RUN rm /etc/apt/sources.list.d/cuda.list && rm /etc/apt/sources.list.d/nvidia-ml.list || true
```
nbren12, line 5: should we just wait for nvidia to fix their bug?

Also, you can simplify this a bit with `rm -f`. It will exit 0 even if the files don't exist.

```suggestion
RUN rm -f /etc/apt/sources.list.d/cuda.list /etc/apt/sources.list.d/nvidia-ml.list
```

Can you add to `docker/fv3fit/Dockerfile` as well?

nbren12 (approved):
Thanks so much for fixing the various diagnostics and looking into the nvidia issue.

docker/prognostic_run/Dockerfile
frodre: I think since we don't need to install any nvidia specific stuff, we can just leave this around, but I'll add a comment as to why it's there and that it can be removed if there's a need to apt-get nvidia libraries.

projects/microphysics/scripts/prognostic_run.py
frodre: I switched to the run-filter style.  I still reference the artifact so the wandb link is there, but nothing is referenced for now.

PR #1749, opened 2022-03-30
fine-res: allow configuring prog run with multiple models

oliverwm1:
We may want to use separate ML models to predict different parts of the ML target. This will allow that.

Refactored public API:
- `runtime.transformers.fv3fit.Config.url` is now `Sequence[str]` instead of `str`


workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -34,7 +34,8 @@ class Adapter:
     timestep: float
 
     def __post_init__(self: "Adapter"):
-        self.model = fv3fit.load(self.config.url)
+        models = [fv3fit.load(url) for url in self.config.url]
+        self.model = MultiModelAdapter(models)  # type: ignore
```
oliverwm1, line 38: `MultiModelAdapter` expects models that have been wrapped with `RenamingAdapter`, but I don't want to bother with that complication here. In practice no one ever uses the renaming feature anyway.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -60,4 +61,4 @@ def partial_fit(self, inputs: State, state: State):
 
     @property
     def input_variables(self) -> Iterable[Hashable]:
-        return self.model.input_variables
+        return list(set(self.model.input_variables) | set(self.config.variables))
```
oliverwm1, line 64: ensure we have both the variables needed for ML prediction (i.e. the ML inputs) and the state variables that we are applying the predictions to.

nbren12 (approved):
approving preemptively...but any tested configs (e.g. in the end to end test and prepare_cofnig tests) will need to be modified.

PR #1750, opened 2022-03-30
Add a MSE-conserving humidity limiter

oliverwm1:
The current specific humidity limiter used in prognostic runs does not conserve MSE. This adds a new limiter that is used for fine-res ML, where it is particularly important to conserve MSE. Furthermore, the 'standard ML' prognostic runs are configured to use the new limiter by default, but there is an option to use the old-style limiter to maintain backwards compatibility. The prognostic run regression test is set to use the old-style, to show that no changes in output occur from this PR if the config is properly set.

New public API:
- `MachineLearningConfig` has new `use_mse_conserving_humidity_limiter` entry, which is true by default

Significant internal changes:
- use new MSE-conserving specific humidity limiter in fine-res ML, and by default in 'standard' prognostic run ML

- [x] Tests added


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -297,7 +297,10 @@ def _get_postphysics_stepper(
             self._log_info("Using MLStepper for postphysics updates")
             model = self._open_model(config.scikit_learn, "_postphysics")
             stepper: Optional[Stepper] = PureMLStepper(
-                model, self._timestep, hydrostatic
+                model,
+                self._timestep,
+                hydrostatic,
+                config.scikit_learn.use_mse_conserving_humidity_limiter,
```
brianhenn, line 303: As an extra precaution, can you log which specific humidity limiter is used at initialization?

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -200,9 +227,17 @@ def __call__(self, time, state):
         dQ1_initial = tendency.get("dQ1", xr.zeros_like(state[SPHUM]))
         dQ2_initial = tendency.get("dQ2", xr.zeros_like(state[SPHUM]))
 
-        dQ1_updated, dQ2_updated = non_negative_sphum(
-            state[SPHUM], dQ1_initial, dQ2_initial, dt=self.timestep,
-        )
+        if self.mse_conserving_limiter:
```
brianhenn, line 230: Could this routine be wrapped into a single `moisture_conserving_limiter` function that takes in `sphum`, `dQ2`, `dt`, `Optional[dQ1]` and returns a dataset of updated `dQ2` and optionally `dQ1`? The point being that the same function could be called from both the machine learning stepper and the fv3fit step transformer, which seems easier to track and test.

workflows/prognostic_c48_run/runtime/loop.py
oliverwm1: It will be recorded in the `fv3config.yml`, but yeah we can also log.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
oliverwm1: Done.

brianhenn (approved):
Thanks @oliverwm1, will be interesting to see how much this helps.

PR #1751, opened 2022-03-30
Log commit SHA in wandb

frodre:
It's useful to know which image runs on Weights & Biases were performed on, so now if available under the environment variable `COMMIT_SHA`, the commit SHA  is logged under `wandb.config["env"]` for all scripts currently logging to wandb.


nbren12 (approved):
Great stuff! Thanks so much.

nbren12:
@AnnaKwa Should we do the same for `fv3fit.train`?

frodre:
> @AnnaKwa Should we do the same for `fv3fit.train`?

I added a SHA log in the `fv3fit.train` routine on this PR.  Anywhere I saw wandb logging happening will log the SHA if it's in the environment.

nbren12:
Oh great. Thanks @frodre.

PR #1755, opened 2022-04-01
Allow updating state in fv3fit transformer

oliverwm1:
This PR allows using fine-res ML predictions as direct state updates in addition to as tendencies. It also allows the prediction of multiple tendencies that are all intended to be applied to the same state variable.

Refactored public API:
- Delete `runtime.transformers.fv3fit.Config.variables`
- Add `runtime.transformers.fv3fit.Config.tendency_predictions` and `runtime.transformers.fv3fit.Config.state_predictions` and note these are a mapping from ML output name to FV3GFS state variable name

Requirement changes:
- Update `fv3gfs-wrapper` submodule to latest master commit to allow setting of `surface_precipitation_rate`

- [x] Tests added

### DIFF
--- a/external/vcm/vcm/data_transform.py
+++ b/external/vcm/vcm/data_transform.py
@@ -32,6 +32,7 @@ TransformName = Literal[
     "Q2_tendency_from_Q2_flux",
     "implied_surface_precipitation_rate",
     "implied_downward_radiative_flux_at_surface",
+    "implied_downward_radiative_flux_at_surface_no_fine_res_t_nudging",
 ]
 
 
@@ -238,6 +239,37 @@ def implied_downward_radiative_flux_at_surface(
     toa_net_flux = ds[DSW_TOA] - ds[USW_TOA] - ds[ULW_TOA]
     if include_temperature_nudging:
         toa_net_flux += ds[COL_T_NUDGE]
+    surface_upward_flux = ds[LHF] + ds[SHF] + ds[USW_SFC] + ds[ULW_SFC]
+    surface_downward_flux = _tendency_to_implied_surface_downward_flux(
+        ds["Qm"], toa_net_flux, surface_upward_flux, ds[DELP], dim="z", rectify=rectify
+    )
+    surface_downward_flux = surface_downward_flux.assign_attrs(
+        units="W/m**2",
+        long_name="Implied downward radiative flux from <Qm> budget closure",
+    )
+    ds["implied_downward_radiative_flux_at_surface"] = surface_downward_flux
+    return ds
+
+
+@register(
+    [
+        "Qm",
+        DELP,
+        DLW_SFC,
+        DSW_SFC,
+        DSW_TOA,
+        ULW_SFC,
+        ULW_TOA,
+        USW_SFC,
+        USW_TOA,
+        LHF,
+        SHF,
+    ],
+    ["implied_downward_radiative_flux_at_surface"],
+)
+def implied_downward_radiative_flux_at_surface_no_fine_res_t_nudging(ds, rectify=True):
+    """Assuming <Qm> = SHF + LHF + R_net + <T_nudge>."""
+    toa_net_flux = ds[DSW_TOA] - ds[USW_TOA] - ds[ULW_TOA]
     surface_upward_flux = ds[LHF] + ds[SHF] + ds[USW_SFC] + ds[ULW_SFC]
     surface_downward_flux = _tendency_to_implied_surface_downward_flux(
         ds["Qm"], toa_net_flux, surface_upward_flux, ds[DELP], dim="z", rectify=rectify
--- a/external/vcm/vcm/data_transform.py
+++ b/external/vcm/vcm/data_transform.py
@@ -32,7 +32,6 @@ TransformName = Literal[
     "Q2_tendency_from_Q2_flux",
     "implied_surface_precipitation_rate",
     "implied_downward_radiative_flux_at_surface",
-    "implied_downward_radiative_flux_at_surface_no_fine_res_t_nudging",
 ]
 
 
@@ -239,37 +238,6 @@ def implied_downward_radiative_flux_at_surface(
     toa_net_flux = ds[DSW_TOA] - ds[USW_TOA] - ds[ULW_TOA]
     if include_temperature_nudging:
         toa_net_flux += ds[COL_T_NUDGE]
-    surface_upward_flux = ds[LHF] + ds[SHF] + ds[USW_SFC] + ds[ULW_SFC]
-    surface_downward_flux = _tendency_to_implied_surface_downward_flux(
-        ds["Qm"], toa_net_flux, surface_upward_flux, ds[DELP], dim="z", rectify=rectify
-    )
-    surface_downward_flux = surface_downward_flux.assign_attrs(
-        units="W/m**2",
-        long_name="Implied downward radiative flux from <Qm> budget closure",
-    )
-    ds["implied_downward_radiative_flux_at_surface"] = surface_downward_flux
-    return ds
-
-
-@register(
-    [
-        "Qm",
-        DELP,
-        DLW_SFC,
-        DSW_SFC,
-        DSW_TOA,
-        ULW_SFC,
-        ULW_TOA,
-        USW_SFC,
-        USW_TOA,
-        LHF,
-        SHF,
-    ],
-    ["implied_downward_radiative_flux_at_surface"],
-)
-def implied_downward_radiative_flux_at_surface_no_fine_res_t_nudging(ds, rectify=True):
-    """Assuming <Qm> = SHF + LHF + R_net + <T_nudge>."""
-    toa_net_flux = ds[DSW_TOA] - ds[USW_TOA] - ds[ULW_TOA]
     surface_upward_flux = ds[LHF] + ds[SHF] + ds[USW_SFC] + ds[ULW_SFC]
     surface_downward_flux = _tendency_to_implied_surface_downward_flux(
         ds["Qm"], toa_net_flux, surface_upward_flux, ds[DELP], dim="z", rectify=rectify
--- a/workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
+++ b/workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
@@ -1,9 +1,12 @@
 import dataclasses
-from typing import Mapping, MutableMapping, Iterable, Hashable, Sequence, Tuple
+from typing import Mapping, MutableMapping, Iterable, Hashable, Sequence, Literal
 
 import xarray as xr
 import fv3fit
-from runtime.steppers.machine_learning import non_negative_sphum_mse_conserving
+from runtime.steppers.machine_learning import (
+    non_negative_sphum_mse_conserving,
+    MultiModelAdapter,
+)
 from runtime.types import State
 from runtime.names import SPHUM, TEMP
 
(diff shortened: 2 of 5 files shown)


external/vcm/vcm/data_transform.py
```diff
@@ -250,6 +251,37 @@ def implied_downward_radiative_flux_at_surface(
...
+        ULW_SFC,
+        ULW_TOA,
+        USW_SFC,
+        USW_TOA,
+        LHF,
+        SHF,
+    ],
+    ["implied_downward_radiative_flux_at_surface"],
+)
+def implied_downward_radiative_flux_at_surface_no_fine_res_t_nudging(ds, rectify=True):
```
nbren12: I realize there are no type hints below, but maybe nice to add in new code.

external/vcm/vcm/data_transform.py
```diff
@@ -250,6 +251,37 @@ def implied_downward_radiative_flux_at_surface(
...
+    """Assuming <Qm> = SHF + LHF + R_net + <T_nudge>."""
+    toa_net_flux = ds[DSW_TOA] - ds[USW_TOA] - ds[ULW_TOA]
```
nbren12: Suggest renaming to `toa_net_flux_down`

external/vcm/vcm/data_transform.py
```diff
@@ -250,6 +251,37 @@ def implied_downward_radiative_flux_at_surface(
...
+    surface_upward_flux = ds[LHF] + ds[SHF] + ds[USW_SFC] + ds[ULW_SFC]
+    surface_downward_flux = _tendency_to_implied_surface_downward_flux(
+        ds["Qm"], toa_net_flux, surface_upward_flux, ds[DELP], dim="z", rectify=rectify
```
nbren12: func takes many args...maybe shouldn't be a function. Would at least refactor any vertical averaging out of there.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -36,31 +35,67 @@ class Config:
     online: bool = True
 
 
+class TendencyOrStateMultiModelAdapter:
```
nbren12, line 37: Would be nice to isolate `TendencyOrState` and `MultiModel` into separate abstrations. These are independent concerns

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -66,7 +71,45 @@ def add_one_to_temperature():
     regression_state(out, regtest)
 
 
-def test_multimodel_adapter(state, tmpdir_factory):
+def test_multimodel_adapter(state):
+    nz = 63
+    outputs = {
+        "Q1": np.full(nz, 1 / 86400),
+        "tendency_of_air_temperature_due_to_nudging": np.full(nz, 1 / 86400),
+        "implied_surface_precipitation_rate": 10 / 86400,
```
nbren12, line 79: pretty complex mocking logic here with hard-coded names. Suggests the underlying APIs are not adequately decoupled.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -129,7 +128,7 @@ def non_negative_sphum_limiter(self, tendencies, inputs):
         return limited_tendencies
 
 
-TOTAL_PRECIP_RATE = "total_precipitation_rate"
+PRECIP_RATE = "surface_precipitation_rate"
```
nbren12: Is this a breaking change for the diagnostics/logged metrics? Not sure why the names needed to change.

nbren12 (changes requested):
Thanks Oli. I don't think it's pre-mature overengineering, but I think the `MulitModelAdapter` does too much and as a result its test is pretty complicated. I have some ideas about how to make the design a little more decoupled that we VC about.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -41,39 +46,47 @@ def __post_init__(self: "Adapter"):
...
             with xr.set_options(keep_attrs=True):
-                state_prediction[variable_name] = (
-                    inputs[variable_name] + tendencies[tendency_name] * self.timestep
+                state_update[variable_name] = (
+                    inputs[variable_name] + prediction[tendency_name] * self.timestep
                 )
-        return state_prediction
+        for variable_name, prediction_name in self.config.state_predictions.items():
+            state_update[variable_name] = prediction[prediction_name]
+        print(state_update)
```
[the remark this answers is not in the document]
oliverwm1: Removed.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -84,3 +97,22 @@ def non_negative_sphum_limiter(self, tendencies, inputs):
...
+
+
+TOTAL_PRECIP_RATE = "total_precipitation_rate"
+TOTAL_PRECIP = "total_precipitation"  # has units of m
+
+
+# TODO: define setter for 'surface_precipitation_rate' in fv3gfs-wrapper
+# so that we do not need to do this conversion here. For now, this function
+# is copied from loop.py
+def _replace_precip_rate_with_accumulation(  # type: ignore
```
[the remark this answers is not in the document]
oliverwm1: Code no longer necessary (or existing...) with the fv3gfs-wrapper update.

external/vcm/vcm/data_transform.py
oliverwm1: I reverted the commit that added this code.

external/vcm/vcm/data_transform.py
oliverwm1: no longer adding this.

external/vcm/vcm/data_transform.py
oliverwm1: no longer adding this.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
oliverwm1: Change no longer in PR.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
oliverwm1: new tests are a bit simpler

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
oliverwm1: Yeah the MultiModel part had no need to be in here. I've modified implementation to focus on the "inversion" of a dataset that is keyed on ML output names to one that is keyed on the target state variable names (and which has done the sum over tendencies if necessary).

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -18,15 +19,19 @@ class Config:
     """
     Attributes:
         url: Sequence of paths to models that can be loaded with fv3fit.load.
-        variables: Mapping from state names to name of corresponding tendency predicted
-            by model. For example: {"air_temperature": "dQ1"}.
+        tendency_predictions: Mapping from state names to name of corresponding tendency
```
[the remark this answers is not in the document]
oliverwm1: I went with something similar to this.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -71,42 +89,41 @@ def add_one_to_temperature():
...
-    )
-    predictor_two = fv3fit.testing.ConstantOutputPredictor(
-        input_variables=["air_temperature", "land_sea_mask"],
-        output_variables=["tendency_of_air_temperature_due_to_nudging"],
-    )
-    predictor_one.set_outputs(**outputs)
-    predictor_two.set_outputs(**outputs)
-    multi_model_predictor = TendencyOrStateMultiModelAdapter(
-        [predictor_one, predictor_two],
+def test_prediction_inverter():
```
nbren12, line 92: This test is still complex because it's testing things like timesteps via an integration test that could be covered by smaller unit tests. It also uses hardcoded names/constants unecessarily.


Here's how I would write the tests to follow the abstractions you've defined. First, the appliers are trivial to test individually:
```
def test_MLOutputApplier_tendnecy():
    x = ...some dataset with "a" and "tendency"
    applier = MLOutputApplier("a", "tendency", timestep=1)
    assert applier.apply(x) - x.a == x.tendency # won't work, but you know what I mean

def test_ML_OutputApplier_state():
     # super easy to test
```
With this unit test done, I now have coverage of input to output name translation, timesteps, tendencies, etc. I would intentionally avoid these concepts when testing `PredictionInverter`. Assuming you refactor the `.invert` method as I suggest, then there is almost nothing to "unit" test" any longer...the `Inverter` mostly just wires some other objects together without changing any data. mypy and other integration tests should cover this wiring logic well enough. Since it doesn't do much, I'd put a high probability on the `Inverter` changing in some future refactor, and it would be nice not to have this test prevent that from happening. OTOH, "MLOutputApplier" will likely be a much more resilient API.

tl;dr: add some unit tests of `MLOuptutApplier` and delete this test.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -1,101 +1,106 @@
...
-        return xr.Dataset(tendencies), xr.Dataset(state_updates)
+        output: MutableMapping[Hashable, xr.DataArray] = {}
+        for ml_output_name, target in self.targets.items():
+            if target.method == "tendency":
+                if target.target_name in output:
+                    output[target.target_name] += predictions[ml_output_name]
+                else:
+                    output[target.target_name] = predictions[ml_output_name]
+            elif target.method == "state":
+                output[target.target_name] = predictions[ml_output_name]
```
nbren12, line 94: There is some strong coupling here. Inverter should not know `tendency.method` it should ask MLOutputApplier for all the information it needs. Simple to fix this by refactoring this to a method of `MLOutputApplier` e.g. `get_target_name`. Once this refactor is done, it's probably possible to delete `PredictorInverter` it's method are basically one-liners at that point.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -66,7 +89,44 @@ def add_one_to_temperature():
...
+    assert PRECIP in inverted_prediction
+    xr.testing.assert_allclose(
+        inverted_prediction[TEMP], xr.DataArray(np.full(NZ, 2 / 86400), dims=("z"))
+    )
+
+
+def test_prediction_inverter_raises_multiple_state_update_error():
+    with pytest.raises(ValueError):
+        PredictionInverter(
+            {Q1: MLOutputApplier(TEMP, "state"), Q2: MLOutputApplier(TEMP, "state")}
```
nbren12, line 118: Can you avoid the globals here:
```suggestion
            {"q1": MLOutputApplier("t", "state"), "q2": MLOutputApplier("t", "state")}
```

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -66,7 +89,44 @@ def add_one_to_temperature():
...
+        )
+
+
+def test_prediction_inverter_raises_state_and_tendency_error():
+    with pytest.raises(ValueError):
+        PredictionInverter(
+            {Q1: MLOutputApplier(TEMP, "tendency"), Q2: MLOutputApplier(TEMP, "state")}
```
nbren12, line 125: same here. using globals opens a back-door to coupling. (e.g. MLOutputApplier also starts using the global unnecessarily).

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -1,36 +1,108 @@
...
+        return output
+
+    def apply(self, inputs: State, predictions: State, timestep: float) -> State:
+        """Apply tendency and state updates to the input state."""
+        apply_methods = {t.target_name: t.apply for t in self.targets.values()}
+        output: MutableMapping[Hashable, xr.DataArray] = {}
+        for name, apply in apply_methods.items():
+            output[name] = apply(inputs, predictions[name], timestep)
+        return output
```
nbren12, line 103:
```suggestion
        return {t.target_name: t.apply(inputs, predictions[t.target_name], timestep) for t in self.targets}
```

nbren12 (changes requested):
Thanks. I think we're getting close. See my comments below. I think with my suggestions you can delete `PredictorInverter` and its test.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -1,101 +1,106 @@
...
+
+    Attrs:
+        target_name: Name of the variable to apply ML predictions to.
+        method: How to apply ML predictions to the target variable.
+    """
+
+    target_name: str
+    method: Literal["tendency", "state"]
+
+    def apply(
```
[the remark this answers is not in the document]
nbren12: nvm. YAGNI.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -1,101 +1,106 @@
...
+    MultiModelAdapter,
+)
 from runtime.types import State
 from runtime.names import SPHUM, TEMP
 
 __all__ = ["Config", "Adapter"]
 
 
+@dataclasses.dataclass
+class MLOutputApplier:
```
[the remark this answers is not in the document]
oliverwm1: Yeah, there is a tension between appropriately naming config parameters and having parameters that dacite is able to interpret. This is the other option I was considering, but didn't go with since it felt like each Applier class should use the same variable name for the target name parameter (e.g. `target_name`):

```
@dataclasses.dataclass
class TendencyApplier:
    tendency_update: str
```

```
@dataclasses.dataclass
class StateApplier:
    state_update: str
```
In this case we could have `output_targets: Mapping[str, Unioin[TendencyApplier, StateApplier]]`.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
oliverwm1: Not totally following here. It is not possible to apply the tendencies one-by-one to the state, because of the desire to limit the **total** tendency. So we need to sum the tendencies for each variable, then pass these summed tendencies to the limiter, and then apply to the state. This test exists primarily to ensure that the tendencies are summed correctly.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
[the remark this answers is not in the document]
nbren12: Let's just leave it as is for now. I think there's room for some pro-active development (maybe a package) for polymorphic configs since dacite-alone is not doing what we want, and this pattern has been showing up elsewhere.

nbren12 (approved):
Approving. The abstractions seem a bit heavy, but I think that's okay for now as long as we're ok removing any tests that block future refactors. The interesting behaviors to "unit" test all seem pretty tautological or can be written as one-liners/dict comprehensions:
- e.g. sum of tendencies = sum of tendencies
- state_after = state_before + timestep * tendency
- group by target variable if a tendency

PR #1760, opened 2022-04-06
offline diags: check for either type of derived model

oliverwm1:
For the input sensitivity figure in the offline diags report, it is necessary to access the "base model" if the provided predictor is a `DerivedModel` or `TransformedPredictor`. Currently the offline diags only checks for a `DerivedModel`.

Significant internal changes:
- if predictor is an instance of `fv3fit.TransformedPredictor` or `fv3fit.DerivedModel`, keep grabbing the `base_model` until we get a model that is not derived.


AnnaKwa (approved):
LGTM

PR #1772, opened 2022-04-13
Enable wandb for offline diags

AnnaKwa:
Added public API:
- `--no-wandb` flag in offline report step adds the option to disable log the metrics from the offline diags report step to wandb.
- added argo train-diags-prog optional parameter `--no-wandb` to set this option for training and offline diags step 

Note that the default behavior for training and offline diags is to log using wandb unless otherwise disabled with the`--no-wandb` flag.

Also added a print statement that prints the offline metrics, model name, and commit sha so that they're available in gcloud logging for ingestion into database.


workflows/argo/offline-diags.yaml
```diff
@@ -33,6 +37,17 @@ spec:
             value: /secret/gcp-credentials/key.json
           - name: CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE
             value: /secret/gcp-credentials/key.json
+          - name: WANDB_ENTITY
+            value: ai2cm
+          - name: WANDB_PROJECT
+            value: "{{inputs.parameters.wandb-project}}"
+          - name: WANDB_TAGS
+            value: "{{inputs.parameters.wandb-tags}}"
+          - name: WANDB_RUN_GROUP
```
nbren12, line 46: This is called `wandb-group` in the microphysics argo workflow: https://github.com/ai2cm/fv3net/blob/c09ea6bebe68cac49f5a30ab52751ea43fb66d99/projects/microphysics/argo/argo.yaml#L30

I like your name better, but I think it would be best to be consistent with the existing precedent for now.  We can change both to wandb-run-group later on.

workflows/argo/offline-diags.yaml
```diff
@@ -56,14 +71,24 @@ spec:
             cat << EOF > training_data.yaml
             {{inputs.parameters.training_data_config}}
             EOF
+
+            if [ "{{inputs.parameters.use-wandb}}" == "false" ];
+            then
+              wandb_flag=" "
```
nbren12, line 77: Would be nice to settle whether these empty strings are necessary or not. If they are then can you add a comment saying why?
```suggestion
              wandb_flag=""
```

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
```diff
@@ -106,6 +107,14 @@ def _create_arg_parser() -> argparse.Namespace:
         default=None,
         help=("Training data configuration yaml file to insert into report"),
     )
+    parser.add_argument(
+        "--wandb",
```
nbren12, line 111: It's a bit confusing to have the argo workflow flip the default. I'd propose having wandb on by default everywhere...e.g. jeremy doesn't use argo.
```suggestion
        "--no-wandb",
```

nbren12: It also looks like you have a test failure related to this:
```
E       AttributeError: 'CreateReportArgs' object has no attribute 'wandb'
```

`CreateReportArgs` is a mock . Can you you refactor the tests to use the real argument parser to create `args`, so that we don't have to worry about the mock getting out of sync again. Either that or re-target the tests at a slightly lower level API that doesn't refer to `args`, but rather the fields in `args`. e.g. `run_report(args.url, args.other_needed_arg, args.output_url)`. I'm not sure if such an api exists, would be nice if it did to allow decoupling the argparse and actual report running logic.

nbren12 (approved):
Thanks! some small suggestions, the only "required" one would be to make sure the default of the CLI and argo workflow is to have wandb on.

workflows/argo/offline-diags.yaml
AnnaKwa: They aren't necessary, I think I just do " " on autopilot sometimes.

workflows/argo/offline-diags.yaml
AnnaKwa: Also changed train-diags-prog and training templates to "run-group"

PR #1773, opened 2022-04-13
Add fv3net/diagnostics/prognostic_run to mypy precommit

oliverwm1:
We use a lot of type hints in the prognostic run diagnostics, but they are not verified as part of linting. This PR adds this code to the mypy files and fixes the existing issues.

EDIT: the existing mypy issues can be seen here: https://app.circleci.com/pipelines/github/ai2cm/fv3net/11717/workflows/8ac94a1b-f715-4361-86ad-4a92030f4061/jobs/104292

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,8 @@ repos:
         workflows/prognostic_c48_run/tests/.+ |
         external/fv3fit/fv3fit/.+ |
         external/loaders/loaders/.+ |
-        workflows/diagnostics/fv3net/diagnostics/offline/.+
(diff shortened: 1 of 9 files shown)


AnnaKwa (approved):
LGTM, thanks for adding this to mypy.

nbren12 (approved):
LGTM. nice to add this mypy. 

Does partial just not propagate the type info?

oliverwm1:
> Does partial just not propagate the type info?

I think the issue was using partial to fix a positional argument that was not the leftmost one, although it only surfaced after I stopped supplying `source` as a keyword argument to `transform` (which mypy was complaining about).

PR #1774, opened 2022-04-15
Add artifact tracking to piggy

nbren12:
It is not possible to find piggy-backed runs by the wandb artifact graph downstream from a given run. This commit fixes that by using the prognostic run artifact from `prognostic_run_diags piggy`.

Also, refactor the prognostic_evaluate.py script to share the code.

Example piggy backed report: https://wandb.ai/ai2cm/microphysics-emulation/runs/33c81kdo/overview?workspace=user-nbrenowitz

I manually tested that `scripts/prognostic_evaluation.py` still works.

@frodre would it make sense to move `prognostic_evaluate.py` to `fv3net/diagnostics` and get some tests coverage? Or maybe just start using the traditional prog run report.


workflows/diagnostics/fv3net/diagnostics/prognostic_run/emulation/query.py
```diff
@@ -0,0 +1,35 @@
...
+        """Return wandb.Run object for this prognostic run"""
+        runs = self.api.runs(
+            filters={"group": self.tag}, path=f"{self.entity}/{self.project}"
+        )
+        prognostic_runs = []
+
+        for run in runs:
+            if run.job_type == "prognostic_run":
+                prognostic_runs.append(run)
+        (run,) = prognostic_runs
```
AnnaKwa, line 30: Is there an assumption here that there is only ever one prognostic run for a given tag? 

If you rerun a prognostic run with a model that was used in a previous prognostic run, is a different tag generated? Or do tags correspond to a model configuration?

AnnaKwa (approved):

workflows/diagnostics/fv3net/diagnostics/prognostic_run/emulation/query.py
nbren12: We discussed this on slack. yes there is this assumption.

PR #1776, opened 2022-04-15
Fix integration test empty WANDB_TAGS var

AnnaKwa:
By default the argo workflow set the env var WANDB_TAGS to an empty string. Wandb raises an error if this is an empty string rather than not defined. I changed the templates only set this env var if non empty.

### DIFF
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -389,3 +389,5 @@ spec:
             value: "24"
           - name: memory-prog
             value: 25Gi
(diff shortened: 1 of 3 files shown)


oliverwm1 (approved):
Would there be a similar issue if the `wandb-group` parameter is left to its default of an empty string? Or does user have to set this parameter?

AnnaKwa:
> Would there be a similar issue if the `wandb-group` parameter is left to its default of an empty string? Or does user have to set this parameter?

Apparently not, it runs fine with an empty group.

PR #1782, opened 2022-04-18
Star dataclass-ser package

nbren12:
A package that helps serialization/deserialization of dataclass based configs. Currently just supports yaml, but could be extended to other formats.

We currently struggle with union types in dataclass configurations. dacite cannot distinguish between config classes if they have the same field. However, many serialization formats such as yaml and protocol buffers natively support polymorphic configs. This package adds direct support for yaml-based serialization. 

I also envision adding support for enum types within dataclass configs since we do not currently cleanly support this.


nbren12:
It would also be nice to add features that could support overriding configs by command line argument (e.g. a.b.c=1).

external/dataclass-ser/dataclass_serialization/yaml.py
```diff
@@ -0,0 +1,170 @@
...
+        self.dumper = dumper
+
+    def visit_dataclass(self, node: "DataclassNode"):
+        if node.cls in self.visited:
+            return
+
+        tag = (
+            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG
+            if node.name
+            else "!" + node.cls.__name__
```
oliverwm1, line 52: how does this work if there are classes defined in different modules with the same name? e.g. `Config` defined all over the place.

oliverwm1 (commented):
Interesting proposal! It would be convenient to be able to handle union types more conveniently, although I don't think the current situation is all that dire and this PR would add complexity.

Nice to leverage dacite (`get_generic`, `is_union` etc). Do you have any concerns about dependence on these functions? Does this leave us open to issues with yaml or dacite version changes?

Also wondering about name conflicts between classes defined in different modules.

external/dataclass-ser/dataclass_serialization/yaml.py
nbren12: they would conflict ;)

nbren12:
>although I don't think the current situation is all that dire and this PR would add complexity.

I'm not sure I would call it "dire", but I have noticed config-wrangling holding us back/leading to custom work-around in a variety of contexts lately. Some patterns I've witnessed lately:
- use string rather than Enum ---> many hardcodes
- encode type info redundantly in field names e.g.: `TendencyApplier.tendency_name`.
- writing `Config` wrappers for simple `tf.keras` objects. E.g. `tf.keras.regularizers.L2.get_config` already does 90% of the work.
- fv3fit has a special way to overload params via CLI with `a.b.c=1`
- add `name/kind` attributes requiring custom handling wherever they are added.

nbren12:
> Do you have any concerns about dependence on these functions? Does this leave us open to issues with yaml or dacite version changes?

Some concerns. This is all public API of yaml which hasn't changed in 10+ years. The dacite code could break, but would be easy to copy paste.

external/dataclass-ser/dataclass_serialization/yaml.py
oliverwm1: +1 for redundancy in module and class names haha

nbren12: yah. this is more just a proof of concept atm. It's probably worth defining a backward-compatible API namespace for our public configs regardless. Whether that namespace is flat or has some nesting is up to us.

PR #1786, opened 2022-04-18
Bump fv3config to v0.9.0

spencerkclark:
This will give us reproducible restarts in prognostic runs.  The fv3config update also requires including the `FSSPEC_GS_REQUESTER_PAYS` environment variable in a few places.

Requirement changes:
- fv3config is now updated to v0.9.0

### DIFF
--- a/tests/end_to_end_integration/argo.yaml
+++ b/tests/end_to_end_integration/argo.yaml
@@ -35,6 +35,9 @@ spec:
               do_sat_adj: false
             gfdl_cloud_microphysics_nml:
               fast_sat_adj: false
+            gfs_physics_nml:
+              fhswr: 1800
+              fhlwr: 1800
           fortran_diagnostics:
             - name: atmos_dt_atmos.zarr
               chunks:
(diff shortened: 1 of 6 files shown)


tests/end_to_end_integration/argo.yaml
```diff
@@ -35,6 +35,9 @@ spec:
               do_sat_adj: false
             gfdl_cloud_microphysics_nml:
               fast_sat_adj: false
+            gfs_physics_nml:
+              fhswr: 1800
+              fhlwr: 1800
```
spencerkclark, line 40: Previously the radiation timesteps were set to their default of 3600 seconds, which is longer than the segment length of this run (1800 seconds).  For reproducible restarts, the radiation timestep must evenly divide the segment length, so I have reduced the radiation timesteps here.

We might consider adding validation for this somewhere -- if the two are not compatible, then the model will crash in an uninformative way at the start of the second segment -- but I think that is outside the scope of this PR.

AnnaKwa (approved):

PR #1788, opened 2022-04-19
Generate emulation training data v4

nbren12:
v3 of the emulation data was generated in November and lacked some important information like cloud after the condensation scheme. The training data also showed some initial condition dependence due to using surface forcings from the initial condtions rather than GRIB files.

* Generate v4 of this training set addressing these issues
* Add  `emulation.ModelConfig.output_start_sec`

cc @frodre 
Probably easiest to review commit-by-commit.

### DIFF
--- a/external/emulation/emulation/config.py
+++ b/external/emulation/emulation/config.py
@@ -2,10 +2,12 @@ import dataclasses
 import datetime
 from functools import partial
 import logging
-from typing import Iterable, Mapping, Optional
+from typing import Iterable, Mapping, Optional, Tuple
+import os
 
 import cftime
 import dacite
+import f90nml
 import yaml
 from emulation._emulate.microphysics import (
     MicrophysicsHook,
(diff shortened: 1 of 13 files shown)


external/emulation/emulation/config.py
```diff
@@ -127,6 +138,38 @@ def factory(keyvals):
         return dataclasses.asdict(self, dict_factory=factory)
 
 
+def _get_storage_hook(storage_config: Optional[StorageConfig]) -> Optional[StorageHook]:
+
+    if storage_config is None:
```
nbren12, line 143: I pulled all this I/O and conditional logic out of `StorageHook.__init__`.

projects/microphysics/create_training/training_fv3config_template.yaml
```diff
@@ -51,6 +51,23 @@ diagnostics:
   - tendency_of_ozone_mixing_ratio_due_to_fv3_physics
   - tendency_of_pressure_thickness_of_atmospheric_layer_due_to_fv3_physics
 namelist:
+  namsfc:
+    # these flags make the coarse run use the GRIB data for snoalb and vegetation
+    # if not included, it will use the restart data (which may differ from GRIB file)
```
spencerkclark, line 56:
```suggestion
    # if not included, it will use the initial condition data (which may differ from GRIB file)
```

spencerkclark (approved):
Looks good to me.  It's remarkable how much of a difference this made.  At some point it might be interesting for us to look at the difference between the relevant fields in the `sfc_data` files of the initial conditions and their counterparts in the GRIB files.

nbren12:
> At some point it might be interesting for us to look at the difference between the relevant fields in the sfc_data files of the initial conditions and their counterparts in the GRIB files.

Indeed. I can't tell if this is a sensitivity we want to present in the paper. If it ends up being relevant for ML it probably is though.

PR #1790, opened 2022-04-19
Option to use first or last n_timesteps for movie

AnnaKwa:
Added public API:
-  Add optional CLI arg to `fv3net.diagnostics.prognostic_run.view.movies`: `--from_end` store_true arg, if True use the last n_timesteps for the movie, if false use the first n_timesteps.

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/movies.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/movies.py
@@ -202,6 +202,14 @@ def register_parser(subparsers):
         type=int,
         help="Number of timesteps to use in movie. If not provided all times are used.",
     )
+    parser.add_argument(
+        "--from_start",
+        action="store_true",
+        help=(
+            "If true, generate the movie of n_timesteps at the start of the run. "
+            "If false, use last n_timesteps of the run."
+        ),
+    )
     add_catalog_and_verification_arguments(parser)
     parser.set_defaults(func=main)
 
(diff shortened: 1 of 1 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/movies.py
```diff
@@ -202,6 +202,14 @@ def register_parser(subparsers):
         type=int,
         help="Number of timesteps to use in movie. If not provided all times are used.",
     )
+    parser.add_argument(
+        "--from_start",
+        action="store_true",
+        help=(
+            "If true, generate the movie of n_timesteps at the start of the run. "
+            "If false, use last n_timesteps of the run."
+        ),
```
oliverwm1, line 211:
```suggestion
        "--from_end",
        action="store_true",
        help="Use end of run for movie instead of start of run.",
```
can we have the default behavior remain unchanged? i.e. can we have the flag be called `from_end` and if it is not provided, then the movie is made from the start of the run.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/movies.py
```diff
@@ -217,10 +225,15 @@ def _merge_prognostic_verification(
...
+def _limit_time_length(
+    ds: xr.Dataset, n_timesteps: int, from_start: bool
+) -> xr.Dataset:
+    """Limit the time dimension of a dataset to the first or last n_timesteps."""
     max_time = min(n_timesteps, ds.sizes["time"])
-    return ds.isel(time=slice(None, max_time))
+    if from_start:
+        return ds.isel(time=slice(None, max_time))
+    else:
+        return ds.isel(time=slice(-max_time, None))
```
oliverwm1, line 236: would need to switch for `from_end` here too.

oliverwm1 (changes requested):
I think the current PR would change the default behavior of the CLI. I'd rather we keep the default the same.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/movies.py
AnnaKwa: Whoops, forgot the default is the opposite of the action. Fixed.

oliverwm1 (approved):
LGTM

PR #1789, opened 2022-04-19
JSON Logging for Emulation Metrics

frodre:
This PR adds json metric logging for emulation diagnostic scripts that can be ingested by gcloud's logging infrastructure.  In addition to this a new regression CronJob (https://github.com/ai2cm/vcm-workflow-control/pull/127) will be run nightly to log important metrics from our training and prognostic runs nightly.

Metrics Added:
- training time
- prognostic run length
- MSE/bias scores on a test dataset for trained models

PR #1791, opened 2022-04-20
micro: Default launches every training config

nbren12:
@frodre not sure if you have something like this already? Feel free to close if so.

### DIFF
--- a/projects/microphysics/experiments/default.py
+++ b/projects/microphysics/experiments/default.py
@@ -11,7 +11,7 @@ def _get_job(config: str, revision):
     config_name = os.path.splitext(os.path.basename(config))[0]
     config = load_yaml(config)
     return EndToEndJob(
-        name=f"{config_name}-{revision[:6]}-v1",
+        name=f"{config_name}-{revision[:6]}-{suffix}",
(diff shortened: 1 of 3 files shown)


frodre:
I updated to the dense, but this will fit nicely in #1789, I think.  I'd like to get one full regression in before switching to the new dataset, so I don't think we need to close this PR.  Maybe this PR can also add in deletions of the precpd cloud water patch in.

frodre:
Merged into the PR [Remove cron.py and add suffix to default.py](https://github.com/ai2cm/fv3net/pull/1789/commits/9a411c8b3d5fbe70d6d888bbe9e9db32762b527f)

nbren12:
>  I'd like to get one full regression in before switching to the new dataset, so I don't think we need to close this PR.

That's a good idea. I may lower the number of epochs here to keep the number of steps similar to the v3 data. I'm getting impatient waiting for this to finish!

nbren12:
https://wandb.ai/ai2cm/microphysics-emulation/runs/2g4snhue/overview?workspace=user-nbrenowitz

PR #1794, opened 2022-04-20
Demonstrate watching k8s metadata with argo-events

nbren12:
We can use argo-events to extract the metadata from argo workflows and
save them in big query.

I tested this using minikube on my laptop.

### DIFF
--- a/workflows/argo/events/example/README.md
+++ b/workflows/argo/events/example/README.md
@@ -1,0 +1,60 @@
+Based on this tutorial: https://argoproj.github.io/argo-events/tutorials/01-introduction/
+
+
+To setup all the stuff
+
+    kustomize build . | kubectl apply -f -
+    # might error on the CRDs...run it again in 5 seconds
+    kustomize build . | kubectl apply -f -
+
+Wait a minute or two. The submit a workflow
+
+    argo submit workflow.yaml
+
+Listing the workflows should show two workflows:
+
+    $ argo list
+    NAME                      STATUS      AGE   DURATION   PRIORITY
+    resource-workflow-5h4fz   Succeeded   49s   45s        0
+    my-workflow54mc5          Succeeded   49s   44s        0
+
+The "resource-workflow-" workflow was launched when the first workflow was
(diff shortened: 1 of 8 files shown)


nbren12:
Will probably move the yamls to `vcm-workflow-control` and we could move the server code to another repo since it is pretty isolatable functionality.

AnnaKwa:
Does using Go have significant advantages over using a Flask app for this or other cases? It's definitely more of a hurdle for reviewing/contributing. My first impression looking over the changes though was that if I were to sit down and try to full understand how this PR worked, most of my time would probably be spent trying to figure out what was going on in the argo events manifests rather than the Go part. If there was a decently big advantage in using Go for this and future stuff then I'd be interested in learning Go to contribute if there was some understanding that initial contributions would take a lot longer and there was time allotted for just learning.

nbren12:
Thanks for the comment anna. I agree the yamls part is probably the more difficult part of this PR.

go doesn't have any huge advantage over python in this case. But it does have some small ones such as the `InferSchema` function which has no counterpart on the python big query language, and more native support for structs etc. The mandatory type-checking makes me feel more confident that this won't fail. 

 The main advantages i see for building go based tooling are the simpler environment management (no venv,pypi,etc, just copy the binary and run). 2) using mandatory static type checking for processes we don't ever want to fail, and 3) the network effects with other parts of the ecosystem (k8s, argo, terraform, etc all written in go). Towards the latter point, I found the go docs for the python sdk were a bit more complete: https://pkg.go.dev/cloud.google.com/go/bigquery.

nbren12:
Some more comments about the overall architecture:
- argo supports saving workflow info to [MySQL/PostGreSQL](https://argoproj.github.io/argo-workflows/workflow-archive/), but not big query
- In theory, one could write a for-loop in the python k8s api to monitor for new argo workflows, but that's what [argo-events](https://argoproj.github.io/argo-events/) already does. the component that does this is the "event source".
- The event source forwards events to a "sensor" that decided what to do with the event (including making a http request).

PR #1801, opened 2022-04-21
Print step metadata

AnnaKwa:
This adds the dataclass `StepMetadata` describing step level metadata which is used in the prognostic run append step, prognostic run diags, training, and offline report scripts to print this information so that it can be pulled into the BigQuery database. I didn't add this for the emulation steps as I am not familiar with their args and dependencies.


workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -101,3 +101,12 @@ def write_chunks(config: UserConfig):
     chunks = get_chunks(diagnostic_file_configs)
     with open("chunks.yaml", "w") as f:
         yaml.safe_dump(chunks, f)
+
+
+def get_model_urls(config_dict: dict) -> List[str]:
```
AnnaKwa, line 106: This function is the iffy-est part of the PR to me as it couples the metadata to the structure of our config object. I'm open to hearing if there is another alternative to get these dependencies.

workflows/prognostic_c48_run/runtime/segmented_run/append.py
```diff
@@ -30,6 +32,16 @@ def read_run_config(run_url):
     return fv3config.load(io.BytesIO(s))
 
 
+def print_step_metadata_json(run_url: str, config: dict):
+    info = StepMetadata(
+        job_type="prognostic_run",
+        commit=os.environ.get("COMMIT_SHA"),
+        url=run_url,
+        dependencies={"ml_models": get_model_urls(config)},
+    )
```
AnnaKwa, line 41: `args` field is left empty for prognostic run since it typically always has the same `fv3config.yml`, temporary run dir args.

workflows/prognostic_c48_run/runtime/config.py
frodre: I think I agree that this location makes sense for this to live here since it's close to the config.  I think we just have to get the return type right so it's generally useful.  I'm thinking maintaining hierarchy would probably be better?  The empty container vs. None probably pertains here too on whether one is easier to query/filter.

```
{
    scikit_learn: <str or None>,  
    prephysics: <mapping of unique items to url or None> 
}

File: workflows/prognostic_c48_run/runtime/segmented_run/append.py [hunk 3]
```diff
@@ -30,6 +32,16 @@ def read_run_config(run_url):
     return fv3config.load(io.BytesIO(s))
 
 
+def print_step_metadata_json(run_url: str, config: dict):
+    info = StepMetadata(
+        job_type="prognostic_run",
+        commit=os.environ.get("COMMIT_SHA"),
+        url=run_url,
+        dependencies={"ml_models": get_model_urls(config)},
```
frodre, line 40: I think this violates the `Mapping[str,str]` usage, but prephysics would have a list of different models anyways?  Any way to expand it out so its a {"<unique_model_name>": "url"}  Or maybe it makes sense to relax the restrictions.  What matters is that these are useful in a query.

frodre (changes requested):
Thanks, @AnnaKwa.  I decided it would be easier if this PR went in before my emulation JSON update, so I can just use the step metadata logging.

I think this is close, just a few main points:
- Making a function that defines the needed step metadata and prints it would have less copy/paste than currently used dataclass
- We should decide whether None or empty containers are preferable for querying

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
```diff
@@ -417,14 +433,8 @@ def create_report(args):
...
-    gcloud_logging = {
-        "json": metrics,
-        "logging.googleapis.com/labels": {
-            "model": metadata["model_path"],
-            "commit_sha": metadata.get("commit", "n/a"),
-        },
-    }
-    print(json.dumps(gcloud_logging))
+    print(json.dumps({"json": metrics}))
+    _print_step_metadata_json(args, metadata)
```
oliverwm1, line 437: do we not want to attach the step metadata to the streamed json as labels?

i.e. just print one dict that is something like

```
{
    'json': metrics,
    'logging.googleapis.com/labels": {k: v for k, v in step_metadata.asdict()}
}
```

workflows/prognostic_c48_run/runtime/config.py
oliverwm1: Yah, this is tricky, there's also possibly emulator/fine-res model dependencies. What about something more general... like search for all GCS urls in the config that have `"trained_models"` in them? (not really sure we want to do this... would only work for models trained via the argo workflows)

nbren12: >as it couples the metadata to the structure of our config object.
I think this is unavoidable since only the config object knows what it's "dependencies" are. 

I can think of a few general solutions that would be more robust than searching for a "hardcoded" string.
1. add a `.dependencies()` method to each class in our dataclass hiearchy that advertises its deps. This is a bit laborious and would still require code to traverse the config tree.
2. Require that all "step" level deps be handled as assets by fv3config. Call `fv3config.config_to_asset_list(config)` and simply return the urls in the asset list. I recently made `fv3config.config_to_asset_list` public API to enable this kind of approach.

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
frodre: I believe all the "facts" will be tied to the step level metadata through `name: pod.metadata.step_name` or `pod.name` as the primary key.  IIRC, labels have some restrictions to make them easily searchable, which might be too constraining for some of this metdata.

@AnnaKwa, maybe the step metadata dump should have a unique key so the payload would look like {"step_metdata": <Metadata JSON>} instead of having it all at the top level.  This would make it easy to query with something like jsonPayload.step_metadata as a field to select on.

workflows/prognostic_c48_run/runtime/segmented_run/append.py
AnnaKwa: I don't think so, since there's no key data in the user config, it just provides models in a list. I can separate prephysics models and the rest of the models out into individual keys.

AnnaKwa: On second thought I think I will leave all the ML models in a single list. For the purposes of tracking dependencies I don't think there should be multiple keys to search on.

frodre (approved):
Thanks for the updates, @AnnaKwa!  LGTM and I'll go ahead and incorporate into my PR.

PR #1805, opened 2022-04-22
Install wrapper from fv3gfs-fortran

nbren12:
The wrapper has been merged into fv3gfs-fortran.

* Remove the fv3gfs-wrapper submodule
* Update the fv3gfs-fortran submodule to a version containing the wrapper

### DIFF
--- a/docker/prognostic_run/Dockerfile
+++ b/docker/prognostic_run/Dockerfile
@@ -84,24 +84,18 @@ RUN cd /tmp/fortran-build/FV3 && \
     make -j 8 && \
     PREFIX=/usr/local make install
 
-COPY docker/prognostic_run/requirements.txt /tmp/requirements.txt
-
-RUN pip3 install -c /tmp/requirements.txt numpy cython jinja2 && make -C /tmp/fortran-build/FV3 wrapper_build
-
-
-#
 # Python Stuff Here
 #
 # Equivalent to passing --no-cache-dir to every pip install
 ENV PIP_NO_CACHE_DIR=off
 
-# do this early on the build since it is upstream of fv3net
-RUN pip3 install -c /tmp/requirements.txt /tmp/fortran-build/FV3/wrapper/dist/*.whl && \
-    python3 -c 'import fv3gfs.wrapper'
-
 # Install dependencies
 COPY docker/prognostic_run/requirements.txt /tmp/requirements.txt
 RUN pip install -r /tmp/requirements.txt
+
+RUN make -C /tmp/fortran-build/FV3 wrapper_build &&\
+    pip3 install --no-dependencies /tmp/fortran-build/FV3/wrapper/dist/*.whl && \
+    python3 -c 'import fv3gfs.wrapper'
 
 COPY external/vcm /fv3net/external/vcm
 RUN pip3 install --no-dependencies /fv3net/external/vcm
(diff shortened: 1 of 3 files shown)


docker/prognostic_run/Dockerfile
```diff
@@ -93,20 +95,14 @@ RUN apt-get update && apt-get install -y  apt-transport-https ca-certificates gn
 # Equivalent to passing --no-cache-dir to every pip install
 ENV PIP_NO_CACHE_DIR=off
 
+# do this early on the build since it is upstream of fv3net
+RUN pip3 install -c /tmp/requirements.txt /tmp/fortran-build/FV3/wrapper/dist/*.whl && \
```
mcgibbon, line 99: Nice.

docker/prognostic_run/Dockerfile
```diff
@@ -78,13 +84,9 @@ RUN cd /tmp/fortran-build/FV3 && \
...
 
-# gcloud
+COPY docker/prognostic_run/requirements.txt /tmp/requirements.txt
 
-#Install gcloud
-RUN apt-get update && apt-get install -y  apt-transport-https ca-certificates gnupg curl gettext && \
-    echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list &&\
-    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && \
-    apt-get update && apt-get install -y google-cloud-sdk jq python3-dev python3-pip kubectl gfortran graphviz
+RUN pip3 install -c /tmp/requirements.txt numpy cython jinja2 && make -C /tmp/fortran-build/FV3 wrapper_build
```
mcgibbon, line 89: Do we need jinja, if we aren't building?

Should numpy and cython go in docker/prognostic_run/requirements.txt instead?

docker/prognostic_run/Dockerfile
```diff
@@ -93,20 +95,14 @@ RUN apt-get update && apt-get install -y  apt-transport-https ca-certificates gn
 # Equivalent to passing --no-cache-dir to every pip install
 ENV PIP_NO_CACHE_DIR=off
 
+# do this early on the build since it is upstream of fv3net
```
mcgibbon, line 98:
```suggestion
# do this early on the build since it is upstream of fv3net
# import as minimal test nothing went wrong with the build/install
```
If this is why you are running the import I think it's worth documenting it, could be hard to tell otherwise.

mcgibbon (approved):
If you can move numpy and cython to the requirements.txt and remove jinja (it shouldn't be required) then please do, otherwise LGTM! Thanks for updating this!

docker/prognostic_run/Dockerfile
nbren12: we are building it.

numpy and cython are in the requiremnts file, but I wanted to isolate the wrapper building to an earlier part of the makefile to allow better caching.

docker/prognostic_run/Dockerfile
```diff
@@ -93,20 +95,14 @@ RUN apt-get update && apt-get install -y  apt-transport-https ca-certificates gn
...
+RUN pip3 install -c /tmp/requirements.txt /tmp/fortran-build/FV3/wrapper/dist/*.whl && \
+    python3 -c 'import fv3gfs.wrapper'
+
 # Install dependencies
 COPY docker/prognostic_run/requirements.txt /tmp/requirements.txt
```
mcgibbon, line 103: I believe you can delete this COPY now since it's moved up higher.

docker/prognostic_run/Dockerfile
nbren12: I added a comment like this.

PR #1807, opened 2022-04-27
Fix emulation scoring OOM

frodre:
The offline scoring of the emulation trained model runs out of memory when tcmalloc tries to allocate necessary memory for eager execution of an RNN model. Switching to .predict(batchsize=...) makes sure the prediction of outputs for scoring happens in a reasonable block size and prevents OOM. 

Additionally, I adjust the training dataset to be the same size as the independent testing dataset for consistency.


frodre:
Split out of #1789 for the sake of clean history.

AnnaKwa (approved):

PR #1808, opened 2022-04-28
Fix bug in step metadata of prog run

AnnaKwa:
I forgot that if the user config omits a key (like scikit_learn) it is present in the final config, but with a None value. This caused a bug when the step metadata function assumed that the key wouldn't be in the final dict.

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/tests/test_config.py
+++ b/workflows/prognostic_c48_run/tests/test_config.py
@@ -1,5 +1,8 @@
+import dacite
 import pytest
-from runtime.config import get_model_urls
+from runtime.config import get_model_urls, UserConfig
+
+dummy_prescriber = {"dataset_key": "data_url", "variables": ["a"]}
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


workflows/prognostic_c48_run/tests/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+                "prephysics": [
+                    {"some_prescribed_data": 0},
+                    {"model": ["prephysics_model_0", "prephysics_model_1"]},
+                ],
+            },
+            ["ml_model_url", "prephysics_model_0", "prephysics_model_1"],
+        ),
+    ],
+)
+def test_get_model_urls(config, model_urls):
```
frodre, line 30: Since it's coupled to the `UserConfig`  for now, maybe use that class in the test so if one-or the other changes, this test will break.

brianhenn (approved):
Approved, seems like @frodre 's suggestion is a good idea though.

PR #1809, opened 2022-04-29
Use common functions for hybrid mapper

oliverwm1:
This PR refactors the "fine res hybrid" mapper to reuse some of the functionality that was developed for the pure fine-res mapper. This allows using the same set of approaches to define the total Q1/Q2, and more flexibility for using nudging and physics tendencies that are not necessarily from the same run.

Deleted public API:
- remove "precomputed" fine-res/hybrid mappers

Refactored public API:
- `open_fine_resolution_nudging_hybrid` now takes the same arguments as `open_fine_resolution` mapper

- [x] Tests added


external/loaders/loaders/mappers/_hybrid.py
```diff
@@ -1,117 +1,85 @@
...
+        additional_dataset_urls: sequence of urls which to zarrs containing additional
+            data to be merged into the resulting mapper dataset
 
     Returns:
         a mapper
     """
     return XarrayMapper(
-        _open_precomputed_fine_resolution_nudging_hybrid_dataset(
-            fine_url=fine_url, nudge_url=nudge_url
+        _open_merged_dataset(
```
brianhenn, line 80: I think for the precomputed hybrid case, the assumption was that only the fine portion is precomputed, and that we still do the "hybrid" computations on the fly. I'd presumably retain that.

brianhenn: This would also work well if the precomputed hybrid and fine res openers were the same. Would just need a flag on whether to call `compute_hybrid_budget` or not, and maybe some careful handling of the wind nudging tendencies.

brianhenn (commented):
Thanks @oliverwm1 for the cleanup of the messy hybrid openers. See comments below based on our discussion about handling the 'precomputed' cases.

external/loaders/loaders/mappers/_fine_res_budget.py
```diff
@@ -75,6 +75,9 @@ class FineResBudget(Protocol):
     SHTFLsfc_coarse: xarray.DataArray
     PRATEsfc_coarse: xarray.DataArray
 
+    def astype(self, dtype):
```
brianhenn, line 78: leftover?

brianhenn (approved):
LGTM

external/loaders/loaders/mappers/_fine_res_budget.py
oliverwm1: needed for type-hinting, since we call `.astype` on an object we type-hint as `FineResBudget`

PR #1819, opened 2022-05-09
Fix the e2e emulation run

frodre:
This PR fixes a few minor things with the E2E emulation workflow
* reverts changes to the prognostic config introduced in #1788
* Address requester pays issue with prognostic run step
* Ensure the prognostic run, piggy-back diags, and prognostic eval appear in in the same wandb run group
* Adds a test mode for faster checks on the E2E job in `default.py`

### DIFF
--- a/projects/microphysics/configs/default.yaml
+++ b/projects/microphysics/configs/default.yaml
@@ -222,8 +222,8 @@ namelist:
   coupler_nml:
     current_date:
     - 2016
-    - $MONTH_INT
-    - 1
+    - 8
+    - 2
     - 0
     - 0
     - 0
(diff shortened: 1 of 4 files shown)


projects/microphysics/configs/default.yaml
```diff
@@ -237,13 +237,13 @@ namelist:
...
     nudge: false
     nudge_qv: false
-    warm_start: false
-    external_ic: true
-    external_eta: false
-    make_nh: true
-    nggps_ic: true
-    mountain: false
-    na_init: 1
+    warm_start: true
```
frodre, line 240: I reset these to the previous configuration values as well, although, I do not know exactly which of them matter.

AnnaKwa (approved):

PR #1827, opened 2022-05-12
Make default 'zonal average' bins symmetric about equator

oliverwm1:
Current default bins for the `vcm.zonal_average_approximate` function are not symmetric about the equator. There is a "[-90, 88]" bin but no "[88, 90]" bin. This PR fixes this issue.


AnnaKwa (approved):
LGTM

Issue #1783, opened 2022-04-18
vcm is hard to pip install


nbren12:
This will make it easier for external collabs to use our code outside the fv3net environment. Also, for us to use vcm e.g. in google collab or another exploratory environments.

@brianhenn I remember you mentioning that you might be interested in setting this up. Probably would want to move the `mappm` fortran stuff into a separate package since the f2py builds often fail in my experience.

nbren12:
Indeed, it is tricky to package compiled components like mappm. [the manylinux](https://github.com/pypa/manylinux) project describes the overall strategy, which is to compile the library against an old version of linux/glibc which will be forward compatible.

nbren12:
something like that yah. mappm is only really used by the restart
recoarsening code (which is never used in notebooks etc) so you could move
all that functionality to a separate package and avoid any need for
optional dependencies.

PR #1828, opened 2022-05-12
Add gfortran to conda requirements

### LINKED ISSUE #1226: Describe use of `make lock_pip` vs `make lock_deps` in developer docs
The docs should probably recommend use of the `make lock_pip` rule if just updating the constraints.txt. Currently the docs only refer to the `make lock_deps` rule, which in the past I've been told to avoid since it causes large changes to the conda lock files.

_Originally posted by @oliverwm1 in https://github.com/VulcanClimateModeling/fv3net/pull/1221#discussion_r636478119_

### LINKED ISSUE #1637: Include gfortran in conda environment
`make create_environment` fails on systems without gfortran installed when building the `vcm.mappm` extension:
```
    warning: build_ext: f77_compiler=None is not available.

    building 'vcm.mappm' extension
    error: extension 'vcm.mappm' has Fortran sources but no Fortran compiler found
```

mcgibbon:
gfortran is required to build vcm.mappm, but was not included in our requirements, causing errors when installing on systems without a system gfortran installed.

This PR updates the conda requirements to include gfortran, so vcm can be built.

It also removes the unused win-64 conda platform, since this does not have a gfortran package. This may be added back in a future PR which provides support for apple m1 (which also does not have gfortran support).

Requirement changes:
- added gfortran to environment.yml
- removed conda-win-64.lock
- updated conda-lock commands to work for latest version of conda-lock
- moved some conda requirements to pip-requirements.txt (some were already there)
- re-generated conda lockfiles
- reduced local package pip install to one command
- added case handling for macbook platforms so an error is given if an m1 macbook is used
- updated requirements documentation to mention lock_pip

Resolves #1637, resolves #1226, partially addresses #1783


environment.yml
```diff
@@ -1,23 +1,19 @@
...
   - cartopy
   - conda-build
   - crcmod
   - ffmpeg
-  - intake
+  - gfortran
   - partd
   - pyquery>=1.4.0
-  # developer tools
-  - isort
```
mcgibbon, line 14: We no longer need to pip install these since we use pre-commit.

nbren12 (approved):
LGTM!

PR #1834, opened 2022-05-13
Make vcm.histogram2d handle different dim orders for inputs

oliverwm1:
Currently `vcm.histogram2d` implicitly assumes that the two supplied data arrays have the same shape/dimension order. This is causing issues in the prognostic report, where `water_vapor_path` and `minus_column_integrated_q2` can have different dim orders (I think because one comes from python diags whereas the other can involve Fortran diags). This PR tests this behavior and makes a fix.

- [x] Tests added


AnnaKwa (approved):
LGTM

PR #1835, opened 2022-05-13
Bring spectral normalization code into fv3fit package

### LINKED ISSUE #1286: Remove spectral normalization?
As far as I can tell this feature isn't used and it incurs a dependency on `tensorflow_addons`. This package is not available in `nix`, which is probably the package manager we will use for future versions of our images, which are stuck on python 3.6.

mcgibbon:
We only use tensorflow_addons for spectral normalization, and that package has compatibility issues with nix. This PR brings in that code to avoid the dependency.

Significant internal changes:
- SpectralNormalization code now exists in fv3fit, copied from tensorflow_addons

Requirement changes:
- fv3fit no longer depends on tensorflow_addons

- [ ] Tests added

Resolves #1286


nbren12 (approved):
LGTM. Is there an easy to extract test for this?

Issue #1802, opened 2022-04-21
Prognostic run tests fail via some entrypoints lacking FSSPEC_GS_REQUESTER_PAYS


brianhenn:
There are several ways to get into the prognostic run container to run the tests (`make enter_prognostic_run && pytest`, `test_prognostic_run`, `image_test_prognostic_run`, `cd workflow/prognostic_c48_run && make test`, etc. Running the tests requires that `FSSPEC_GS_REQUESTER_PAYS=vcm-ml` be set, but currently only `image_test_prognostic_run` sets this variable upon entering the container, so the others fail. We should simplify the number of entrypoints and ensure that the variables needed for the tests always get set.

mcgibbon:
Resolved by #1837.

Issue #1427, opened 2021-10-18
Make prognostic run dev environment work with application default credentials


nbren12:
The prognostic run development enviroment (docker-compose) needs service account credentials. It should be possible to use with the keys configured by `google auth application-default login`.

mcgibbon:
Is this resolved by deleting the docker-compose.yml?

nbren12:
The solution is to bind-mount in these keys: `~/.config/gcloud/application_default_credentials.json` and point to it with the GOOGLE_APPLICATION_CREDENTIALS. Either that or bind-mount in the whole `~/.config/gcloud` folder.

mcgibbon:
OK, the top-level docker system already does that so this should be handled by deleting docker-compose.yml in #1837 and replacing it with the top-level enter target.

mcgibbon:
Resolved by #1837.

PR #1837, opened 2022-05-16
Reduce duplicated prognostic run docker code

### LINKED ISSUE #1821: Too many ways to entrypoints to prognostic run dev env
There are a lot of ways to get into the prognostic run dev environment and not all of them work.

#1427

mcgibbon:
Currently there are four ways to enter the prognostic run image, and not all of them are tested. This PR removes two of those entrypoints and reduces code duplication between the remaining two.

In particular, there are currently:

- `make test_prognostic_run` (unused)
- `make image_test_prognostic_run`
- `make enter_prognostic_run`
- docker-compose entrypoints in `workflows/prognostic_c48_run`
- docker-compose entrypoints in `projects/microphysics`

This PR keeps only `image_test_prognostic_run` and `enter_prognostic_run`.

Refactored public API:
- Deleted `make_test_prognostic_run`
- Deleted `workflows/prognostic_c48_run/docker_compose.yml` and related `workflows/prognostic_c48_run/Makefile`
- Deleted `projects/microphysics/docker_compose.yml`, updated microphysics documentation to refer to top-level `make enter_prognostic_run`
- All docker image entrypoints in the top level Makefile that pass authentication arguments now also receive the environment variable `FSSPEC_GS_REQUESTER_PAYS=vcm_ml`

Requirements changes:
- removed tensorflow_addons from environment (no longer required since #1835)

Significant internal changes:
- Top level Makefile targets now re-use DOCKER_AUTH_ARGS

Resolves #1821, #1802, #1427


Makefile
```diff
@@ -12,6 +12,23 @@ BEAM_VERSION = 2.37.0
 UBUNTU_IMAGE = ubuntu@sha256:9101220a875cee98b016668342c489ff0674f247f6ca20dfc91b91c0f28581ae
 # prognostic base image is updated manually, not on every commit
 PROGNOSTIC_BASE_VERSION = 1.0.0
+DOCKER_AUTH_ARGS = \
+	-v ${GOOGLE_APPLICATION_CREDENTIALS}:/tmp/key.json \
+	-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/key.json \
+	-e FSSPEC_GS_REQUESTER_PAYS=vcm-ml
+
+DOCKER_INTERACTIVE_ARGS = \
```
mcgibbon, line 20: I could put these below in the one make target where they're used, it just felt "cleaner" this way.

docker/prognostic_run/Dockerfile
```diff
@@ -26,19 +26,19 @@ RUN make -C /tmp/fortran-build/FV3 wrapper_build &&\
     python3 -c 'import fv3gfs.wrapper'
 
 COPY external/vcm /fv3net/external/vcm
-RUN pip3 install --no-dependencies /fv3net/external/vcm
+RUN pip3 install --no-dependencies -e /fv3net/external/vcm
```
mcgibbon, line 29: I also modified PYTHONPATH below. Did this for redundancy, since if a user tries to modify the PYTHONPATH it's likely to break things in hard-to-detect ways.

docker/prognostic_run/Dockerfile
```diff
@@ -26,19 +26,19 @@ RUN make -C /tmp/fortran-build/FV3 wrapper_build &&\
...
 
 COPY external/artifacts /fv3net/external/artifacts
-RUN pip3 install --no-dependencies /fv3net/external/artifacts
+RUN pip3 install --no-dependencies -e /fv3net/external/artifacts
```
brianhenn, line 32: Useful to have these other packages be installed in development mode.

pip-requirements.txt
```diff
@@ -9,7 +9,6 @@ sphinx-argparse
 sphinx_rtd_theme
 sphinx-gallery
 tensorflow>=2.3
-tensorflow-addons
```
brianhenn, line 12: Can you note in the PR description that there are requirement changes in it (and maybe note what PR removing this package this is in response to?)

brianhenn (commented):
Thanks @mcgibbon the prognostic run part looks good to me; @frodre has a better handle on the emulation entrypoints so I'll let him look at that.

pip-requirements.txt
mcgibbon: Added, I forgot to do this step in my last PR.

frodre (approved):
LGTM! `.env` works and all the pieces for emulation related testing seem good :)

docker/prognostic_run/Dockerfile
mcgibbon: Yes, the `-e` will not work unless you have a `egg-info` folder created in the directory which happens if you `pip install -e` that directory, in which case PYTHONPATH will matter. But if a user does `PYTHONPATH=/my/custom/code` it would break all the packages if not for the redundant things-are-develop-installed. So this way there would need to be two failures for the installs to break.

PR #1840, opened 2022-05-16
Refactor poetry packages to use setuptools

### LINKED ISSUE #1012: Consistent python packaging strategy in fv3net
Possibly. I think we need a holistic approach to the fv3net build system, since right now things are scattered. It is clear that poetry doesn't work well with fv3net:
1. It doesn't allow building environments which sub-select dependencies from a global set of constraints. I tried to use the fix the poetry environment for the prog run diags, but hit many of the version problems (e.g. chardet) that have been solved in `fv3net/constraints.txt`. 
2. No in-place builds
3. Can't build sdists needed by the dataflow jobs.

_Originally posted by @nbren12 in https://github.com/VulcanClimateModeling/fv3net/pull/1002#discussion_r578737324_

mcgibbon:
From #1012:

> It is clear that poetry doesn't work well with fv3net:
> 
> - It doesn't allow building environments which sub-select dependencies from a global set of constraints. I tried to use the fix the poetry environment for the prog run diags, but hit many of the version problems (e.g. chardet) that have been solved in fv3net/constraints.txt.
> - No in-place builds
> - Can't build sdists needed by the dataflow jobs.

This PR refactors poetry packages to use setuptools.

Significant internal changes:
- Poetry packages are refactored to include a `setup.py` file and use setuptools

Requirement changes:
- Removed `poetry` from requirements

Resolves #1012


external/fv3viz/setup.cfg
```diff
@@ -0,0 +1,32 @@
+[metadata]
+name = fv3viz
+version = attr: fv3viz.__version__
+description = ""
+long_description = file: README.md
+author = Vulcan Technologies LLC
+author_email = oliwm@allenai.org
```
oliverwm1, line 7:
```suggestion
author = Allen Institute for Artificial Intelligence
author_email = oliverwm@allenai.org
```

oliverwm1 (approved):
Thanks for the cleanup. Just one small change.

PR #1841, opened 2022-05-16
Add level masking to emulation

frodre:
Our microphysics emulation experiments are currently experiencing large biases in the top-most levels of the atmosphere, even for our dense-local model.  Ideally, the ML would ascertain that the relative humidity is such that there's no condensation at the TOA, but in the interim, we should just ignore the predictions up there.

This PR adds a mask which substitutes the Fortran ZC-microphysics prediction at the selected levels during the simulation.

- [x] tests added


external/emulation/emulation/config.py
```diff
@@ -78,6 +87,9 @@ def _build_masks(self) -> Iterable[Mask]:
         for key, range in self.ranges.items():
             yield RangeMask(key, min=range.min, max=range.max)
 
+        for key, _slice in self.mask_emulator_levels.items():
```
frodre, line 90: I wonder if this should be placed at the very end?  As is, it means the conservative clouds will be determined by the blended humidity state.  Probably not a big deal since the fortran levels should be "correct".

frodre: Moved it.

frodre:
Masking working as intended. [wandb](https://wandb.ai/ai2cm/microphysics-emulation/groups/gscond-only-dense-nfile1980-mask-upper5-2d-v3-online?workspace=user-aperkins)

image

.gitignore
```diff
@@ -127,3 +127,9 @@ tests/end_to_end_integration/submit_e2e_job_k8s.yml
 
 # redundant lock files that should be regenerated before builds
 docker/*/requirements.txt
+
+# wandb
+**/wandb
```
nbren12, line 132: i worry this will hide code directories named `wandb`. can you move these additions to `projects/microphysics/.gitignore`?

external/emulation/emulation/config.py
```diff
@@ -44,6 +44,12 @@ class Range:
     max: Optional[float] = None
 
 
+@dataclasses.dataclass
+class LevelSlice:
```
nbren12, line 48: Looks good. Would be nice to use yaml tags to do this in the future:

```
mask_emulator_levels: !!slice start:0, stop:0
```

external/emulation/emulation/masks.py
```diff
@@ -37,3 +37,17 @@ def __call__(self, state: FortranState, emulator: FortranState) -> FortranState:
...
+    def __init__(self, key: str, start: Optional[int], stop: Optional[int]):
+        self.key = key
+        self.start = start
+        self.stop = stop
+
+    def __call__(self, state: FortranState, emulator: FortranState) -> FortranState:
+        out = {**emulator}
+        use_fortran_state = slice(self.start, self.stop)
+        # Fortran state TOA is index 79, and dims are [z, sample]
+        out[self.key][use_fortran_state] = state[self.key][use_fortran_state]
```
nbren12, line 52: This will actually mutate `state[key]` since it points to the same numpy array as `out[self.key]`. Either do a copy or use `np.where`.

nbren12 (approved):
I have a couple of suggestions, but otherwise this looks good. Why did it take so many commits for this PR, btw? seems like the code is pretty straightforward.

nbren12:
gotcha. i guess the task was on the backlog for a while.

external/emulation/emulation/masks.py
frodre: Updated to remove the input argument mutation.

PR #1844, opened 2022-05-18
Use consistent bins for 1D and 2D q2 histograms

oliverwm1:
This will remove some plotting artifacts in the water vapor path versus column Q2 plots in the offline report. Currently, because different bins for <Q2> are used for the 1D and 2D histograms there end up being many missing values in the 2D histogram which results in plotting artifacts. See example attached.

hist2d_wvp_vs_q2-1


AnnaKwa (approved):
LGTM

PR #1849, opened 2022-05-19
Prognostic run: use fv3gfs-fortran with fv_sat_adj diags

### LINKED ISSUE #1825: prognostic run: fast saturation adjustment tendencies not output
We currently don't have the ability to output the tendencies from the `fv_sat_adj` Fortran routine if it's turned on. Since it's called from the dynamics step of the model loop, we can capture the combined dynamics + fast saturation tendencies, but having them lumped together is not consistent with how we tend to consider the fast saturation a "physics" process. It makes it hard to informatively close budgets for runs with it turned on. 

@spencerkclark added the ability to write out these tendencies separately in the SHiELD code. Could we add those as diagnostics to the `fv3gfs-fortran` submodule in fv3net?

brianhenn:
Updates the fv3gfs-fortran submodule to allow outputting `fv_sat_adj` tendency diagnostics. No code changes needed in the prognostic run report because because they already accounted for in the 2D outputs, but this allow for computing 3D diagnostic budgets and seeing the `fv_sat_adj` tendencies specifically. See notebook with these outputs [here](https://github.com/ai2cm/explore/blob/master/brianh/2022-05-06-N2F-40-day-Q-comparison/2022-05-16-debug-fast-sat-adj-diags.ipynb).

Requirement changes:
- fv3gfs-fortran update to latest master with added diagnostic options

Resolves #1825


oliverwm1 (approved):
LGTM

PR #1851, opened 2022-05-19
Separate mappm into its own package

### LINKED ISSUE #1783: vcm is hard to pip install

mcgibbon:
Currently `vcm` is difficult to install and can crash on import due to its compiled component, mappm. This PR separates that component into its own package, and treats it as an optional dependency so that `vcm` may be used independent of its remapping functionality.

Significant internal changes:
- `mappm` is moved into its own package in `external/mappm`, and if not installed will only raise an exception when used

Resolves #1783


.environment-scripts/install_local_packages.sh
```diff
@@ -19,6 +19,7 @@ pip install -c constraints.txt \
   -e external/report \
   -e external/synth \
   -e external/vcm \
+  -e external/mappm \
```
mcgibbon, line 22: If and when I commit m1 environment files, we can put this install inside bash logic.

external/vcm/tox_install.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e -x
+
+pip install -c ../../constraints.txt numpy
```
mcgibbon, line 5: tox really doesn't want to do multiple install commands, which meant I had to use this script. For some reason it also switched to doing a "nodeps" install of vcm, which necessitated the final `pip install .`. Took about an hour of trial-and-error to get this right.

mcgibbon:
All integration tests passed.

external/vcm/tox.ini
```diff
@@ -5,13 +5,15 @@
 
 [tox]
 envlist = py3
+skipsdist = true
```
AnnaKwa, line 8: What does this new flag do?

AnnaKwa (approved):
LGTM

external/vcm/tox.ini
mcgibbon:
```suggestion
```

PR #1856, opened 2022-05-20
Add tapering transform

AnnaKwa:
This adds transforms for tapering dQ1 and dQ2 tendencies. This is the most straightforward way of configuring a model to output tendency predictions that are tapered. Note that in this implementation of tapering, if using the standard scaling of model levels in the loss, the upper levels still have the same contribution to the loss function as if they were not tapered at all.

An additional update (in progress) is still needed to the prognostic run in order for these models to be usable on line. Currently the prognostic run defaults to using outputs named `dQ1, dQ2` as the temperature and humidity tendency updates, so it would not recognize the transformed outputs `tapered_dQx` as the tendencies to apply. The upcoming PR would add a configuration option to specify which outputs to use as updates and whether they were either tendencies or state updates.

(Delete unused sections)
Added public API:

Refactored public API:

Significant internal changes:

Requirement changes:

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -232,7 +232,7 @@ class TransformedPredictor(Predictor):
     def load(cls, path: str) -> "TransformedPredictor":
         with fsspec.open(os.path.join(path, cls._CONFIG_FILENAME), "r") as f:
             config = yaml.safe_load(f)
-        base_model = io.load(config["base_model"])
+        base_model = io.load(os.path.join(path, cls._BASE_MODEL_SUBDIR))
         transform_configs = [
             dacite.from_dict(vcm.DataTransform, x) for x in config["transforms"]
         ]
--- a/external/vcm/vcm/calc/calc.py
+++ b/external/vcm/vcm/calc/calc.py
@@ -55,3 +55,10 @@ def weighted_average(
         kwargs=kwargs,
         dask="allowed",
     )
+
+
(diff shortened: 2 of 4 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -232,7 +232,7 @@ def dump(self, path: str):
     def load(cls, path: str) -> "TransformedPredictor":
         with fsspec.open(os.path.join(path, cls._CONFIG_FILENAME), "r") as f:
             config = yaml.safe_load(f)
-        base_model = io.load(config["base_model"])
```
AnnaKwa, line 235: This fixes a bug where TransformedPredictor.load can't find the base model directory if the model being loaded is not in the current working directory.

external/vcm/vcm/calc/calc.py
```diff
@@ -55,3 +55,10 @@ def weighted_average(
         kwargs=kwargs,
         dask="allowed",
     )
+
+
+def vertical_scale_factors(n_levels: int, cutoff: int, rate: float):
```
spencerkclark, line 60: Adding "tapering" to the name might help make the purpose of this function clearer.
```suggestion
def vertical_tapering_scale_factors(n_levels: int, cutoff: int, rate: float):
```

spencerkclark (approved):
Thanks @AnnaKwa.  Just a minor naming suggestion; otherwise this looks good to me.

PR #1858, opened 2022-05-23
Collapse offline report metadata

AnnaKwa:
This collapses the configuration metadata so that you don't have to scroll past a long list of timesteps to get to the meat of the report. ex. https://storage.googleapis.com/vcm-ml-public/annak/2022-05-23/test-html-collapsible-3/index.html

### DIFF
--- a/external/report/report/create_report.py
+++ b/external/report/report/create_report.py
@@ -28,15 +28,22 @@ HTML_TEMPLATE = Template(
     Report created {{now}}
 
     {% if metadata is not none %}
-        <h2>Metadata </h2>
-    <details>
-        <summary> Click to expand </summary>
-        <p class="abstract-text" style="font-size:.90em">
+    <h2>Metadata </h2>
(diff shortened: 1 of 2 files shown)


oliverwm1:
Is it possible to have the metadata be collapsed only if it is above a certain length? Right now in the prognostic run report, the metadata is usually short (<10 lines) but contains very useful stuff like movie links. Would prefer it be visible by default.

AnnaKwa:
> Is it possible to have the metadata be collapsed only if it is above a certain length? Right now in the prognostic run report, the metadata is usually short (<10 lines) but contains very useful stuff like movie links. Would prefer it be visible by default.

Total lines of the nested dict config is a bit messy to code in, instead I'll just make an optional kwarg to collapse the metadata and have it be False by default since it sounds like we'll always want this collapsed in the offline report but not in the prog run report.

oliverwm1 (approved):
LGTM

PR #1864, opened 2022-06-01
Add gscond classification transform

frodre:
This PR adds a classification transform that adds boolean fields to the state  discerning positive tendency (creating cloud), negative tendency (removing _some_ cloud), zero cloud (destruction of a cloud), and zero tendency (no cloud water change).   We will be using these to build cloud classifiers in a follow-on PR.

- [x] Tests added

### DIFF
--- a/external/fv3fit/tests/emulation/test_train_microphysics.py
+++ b/external/fv3fit/tests/emulation/test_train_microphysics.py
@@ -1,15 +1,16 @@
-from pathlib import PosixPath
-import sys
-from fv3fit.dataclasses import asdict_with_enum as asdict
-from unittest.mock import Mock
-
-from fv3fit.emulation.losses import CustomLoss
-
 import pytest
 import tensorflow as tf
+import sys
+from pathlib import PosixPath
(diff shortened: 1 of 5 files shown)


nbren12:
Test failure seems unrelated. I think this needs a test before merging.

external/fv3fit/tests/emulation/test_train_microphysics.py
```diff
@@ -177,3 +178,27 @@ def test_TrainConfig_build_loss():
     loss_value, _ = loss(data, data)
     assert 0 == pytest.approx(loss_value.numpy())
     transform.forward.assert_called()
+
+
+def test_TrainConfig_GscondClassesV1():
+    timestep = 1.0
+    # thresholds set to 1e-15 at time of writing tests
+    data = {
```
nbren12, line 186: I would consider targeting the "sums to one" at the lower level unit `classes` where that behavior is defined and perhaps just test for presence of the required names here explicitly.

nbren12 (approved):
LGTM. most of this is done by pair programming I have a suggestion about the test below.

Issue #1862, opened 2022-05-31
fv3net quickstart recommends service account based auth


oliverwm1:
The [fv3net quickstart](http://vulcanclimatemodeling.com/docs/fv3net/quickstarts.html#cloud-authentication) suggests using service account key-based auth, which we have moved away from using since moving to AI2. The quickstart should be updated to the latest recommended auth strategy.

oliverwm1:
Closed by #1866

PR #1866, opened 2022-06-02
Update auth quickstart section

AnnaKwa:
Updates the auth quickstart for instructions that don't require a service account.

Resolves #< #1862 >

### DIFF
--- a/docs/quickstarts.rst
+++ b/docs/quickstarts.rst
@@ -29,26 +29,23 @@ Once generated, run ``cp .env_template .env`` in the fv3net root and edit the ``
 Cloud Authentication
 --------------------
 
-The ``fv3net`` project currently utilizes Google Cloud to deploy workflow items to services such as Kubernetes and Dataflow.  Authentication requires an installation of `Google Cloud SDK <https://cloud.google.com/sdk/docs/install>`_.
+The ``fv3net`` project currently utilizes Google Cloud to deploy workflow items to services such as Kubernetes and Dataflow.
+Authentication requires an installation of `Google Cloud SDK <https://cloud.google.com/sdk/docs/install>`_.
 
-Authentication obtained via ``gcloud auth login`` does not work well with secrets management and is not used by many APIs. Service account key-based authentication works much better, because the service account key is a single file that can be deployed in a variety of contexts (K8s cluter, VM, etc) (`Google docs on service accounts <https://cloud.google.com/iam/docs/service-accounts>`_). Many Python APIs can authenticate with google using the ``GOOGLE_APPLICATION_CREDENTIALS`` environmental variable `(See Google authentication details) <https://cloud.google.com/sdk/docs/authorizing>`_.
+Authentication obtained via ``gcloud auth login`` does not work well with secrets management and is not used by many APIs;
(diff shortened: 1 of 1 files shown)


AnnaKwa:
I fixed the PR to leave the original instructions for people authenticating on a VM. I'm not sure what the current best way to save the SA credentials is though as there's some permissions errors that pop up with that command (I assume those would also prevent you from getting the key from the console).

docs/quickstarts.rst
```diff
@@ -29,26 +29,36 @@ Once generated, run ``cp .env_template .env`` in the fv3net root and edit the ``
...
+
+    > gcloud auth application-default login
+
+    If you are working in a docker container, you can bind mount in the necessary credentials location in your `docker-compose.yaml`::
+
+      volumes:
+        - ~/.config/gcloud:/root/.config/gcloud
+
+* VM
+
```
frodre, line 53: I would just remove the service account key information for now and say something along the lines of.

```
VMs are currently automatically provisioned with an active service account specific to the VM user.  No 
further action should be required, but we are looking into also providing a key file referenced by
GOOGLE_APPLICATION_CREDENTIALS=... to simplify integration with many tools that use the environment
variable API.
```

frodre (approved):
LGTM

docs/quickstarts.rst
```diff
@@ -29,26 +29,23 @@ Once generated, run ``cp .env_template .env`` in the fv3net root and edit the ``
...
+Authentication requires an installation of `Google Cloud SDK <https://cloud.google.com/sdk/docs/install>`_.
 
-Authentication obtained via ``gcloud auth login`` does not work well with secrets management and is not used by many APIs. Service account key-based authentication works much better, because the service account key is a single file that can be deployed in a variety of contexts (K8s cluter, VM, etc) (`Google docs on service accounts <https://cloud.google.com/iam/docs/service-accounts>`_). Many Python APIs can authenticate with google using the ``GOOGLE_APPLICATION_CREDENTIALS`` environmental variable `(See Google authentication details) <https://cloud.google.com/sdk/docs/authorizing>`_.
+Authentication obtained via ``gcloud auth login`` does not work well with secrets management and is not used by many APIs;
+to use user credentials for API calls, you should also also generate credentials for this with ``gcloud auth application-default login``.
 
-* If gcloud is a fresh install, initialize and grab a keyfile::
+* If gcloud is a fresh install,
 
     > gcloud init
```
frodre, line 37: These seem to be interpreted as "quotes" by Githubs auto rendering.  I'd replace with a ~ or $.

docs/quickstarts.rst
```diff
@@ -32,19 +32,32 @@ Cloud Authentication
 The ``fv3net`` project currently utilizes Google Cloud to deploy workflow items to services such as Kubernetes and Dataflow.
 Authentication requires an installation of `Google Cloud SDK <https://cloud.google.com/sdk/docs/install>`_.
 
-Authentication obtained via ``gcloud auth login`` does not work well with secrets management and is not used by many APIs;
-to use user credentials for API calls, you should also also generate credentials for this with ``gcloud auth application-default login``.
-
 * If gcloud is a fresh install,
 
     > gcloud init
     > gcloud auth login
```
frodre, line 38: Move `auth login` to the Local Machine section

PR #1869, opened 2022-06-03
Make vcm compatible with py37

### LINKED ISSUE #1853: vcm doesn't work with python 3.7
Google colab is still on python 3.7, but vcm has some dependencies that aren't. It would be nice to support 3.7 for now to allow using vcm in more context. This is what I see when trying to use vcm from google colab.

```
      1 import dataclasses
      2 from toolz.functoolz import curry
----> 3 from typing import Callable, Literal, MutableMapping, Sequence, Set
      4 import xarray as xr
      5 import vcm

ImportError: cannot import name 'Literal' from 'typing' (/usr/lib/python3.7/typing.py)
```

Steps to reproduce the problem:
```
conda create -n py37 python=3.7 pip
conda run -n py37 pip install external/vcm
pip install external/vcm
python -c 'import vcm'
```

Potential solution: use `Literal` from typing_extensions rather than `typing`.

cc @mcgibbon

nbren12:
Some slight adjustments to make vcm usable from google colab.

Here is a working example: https://colab.research.google.com/drive/1jLCfdUy667UQxJHfeTXhp1NmPJzumHN2?usp=sharing

Resolves #1853

I tried adding some testing infrastructure for python 3.7, but couldn't get it to work right away. We can add it later if we start using python3.7 environments more often.

### DIFF
--- a/external/vcm/setup.py
+++ b/external/vcm/setup.py
@@ -29,6 +29,7 @@ install_requirements = [
     "pytest-regtest",
     "h5netcdf>=0.8",
     "intake-xarray>=0.3.1",
+    "typing_extensions",
     "dacite",
 ]
 
@@ -44,6 +45,7 @@ setup(
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
     ],
     description="vcm contains general purposes tools for analyzing FV3 data",
     install_requires=install_requirements,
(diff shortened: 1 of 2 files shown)


external/vcm/setup.py
```diff
@@ -24,11 +24,12 @@
     "xgcm",
     "cftime",
     "pytest",
-    "google-cloud-storage>=1.28.0",
+    "google-cloud-storage>=1.18.1",
```
mcgibbon, line 27: Do you need to re-make the constraints.txt?

nbren12: No. The new version range contain the old one. so any lock file satisfying the old will also satisfy the new.

mcgibbon (approved):
LGTM, thanks!

PR #1875, opened 2022-06-08
Automatically link coverage reports in PR description

### LINKED ISSUE #461: coverage statistics are hard to view
It would be nice to add some code coverage tools to this repo, so we can measure the impact a PR has on the coverage, and highlight areas of concern.

mcgibbon:
We currently have automatically generated code coverage reports created when unit tests are run, but these coverage reports are difficult to find and access. This PR updates the circleci workflow so that it automatically updates the PR description with a link to the latest coverage reports for each unit testing workflow.

Resolves #461

- test_unit: [83%](https://output.circle-artifacts.com/output/job/2d1aaf72-fe9c-450c-b94e-b233bdde7d69/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


mcgibbon:
Confirmed initial report shows up as expected, re-launched the tests to see if the report link is updated correctly when the report is already linked. Current link is to [this report](https://output.circle-artifacts.com/output/job/251e77a5-8f10-4292-b80a-6c2bdf848633/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html).

mcgibbon:
The PR "works" but I'm having trouble getting the way newline characters are represented to be consistent between when I run the script locally and on CI. I may try refactoring the script so it uses `gh pr view` and extracts the PR body, instead of using curl (which seems to modify special characters).

mcgibbon:
Reading the coverage report itself I made another issue #1880

mcgibbon:
Follow-on issue should be fixed by #1882.

.circleci/update_coverage_description.sh
```diff
@@ -0,0 +1,21 @@
...
+
+report_type=$1
+
+if [[ -f htmlcov/index.html ]]; then
+    pr_number=${CIRCLE_PULL_REQUEST##*/}
+    body=$(gh pr view | sed -e '1,13d')
+    coverage_label=$(grep -oP '(?<=<span class="pc_cov">).+(?=<\/span>)' htmlcov/index.html)
+
+    if [[ "$body" == *"${report_type}:"* ]]; then
+        body=$(echo "$body" | sed -e "s/${report_type}: [.*](https:\/\/.*\/index.html)/${report_type}: [${coverage_label}](https:\/\/output.circle-artifacts.com\/output\/job\/${CIRCLE_WORKFLOW_JOB_ID}\/artifacts\/0\/tmp\/coverage\/htmlcov-${report_type}\/index.html)/g")
```
frodre, line 13: I would move the replacing string with the artifact url into a variable with a name that's easier to understand, and then reuse it in the sed command.  (it looks like they are identical)

e.g., 
`ci_report_artifact="https:\/\/output.circle-artifacts.com\/output\/job\/${CIRCLE_WORKFLOW_JOB_ID}\/artifacts\/0\/tmp\/coverage\/htmlcov-${report_type}\/index.html)"`

.github/pull_request_template.md
```diff
@@ -20,3 +20,5 @@ Requirement changes:
 - [ ] Tests added
 
 Resolves #<github issues> [JIRA-TAG]
+
+Coverage reports (updated automatically):
```
frodre, line 24: Might want to add a small note here to not delete since the script seems to look for this for inserting coverage.

frodre (approved):
lgtm!  Thanks Jeremy.  I was expecting there might be a few different sections for the different unit test parameters in CI, but it looks like you just combine them in the follow-on PR.

.circleci/update_coverage_description.sh
mcgibbon: Great idea.

mcgibbon:
Updated, deleted the placeholder to check it works.

mcgibbon:
Worked with no placeholder. Double-checking the with-placeholder case also still works, since I edited that if-clause as well.

mcgibbon:
Worked. Auto-merge should complete in 10-20 minutes when the SSH session I started closes.

Issue #1876, opened 2022-06-08
Fortran physics diagnostics not always saved out


AnnaKwa:
I ran into an issue in which some recent prognostic runs completed running over an entire year but only had `sfc_dt_atmos` data saved for the first few segments. 

ex. 
![image](https://user-images.githubusercontent.com/16710132/172728967-35ef0869-facb-41e5-83ea-35f68a829034.png)

When we looked at the logs it appeared that the fortran diagnostics for the sfc_dt_atmos netcdfs was not being output after the initial first 5 segments. 

I added [an example in this debugging branch](https://github.com/ai2cm/fv3net/tree/debug/prog-diags-saving/workflows/prognostic_c48_run/reproduce-issue) that reproduces the issue: `make enter_prognostic_run`, `cd workflows/prognostic_c48_run/reproduce-issue`, `./test.sh`. This executes two prognostic runs: one using the IC from one segment into the longer prognostic run, which successfully saves out the `sfc_dt_atmos.tile*.nc` diagnostics, and another which uses the IC from later on after this data stops being saved, which does not save these netcdfs. The fv3config.yml in the reproducible example is copied directly from the prognostic run used for the ICs, with the exception of changing the layout to (1,1) and shortening the total run time.

Another thing to note is that multiple prognostic runs using different ML models all stopped saving the sfc_dt_atmos netcdfs at the same segment of the run.

spencerkclark:
@AnnaKwa ah I think I may know what's going on.  Try setting `atmos_model_nml.fhmax` to something larger than `1024.0` (ideally it should be at least the number of hours you plan to run the entire simulation for, not just a single segment).  I believe the fortran model will stop writing out physics diagnostics once it has run `fhmax` total hours (this is consistent with when things stopped in your run).  I think the reason we did not see this before was https://github.com/ai2cm/fv3config/issues/147. 

Assuming that works, we should probably pick a better default for that parameter.

AnnaKwa:
Great, thanks- it was exactly that. I'll update the defaults to use a bigger number.

spencerkclark:
See ai2cm/fv3gfs-fortran#302 for the implementation of a flag that would allow for the indefinite output of physics diagnostics at a regular frequency.

PR #1877, opened 2022-06-09
Extend fhmax default

AnnaKwa:
Fortran physics diags weren't saving for runs extending past 1024 hrs after the IC, this number is controlled by the namelist parameter `fhmax`. This didn't pop up earlier because we were writing the diag table differently prior to [these changes to address the reproducible restarts issue](https://github.com/ai2cm/fv3config/issues/147).

Resolves #<https://github.com/ai2cm/fv3net/issues/1876>


spencerkclark (commented):
Thanks @AnnaKwa -- I think the regression tests for `prepare_config` need to be update to reflect this change.  Also did you intend to commit the config and script to reproduce this issue?

At some point may we want to bump [this parameter](https://github.com/ai2cm/fv3gfs-fortran/blob/075d5150dfdcad5e8165d2b7ed5771e78ae40095/FV3/atmos_model.F90#L183) up in the fortran as well, since it controls the length of the possible diagnostics array (or perhaps rewrite things in a way such that there is not a limit).  For now I think this works since you can still output at a frequency up to 3 hourly for the full new `fhmax` and stay under the current cap of `65536`.

spencerkclark:
> At some point may we want to bump [this parameter](https://github.com/ai2cm/fv3gfs-fortran/blob/075d5150dfdcad5e8165d2b7ed5771e78ae40095/FV3/atmos_model.F90#L183) up in the fortran as well, since it controls the length of the possible diagnostics array (or perhaps rewrite things in a way such that there is not a limit). For now I think this works since you can still output at a frequency up to 3 hourly for the full new `fhmax` and stay under the current cap of `6553`

See ai2cm/fv3gfs-fortran#301 for a more complete description of this issue and a possible more robust solution.

spencerkclark (approved):
Great!  Looks good now.

PR #1878, opened 2022-06-09
Add rectified bias to offline training scores

frodre:
The bias of the gscond-only models does not tell the whole story when the subsequent fortran call removes negative cloud water values.  This PR adds a "rectified" bias term which compares the prediction of condensate after gscond where all values < 0 = 0 to a target.

- [x] tested


nbren12 (approved):
lgtm. pair programmed.

PR #1879, opened 2022-06-09
Revert "Extend fhmax default (#1877)"

AnnaKwa:
The last PR caused seg faults for prognostic runs with lower physics diagnostics save intervals- they ran out of space in their `fdiag` arrays. Reverting to the previous default for fhmax, and [the issue with physics diags not saving after ~40d](https://github.com/ai2cm/fv3net/issues/1876) will instead be fixed on the fortran side (solution proposed [here](https://github.com/ai2cm/fv3gfs-fortran/issues/301)). 

Instead of changing the default, in the meantime if we are saving data for longer than 1024 hr, fhmax can be manually updated to be larger in the user config.

### DIFF
--- a/external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
+++ b/external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
@@ -23,7 +23,7 @@ namelist:
     chksum_debug: false
     dycore_only: false
     fdiag: 0.0
-    fhmax: 99999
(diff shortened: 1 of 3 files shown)


spencerkclark:
Thanks @AnnaKwa.  I anticipated this issue, yet somehow it didn't occur to me that we output physics diagnostics at every timestep in most of our other workflows.  Perhaps I should prioritize ai2cm/fv3gfs-fortran#301.

PR #1886, opened 2022-06-13
Use smaller version of sample prog run data

### LINKED ISSUE #1885: Random failures of prognostic run report test
The test_prognostic_run_report test is often [randomly failing on CI](https://app.circleci.com/pipelines/github/ai2cm/fv3net/12157/workflows/88eccee3-8ca8-4980-9016-1a10c3fd099b/jobs/109193) (about 10-30% of the time), possibly due to running out of memory.

This might be solvable by decreasing the size of the test dataset, though it depends on what is actually causing the random failure.

oliverwm1:
The prognostic run report test is slow and sometimes fails (we think due to OOM issues). This PR updates that test to use a lighter-weight version of the test data. Specifically the new data has a time dimension that is about 1/4 as long, and it has fewer data variables. It should hopefully resolve #1885

### DIFF
--- a/workflows/diagnostics/tests/prognostic/test_integration.sh
+++ b/workflows/diagnostics/tests/prognostic/test_integration.sh
@@ -13,7 +13,7 @@ cd workflows/diagnostics
 
 # compute diagnostics/mterics for a short sample prognostic run
 mkdir -p /tmp/$random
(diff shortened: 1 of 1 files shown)


oliverwm1:
On latest commit, RAM usage is down to 50% from around 90%. Also runtime is down to ~8 minutes.

PR #1888, opened 2022-06-14
Enable the indefinite output of physics diagnostics by default

### LINKED ISSUE #1876: Fortran physics diagnostics not always saved out

spencerkclark:
This PR bumps the version of the fortran model to enable outputting physics diagnostics indefinitely by default.  It resolves #1876.  For reference see also ai2cm/fv3gfs-fortran#301 and ai2cm/fv3gfs-fortran#302.

- test_unit: [83%](https://output.circle-artifacts.com/output/job/aba1a6a5-30ed-4276-a9ce-07086843c827/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


AnnaKwa (approved):
Thanks!

spencerkclark:
Despite the timeout in CircleCI, the integration tests appear to have passed.  I looked at the job from the command line, e.g.:
```
$ argo get integration-test-80f039e-dd0f
```
Therefore I will go ahead and merge.

PR #1890, opened 2022-06-20
Add TendencyToFlux and MSE transforms

oliverwm1:
To enable predicting fluxes while optimizing tendencies—or adding flux terms like precipitation to the loss function—this PR adds tensor transforms to compute fluxes from tendencies, and to compute the MSE tendency from temperature and specific humidity tendency.

Added public API:
- `TendencyToFlux` and `MoistStaticEnergyTransform`
- to `vcm.DerivedMapping`: net TOA radiative flux and upward surface flux (sensible + radiation)

Significant internal changes:
- loss variables are now used when determining required model variables
- pressure thickness is manually added when using `TendencyToFlux` transform (since this operates on outputs)

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/2437b75a-f4aa-45de-bcba-f137056acbef/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/vcm/vcm/derived_mapping.py
```diff
@@ -360,3 +360,43 @@ def water_vapor_path(self):
     return da.assign_attrs(
         {"long_name": "column integrated water vapor", "units": "mm"}
     )
+
+
+@DerivedMapping.register(
```
nbren12, line 365: Are these only used for your specific model in`fv3fit`? If so, I would consider isolating these near your fv3fit code. You can call `DerivedMapping.register` from any file...doesn't have to be here.

external/vcm/vcm/derived_mapping.py
```diff
@@ -360,3 +360,43 @@ def water_vapor_path(self):
...
+    "total_sky_net_radiative_flux_at_top_of_atmosphere",
+    required_inputs=[
+        "total_sky_downward_shortwave_flux_at_top_of_atmosphere",
+        "total_sky_upward_shortwave_flux_at_top_of_atmosphere",
+        "total_sky_upward_longwave_flux_at_top_of_atmosphere",
+    ],
+)
+def total_sky_net_radiative_flux_at_top_of_atmosphere(self):
```
nbren12, line 373: Is "sky" implied by "top of atmosphere"? Or is this some CF name that you are using?
```suggestion
def total_net_radiative_flux_at_top_of_atmosphere(self):
```

external/vcm/vcm/derived_mapping.py
```diff
@@ -360,3 +360,43 @@ def water_vapor_path(self):
...
+    result = (
+        self["total_sky_downward_shortwave_flux_at_top_of_atmosphere"]
+        - self["total_sky_upward_shortwave_flux_at_top_of_atmosphere"]
+        - self["total_sky_upward_longwave_flux_at_top_of_atmosphere"]
+    )
+    if "storage_of_internal_energy_path_due_to_fine_res_temperature_nudging" in self:
```
nbren12, line 379: This seems like a modeling assumption specific to your ML model. Is this appropriate for the general use implied by including it in `vcm`?

nbren12 (approved):
Overall the implementations look good but the new transform and derived variables seem to make assumptions specific to your particular ML method/dataset. I have a high level request to unify all these additions into a single directory tree (`fv3fit/emulation/flux`?) which can then be imported where needed. I realize we have "zhao carr" specific code scattered around the code base, but I've been wanting to clean that up. Since you are just getting started I think you start on a better foot.

external/vcm/vcm/derived_mapping.py
oliverwm1: "total sky" is in contrast with "clear sky" (i.e. no clouds). I'm following convention established here: 

https://github.com/ai2cm/fv3gfs-fortran/blob/cf1eb54ab439059dc7796e3b0b16404a5dda4966/FV3/wrapper/fv3gfs/wrapper/physics_properties.json#L52

external/vcm/vcm/derived_mapping.py
oliverwm1: Good idea.

oliverwm1: Hmm realizing this won't actually work, since I want the data provided by `loaders` to include these derived quantities. Other than the possible usage of the fine-res temperature nudging tendency (which is not required) they are generally useful.

external/vcm/vcm/derived_mapping.py
oliverwm1: Can't actually register in fv3fit. Alternatively I could define this quantity in the fine-res loader.

external/vcm/vcm/derived_mapping.py
oliverwm1: I left the `upward_heat_flux_at_surface` derived quantity here since it is general (even if not currently used elsewhere) and moved the TOA net flux calculation to the fine-res loader.

external/vcm/vcm/derived_mapping.py
nbren12: Are you sure this won't work? We just need to sure the "registration" is performed when needed (prog run, offline, training). Importing `fv3fit` should be enough to register these. Alternatively you could add it to the mapper (or define them as a tensor_transform).

oliverwm1: I did try it, and at least when using the `DerivedMapping` (imported from vcm) within a pytest in fv3fit, it did not have the quantities registered within fv3fit available. Maybe I needed to do some additional fv3fit import. However, I'm not sure that import would be guaranteed to happen if just using loaders anyway. I moved the TOA flux calculation to the fine-res mapper.

nbren12: Maybe it's a moot point, but I think the pattern should work...I've used it before. If you are interested, I could help you debug if you point me to the test code. The module containing the code was probably not being imported.

oliverwm1: I deleted the test code. I think I'm just going to merge as is and move on.

nbren12: No problem.

PR #1896, opened 2022-06-24
Online prognostic run with classifier based masking

frodre:
Add options to use a classifier model to mask zero_cloud or zero_tendency cloud water output cases.

- test_unit: [83%](https://output.circle-artifacts.com/output/job/67aec395-4c79-4ac6-9cfc-7e88f90d2268/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/emulation/emulation/_emulate/microphysics.py
+++ b/external/emulation/emulation/_emulate/microphysics.py
@@ -140,9 +140,9 @@ class MicrophysicsHook:
                 dimensions or [sample]
         """
 
-        model_outputs = predict(self.model, state)
+        model_outputs = _predict(self.model, state)
         if self.classifier is not None:
-            classifier_outputs = predict(self.classifier, state)
+            classifier_outputs = _predict(self.classifier, state)
             model_outputs.update(classifier_outputs)
 
         # fields stay in global state so check overwrites on first step
--- a/external/emulation/emulation/_emulate/microphysics.py
+++ b/external/emulation/emulation/_emulate/microphysics.py
@@ -79,7 +79,7 @@ def always_emulator(state: FortranState, emulator: FortranState):
     return emulator
 
 
-def predict(model: tf.keras.Model, state: FortranState) -> FortranState:
+def _predict(model: tf.keras.Model, state: FortranState) -> FortranState:
     # grab model-required variables and
     # switch state to model-expected [sample, feature]
     inputs = {name: state[name].T for name in model.input_names}
--- a/external/emulation/emulation/_emulate/microphysics.py
+++ b/external/emulation/emulation/_emulate/microphysics.py
@@ -141,9 +141,8 @@ class MicrophysicsHook:
         """
 
         model_outputs = predict(self.model, state)
-
         if self.classifier is not None:
-            classifier_outputs = predict(self.model, state)
+            classifier_outputs = predict(self.classifier, state)
             model_outputs.update(classifier_outputs)
 
         # fields stay in global state so check overwrites on first step
--- a/external/emulation/emulation/_emulate/microphysics.py
+++ b/external/emulation/emulation/_emulate/microphysics.py
@@ -25,21 +25,27 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
-@print_errors
 def _load_tf_model(model_path: str) -> tf.keras.Model:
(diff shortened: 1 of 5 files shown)


external/emulation/emulation/config.py
```diff
@@ -85,9 +86,13 @@ class ModelConfig:
     mask_gscond_identical_cloud: bool = False
     mask_gscond_zero_cloud: bool = False
     enforce_conservative: bool = False
+    mask_gscond_zero_cloud_classifier: bool = False
```
nbren12, line 89: With all these gscond specific options we should probably make a GscondModel subclass soon.

external/emulation/emulation/zhao_carr.py
```diff
@@ -90,10 +95,36 @@ def mask_where_fortran_cloud_identical(state, emulator):
         state[Input.cloud_water],
         emulator[GscondOutput.cloud_water],
     )
-    net_condensation = cloud_out - state[Input.cloud_water]
```
nbren12, line 93: do we have any coverage of these additions?

external/emulation/emulation/zhao_carr.py
```diff
@@ -90,10 +95,36 @@ def mask_where_fortran_cloud_identical(state, emulator):
...
-    return {**emulator, **apply_condensation_liquid_phase(state, net_condensation)}
+    return _update_with_net_condensation(cloud_out, state, emulator)
+
+
+def _get_classify_output(emulator, class_name):
+    names = sorted(CLASS_NAMES)
+    idx = names.index(class_name)
+    logit_classes = emulator["gscond_classes"]
+    # TODO this should probably be in transforms
+    one_hot = logit_classes == np.max(logit_classes, axis=0, keepdims=True)
```
nbren12, line 106: It's a bit confusing that this is transposed from the fv3fit conventions where the one hot encoded dimension is the last one. Do you think it would be clearer to pass axis as a function argument (e.g. one_hot_axis=0)?

external/emulation/emulation/zhao_carr.py
```diff
@@ -90,10 +95,36 @@ def mask_where_fortran_cloud_identical(state, emulator):
         state[Input.cloud_water],
         emulator[GscondOutput.cloud_water],
     )
-    net_condensation = cloud_out - state[Input.cloud_water]
-    return {**emulator, **apply_condensation_liquid_phase(state, net_condensation)}
+    return _update_with_net_condensation(cloud_out, state, emulator)
+
+
+def _get_classify_output(emulator, class_name):
```
nbren12, line 101: One disadvantage of passing the class_name separately is that this routine will not advertise what class names are available. What do you think about refactoring to `_get_classify_output(emulator) -> Mapping[str, tf.Tensor` where the `str` keys are the class names? This would make it compatible with the ["routing" code](https://github.com/ai2cm/fv3net/pull/1898/commits/3abd6d4254052f3f3e4547f0b60ca308e4ac65e4) I have in my PR.

external/emulation/emulation/_emulate/microphysics.py
```diff
@@ -73,6 +79,18 @@ def always_emulator(state: FortranState, emulator: FortranState):
     return emulator
 
 
+def predict(model: tf.keras.Model, state: FortranState) -> FortranState:
```
nbren12, line 82:
```suggestion
def _predict(model: tf.keras.Model, state: FortranState) -> FortranState:
```
Suggest keeping this private.

nbren12 (approved):
LGTM, but it would be nice to make sure we have some coverage of the new code paths. I think we can minimize the testing surface area a bit with my suggestions for `_get_classify_output`.

external/emulation/emulation/zhao_carr.py
frodre: I switched it to output all available classes with names as keys.

external/emulation/emulation/zhao_carr.py
frodre: Agree, thanks for the general discussion about the fv3fit/prognostic run ordering conventions.  I added the argument for now to add a bit of clarity.

external/emulation/emulation/zhao_carr.py
frodre: No coverage unfortunately.  Know the classification works based on the testing right now, but probably regtesting is the simplest approach.

PR #1899, opened 2022-06-27
Allow scaling predicted tendencies via config in prog run

mcgibbon:
This PR adds the ability to scale the tendencies produced by our ML models during prognostic runs

Refactored public API:
- The prognostic run configuration now has an optional `scaling` field in its machine learning configuration which can be set to a mapping from ML tendency name to scaling factor.

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/5d4b688b-465b-4604-8473-9e94f9badd59/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -343,13 +343,17 @@ def load(cls, path: str) -> "SklearnWrapper":
...
         packer_config = dacite.from_dict(PackerConfig, packer_config_dict)
 
         obj = cls(input_variables, output_variables, model, packer_config=packer_config)
         obj.target_scaler = scaler_obj
+        output_features_dict = metadata["output_features"]
+        if isinstance(output_features_dict, dict):
+            output_features_ = dacite.from_dict(PackingInfo, data=output_features_dict)
+        else:
+            # older model format, uses tuple instead
+            output_features_ = PackingInfo.from_tuples(output_features_dict[1])
```
mcgibbon, line 356: This was necessary to be able to load @spencerkclark 's older models.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -85,6 +86,44 @@ def get_momentum_diagnostics(self, state, tendency) -> Diagnostics:
...
+        """Label used for naming diagnostics.
+        """
+        return self._stepper.label
+
+    def __call__(self, time, state) -> Tuple[Tendencies, Diagnostics, State]:
+        tendencies, diagnostics, state = self._stepper(time, state)
+        for name, scale in self._scales.items():
+            if name in tendencies:
+                tendencies[name] *= scale
+        return tendencies, diagnostics, state
```
oliverwm1, line 116: What if there is some item in `diagnostics` that is computed from something in `tendencies`? (like a column-integrated tendency). This would lead to an incorrect diagnostic.

workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -63,6 +63,7 @@ class UserConfig:
     online_emulator: Optional[
         Union[runtime.transformers.emulator.Config, runtime.transformers.fv3fit.Config]
     ] = None
+    scale_tendencies: Mapping[str, float] = dataclasses.field(default_factory=dict)
```
oliverwm1, line 66: Feels odd for this to be a top-level config item. Why not add it as part of `MachineLearningConfig` since that is where it is intended to be used (if I understand correctly).

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -315,6 +356,8 @@ def _get_postphysics_stepper(
         else:
             self._log_info("Performing baseline simulation")
             stepper = None
+        if len(config.scale_tendencies) > 0 and stepper is not None:
+            stepper = TendencyScalingStepper(stepper, config.scale_tendencies)
```
oliverwm1, line 360: Is the intent to allow scaling nudging tendencies too? If not, it would be better to scope this more narrowly to the PureMLStepper.

oliverwm1 (changes requested):
I think this change would be better scoped more narrowly to the `PureMLStepper`—let me know if you think otherwise.

Also with current implementation, some diagnostics will be inconsistent with the scaled tendencies.

oliverwm1:
One simple way to implement would be to add an `OutputScalingAdapter` similar to the `MultiModelAdapter`.

mcgibbon:
> One simple way to implement would be to add an `OutputScalingAdapter` similar to the `MultiModelAdapter`.

This would be reasonable to do, instead I put the functionality into MultiModelAdapter because the typing specifies MultiModelAdapter in particular and not some kind of "ModelAdapter" ABC/protocol I could implement another version of. I can add that and split out the class if you like.

oliverwm1 (approved):
LGTM, thanks for the changes.

PR #1900, opened 2022-06-27
Drop tile-only variables before coarsening

AnnaKwa:
If time averaging is set for the fortran diagnostics in the user config, the fortran model inserts a few additional variables `average_T1, average_T2, average_DT` into the atmos diags which only have tile and time dimensions. This causes the coarsening step to choke during the prognostic run diags. This PR hard codes these variables as a list to drop before coarsening. I did this instead of automatically dropping any variables without the full {x, y, tile} dims because I wasn't sure if it was always the case that the coarse run data has all these dims and didn't want to accidentally drop more variables than this set.

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/constants.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/constants.py
@@ -3,6 +3,10 @@ from typing import Mapping, Sequence
 
 MovieUrls = Mapping[str, Sequence[str]]
 
+# Added to atmos diags by fortran model if interval-averaged.
+# Removed before coarsening run data to avoid errors.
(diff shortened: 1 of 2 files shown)


oliverwm1 (changes requested):
We already do some variable-dropping before the coarsening step. I think it would be better if we called `_drop_tile_only_vars` in the same place instead of in the coarsening function. Or just add the new sequence of vars to the existing drop call.

https://github.com/ai2cm/fv3net/blob/c0f7a54d6396c57dd894afe29e8d443c0836a72f/workflows/diagnostics/fv3net/diagnostics/prognostic_run/load_run_data.py#L129-L131

oliverwm1 (approved):
LGMT, thanks!

PR #1901, opened 2022-06-28
Drop fortran time averaging info variables

AnnaKwa:
If fortran diags are time averaged, additional variables `average_T1, average_T2, average_DT` are saved out. These cause issues later down in the online diagnostics pipelines as they are saved for each time and tile but not x/y. The information is also pretty redundant as the only useful part is the value of DT (which can be inferred from the time dimension anyhow). 
At first I tried just dropping these in the prog run diags loading function, but since these variables also cause issues in other code like the regridding, I'm moving the fix to the source and just dropping these variables during postprocessing. The dt information is saved in the dataset attribute `time_averaging_interval` so that there is still some record in the dataset that values are time averaged.

- [x] Tests added
 

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/constants.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/constants.py
@@ -3,10 +3,6 @@ from typing import Mapping, Sequence
 
 MovieUrls = Mapping[str, Sequence[str]]
 
-# Added to atmos diags by fortran model if interval-averaged.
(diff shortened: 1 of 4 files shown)


AnnaKwa:
Attempted to keep these with just the time dimension for the case Oli mentioned where dt might not be constant, but for some reason `average_T1, T2` still choke in coarsening while `average_DT` does not despite similar dims. I'll try keeping these variables in preprocessing with time dim only, and dropping in the data loading function.

PR #1904, opened 2022-06-29
Add end of epoch checkpoint custom callback

AnnaKwa:
- Add a registry for custom callbacks so that if any new ones are added to `fv3fit.keras._models.shared.callbacks` they can be registered with `@register_custom_callback` to be available for use.

- Add custom callback `EpochModelCheckpoint` which inherits from `keras.callbacks.ModelCheckpoint` but has an additional kwarg `period` which allows checkpoints to be saved every `period` epochs instead of every epoch. This used to be an option in `keras.callbacks.ModelCheckpoint`  but was deprecated in favor of only being able to specify the save interval in number of samples.

- Removed the defunct option `save_model_checkpoints` from `DenseHyperparameters`. This was used before the refactor of the training loop allowed built in keras callbacks to be used.

- test_unit: [83%](https://output.circle-artifacts.com/output/job/05b412de-f78e-4ff3-974f-2486e3d8dd6e/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/keras/_models/shared/callbacks.py
```diff
@@ -2,8 +2,21 @@
 import tensorflow as tf
 import wandb
 
+# Add third party callbacks here to make them available in training
+ADDITIONAL_CALLBACKS = {"WandbCallback": wandb.keras.WandbCallback}
```
mcgibbon, line 6: Why was this renamed? Our callbacks are also third party callbacks (from the standpoint of Keras), and the error message describing these as "third party callbacks" wasn't updated. I'd suggest either updating that error message or reverting the variable name change.

external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -81,7 +79,6 @@ class DenseHyperparameters(Hyperparameters):
         default_factory=TrainingLoopConfig
     )
     loss: LossConfig = LossConfig(scaling="standard", loss_type="mse")
-    save_model_checkpoints: bool = False
```
mcgibbon, line 84: As long as you're sure this breaking API change is not important.

Is it easy enough to figure out how to set these callbacks for use in config? Wherever we document that argument, the documentation should probably be updated to say it can include any option registered using your new registration function.

external/fv3fit/fv3fit/keras/_models/shared/callbacks.py
```diff
@@ -2,8 +2,21 @@
...
 
-THIRD_PARTY_CALLBACKS = {"WandbCallback": wandb.keras.WandbCallback}
+
+def register_custom_callback(name: str):
+    """
+    Returns a decorator that will register the given training function
+    to be usable in training configuration.
```
mcgibbon, line 12: Decorators are generally confusing when describing their api usage in English, can you include an "examples" section in this docstring with a usage example?

external/fv3fit/tests/training/test_main.py
```diff
@@ -506,15 +506,7 @@ def cli_main(args: argparse.Namespace):
...
-        pytest.param(
-            "dense", {"save_model_checkpoints": False}, ["dQ1", "dQ2"], id="dense"
-        ),
-        pytest.param(
-            "dense",
-            {"save_model_checkpoints": True},
-            ["dQ1", "dQ2"],
-            id="dense_with_checkpoints",
-        ),
+        pytest.param("dense", {}, ["dQ1", "dQ2"], id="dense"),
```
mcgibbon, line 509: I think I'm OK approving without tests for this new feature, since the registration pattern is simple and testing this could be very complex. Consider whether there are any tests you'd like to add.

mcgibbon (approved):

external/fv3fit/fv3fit/keras/_models/shared/callbacks.py
AnnaKwa: Sure. I thought 'third party' mean third party relative to the us+keras.

mcgibbon: That's not an unreasonable interpretation, renaming these could be fine.

PR #1908, opened 2022-07-01
Min-max novelty detector and tests

chsanford:
This PR creates the blueprint for novelty detection algorithms by introducing the simplest novelty detection predictor, which identifies whether a column is out-of-sample by comparing coordinate-wise to the maximum and minimum values. It also introduces new unit tests for this that can be applied for all such predictors.

Significant internal changes:
- Added MinMaxNoveltyDetector, a Predictor that identifies out-of-sample points.

- [ ] Tests for novelty detectors in test_train_novelty_detection.py that ensures that outputs are properly configured and that  extreme scalings of toy inputs are classified as novelties.
- [ ] Segments tests in test_train.py into two categories: those that apply to all learning algorithms (supervised and unsupervised) and those that only apply to supervised tasks. That way, tests of things like the ability to properly save and load models extend to all, but explicitly supervised tests are not applied to the novelty detection algorithms. If this isn't organized the way others would like, I'm happy to rearrange.

### DIFF
--- a/external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
+++ b/external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
@@ -1,0 +1,131 @@
+from fv3fit._shared import stacking
+from .. import _shared
+from .._shared import (
+    match_prediction_to_input_coords,
+    pack,
+    pack_tfdataset,
+    Predictor,
+    register_training_function,
+    SAMPLE_DIM_NAME,
+    stack_non_vertical,
+)
+from .._shared.config import MinMaxNoveltyDetectorHyperparameters, PackerConfig
+
+import dataclasses
+import dacite
+import fsspec
+from fv3fit import tfdataset
+from fv3fit._shared.packer import PackingInfo, clip_sample, unpack
+from fv3fit.tfdataset import apply_to_mapping, ensure_nd
+from fv3fit.typing import Batch
+import io
+import joblib
+import logging
+import numpy as np
+from sklearn.preprocessing import MinMaxScaler
+import tensorflow as tf
+from typing import Optional, Iterable, Sequence
+from vcm import safe
+import xarray as xr
+import yaml
+
+@register_training_function("min_max_novelty_detector", MinMaxNoveltyDetectorHyperparameters)
+def train_min_max_novelty_detector(
+    hyperparameters: MinMaxNoveltyDetectorHyperparameters,
+    train_batches: tf.data.Dataset,
+    validation_batches: tf.data.Dataset
+):
+    train_batches = train_batches.map(
+        tfdataset.apply_to_mapping(tfdataset.float64_to_float32)
+    )
+    
+    model = MinMaxNoveltyDetector(hyperparameters)
+    model.fit(train_batches)
+    return model
+
+@_shared.io.register("minmax")
+class MinMaxNoveltyDetector(Predictor):
+
+    _PICKLE_NAME = "minmax.pkl"
+    _METADATA_NAME = "metadata.bin"
+
+    _NOVELTY_OUTPUT_VAR = "is_novelty"
+
+    dims = ["x", "y", "tile", "time"]
+
+    def __init__(
+        self,
+        hyperparameters: MinMaxNoveltyDetectorHyperparameters
+    ):
+        output_variables = [self._NOVELTY_OUTPUT_VAR]
+        super().__init__(hyperparameters.input_variables, output_variables)
+        self.packer_config = hyperparameters.packer_config
+
+    def fit(self, batches: tf.data.Dataset):
+        logger = logging.getLogger("MinMaxNoveltyDetector")
+        logger.info(f"Fitting min-max novelty detector.")
+
+        batches = batches.map(apply_to_mapping(ensure_nd(2)))
+        batches_clipped = batches.map(clip_sample(self.packer_config.clip))
+        x_dataset, self.input_features_ = pack_tfdataset(
+            batches_clipped, [str(item) for item in self.input_variables]
+        )
+        x_dataset = x_dataset.unbatch()
+
+        total_samples = sum(1 for _ in iter(x_dataset))
+        X: Batch = next(iter(x_dataset.batch(total_samples)))
+
+        self.scaler = MinMaxScaler()
+        self.scaler.fit(X)
+
+        logger.info(f"Min-max novelty detector done fitting.")
+
+    def predict(self, data: xr.Dataset) -> xr.Dataset:  
+        stack_dims = [dim for dim in data.dims if dim not in stacking.Z_DIM_NAMES]  
+        stacked_data = data.stack({SAMPLE_DIM_NAME: stack_dims}).transpose(SAMPLE_DIM_NAME, ...)
+
+        X, _ = pack(stacked_data[self.input_variables], [SAMPLE_DIM_NAME], self.packer_config)
+        scaled_X = self.scaler.transform(X)
+        larger_than_max = np.where(scaled_X.max(axis=1) > 1, 1, 0)
+        smaller_than_min = np.where(scaled_X.min(axis=1) < 0, 1, 0)
+        stacked_is_novelty = larger_than_max + smaller_than_min
+
+        new_coords = {k: v for (k, v) in stacked_data.coords.items() if k not in stacking.Z_DIM_NAMES}
+        stacked_is_novelty = xr.DataArray(stacked_is_novelty, dims=[SAMPLE_DIM_NAME], coords=new_coords)
+        is_novelty = stacked_is_novelty\
+            .to_dataset(name=self._NOVELTY_OUTPUT_VAR)\
+            .unstack(SAMPLE_DIM_NAME)
+
+        return match_prediction_to_input_coords(data, is_novelty)
+
+    def dump(self, path: str) -> None:
+        fs: fsspec.AbstractFileSystem = fsspec.get_fs_token_paths(path)[0]
+        fs.makedirs(path, exist_ok=True)
+
+        mapper = fs.get_mapper(path)
+
+        f = io.BytesIO()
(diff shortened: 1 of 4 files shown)


external/fv3fit/tests/training/test_train.py
```diff
@@ -126,23 +146,28 @@ def _get_dataset_precipitative(sample_func):
...
-    i_3d_input = input_variables.index("var_in_3d")
-    output_values = [input_values[i_3d_input]]
+    if labeled:
+        output_variables = ["var_out"]
+        i_3d_input = input_variables.index("var_in_3d")
+        output_values = [input_values[i_3d_input]]
+        processed_dataset = _process(input_variables, input_values, output_variables, output_values)
+    else:
+        output_variables = []
+        processed_dataset = _process_unlabeled(input_variables, input_values)
```
AnnaKwa, line 161: What breaks if the new model is given the original dataset default with the output variables? 
I think typically our models should be able to fit/predict on any dataset that has additional variables on top of the required inputs/outputs.

AnnaKwa (commented):
Looks good, nice job implementing the new detector! My main suggestion is to add a `NoveltyDetector(Predictor)` class from which the min max detector and future detectors would fall under (see comments).
The fortran submodule likes to sneak into PRs, can you remove it?

external/fv3fit/tests/training/test_train.py
```diff
@@ -27,18 +27,32 @@
...
 
 
 # automatically test on every registered training class
-@pytest.fixture(params=GENERAL_TRAINING_TYPES)
+@pytest.fixture(params=GENERAL_TRAINING_TYPES+UNLABELED_TRAINING_TYPES)
 def model_type(request):
     return request.param
 
+# automatically test on every registered training class with labeled data
+@pytest.fixture(params=GENERAL_TRAINING_TYPES)
```
AnnaKwa, line 52: To avoid confusion with naming what is now `GENERAL_TRAINING_TYPES` would more accurately be named `SUPERVISED_... or LABELED_...`  and  `GENERAL_... = LABELED.. + NONLABELED_`

nbren12 (commented):
@mcgibbon  Can we make the "learns the identity" test an opt-in? It is not relevant to all ML problems and when it fails it is often a false positive or issue with some tolerance. (i ran into the same challenges as clayton did here when adding the emulation training function).

external/fv3fit/tests/training/test_train.py
```diff
@@ -27,18 +27,32 @@
     "dense",
     "transformed",
 ]
+
+# unlabeled models that can be tested on all model configuration tests, but not those
+# that actually depend on the quality of supervised learning that occurs
+UNLABELED_TRAINING_TYPES = [
```
nbren12, line 33: Classification problems are "labeled" but will also probably fail some of the tests. Is there a name for this list that will also included classifiers? (e.g. NON_REGRESSION_TYPES)

mcgibbon:
> @mcgibbon Can we make the "learns the identity" test an opt-in? It is not relevant to all ML problems and when it fails it is often a false positive or issue with some tolerance. (i ran into the same challenges as clayton did here when adding the emulation training function).

"learns the identity" is a bit of a misnomer for that test and it could be renamed, really it supports any supervised example you think your model should be able to predict. For example on a classifier you might give data uniformly distributed between 0 and 1 as input, with the output being `(input > 0.5).astype(np.int)`. Especially since your (very good) updates to the data fixture, it is very straightforward to independently define the target "problem" (i.e. data) for each model's test.

Of course, having a separate test file for a model is also perfectly fine. What's important to me is that some kind of supervised training problem is tested.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(
+        self,
+        hyperparameters: MinMaxNoveltyDetectorHyperparameters
+    ):
+        output_variables = [self._NOVELTY_OUTPUT_VAR]
+        super().__init__(hyperparameters.input_variables, output_variables)
+        self.packer_config = hyperparameters.packer_config
+
+    def fit(self, batches: tf.data.Dataset):
```
mcgibbon, line 64: Feel free to ignore this, but just an FYI from previous iterations of code like this.

Having `fit` as a method on the predictor is something we've had on our predictors before, and makes the APIs a bit more complicated. For example, it's possible to call `predict` before `fit`, which means you either need error message logic or you'll give a confusing error message, and it's possible to call `fit` after `load` or `dump` before `fit`, etc. It's also possible to call `fit` multiple times, in which case it's not obvious what should happen (should the training be re-done? updated incrementally?).

These issues are avoided if you make a `fit(batches) -> Predictor` function. For example, if you make `input_features_` and `scaler` arguments of `__init__` for this class, you don't need a fit method because the fitting happens before constructing the predictor.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
```diff
@@ -0,0 +1,131 @@
...
+
+        total_samples = sum(1 for _ in iter(x_dataset))
+        X: Batch = next(iter(x_dataset.batch(total_samples)))
+
+        self.scaler = MinMaxScaler()
+        self.scaler.fit(X)
+
+        logger.info(f"Min-max novelty detector done fitting.")
+
+    def predict(self, data: xr.Dataset) -> xr.Dataset:
```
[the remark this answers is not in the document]
chsanford: Thank you, that's a great suggestion! The only ways in which I've currently deviated are (1) having the output of `.predict` be an `xr.Dataset` rather than the direct output of the sklearn class (so that we can specify the output for every novelty detector and that we can use the same `predict_novelty` method for each) and (2) have 1/0 outputs rather than Booleans, since it seems somewhat easier to deal with for aggregation. I'm flexible on either.

external/fv3fit/tests/training/test_train.py
chsanford: Just eliminated this difference! You're right that we can just ignore the output variables.

external/fv3fit/tests/training/test_train.py
chsanford: Regarding this and @nbren12 's comment below, I'm currently going with `REGRESSION_TRAINING_TYPES` and `NON_REGRESSION_TRAINING_TYPES`.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
chsanford: I'll make this change once I get back; it should be easy. My only hesitation was that `predict` and `fit` the way they're currently defined are the way things are done with `sklearn` classes (albeit with exceptions if you call predict on an untrained model) and how `_random_forest.py` is configured too. Would you like me to change random forest to make it too match this convention?

chsanford:
Thank you all for the helpful comments! FYI, I'm still not sure why the circleci tests aren't running.

nbren12:
> Thank you all for the helpful comments! FYI, I'm still not sure why the circleci tests aren't running.

I'll look into it.

nbren12:
@chsanford I submitted a ticket with circle ci. Can you try opening up another dummy PR with an empty commit to see if you can get the tests running? Maybe this one ended up in limbo somehow.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
mcgibbon: Refactoring the random forest class to behave like this has been on my to-do list for almost a year, just isn't high enough priority since the code works and we haven't needed to make any changes to it in that time. I wouldn't ask you to do it since you only have so much time with us, but if you want to do this refactor ask your mentors so they can help you prioritize. If you refactor the RF class I'd suggest doing it in a different PR to separate the changes.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
```diff
@@ -0,0 +1,161 @@
...
+
+        stack_dims = [dim for dim in data.dims if dim not in stacking.Z_DIM_NAMES]
+        stacked_data = data.stack({SAMPLE_DIM_NAME: stack_dims})
+        stacked_data = stacked_data.transpose(SAMPLE_DIM_NAME, ...)
+
+        X, _ = pack(
+            stacked_data[self.input_variables], [SAMPLE_DIM_NAME], self.packer_config
+        )
+        scaled_X = self.scaler.transform(X)
+        scores_larger_than_max = np.maximum(scaled_X.max(axis=1) - 1, 0)
```
AnnaKwa, line 105: It might be convenient to have another method apart from predict that returns just the scores normalized to the min/max in dataset form without the z dimension collapse. ex. you could see which level(s) were going out of range. Up to you if you think that is useful enough to be added as a method - the same information could be manually calculated with `MinMaxNoveltyDetector.scaler`. If you decide to add that it could also be called first in `predict` and the rest of `predict` could just transform the dataset into the stacked scores.

external/fv3fit/tests/training/test_train.py
```diff
@@ -20,29 +21,42 @@
...
 # automatically test on every registered training class
-@pytest.fixture(params=GENERAL_TRAINING_TYPES)
+@pytest.fixture(params=REGRESSION_TRAINING_TYPES + NON_REGRESSION_TRAINING_TYPES)
 def model_type(request):
     return request.param
 
 
+# automatically test on every registered training class with labeled data
+@pytest.fixture(params=REGRESSION_TRAINING_TYPES)
+def labeled_model_type(request):
```
AnnaKwa, line 53: Could you also update the name here and in references to `regression_model_type`?

AnnaKwa (approved):
Looks good! I had a minor optional suggestion and additional edit to go with the naming change in the test. I don't need to rereview those changes. Also, the fortran submodule is still showing up in the PR.

PR #1911, opened 2022-07-01
Routed regression (rebased)

nbren12:
This PR contains the routed regression training and prognostic run code (external/emulation). The main changes are
1. allow masking a function in the loss
2. add a transform for modifying the classes with ML outputs ('GscondRoute')
3. Ensure the regression model (and backwards steps) use the classifiers predictions of the mask.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/5ca4859c-c2d9-48af-bc82-a95277017eb0\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/fv3fit/fv3fit/emulation/models/transformed_model.py
```diff
@@ -1,9 +1,116 @@
...
 from fv3fit.keras.adapters import ensure_dict_output
 from fv3fit.emulation.transforms import TensorTransform
+from fv3fit.emulation.types import TensorDict
 import tensorflow as tf
 
-__all__ = ["transform_model"]
+__all__ = ["transform_model", "TransformedModel"]
+
+
+class TransformedModel(tf.keras.Model):
```
nbren12, line 10: I also considered simply saving separate keras Functional models for the forwards and backwards transformations, but ended up using a custom model instead. The use of tf.functions below lead to some complexity down the road, since they can "memorize" the wrong inputs.

external/emulation/emulation/models.py
```diff
@@ -0,0 +1,73 @@
...
+        return model_outputs
+
+
+class TransformedModelWithClassifier:
+    def __init__(self, model, classifier):
+        self.model = model
+        self.classifier = classifier
+
+    def __call__(self, state: FortranState) -> FortranState:
+        transformed_inputs = _predict(self.model.forward, state)
```
nbren12, line 33: Note to self. Need to change this implementation.

external/fv3fit/fv3fit/emulation/transforms/factories.py
```diff
@@ -142,6 +173,22 @@ def backward_names(self, requested_names: Set[str]) -> Set[str]:
             requested_names = factory.backward_names(requested_names)
         return requested_names
 
+    def backward_input_names(self) -> Set[str]:
```
frodre, line 176: Little worried that this is doubling down on an implementation that's in the "debt" column.

frodre (commented):
Thanks @nbren12.  

Overall, it's cool that you got this to work!  Looking forward to seeing how the prognostic run pans out.  The only thing I'm hesitant about is that the changes to the transforms adds a fair amount of complexity to the tensor transforms, which have been noted as an area to be remedied.

I think this PR allows us to overwrite a specific variable by inclusion in the requested model inputs, and a two-step model prediction (classifier -> regression).  It would be nice if we could overwrite any at any time vs. having to rely on correct placement in forward and backward calls.

With that said, I think if the prog-run works well we can merge this in.  If not, maybe it's worth leaving the fv3fit code we've used for the simple classifier work, add relevant emulation updates, and holding onto this branch.

external/fv3fit/fv3fit/emulation/transforms/factories.py
nbren12: Hmm. Oli had to work around the inability to specify inputs for the backward transforms too, so I think it is a flexibility that we need as long as we want to allow transforms to advertise their own inputs/outputs. For the typical transform, these methods are simpler than the "backwards_names" and simply list ins/outs.

Thinking ahead, I think the path forward to pay down the debt is to separate the backwards and forwards transforms into separate objects with this signature `transform.{__call__/output_names/input_names}`. This current design would be easy to refactor to that setup simply by deleting the "backward_".

external/fv3fit/fv3fit/emulation/zhao_carr/loss.py
```diff
@@ -0,0 +1,66 @@
...
+            loss_variables = [
+                "temperature_gscond_difference_tscaled",
+                "humidity_gscond_difference_tscaled",
+            ]
+            for v in loss_variables:
+                truth = x[v]
+                pred = y[v]
+
+                mask_casted = tf.cast(mask, truth.dtype)
+                loss_value = tf.keras.losses.MSE(
```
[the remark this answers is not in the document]
nbren12: I initially tried `sum((x-y)**2 * mask )/ sum(mask)` which I think is equivalent to what you are suggesting. Which is equivalent to making some batches more important than others. Sometimes `sum(mask)` is zero though and I kept running into NaNs. Ultimately, I felt it better to weight each `(sample,z)` grid point equally even if they are in different batches. 

I checked that the loss was O(1) on initial tuning/testing.

external/fv3fit/fv3fit/emulation/transforms/zhao_carr.py
```diff
@@ -80,6 +96,75 @@ def forward(self, x: TensorDict) -> TensorDict:
...
+    cloud_after = cloud_before + condensation
+
+    # apply no change case
+    cloud_out = tf.where(zero_tendency, cloud_before, cloud_after)
+    t_out = tf.where(zero_tendency, t_before, t_after)
+    q_out = tf.where(zero_tendency, qv_before, qv_after)
+
+    # apply zero cloud case
+    # TODO fix these constants
+    latent_heat = 2.51e6
```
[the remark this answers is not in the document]
nbren12: we do have some constant there already. I guess I preferred copy-paste over the explicit coupling in this instance since it's an approximate value not accounting for ice anyway. I kind of want to gather all things with "zhao carr" in the name into a common directory structure.

nbren12:
> With that said, I think if the prog-run works well we can merge this in. If not, maybe it's worth leaving the fv3fit code we've used for the simple classifier work, add relevant emulation updates, and holding onto this branch

I guess I don't see this as a big step change in terms of complexity and would like to merge if possible since leaving things on branches doesn't really work. I added an ability to mask the loss for this specific problem and some input/output metadata to the backwards transforms (a missing functionality which Oli was actively hacking around). I also paid down some technical debt with the MicrophysicsHook by removing the I/O from its init.

I would appreciate some concrete thoughts on how to make it simpler though... I went through 2 phases of design...and this is MUCH simpler than my first attempt.

nbren12:
> It would be nice if we could overwrite any at any time vs. having to rely on correct placement in forward and backward calls.

I considered breaking it apart more, but felt this would lead to excessive coupling between the prog run and ml code. If there are n transforms, and 1 model, then there 2n places the state could be inserted. I thought it simpler if the prog run interprets the saved ml regressor as a single black box (which is why this code worked by only changing the order of the calls to the classifier/regressor).

frodre:
Thanks for your thoughts on this.  I agree it's not a huge step change in complexity, and you mentioning that this fixes something that Oli has had to work around recently makes this more important to have in the interim.  My hesitancy in the increased management of forward-backward requirement harvesting is that it was already hard to debug the key error if/when I mess up the long list of tensor transforms.  It doesn't happen often, but I remember it being a painful debug last time around.  I guess we could have some introspection when key errors occur that would point out the source of the variable ask.

>If there are n transforms, and 1 model, then there 2n places the state could be inserted. I thought it simpler if the prog run interprets the saved ml regressor as a single black box (which is why this code worked by only changing the order of the calls to the classifier/regressor).

Not sure I follow.  Couldn't the model just be applied and serialized as a transform?  I guess there would need to be a dummy/pass option or separate definitions for applying transforms for target data creation.  I agree that the prognostic run shouldn't know anything about the transforms happening under the hood.

Overall, I am okay merging.  I just wanted to have a conversation about my concerns here so thanks for indulging.  Medium term,  splitting the forward/backward into two separately timed "forward" calls makes sense and lowers the cognitive load of transform placement.  I think long-term, lowering the configuration overhead for defining transforms in YAML would be great, which I think we've discussed previously but who knows when we'll have the ambition for, haha.

external/fv3fit/fv3fit/emulation/transforms/factories.py
frodre: Yeah, I think I agree with this path forward if I understand the change proposed.  Essentially, we have two separate transform groups.  One before model, one after model, and there's only a single "direction".  This would've solved my problem from before the v4 dataset switch as well.   Thanks for the extra context too, that Oli has had issues with this too.

frodre (approved):
Approved.  My only request for this PR is the GSCondRoute naming consistency.

external/fv3fit/fv3fit/emulation/transforms/zhao_carr.py
```diff
@@ -80,6 +96,75 @@ def forward(self, x: TensorDict) -> TensorDict:
...
+    q_out = tf.where(zero_cloud, qv_before + cloud_before, q_out)
+    t_out = tf.where(
+        zero_cloud, t_before - cloud_before * latent_heat / specific_heat, t_out
+    )
+
+    return cloud_out, t_out, q_out
+
+
+@dataclasses.dataclass
+class GSCondRoute(TensorTransform):
```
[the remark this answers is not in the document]
nbren12: done

PR #1918, opened 2022-07-07
Precpd and gscond classifier for full ZC emulation

frodre:
This PR adds generalizes the name GscondClassesV1  to MicorphysicsClassesV1 since we're now testing gscond or precpd and adds a few tests.  

- test_unit: [83%](https://output.circle-artifacts.com/output/job/86f603e4-5939-48fb-a68d-4a2edba9b1bd/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/tests/emulation/test_zhao_carr.py
```diff
@@ -16,3 +22,50 @@ def test_classify():
...
+
+    state = _get_cloud_state()
+
+    transform = classify_class()
+    transform = transform.build(state)
+    assert transform
+
+
+def test_classify_classes_v1():
+    state = _get_cloud_state()
```
frodre, line 47: I didn't know sets were mutable, so I got a few fun errors related to that when using "CLASS_NAMES" as the requested variables.

nbren12: https://docs.python.org/3.8/library/stdtypes.html?highlight=frozenset#frozenset

projects/microphysics/argo/end_to_end.py
```diff
@@ -121,25 +121,21 @@ def _label_args(labels):
 
 
 def set_prognostic_emulation_model(
```
nbren12, line 123: Over the past couple weeks, I've noticed that this function adds a layer of indirection that made it harder for me to understand the configuration. It appears that you needed to make some changes for this PR, suggesting the abstraction is not very robust.

What do you think about delete this and move the appropriate settings into a yaml file? I think copy/pasting yamls will be clearer in this case.

nbren12 (approved):
Thanks. Some suggestions below.

My main priority would be to add a yaml containing the ML settings for your best prognostic run. Currently these settings are difficult to infer because of the `set_prognostic_emulation_model` function.

external/emulation/emulation/zhao_carr.py
```diff
@@ -104,7 +104,7 @@ def mask_where_fortran_cloud_identical(state, emulator):
     return _update_with_net_condensation(cloud_out, state, emulator)
 
 
-def _get_classify_output(emulator, one_hot_axis=0):
+def _get_classify_output(emulator, one_hot_axis=0, class_key="gscond_classes"):
```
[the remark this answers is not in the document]
frodre: I think this is the right way.  Note that it did require a specification in `ModelWithClassifier` since the usage of "gscond_classes" was implicit with the function call.

PR #1929, opened 2022-07-12
OCSVM novelty detector

chsanford:
Creates a one-class SVM novelty detector, which was widely explored in an offline manner so far. This follows the blueprint created in [this PR](https://github.com/ai2cm/fv3net/pull/1908) and passes the same tests.

Significant internal changes:
- Added OCSVMNoveltyDetector, a Predictor that identifies out-of-sample points with sklearn's OneClassSVM. Implements the Novelty Detector Class.

Coverage reports (updated automatically):
- test_unit: [83%](https://output.circle-artifacts.com/output/job/14622f37-542e-4618-9a40-3d414c3a975d/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

PR #1930, opened 2022-07-12
Out-of-sample Predictor class for masking a predictor with a novelty detector

chsanford:
Creates a new `OutOfSamplePredictor` class, which combines a `Predictor` and a `NoveltyDetector` to a hybrid predictor that predicts 0 for a sample if deemed a novelty. This approach can use the new `MinMaxNoveltyDetector` and `OCSVMNoveltyDetector` from [these](https://github.com/ai2cm/fv3net/pull/1908) [two](https://github.com/ai2cm/fv3net/pull/1929) PRs.

Significant internal changes:
- New class `OutOfSamplePredictor` that implements `Predictor` and is defined analogously to `EnsemblePredictor`.

- [ ] New suite of tests to verify that the output is correct (unchanged if not novelty,  zero if novelty) on synthetic data and that `dump` and `load` functionality works properly.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -8,6 +8,8 @@ import numpy as np
 import xarray as xr
 import vcm
 
+from fv3fit._shared.novelty_detector import NoveltyDetector
+
 from . import io
 from .predictor import Predictor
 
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -238,3 +240,76 @@ def load(cls, path: str) -> "TransformedPredictor":
...
+        masked_predict = base_predict.copy()
+        for output_variable in self.output_variables:
+            masked_predict[output_variable] = xr.where(
+                novelties[NoveltyDetector._NOVELTY_OUTPUT_VAR] == 0,
+                base_predict[output_variable],
+                0,
+            )
+        return masked_predict
+
+    def dump(self, path: str):
```
AnnaKwa, line 290: For models that have to be manually constructed like this and the `EnsembleModel` I can't think of a use case where the dump method is needed- unless there's one you can think of, I'd suggest raising a `NotImplementedError` that points to the documentation for how to write the manual configuration, e.g. https://github.com/ai2cm/fv3net/blob/9b2b61d304d6494c3889578a211ed9da5d5d89b1/external/fv3fit/fv3fit/_shared/models.py#L168

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -238,3 +240,76 @@ def load(cls, path: str) -> "TransformedPredictor":
...
+        io.dump(self.novelty_detector, novelty_detector_path)
+        with fsspec.open(os.path.join(path, self._CONFIG_FILENAME), "w") as f:
+            yaml.safe_dump(options, f)
+
+    @classmethod
+    def load(cls, path: str) -> "OutOfSampleModel":
+        with fsspec.open(os.path.join(path, cls._CONFIG_FILENAME), "r") as f:
+            config = yaml.safe_load(f)
+
+        base_model = io.load(os.path.join(path, cls._BASE_MODEL_SUBDIR))
```
AnnaKwa, line 308: You can remove the `_BASE_MODEL_SUBDIR, _NOVELTY_DETECTOR_SUBDIR` and instead have fields for the component model paths in the config file, i.e. your config would look like
```
cutoff: 0
novelty_detector: gs://vcm-ml-scratch/claytons/novelty_detector
base_model: gs://vcm-ml-scratch/claytons/base_model
```
and then this `load` method would call `io.load` on those paths from the config. Then you could cut out the step of having to copy the component models into the novelty detector parent directory.

AnnaKwa (changes requested):
Nice! I think you could further simplify loading, see comment. Also, could you add a short section to the docs [like this one](http://vulcanclimatemodeling.com/docs/fv3fit/ensembles.html) explaining how to manually construct an `OutOfSampleModel`?

AnnaKwa (approved):
LGTM! I have a couple remaining requests regarding documentation but don't need to review those- just make sure the docs build ok before merging.

PR #1931, opened 2022-07-12
Offline diags: use proper zonal mean in zonal mean variance

AnnaKwa:
The offline diags had a bug where the mean being used in the zonally averaged variance calculation was the global mean, not the zonal avg mean. This inflates zonal average variance, which in turn inflates R^2. dQ2 at the poles is particularly affected due to its range across multiple orders of magnitude.

ex. before and after bug fix
image
image

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/a73fede4-9b98-4bbb-9885-cf1f691f4587\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
@@ -372,7 +372,8 @@ for mask_type in ["global", "sea", "land"]:
         zonal_mean_interp = xr.Dataset()
         for var in zonal_mean:
             da = zonal_mean[var]
-            interp_func = interp1d(da.lat, da.transpose(..., "lat"))
+            # Ignore extrapolation errors for test case with randomized lat data
+            interp_func = interp1d(da.lat, da.transpose(..., "lat"), bounds_error=False)
             zonal_mean_interp[var] = xr.DataArray(
                 interp_func(grid.lat.transpose(*HORIZONTAL_DIMS)),
                 dims=("pressure", *HORIZONTAL_DIMS),
--- a/workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
@@ -10,6 +10,7 @@ from fv3net.diagnostics._shared.constants import (
 import logging
 import numpy as np
 import pandas as pd
+from scipy.interpolate import interp1d
 from typing import Sequence, Tuple, Dict
 import xarray as xr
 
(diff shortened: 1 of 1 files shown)


nbren12:
Exactly why I dislike R^2...funny how often this comes up. 

Would you be open to generalizing this function: https://github.com/ai2cm/fv3net/blob/9b2b61d304d6494c3889578a211ed9da5d5d89b1/external/vcm/vcm/calc/metrics.py#L18? 

@YakelynRJ and I wrote this r2_score function that works with arbitrary averaging operators: https://github.com/ai2cm/fv3net/blob/9b2b61d304d6494c3889578a211ed9da5d5d89b1/projects/cloud_ml/scripts/scoring.py#L12. It has a similar signature to the other scoring functions in `vcm.calc.metrics`.

workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
```diff
@@ -364,11 +365,22 @@ def variance_zonal_avg_3d(diag_arg, mask_type=mask_type):
...
+        # Fill cubedsphere with zonal mean, interpolating between integer latitudes
+        zonal_mean_interp = xr.Dataset()
+        for var in zonal_mean:
+            da = zonal_mean[var]
+            # Ignore extrapolation errors for test case with randomized lat data
+            interp_func = interp1d(da.lat, da.transpose(..., "lat"), bounds_error=False)
+            zonal_mean_interp[var] = xr.DataArray(
+                interp_func(grid.lat.transpose(*HORIZONTAL_DIMS)),
+                dims=("pressure", *HORIZONTAL_DIMS),
+            )
```
spencerkclark, line 380: You should be able to do this without interpolating if you compute the variance as the mean of the squares minus the square of the means.

AnnaKwa: Thanks good point. I'll just use the cloud_ml function above.

AnnaKwa:
> Exactly why I dislike R^2...funny how often this comes up.
> 
> Would you be open to generalizing this function:
> 
> https://github.com/ai2cm/fv3net/blob/9b2b61d304d6494c3889578a211ed9da5d5d89b1/external/vcm/vcm/calc/metrics.py#L18
> 
> ?
> @YakelynRJ and I wrote this r2_score function that works with arbitrary averaging operators:
> 
> https://github.com/ai2cm/fv3net/blob/9b2b61d304d6494c3889578a211ed9da5d5d89b1/projects/cloud_ml/scripts/scoring.py#L12
> 
> . It has a similar signature to the other scoring functions in `vcm.calc.metrics`.

Since I would want to replace the R2 calculation in the offline diags with this function being added to the computation registry, I'll just do that separately and limit this PR to just the bug fix for variance.

spencerkclark (approved):
Thanks @AnnaKwa; this looks good to me.

PR #1932, opened 2022-07-12
Add radiation physics standalone

nbren12:
Add DSL's python port of the radiation scheme. I deleted some large files, which are not needed to follow the quickstart instructions. I also added a `shell.nix` file that can be used to bootstrap a development environment.

One issue is that this code is currently LGPL (inherited from gt4py). This is incompatible with fv3net's broader license. We will need to remove gt4py using code if we want to use this from MIT licensed subprojects of fv3net.

cc @elynnwu @YakelynRJ @brianhenn 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/6d432004-8a91-46b7-a1b9-eabbfc5394b1\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


elynnwu:
I believe you should be able to skip over all the files with `*_gt4py`.

brianhenn (approved):
@nbren12 thanks!

PR #1937, opened 2022-07-15
radiation: improve performance by extracting I/O etc

YakelynRJ:
Improved performance by:

1. extracting i/o
2. removing unnecessary loops and vectorized them

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/06a8cd9f-7fe7-4ddb-bc41-5d683f56deff\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/radiation/python/radlw/radlw_main.py
+++ b/external/radiation/python/radlw/radlw_main.py
@@ -5,6 +5,7 @@ import sys
 import os
 import time
 import warnings
+from numba import jit
 
 sys.path.insert(0, "..")
 from radphysparam import (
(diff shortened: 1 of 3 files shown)


YakelynRJ:
Please review the last commit: extract data loading to lwrad(). All data loading occurs inside lwrad()

nbren12 (approved):
LGTM! Thanks. I manually ran `test_driver.py` to test this myself and the code changes seemed ok, but given the number of LOC I couldn't verify it 100%. 

For the next PR, I would focus on refactoring the redundant computation of the lookup table from https://github.com/ai2cm/fv3net/blob/4684e655611cbfa7d405b6851227ab55a31de345/external/radiation/python/funcphys.py#L48. This was responsible for a large percent of the remaining cost. There are a couple of paths here...you could either compute the lookup table once on initialization and pass it into fvps....or you could compute just the two neighboring lookup table points needed for the linear interpolation.

Just as a reminder, in the future, I would aim for one commit per function/refactor and make sure not to commit any commented code :)

PR #1945, opened 2022-07-20
Draft PR of tapering design

chsanford:
- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/91a6cf87-da7a-4aad-adfa-d0e78f412ba8\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -9,8 +9,7 @@ import xarray as xr
 import vcm
 
 from fv3fit._shared.novelty_detector import NoveltyDetector
-from fv3fit._shared import taper_function
-from fv3fit._shared.taper_function import get_taper_function
+from fv3fit._shared.taper_function import get_taper_function, taper_mask
 
 from . import io
 from .predictor import Predictor
@@ -278,7 +277,7 @@ class OutOfSampleModel(Predictor):
         self.novelty_detector = novelty_detector
         self.cutoff = cutoff
         self.taper = taper or get_taper_function(
-            taper_function._MASK_NAME, {"cutoff": cutoff}
+            taper_mask.__name__, {"cutoff": cutoff}
         )
 
         base_inputs = set(base_model.input_variables)
@@ -324,24 +323,21 @@ class OutOfSampleModel(Predictor):
 
         base_model = io.load(config["base_model_path"])
         novelty_detector = io.load(config["novelty_detector_path"])
-        cutoff = 0
-        if "cutoff" in config:
-            cutoff = config["cutoff"]
+        cutoff = config.get("cutoff", 0)
 
         assert isinstance(novelty_detector, NoveltyDetector)
 
         default_tapering_config = {
-            "name": taper_function._MASK_NAME,
+            "name": taper_mask.__name__,
             "cutoff": cutoff,
             "ramp_min": cutoff,
             "ramp_max": 1 if cutoff == 0 else max(cutoff * 2, cutoff / 2),
             "threshold": cutoff,
-            "rate": 0.5,
         }
-        if "tapering_function" in config:
-            tapering_config = {**default_tapering_config, **config["tapering_function"]}
-        else:
-            tapering_config = default_tapering_config
+        tapering_config = {
+            **default_tapering_config,
+            **config.get("tapering_function", {}),
+        }
 
         taper = get_taper_function(tapering_config["name"], tapering_config)
 
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -1,5 +1,5 @@
 import dataclasses
-from typing import Iterable, Optional, Set, Hashable, Sequence
+from typing import Callable, Iterable, Optional, Set, Hashable, Sequence
 import dacite
 import fsspec
 import yaml
@@ -10,7 +10,7 @@ import vcm
 
 from fv3fit._shared.novelty_detector import NoveltyDetector
 from fv3fit._shared import taper_function
-from fv3fit._shared.taper_function import MaskTaperFunction, TaperFunction
+from fv3fit._shared.taper_function import get_taper_function
 
 from . import io
 from .predictor import Predictor
@@ -255,8 +255,8 @@ class OutOfSampleModel(Predictor):
         self,
         base_model: Predictor,
         novelty_detector: NoveltyDetector,
-        cutoff: Optional[float] = None,
-        taper: Optional[TaperFunction] = None,
+        cutoff: float = 0,
+        taper: Optional[Callable[[xr.DataArray], xr.DataArray]] = None,
     ):
         """
         Args:
@@ -276,8 +276,10 @@ class OutOfSampleModel(Predictor):
         """
         self.base_model = base_model
         self.novelty_detector = novelty_detector
-        self.cutoff = cutoff or novelty_detector._get_default_cutoff()
-        self.taper = taper or MaskTaperFunction(self.cutoff)
+        self.cutoff = cutoff
+        self.taper = taper or get_taper_function(
+            taper_function._MASK_NAME, {"cutoff": cutoff}
+        )
 
         base_inputs = set(base_model.input_variables)
         base_outputs = set(base_model.output_variables)
@@ -292,11 +294,11 @@ class OutOfSampleModel(Predictor):
 
     def predict(self, X: xr.Dataset) -> xr.Dataset:
         base_predict = self.base_model.predict(X)
-        novelties = self.novelty_detector.predict_novelties(X, cutoff=self.cutoff)
-        taper_values = self.taper.get_taper_value(
-            novelties[NoveltyDetector._SCORE_OUTPUT_VAR]
+        centered_scores, diagnostics = self.novelty_detector.predict_novelties(
+            X, cutoff=self.cutoff
         )
-        novelties[self._TAPER_VALUES_OUTPUT_VAR] = taper_values
+        taper_values = self.taper(centered_scores)
+        diagnostics[self._TAPER_VALUES_OUTPUT_VAR] = taper_values
 
         tapered_predict = xr.Dataset(
             coords=base_predict.coords, attrs=base_predict.attrs
@@ -306,7 +308,7 @@ class OutOfSampleModel(Predictor):
                 base_predict[output_variable] * taper_values
             )
 
-        return xr.merge([tapered_predict, novelties])
+        return xr.merge([tapered_predict, diagnostics])
 
     def dump(self, path):
         raise NotImplementedError(
@@ -322,19 +324,26 @@ class OutOfSampleModel(Predictor):
 
         base_model = io.load(config["base_model_path"])
         novelty_detector = io.load(config["novelty_detector_path"])
+        cutoff = 0
         if "cutoff" in config:
             cutoff = config["cutoff"]
-        else:
-            cutoff = None
 
         assert isinstance(novelty_detector, NoveltyDetector)
 
-        taper = None
+        default_tapering_config = {
+            "name": taper_function._MASK_NAME,
+            "cutoff": cutoff,
+            "ramp_min": cutoff,
+            "ramp_max": 1 if cutoff == 0 else max(cutoff * 2, cutoff / 2),
+            "threshold": cutoff,
+            "rate": 0.5,
+        }
         if "tapering_function" in config:
-            tapering_config = config["tapering_function"]
-            taper = getattr(taper_function, tapering_config["name"]).load(
-                tapering_config, novelty_detector
-            )
+            tapering_config = {**default_tapering_config, **config["tapering_function"]}
+        else:
+            tapering_config = default_tapering_config
+
+        taper = get_taper_function(tapering_config["name"], tapering_config)
 
         model = cls(base_model, novelty_detector, cutoff=cutoff, taper=taper)
         return model
--- a/external/fv3fit/fv3fit/_shared/novelty_detector.py
+++ b/external/fv3fit/fv3fit/_shared/novelty_detector.py
@@ -1,5 +1,5 @@
 import abc
-from typing import Hashable, Iterable, Optional
+from typing import Hashable, Iterable, Tuple
 from fv3fit._shared.predictor import Predictor
 import xarray as xr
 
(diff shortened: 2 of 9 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -281,7 +284,9 @@ def __init__(
 
     def predict(self, X: xr.Dataset) -> xr.Dataset:
         base_predict = self.base_model.predict(X)
-        novelties = self.novelty_detector.predict_novelties(X, self.cutoff)
+        novelties = self.novelty_detector.predict_novelties(
+            X, cutoff=self.cutoff, taper_function=self.taper_function
+        )
```
AnnaKwa, line 289: (see comment in NoveltyDetector about moving the taper function call out to here)
I think you should keep the previous behavior of having `self.novelty_detector.predict_novelties(X, self.cutoff)` just return information about novelty score and label, and have a separate call after that to 
```
taper_values = self.taper_function(novelties[whatever_the_score_var_is])
```
You can merge that into the final dataset at the end.

The one part that looks like it is NoveltyDetector dependent is the default cutoff- I think you could leave the `get_default_cutoff` method on the NoveltyDetectors, the usage for setting up the default taper functions in their .load looks fine.

AnnaKwa (commented):
I didn't go into the details to review the calculations of specific functions, but looked at the overall outline. I like the setup and instantiation of the different TaperFunctions, that looks good. My main request is to separate the concerns of scoring the novelties from the concern of altering the output tendencies. LMK if you want to talk more about the comments there or they don't make sense.

external/fv3fit/fv3fit/_shared/taper_function.py
```diff
@@ -0,0 +1,106 @@
+import abc
+import numpy as np
+import xarray as xr
+
+from fv3fit._shared.novelty_detector import NoveltyDetector
+
+
+class TaperFunction(abc.ABC):
+    @abc.abstractmethod
+    def get_taper_value(self, novelty_score: xr.DataArray) -> xr.DataArray:
```
nbren12, line 10: This is an interface with a single method. Often this implies that it should just be a function instead. Would it be possible to implement all the TaperFunctions as actual functions? They look nice and tidy that way. For example:

```
def taper_mask(novelty_score, cutoff):
    return xr.where(novelty_score > cutoff, 0, 1)

def taper_ramped(novelty_score, max, min):
    unclipped = (max - novelty_score) / (max - min)
    return np.clip(unclipped, 0, 1)
```


Then you can put all the instantiation logic into a single factory function:

```
def get_taper_function(config: dict) -> Callable[[xr.DataArray], xr.DataArray]:
    if "cutoff" in config:
        return lambda x: taper_mask(x, **config)
    elif "ramp_min" in config:
        return lambda x: taper_ramped(x, **config)
    ...
```

The advantage of this is that there is less coupling between the argument binding/config and the implementations themselves. Also, you don't need to worry about the name `taper_func.get_taper_value(x)` wherever you use it...which is just more typing and reasons to change compared to `taper_func(x)` IMO.

external/fv3fit/fv3fit/_shared/taper_function.py
```diff
@@ -0,0 +1,106 @@
+import abc
+import numpy as np
+import xarray as xr
+
+from fv3fit._shared.novelty_detector import NoveltyDetector
```
nbren12, line 5: Is this a circular import? It seems like `NoveltyDetector` is the parent object here...so the `.load` methods below probably shouldn't need access to it. It seems like they only use it to get default values. Maybe define the defaults in this module instead?

external/fv3fit/fv3fit/_shared/taper_function.py
chsanford: Thank you, this is a great suggestion! Working on it now.

external/fv3fit/fv3fit/_shared/taper_function.py
chsanford: I don't think this is a circular import, since the parent object is `OutOfSampleModel` (which also contains a `NoveltyDetector` object). If possible, I think I'd prefer to keep the default logic in the individual implementations of `NoveltyDetector`, since the default cutoffs are computed in very different ways. (e.g. minmax is always zero, OCSVM is a negative number determined by the maximum score of all training samples, and the ensemble approach (if implemented) will be a positive variance, likely also the maximum observed scores.). Since it isn't a circular import, is it fine if I have something like this, which populates a config file based on the default values and passes it to the other taper function? 

```
def get_taper_function(name: str = _MASK_NAME, config: dict = {}) -> Callable[[xr.DataArray], xr.DataArray]:
    if name == _MASK_NAME:
        return lambda x: taper_mask(x, **config)
    elif name == _RAMP_NAME:
        return lambda x: taper_ramp(x, **config)
    elif name == _DECAY_NAME:
        return lambda x: taper_decay(x, **config)
    else:
        raise ValueError("Incorect")


def get_taper_function_from_novelty_detector(novelty_detector: NoveltyDetector, name: str = _MASK_NAME) -> Callable[[xr.DataArray], xr.DataArray]:
    cutoff = novelty_detector._get_default_cutoff()
    config = {
        "cutoff": cutoff,
        "ramp_min": cutoff,
        "threshold": cutoff
    }
    if cutoff > 0:
        config["ramp_max"] = 2 * cutoff
    elif cutoff < 0:
        config["ramp_max"] = cutoff / 2
    else:
        config["ramp_max"] = 1
    return get_taper_function(name, config)
```

chsanford: Actually, nevermind, it might be easiest to do this in `OutOfSampleModel.load`. I'll mock something up in a commit in a bit.

nbren12: >  I think I'd prefer to keep the default logic in the individual implementations of NoveltyDetector, since the default cutoffs are computed in very different ways. (e.g. minmax is always zero, OCSVM is a negative number determined by the maximum score of all training samples, and the ensemble approach (if implemented) will be a positive variance, likely also the maximum observed scores.)

Interesting. It sounds like `NoveltyDetector` basically owns the concept of `TaperFunction` and could potentially hide this concept from other objects... in particular the `OutOfSampleModel`. Currently  `NoveltyDetector` and `OutOfSampleModel` and `TaperFunction` all know about tapering...which seems like a poor separation of concerns.

There are a couple of options to reduce this coupling:
1. Remove the concept of `OutOfSampleModel` and simply call the appropriate tapering logic in your `NoveltyDetector` implementations. 
2. Make `NoveltyDetector.predict_novelties` return a score  `alpha` from 0 to 1 that reflects the amount of tapering you want to apply. Then `OutOfSampleModel.predict` can use the score like this:
```
        base_predict = self.base_model.predict(X)
        # modify .precit_novelties to return a single dataarray of scores
        # score = 1 is OOS, score=0 is not.
        score = self.novelty_detector.predict_novelties(X, self.cutoff)
        masked_predict = base_predict.copy()
        for output_variable in self.base_model.output_variables:
            masked_predict[output_variable] =  (1-score) * base_predict[output_variable]
        return xr.merge([masked_predict, novelties])
```

nbren12: The second option is most direct generalization of what the code on master currently does IMO.

chsanford: I [talked](https://github.com/ai2cm/fv3net/pull/1945#discussion_r925891537) with @AnnaKwa and @oliverwm1 about using this kind of approach (and that's actually what was originally implemented in the first commit of the PR), but we were primarily thinking that tapering is something specific to the `OutOfSampleModel` rather than something intrinsic to a `NoveltyDetector`, since novelty detection alone isn't directly linked to the online predictions unless we specify that by creating an `OutOfSampleModel`. I'm ok with either way

chsanford: So one potential option would be to shift the score to make the default cutoff for each be zero, but I'm slightly hesitant to do that rn since each score is meaningful in the context of the novelty detector that produces it; even if the score is always negative for OCSVM and always non-negative for min-max, both are meaningful quantities that I would want to look at directly, given the context of the model. 

Also, there are already a bunch of saved runs with recorded novelty_scores as they're currently defined. To avoid inconsistency, I'd rather add a new field like `centered_score` than replace the old one.

nbren12: > since each score is meaningful in the context of the novelty detector that produces it

This makes sense. However, at the same time, we should hide this kind of detail from other components whenever possible to avoid coupling. I like the idea of adding a centered_score output. 

Perhaps you can modify `NoveltyDetector.predict_novelties` to return a tuple of (centered_score, diagnostics). centered_score will be a dataarray which means you can remove the reference to `SCORE_OUTPUT_VAR` (one less point of coupling). `diagnostics` is datasets/dict of dataarrays that can simply be merged into the output without inspecting its contents at all.

AnnaKwa: > NoveltyDetector basically owns the concept of TaperFunction

I was kind of confused by this statement as I think the most recent iteration of the PR moved the tapering out of the NoveltyDetector. Although it is true that there's still some coupling between the tapering and novelty detectors via each detector's get_default_cutoff method. The suggestion to return the tuple of centered_score and diagnostics sounds like a good way to avoid that.

chsanford: This all sounds good! I'll add the centered score and have a new commit shortly.

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -305,12 +324,26 @@ def load(cls, path: str) -> "OutOfSampleModel":
 
         base_model = io.load(config["base_model_path"])
         novelty_detector = io.load(config["novelty_detector_path"])
+        cutoff = 0
```
AnnaKwa, line 327:
```suggestion
        cutoff = config.get("cutoff", 0)
```
then can omit following if "cutoff" statement

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -305,12 +324,26 @@ def load(cls, path: str) -> "OutOfSampleModel":
...
         if "cutoff" in config:
             cutoff = config["cutoff"]
-        else:
-            cutoff = None
 
         assert isinstance(novelty_detector, NoveltyDetector)
 
-        model = cls(base_model, novelty_detector, cutoff)
+        default_tapering_config = {
```
AnnaKwa, line 333: Can these default just be handled in the definition of the individual tapering functions and removed from the OOS model?

AnnaKwa (changes requested):
Thanks for the edits Clayton, I think the overall structure looks good to me. I have some smaller requests. The main ones are to omit/move some of the new tests that combine a couple checks, and leave the default behavior to the tapering functions instead of the OOS model.

external/fv3fit/fv3fit/_shared/models.py
chsanford: That should work, as long as it's fine that `cutoff` gets passed to `get_taper_function`.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
```diff
@@ -113,6 +110,9 @@ def predict(self, data: xr.Dataset) -> xr.Dataset:
         score_dataset = stacked_scores.to_dataset(name=self._SCORE_OUTPUT_VAR).unstack(
             SAMPLE_DIM_NAME
         )
+        score_dataset[self._CENTERED_SCORE_OUTPUT_VAR] = score_dataset[
```
nbren12, line 113: Maybe a bit simpler to also return a tuple here and avoid the `_CENTERED_SCORE_OUTPUT_VAR`. This will also allow mypy to check that the data is being passed correctly between routines which will make this code more maintainable.

external/fv3fit/fv3fit/_shared/novelty_detector.py
```diff
@@ -19,24 +19,24 @@ class NoveltyDetector(Predictor, abc.ABC):
...
-        self, X: xr.Dataset, cutoff: Optional[float] = None
-    ) -> xr.Dataset:
-        if cutoff is None:
-            cutoff = self._get_default_cutoff()
+        self, X: xr.Dataset, cutoff: float = 0
+    ) -> Tuple[xr.DataArray, xr.Dataset]:
+        diagnostics = self.predict(X)
 
-        score_dataset = self.predict(X)
+        is_novelty = xr.where(
```
nbren12, line 37: Should this be moved to `OutOfSampleModel`?  It seems like`OutOfSampelModel` now owns the idea of cutoff and should keep it as private as possible. The cutoff is only used to compute the `is_novelty` output...which I think is only used for diagnostics. Since the `OutOfSampleModel` can compute this on its own, it's probably best to do so.

Or maybe just delete the code since it can be easily diagnosed from `centered_score` anyway.

tl;dr;. NoveltyDetector returns a centered score. OutOfSampleModel decides if it is a "novelty" or not based on it's own configured cutoff.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
chsanford: Would that break the type signature of the `Predictor` class's `predict` function? It seems to call for a single dataset having each element of `output_variables`. Is your proposal for a pair of data arrays, one for the regular scores and one for the centered score?

external/fv3fit/fv3fit/_shared/novelty_detector.py
AnnaKwa: That sounds good, I think it'd be cleaner to refactor and change the novelty detector's tests in a follow-up PR though and try keep this one mostly related to adding the taper function.

chsanford: Sgtm too. Once we do that, there probably isn't a reason for `predict_novelties` to exist as a function, and the only point of `NoveltyDetector` would be to specify the output variables of any `Predictor` implementing it.

external/fv3fit/fv3fit/sklearn/_min_max_novelty_detector.py
nbren12: Oh...i'm not proposing changing the Predictor API. I didn't see that the `NoveltyDetector` ABC is a subclass of Predictor too. I was confused since `OutOfSampleModel` is also a `Predictor`.

Anyway, feel free to ignore this comment...just feels like some layers/inheritance/methods could be removed as you suggest [here](https://github.com/ai2cm/fv3net/pull/1945#discussion_r927208280)...either in this or a future PR.

AnnaKwa (approved):

PR #1946, opened 2022-07-20
 improve performance of funcphys.py

YakelynRJ:
Changes in funcphys:
- fpvs was modified to take 2D arrays 

Changes in radiation_driver:
- The section that uses fpvs was vectorized. 

Computing time is reduced from 205 to 51 seconds

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/d24b81d2-975e-40fd-bceb-0e255d761040\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/radiation/python/funcphys.py
+++ b/external/radiation/python/funcphys.py
@@ -45,6 +45,7 @@ def fpvs(t):
     c2xpvs = 1.0 / xinc
     c1xpvs = 1.0 - xmin * c2xpvs
 
+    ## this line can be call only one time when initialized 
     tbpvs = np.array([ fpvsx(xmin + jx * xinc) for jx in range(nxpvs)])
     
     xj = np.minimum(np.maximum(c1xpvs + c2xpvs * t, 1.0), nxpvs)
--- a/external/radiation/python/funcphys.py
+++ b/external/radiation/python/funcphys.py
@@ -1,7 +1,6 @@
 import numpy as np
 from phys_const import *
 
-
 def fpvs(t):
     # $$$     Subprogram Documentation Block
     #
@@ -25,7 +24,8 @@ def fpvs(t):
     # Usage:   pvs=fpvs(t)
     #
     #   Input argument list:
-    #     t          Real(krealfp) temperature in Kelvin
+    #     t        2 dimensional array Real(krealfp) temperature in Kelvin
+    #              [npoints, nlevels]
     #
     #   Output argument list:
     #     fpvs       Real(krealfp) saturation vapor pressure in Pascals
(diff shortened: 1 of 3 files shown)


YakelynRJ:
Changes in funcphys:
Screen Shot 2022-07-20 at 6 47 27 AM


fpvs was modified to take 2D arrays
Changes in radiation_driver:

The section that uses fpvs was vectorized.
Computing time is reduced from 205 to 51 seconds

external/radiation/python/funcphys.py
```diff
@@ -45,18 +45,15 @@ def fpvs(t):
     c2xpvs = 1.0 / xinc
     c1xpvs = 1.0 - xmin * c2xpvs
 
-    for jx in range(nxpvs):
-        x = xmin + jx * xinc
-        tt = x
-        tbpvs[jx] = fpvsx(tt)
+    tbpvs = np.array([ fpvsx(xmin + jx * xinc) for jx in range(nxpvs)])
```
nbren12, line 48: I'm surprised you got such a large speedup since this code is still computing fpvsx many more times than it needs to. How is your current code different than before? Seems like you mostly replaced `np.maximum` with `max` and used a list comprehension rather than a for-loop, but it's not vectorized since these are still python scalars. 

This code is an attempt to speedup `fvpsx` with a lookup table. `fvpsx` has exponentials in it, so is slow. Precomputing some values and then interpolating between them is a common speedup technique. When andrew ported this code, it looks like he inlined the pre-computation in here to keep things simple...but really it should only be happening once while the model is initialized.  It computes a lookup table `tbpvs` with around 7000  points...in order to find the 2  points closest to the provided `t` and interpolate linearly between them. This is 7000x slower than just calling the `fpvsx` directly...so the code is rather silly.

Here is my suggested refactor:
1. Revert your changes to these lines
2. Move this for loop into a function `initialize_fpvsx_lookup_table` which returns `tbpvs`
3. Call `initialize_fpvsx_lookup_table` during the lw radiation initiialization process.
4. Pass the `tbpvs` down here.

Please let me know if you have any questions and we can discuss over VC.

nbren12 (changes requested):

external/radiation/python/funcphys.py
nbren12: we can defer this refactor

nbren12 (approved):
LGTM! Really nice speedup even if there is still some redundant computation.

It seems like the next PR should focus on @jitting radsw_main.py spcvrtm. It looks pretty long, so I think it will be challenging to vectorize easily.

PR #1948, opened 2022-07-20
Cloud ML: Rad port validated in CI

brianhenn:
Puts validation of the Python radiation port with the example dataset in CI. See example [output](https://app.circleci.com/pipelines/github/ai2cm/fv3net/12498/workflows/c46e4dba-40f3-4c8b-98e4-effbacee9b8a/jobs/112867/parallel-runs/0/steps/0-109).

- test_unit: [83%](https://output.circle-artifacts.com/output/job/1d37714c-3ab9-4193-a0ec-822f34ca25c4/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,6 +225,7 @@ jobs:
           name: "Get data"
 #                       nix-shell --command "gcloud auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS"
           command: |
+            . /home/circleci/.nix-profile/etc/profile.d/nix.sh
             nix-shell --command "cd external/radiation && ./get_data.sh"
       - run:
           name: "Validate radiation port"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,15 +215,17 @@ jobs:
           name: "gcloud auth"
           command: echo $ENCODED_GOOGLE_CREDENTIALS | base64 -d > $GOOGLE_APPLICATION_CREDENTIALS
       - run:
-          name: "Get data"
-          command: cd external/radiation && ./get_data.sh
-      - run:
           name: "Install nix and cachix"
           command: |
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -202,6 +203,34 @@ jobs:
...
+
+  validate_radiation_port:
+    machine:
+      image: ubuntu-2004:202111-02
+    environment:
+      GOOGLE_PROJECT_ID: vcm-ml
+      GOOGLE_APPLICATION_CREDENTIALS: /tmp/key.json
+    steps:
+      - checkout
+      - gcp-cli/install
```
nbren12, line 215: It's probably more elegant to manage all dependencies with nix. This will ensure the CI environment remains as close as possible to the dev environment.  

Currently, the shell.nix doesn't include gsutil, but it probably should. You can do this by adding 
```
pkgs.google-cloud-sdk
```
near the other deps in `shell.nix`.

99% of things we would want to install are available in nix. you can search for them here: https://search.nixos.org/packages?channel=22.05&show=google-cloud-sdk&from=0&size=50&sort=relevance&type=packages&query=gsutil

.circleci/config.yml
```diff
@@ -202,6 +203,34 @@ jobs:
...
+      - run:
+          name: "gcloud auth"
+          command: echo $ENCODED_GOOGLE_CREDENTIALS | base64 -d > $GOOGLE_APPLICATION_CREDENTIALS
+      - gcp-cli/initialize
+      - run:
+          name: "Get data"
+          command: cd external/radiation && ./get_data.sh
+      - run:
```
nbren12, line 223: I suggest adding cachix here.

nbren12 (approved):
Thanks for putting this together! This will be super helpful.

The purist in me feels it is a bit odd to install `gsutil` using an orb when we already are using the best package manager in the universe ;)

PR #1949, opened 2022-07-21
Jitting radlw

YakelynRJ:
This is a long PR that:
1. jitted cldprop
2. jitted rtrnmc
3. jitted rtrnmr 
4. removed dataarrays as inputs from taugb* functions and replaced them with np.arrays
5. moved taugb* functions to a different file to reduce radlw_main.py size 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/dec3d4d4-b07e-4f8d-891d-9f1302022917\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


nbren12:
I hadn't realized that radlw_main is nearly 250kb! This means that each commit editing it adds 250kb to the repo. It's good that you moved the taugb functions to a new file.

To avoid adding too many commits with the large file I think we should squash merge this one. Or you can rewrite your history to bundle the first 3 commits and the taumol commits into 2 commits total if you want to keep the history distinct.

nbren12 (approved):
LGTM! I'm rubber stamping since this is mostly moving functions around and we are relying on the test to cover the rest of it.

PR #1950, opened 2022-07-22
Add vertically integrated tracer diagnostics

nbren12:
This will be computed every timestep and printed to the console with the other
diagnostics.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/f29fd349-70cd-41e3-af99-0fe8731ae7f8\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


nbren12:
I'm not sure if it makes more sense to put this at the end or beginning of the timestep.

brianhenn (approved):
LGTM.  Up to you about when this gets called; I could see putting it at the end so it represents the state after physics instead of the state after the previous time's physics.

PR #1963, opened 2022-07-26
Emulation conservative cloud limiter

nbren12:
The cloud limiter used by some of gscond models does not conserve heat or water.

Implement a conservative cloud limiter and general framework for using tensor transforms in prognostic runs.

Experiment in progress: https://wandb.ai/ai2cm/microphysics-emulation/groups/precpd-diff-only-rnn-combined-cloudlimit-v2-online/workspace?workspace=user-nbrenowitz

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/8c3f5c9b-9291-4eba-80db-329be76eeb61\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/fv3fit/fv3fit/emulation/transforms/zhao_carr.py
```diff
@@ -46,6 +50,61 @@
 QV_PRECPD = "specific_humidity_after_precpd"
 
 
+def limit_negative_cloud(
+    *, cloud: tf.Tensor, humidity: tf.Tensor, temperature: tf.Tensor,
```
frodre, line 54: What is the extra args inclusion, `*`, for?

frodre (approved):
Looks good, @nbren12 .  Thanks for the `train_microphysics` cleanup and the warning suppression.  

My only comment is that the use of `_predict` for transposes in our emulation model plumbing is getting convoluted.  It'd be nice if that could be consolidated or simplified.

external/fv3fit/fv3fit/emulation/transforms/zhao_carr.py
nbren12: It suppresses the use of positional arguments.

external/emulation/emulation/models.py
```diff
@@ -32,6 +32,18 @@ def __call__(self, state: FortranState):
         return model_outputs
 
 
+def transform_model(
+    model: Callable[[FortranState], FortranState], transform: Any
+) -> Callable[[FortranState], FortranState]:
+    def combined(x: FortranState) -> FortranState:
+        x_transformed = _predict(transform.forward, x)
+        x_transformed.update(model(x_transformed))
```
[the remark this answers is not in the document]
nbren12: Yah. the type hints might clarify things a bit. I agree we could pull the transpose operation to a higher level so that we can work in the more predominant convention.

PR #1965, opened 2022-07-28
Allow regularization for emulation dense model

oliverwm1:
Useful to be able to configure default regularization strategies (e.g. L2) for the emulation dense model, which I am currently using for the fine-only experiments.

Refactored public API:
- `ArchitectureConfig` now takes an optional `kernel_regularizer` parameter, which is used for the hidden layers and output layer for the dense model. Other architecture types cannot use the regularizer currently.

Significant internal changes:
- To enable this, some refactoring was necessary to avoid a circular import. Specifically a new `fv3fit._shared.training_config` module was created to contain code related to registering training functions.

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/79a7a51a-7026-4e8a-8bab-8a6a848aa6b4/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/emulation/layers/architecture.py
```diff
@@ -295,17 +308,31 @@ def NoWeightSharingSLP(
 class StandardOutput(tf.keras.layers.Layer):
     """Uses densely-connected layers w/ linear activation"""
 
-    def __init__(self, feature_lengths: Mapping[str, int], *args, **kwargs):
+    def __init__(
+        self,
+        feature_lengths: Mapping[str, int],
+        kernel_regularizer: Optional[Mapping[str, Any]] = None,
```
nbren12, line 314: `kernel_regularizer` or `RegularizerConfig` are two new coupling points that could give this class a reason to change. The `dacite.from_dict`call  in  particular is new concern that is liable to change with the configuration frontend. 

A couple of options:
1. pass in the instantiated `self._kernel_regularizer.instance` in directly (this at least addresses the concern of coupling to the front-end configs).
2. Pass in a `dense_layer_factory` as a keyword argument with a default of `tf.keras.layers.Dense`. Build the dense_layer_factory after the configs are built using the kernel regularizer. You can use `dense_layer_factory` above too.

2 is the most testable design and completely removes "regularization" from the list of things these classes need to worry about and it reduces repetition and potentially inconsistent handling of the configs.

nbren12 (changes requested):
This will be a very useful feature. Thanks for putting it together. I have some big picture concerns about coupling the low level classes to the dataclass configuration frontend.

external/fv3fit/fv3fit/emulation/layers/architecture.py
oliverwm1: Thanks that's a good point.

> pass in the instantiated self._kernel_regularizer.instance in directly (this at least addresses the concern of coupling to the front-end configs).

I was worried this would cause issues for serialization (in particular adding the keras regularizer object to the `get_config` call) but that seems to not be the case.

> Pass in a dense_layer_factory as a keyword argument with a default of tf.keras.layers.Dense. Build the dense_layer_factory after the configs are built using the kernel regularizer. You can use dense_layer_factory above too.

Agreed this would clean things up. So I should also  remove `width` and `activation` from the MLPBlock args, since these would be handled by the dense_layer_factor?

nbren12: > Agreed this would clean things up. So I should also remove width and activation from the MLPBlock args, since these would be handled by the dense_layer_factor?

Hmm.. that feels a bit too far for me but it's a subjective call. Up to you.

nbren12 (approved):
LGTM

PR #1966, opened 2022-07-28
Simple script to prepare out-of-sample experiment configs

chsanford:
I got tired of copying config files over and over and uploading them individually to Google Cloud. This python script takes one big config file that specifies a base model and a collection of novelty detection models (with different cutoffs) and (1) generates a collection of `out_of_sample_model.yaml` config files with `name` files for each novelty detector on GCS, (2) creates local `prognostic-run.yaml` config files for the prognostic runs, and (3) makes one `run.sh` script that launches all of the argo jobs.

- test_unit: [83%](https://output.circle-artifacts.com/output/job/cc7423da-91fe-4e41-b3d9-973c1b468edc/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


projects/out_of_sample/scripts/experiment-config/prep_oos_experiments.py
```diff
@@ -0,0 +1,168 @@
...
+            argo_submit_str, job_name, output_path = make_argo_submit_command(
+                bucket,
+                project,
+                experiment,
+                trial,
+                os.path.join(prognostic_config_path_dir_name, "prognostic-run.yaml"),
+                segments,
+            )
+            run_str += argo_submit_str
+            print(f"Experiment #{n}: {nd_name} with cutoff={cutoff}")
```
AnnaKwa, line 149: It may be useful to print these lines to file so you have an easily accessible list of the run names and their configurations

projects/out_of_sample/scripts/experiment-config/prep_oos_experiments.py
```diff
@@ -0,0 +1,168 @@
...
+    for novelty_detector in all_experiment_config["novelty_detectors"]:
+        nd_name = novelty_detector["nd_name"]
+        nd_path = novelty_detector["nd_path"]
+        for cutoff in novelty_detector["cutoffs"]:
+            n += 1
+            model_config_paths = []
+            for tendency, tendency_path in tendency_paths.items():
+                tendency_id = f"{nd_name}-{cutoff}-{tendency}"
+                model_config_path = os.path.join(model_config_dir_path, tendency_id)
+                temp_config_path = os.path.join(temp_dir.name, tendency_id)
```
AnnaKwa, line 116: FYI If the temporary directory only needs to be referenced in one bit of the code rather than throughout, you can just use it within the context 
```
with tempfile.TemporaryDirectory() as tmpdirname:
    do stuff with the tmpdir and it gets deleted after this block
```
so you don't have to manually call cleanup or do the atexit stuff

AnnaKwa:
I mostly looked at the prep OOS script, it looks good to me. You might want to test it out on a config with a single entry to just manually verify that the correct fields are updated in the config that will be submitted.

chsanford:
I verified it on a config file for a pair of tests and verified that they indeed launched.

chsanford:
I verified it on a config file for a pair of tests and verified that they indeed launched.

AnnaKwa (approved):

PR #1970, opened 2022-07-29
Fix bug where segment 2 is run repeatedly

nbren12:
When using the python segemented run api, prognostics runs of more than
two segements repeat the second segment indefinitely. This occurs due to
a bug in read_last_segment. fsspec file listings share a write-through
cache that is not updated when GCS is modified outside of fsspec. This
means that listing the artifacts directory will not show the
folders uploaded by the gsutil part of append.

The first call to read_last_segment returns None and doesn't call
fsspec.ls. The second call succeeds, and the third call returns the
cached contents of the second call.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/5b9627a9-de5d-4de2-8e4a-c3f191f6b92e\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


spencerkclark (approved):
Looks good to me.

PR #1971, opened 2022-07-29
fv3kube: fix v0.7 base yaml SST flag

### LINKED ISSUE #1972: Prognostic run: erroneous SSTs in v0.7 base yaml
The latest (v0.7) fv3config base yaml has `gfs_physics_nml.use_analysis_sst: true`, which results in SSTs being set to 300K apparently, because dycore nudging (the intended usage for `use_analysis_sst`) is also turned off in this base yaml. See example report [here](https://storage.googleapis.com/vcm-ml-public/argo/2022-07-29-19b59a9385d5/maps.html) of the base yaml behavior. Overriding SSTs fixes this but the 300K SSTs remain in the diagnostics. Solution is to set `use_analysis_sst: false` for applications that aren't nudge to obs.

brianhenn:
Fixes a misconfiguration in the fv3kube v0.7 base yaml that resulted in SSTs being set to wrong values for non nudge-to-obs runs, unless they were overridden (but even then the wrong values remained in the diagnostics). 

Refactored public API:
- changes `gfs_physics_nml.use_analysis_sst` to false in v0.7 base yaml; also removes analysis data provisioning not used outside of nudge-to-obs

Resolves #1972 

- test_unit: [83%](https://output.circle-artifacts.com/output/job/859f6d17-28e6-492e-9887-84d04607ebf9/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
+++ b/external/fv3kube/fv3kube/base_yamls/v0.7/fv3config.yml
@@ -280,7 +280,7 @@ namelist:
     shal_cnv: true
     swhtr: true
     trans_trac: true
-    use_analysis_sst: true
+    use_analysis_sst: false
     use_ufo: true
   interpolator_nml:
     interp_method: conserve_great_circle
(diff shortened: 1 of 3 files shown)


oliverwm1 (approved):
LGTM

PR #1974, opened 2022-07-29
Extract all io

YakelynRJ:
This PR extract i/o from: 
- main long/short wave radiation classes
- radiaton_aerosols
- radiation_sfc 
- radiation_gases

All data loading occurs in getdata.py 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/f3118c3c-571d-4bc0-a60e-39b170078dd5\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/radiation/python/getdata.py
```diff
@@ -0,0 +1,240 @@
+import numpy as np
+from config import *
+import os 
+import xarray as xr 
+import warnings
+
+
+def random_numbers(LOOKUP_DIR, me):
```
nbren12, line 8: nit: can you use lower case for the function arguments in this file. Typically we reserve all caps for module level constants. Also some type hints would be helpful since it isn't clear what "me" is.
```suggestion
def random_numbers(lookup_dir: str, me: int):
```

external/radiation/python/getdata.py
```diff
@@ -0,0 +1,240 @@
+import numpy as np
```
nbren12, line 1: Since this is a new file...it would be nice to add it to our linting rules. These are set here: https://github.com/ai2cm/fv3net/blob/279bdb782fa3232e2cc792f7bb97fcf86c050475/.pre-commit-config.yaml#L18

external/radiation/python/getdata.py
```diff
@@ -0,0 +1,240 @@
...
+        ## file names needed in lwrad()
+    
+    lw_dict = {}
+    dfile = os.path.join(LOOKUP_DIR, "totplnk.nc")
+    pfile = os.path.join(LOOKUP_DIR, "radlw_ref_data.nc")
+    lw_dict['totplnk'] = xr.open_dataset(dfile)["totplnk"].values
+    lw_dict['preflog'] = xr.open_dataset(pfile)["preflog"].values
+    lw_dict['tref'] = xr.open_dataset(pfile)["tref"].values
+    lw_dict['chi_mls'] = xr.open_dataset(pfile)["chi_mls"].values
+    del(dfile, pfile)
```
nbren12, line 32: I think you can probably delete all the manual `del` calls in this file. I'm not sure why they are needed.

external/radiation/python/getdata.py
```diff
@@ -0,0 +1,240 @@
...
+                      10:['selfref','forref','absa','absb','fracrefa','fracrefb'],
+                      11:['selfref','forref','absa','absb','fracrefa','fracrefb','ka_mo2','kb_mo2'],
+                      12:['selfref','forref','absa','fracrefa'],
+                      13:['selfref','forref','absa','fracrefa','fracrefb','ka_mco2','ka_mco','kb_mo3'],
+                      14:['selfref','forref','absa','absb','fracrefa','fracrefb'],
+                      15:['selfref','forref','absa','fracrefa','ka_mn2'],
+                      16:['selfref','forref','absa','absb','fracrefa','fracrefb'],
+                    }
+
+    for nband in range(1, 17):
```
nbren12, line 63: Would probably be safer to use the keys in varname_bands
```suggestion
    for nband in varname_bands: 
```

external/radiation/python/getdata.py
```diff
@@ -0,0 +1,240 @@
...
+                    23:['selfref','forref','absa','rayl','givfac'],
+                    24:['selfref','forref','absa','absb','abso3a','abso3b','rayla','raylb'],
+                    25:['absa','abso3a','abso3b','rayl'],
+                    26:['rayl'],
+                    27:['absa','absb','rayl'],
+                    28:['absa','absb','rayl'],
+                    29:['forref','absa','absb','selfref','absh2o','absco2','rayl']
+                    }
+                  
+    for nband in range(16, 30):
```
nbren12, line 124:
```suggestion
    for nband in varnames_bands: 
```

external/radiation/python/getdata.py
nbren12, line 8: We typically reserve upper case for module level constants. Can you modify all the function arguments in this file to lower case?

Also, `me` is not an informative name. I would change to `tile` and add type hints to clarify what it is.
```suggestion
def random_numbers(lookup_dir: str, tile: int):
```

You should be able to make similar edits in the other functions.

nbren12 (changes requested):
Looks good. Just some style nits below. 

It would be nice to enable linting for the new files in this PR...but perhaps we should just lint/reformat everything in a follow-up PR. I excluded this sub-directory from our formatting requirements to avoid creating too big a diff, but your refactors have touched all the code anyway.

external/radiation/python/getdata.py
YakelynRJ: I will allow lint/reformat in a next PR

external/radiation/python/getdata.py
YakelynRJ: removed.

external/radiation/python/getdata.py
YakelynRJ: addressed

external/radiation/python/getdata.py
YakelynRJ: addressed

nbren12 (approved):

PR #1976, opened 2022-07-29
Allow prog diags to run if verification time range is non overlapping

AnnaKwa:
If the verification time range doesn't overlap with the prognostic data, differencing the two datasets leads to a zero-length time dimension, which in turn leads to various errors in other functions that take in the bias data. The "DivisionByZero" error that pops up in the zonal mean bias is due to this issue. This PR adds some if/else statements in diagnostic functions that used to assume a nonzero time dimension.

Some of the repeated if/else checks could be avoided if the bias function used `with xr.set_options(arithmetic_join="outer")`. However, I think this would result in a lot of unnecessary reads of both the prognostic and verification data to just end up with a final diagnostic full of NaNs.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/69b02a68-3f87-4478-b774-dbdd403cc1b1\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/_shared/transform.py
+++ b/workflows/diagnostics/fv3net/diagnostics/_shared/transform.py
@@ -157,10 +157,14 @@ def daily_mean(split: timedelta, arg: DiagArg) -> DiagArg:
         arg: input arguments to transform prior to the diagnostic calculation
     """
     prognostic, verification, grid = arg.prediction, arg.verification, arg.grid
-    split_time = prognostic.time.values[0] + split
-    prognostic = _resample_end(prognostic, split_time, "1D")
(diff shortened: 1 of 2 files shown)


nbren12 (approved):
LGTM. Though maybe it would be cleaner to simply not call the `bias` and `rmse` functions if the verification is empty. Would be nice to have a warning emitted e.g. when the data first loads.

AnnaKwa:
> maybe it would be cleaner to simply not call the bias and rmse functions if the verification is empty


I think right now there's no option for 'no verification'- if the flag isn't provided it defaults to the 40 day SHiELD verification data. Maybe the default should be changed to an empty verification.

PR #1977, opened 2022-07-30
e2e test: restore nudge to obs entries in prog run config

brianhenn:
#1971 broke the e2e test, because it removed from the fv3kube base yaml necessary entries for nudge to obs (which is done in the e2e prog run) without adding them to the e2e prog run config. This restores them in the latter location.

- test_unit: [83%](https://output.circle-artifacts.com/output/job/5b0ae247-75b5-4127-a966-bb4044614cdb/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


AnnaKwa (approved):
LGTM, could you add a comment in the prognostic run config that the nudging to obs is enabled in the 'train-diags-prog' workflow just for code coverage?

PR #1979, opened 2022-08-01
Reformat and enable linting of radiation

YakelynRJ:
This PR is to reformat and lint radiation files changed in previous PRs.
I also deleted files not used in test_driver.py

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/6eac23fb-9f17-493f-b844-8cde59a0193d\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/radiation/python/radiation_clouds.py
```diff
@@ -600,14 +600,6 @@ def progcld2(
                     qcwat[i, k] = 0.0
                     clw2[i, k] = clwf[i, k]
 
-        # -# Call module_microphysics::rsipath2(), in Ferrier's scheme, to
-        # compute layer's cloud liquid, ice, rain, and snow water condensate
-        # path and the partical effective radius for liquid droplet, rain drop,
-        # and snow flake.
-        cwp, cip, crp, csp, rew, rer, res, rsden = rsipath2(
```
nbren12, line 607: If this is not used, what about the surrounding code blocks? It would be less confusing to delete the whole block of unused code rather than just this subroutine call which happened to give the linting error.

external/radiation/python/radiation_astronomy.py
```diff
@@ -1,6 +1,3 @@
-import sys
-
-sys.path.insert(0, "..")
```
nbren12, line 6: Are these path modifications really not needed? Is this code executed?

external/radiation/python/radiation_driver.py
```diff
@@ -345,13 +345,13 @@ def GFS_radiation_driver(
         LM = Model["levr"]
         LEVS = Model["levs"]
         IM = Grid["xlon"].shape[0]
-        NFXR = Model["nfxr"]
+        # NFXR = Model["nfxr"] # never used according to lint
```
nbren12, line 348: Please delete rather than comment out.

external/radiation/python/radiation_gases.py
```diff
@@ -137,42 +137,43 @@ def gas_update(self, iyear, imon, iday, ihour, loz1st, ldoco2, me, data_gases):
...
-                    k2oz = (imon % 12) + 1
-                    midp = mdays[k2oz] / 2 + 1 + mdays[k1oz]
-
-            if iday < midmon:
-                id = iday + mdays[k1oz]
-            else:
-                id = iday
-
-            facoz = float(id - midm) / float(midp - midm)
+        # if self.ioznflg == 0:
```
nbren12, line 155: Please delete rather than comment out.

external/radiation/python/radiation_sfc.py
```diff
@@ -365,18 +364,19 @@ def setalb(
...
-                    # compute beta distribution parameters for all 4 albedos
-                    m = sfcalb[i, kk]
-                    s = pertalb[0] * m * (1.0 - m)
-                    alpha = m * m * (1.0 - m) / (s * s) - m
-                    beta = alpha * (1.0 - m) / m
-                    # compute beta distribution value corresponding
-                    # to the given percentile albPpert to use as new albedo
-                    albtmp = ppfbet(albPpert[i], alpha, beta, iflag)
-                    sfcalb[i, kk] = albtmp
+        # what's below needs a subrutine called ppfbet() but that is not defined
```
nbren12, line 367: Please delete rather than comment out.

nbren12 (changes requested):
Thanks. This will simplify code review in the future, but I am concerned about commenting/deleting code in commits innocuously titled "fix lint errors". This should be done in a separate commit so that it is easy to revert....and clearly labeled with e.g. "delete dead code"...perhaps it would be best to first make a separate PR just deleting the dead code and then rebase this one on top of that (e.g. make this PR only contain formatting changes). Alternatively, you could also restore the deleted code and ignore the errors inline like this: https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors.

nbren12:
Closing in favor of #1985

PR #1987, opened 2022-08-03
Train models with high antarctic points only

nbren12:
Changes to allow training emulators for just the antarctic high plateau. Hopefully we have enough training data there.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/5f62559b-e145-4aeb-a51d-47bd20970262\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
+++ b/external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
@@ -1,4 +1,10 @@
+"""Filters to be used with unbatched tensorflow datasets
+
+These should be callables of a single column of data (i.e. no batch dimension)
+it must return a scalar boolean.
+"""
 from typing import Set
+import tensorflow as tf
 from fv3fit.emulation.types import TensorDict
 import dataclasses
 import numpy as np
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
```diff
@@ -0,0 +1,18 @@
...
+@dataclasses.dataclass
+class HighAntarctic:
+    high_antarctic_only: bool = True
+
+    @property
+    def input_variables(self) -> Set[str]:
+        return {"latitude", "surface_air_pressure"}
+
+    def __call__(self, elm: TensorDict):
+        lat = elm["latitude"] < np.deg2rad(-60)
```
frodre, line 16: What does `elm` stand for?

external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
```diff
@@ -0,0 +1,18 @@
...
+        ps = elm["surface_air_pressure"] < 700e2
+        return (ps & lat)[0]
```
frodre, line 18: `[0]` is to squeeze dimensions, I assume.  Maybe add a module level comment  or protocol that returns should be applicable only to a single element if we don't want broadcasts for the unbatched `dataset.filter` operation.

frodre (approved):
lgtm

external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
nbren12: element...sorry.

external/fv3fit/fv3fit/emulation/zhao_carr/filters.py
nbren12: Correct. it needs to return a scalar bool. I documented this.

PR #1988, opened 2022-08-03
Remove inout-like parameters to GFS_radiation_driver

YakelynRJ:
This PR defined a method called _GFS_radiation_driver() that does not take zero array inputs. 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/3fc45540-4060-4207-a6ea-85734af1895f\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/radiation/python/radiation_driver.py
+++ b/external/radiation/python/radiation_driver.py
@@ -344,7 +344,6 @@ class RadiationDriver:
             Coupling,
             Grid,
             Tbd,
-            Radtend,
             randomdict,
             lwdict,
             swdict,
@@ -358,7 +357,6 @@ class RadiationDriver:
         Coupling,
         Grid,
         Tbd,
-        Radtend,
         randomdict,
         lwdict,
         swdict,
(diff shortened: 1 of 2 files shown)


external/radiation/python/radiation_driver.py
```diff
@@ -423,6 +421,18 @@ def _GFS_radiation_driver(
 
         scmpsw = dict()
         Diag = dict()
+        Radtend = dict()
```
nbren12, line 408: No need to change this, but usually to initialize a dict one can do this: `Radtend = {}`

external/radiation/python/radiation_driver.py
```diff
@@ -403,21 +403,11 @@ def _GFS_radiation_driver(
         cldtausw = np.zeros((IM, Model["levr"] + self.LTP))
 
         Coupling = dict()
```
nbren12, line 405: Perhaps we should add some regression tests for the outputs of `Coupling` soon. It is a gap in our test coverage. Maybe you can work with @elynnwu on this.

nbren12 (approved):
Looks perfect. So is this function completely pure now or does it still modify some of its inputs in place?

YakelynRJ:
@nbren12 yes, it is a pure function.

PR #1989, opened 2022-08-04
Remove tbd parameters gfs radiation driver

YakelynRJ:
This PR removes Tbd i/o because the radiation driver module does not use them. 

phy_f3d is only used for flags and values that are set to False or do not apply. 
icsdsw and icsdlw are random generated seeds that were called in mcica_subcol() but they were not used. 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/ff889515-422b-4fb9-a222-424a2f50b819\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/radiation/python/radiation_driver.py
```diff
@@ -747,31 +747,9 @@ def _GFS_radiation_driver(
                     if ccnd[i, k, 0] < self.EPSQ:
                         ccnd[i, k, 0] = 0.0
 
-        if Model["uni_cld"]:
```
nbren12, line 750: Which of these options is not implemented? 

It appears Andrew only partially ported the `Model["effr_in"]` logic, since effrl, etc are not used anywhere. This PR would be clearer if it deleted all the dead code e.g. references to the `effrl`, `effrr`, etc and raised an error near the top of the function if `Model["effr_in"]` is true.

external/radiation/python/radiation_driver.py
```diff
@@ -747,31 +747,9 @@ def _GFS_radiation_driver(
...
-                        cldcov[i, k1] = Tbd["phy_f3d"][i, k, Model["indcld"] - 1]
-                        effrl[i, k1] = Tbd["phy_f3d"][i, k, 1]
-                        effri[i, k1] = Tbd["phy_f3d"][i, k, 2]
-                        effrr[i, k1] = Tbd["phy_f3d"][i, k, 3]
-                        effrs[i, k1] = Tbd["phy_f3d"][i, k, 4]
-            else:
-                for k in range(LM):
-                    k1 = k + kd
-                    for i in range(IM):
-                        cldcov[i, k1] = Tbd["phy_f3d"][i, k, Model["indcld"] - 1]
```
nbren12, line 764: Are you sure this line isn't used? Is it always overwritten by L754 below?

nbren12 (changes requested):
I'm worried we may want to restore the dead code at some point. It would be clearer to delete all the dead code rather than leaving it in a half-deleted state.

external/radiation/python/radlw/radlw_main.py
```diff
@@ -1726,14 +1723,6 @@ def lwrad(
         if verbose:
             print("Beginning lwrad . . .")
             print(" ")
-
-        if self.isubclw == 1:
```
brianhenn, line 1730: Similar to Noah's comment about `effr`, I would add raising an error if either `isubclw` or `isubcsw` isn't 2. Those two seem like the more commonly-changed flags that weren't implemented/validated in the port, as far as I know.

external/radiation/python/radiation_driver.py
YakelynRJ: Model["effr_in"] is set to False but I added exception error for when it is set to True.

external/radiation/python/radiation_driver.py
YakelynRJ: Yes. Model["uni_cld"] is to to False and therefore everything inside this loop is not used.

external/radiation/python/radlw/radlw_main.py
YakelynRJ: Added raise error for when effr_in is set to True. 

isubclw and isubcsw are called at initialization so I added rise error for when they are set  to 2.

external/radiation/python/radiation_driver.py
```diff
@@ -118,6 +118,7 @@ def radinit(
                     f"- ISUBCLW={isubclw}, Use McICA with random ",
                     f"permutation seeds for LW random number generator",
                 )
+                raise ValueError('isubclw == 2 is not implemented')
```
brianhenn, line 121: This should raise an error if `isubclw` is _not_ 2. Same issue with `isubcsw`. You could keep the print statements for options 0 and 1 and just add `raise` statements after them.

nbren12 (approved):
LGTM. I am going to squash merge this since the history is a bit messy with linting etc mixed in.

PR #1986, opened 2022-08-02
Implement pytorch graph-based network training in fv3fit

EbrahimNabizadeh:
This PR adds a new graph-based network to fv3fit for the sequence climate data, written in Pytorch. The following changes and packages are required to build/train the network.  

Added public API:

Added a new “graph” training function for PyTorch graph-based neural network training to fv3fit, including the following tools: 
- fv3fit.pytorch.training_loop, optimizes PyTorch models based on configuration
- fv3fit.pytorch.loss, measures the loss of model output and the target value based on configuration. It also contains a multistep loss function to be used during training. 
- fv3fit.pytorch.graph.graph_config, includes configuration for a graph-based model  
- fv3fit.pytorch.optimizer, a Pytorch optimization configuration
- fv3fit.pytorch.predict, prediction for sequence data (load/dump need to be written in another PR))
- fv3fit.pytorch.graph.graph, returns the fv3fit predictor based on PyTorch

Significant internal changes:

- graph_builder designs the graph structure using  geopy, which will then use in training a graph network 

Requirement changes:
- The fv3fit setup.py has been changed to include the extra packages that are required to build a graph network in Pytorch. These changes include the following packages:
  - torch (a machine learning library),
  - torchvision (a package consisting of popular datasets, model architectures, and common data transformations),
  - dgl (an API for graph-based networks),
  - geopy (use for building graphs based on lat-lon coordinate),
  - tensorflow-datasets (use to transfer tf.dataset to PyTorch tensor).

- [x] Ran `make lock_pip` after adding the above packages to setup.py (https://vulcanclimatemodeling.com/docs/fv3net/dependency_management.html#dependency-management)
- [x] Add PR review with license info for any additions to `constraints.txt`
  ([example](https://github.com/ai2cm/fv3net/pull/1218#pullrequestreview-663644359))

- [x] Tests added
Identity test: quantify the ability of the model to determine the identity function

Coverage reports (updated automatically):
- test_unit: [84%](https:\/\/output.circle-artifacts.com\/output\/job\/66187e54-bb1d-4538-9e2a-030857247a41\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

PR #1991, opened 2022-08-08
Fast graph generation in fv3fit

mcgibbon:
The current graph generation code in fv3fit is very slow, and requires an external dependency. This PR uses indexing-based logic on a cubed sphere instead to greatly reduce the time needed to generate the graph, and remove the external dependency.

Based on and may be merged into #1986 .

Significant internal changes:
- Greatly reduced build_graph execution time and removed dependency on geopy

Requirement changes:
- Removed dependency of fv3fit on geopy

- [x] Tests added

- test_unit: [84%](https://output.circle-artifacts.com/output/job/c2609ef8-59d9-4bf0-90cf-7c9339cbeec6/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/tests/training/test_graph.py
```diff
@@ -115,3 +118,14 @@ def sample_func():
         )
 
     return sample_func
+
+
+def test_graph_builder():
+    graph = build_graph(2)
+    edges = list(zip(graph[0], graph[1]))
+    assert (0, 1) in edges
```
frodre, line 126: Add a comment after each assertion which connection it is testing or add intermediate variables with the name.

external/fv3fit/fv3fit/pytorch/graph/graph_builder.py
```diff
@@ -1,46 +1,57 @@
...
+                dims=["tile", "x", "y", "z"],
+            )
+        }
+    )
+    ds = append_halos(ds, n_halo=1).squeeze("z")
+    index = ds["index"].values
+    total_edges = n_points * 5
+    out = np.empty((total_edges, 2), dtype=int)
+    # left connections
+    out[:n_points, 0] = index[:, 1:-1, 1:-1].flatten()
```
frodre, line 42: Nice! I wanted to draw this out but I didn't have the tile order memorized :p.  Makes sense via the indexing slices though and thanks for the comments to categorize

frodre (approved):
Nice work.  LGTM

EbrahimNabizadeh (approved):
looks good to me.

external/fv3fit/tests/training/test_graph.py
mcgibbon: Added comments.

external/fv3fit/fv3fit/pytorch/graph/graph_builder.py
```diff
@@ -1,46 +1,57 @@
...
-    nodes = []
-    edges = []
-
-    grid = catalog[config.resolution].read()
-    lat = grid.lat.load()
-    lon = grid.lon.load()
-
-    lat = lat[:, :: config.coarsen, :: config.coarsen].values.flatten()
-    lon = lon[:, :: config.coarsen, :: config.coarsen].values.flatten()
+def build_graph(nx_tile: int) -> Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
mcgibbon: Unfortunately the number of edges can't change - it can't be fewer, because then a direction must be missing. If it is more, the increased edges must be symmetric, which would mean adding 4 corners to 9 edges, but for corner points there are only 8 neighbor points, not 9. If DGL supports having different numbers of edges per node, we can add an `include_corners: bool` argument for this later.

PR #1994, opened 2022-08-10
Prognostic run: restore timings and intermediate restarts

### LINKED ISSUE #1785: prognostic run: python timings not output
We have [python wrapper timing](https://github.com/ai2cm/fv3net/blob/8857cd27a4e957514b8311275055911186a02226/workflows/prognostic_c48_run/runtime/loop.py#L383) in the prognostic run loop, but the function to log it is not currently called. It looks like we switched from calling `loop.cleanup()` (includes logging python timing) to `wrapper.cleanup()` (doesn't include it) at some point to avoid a gc/mpi bug; we should refactor timing logging call to somewhere appropriate to bring it back.

brianhenn:
We currently do not have the ability to log wrapper timing and write out intermediate restarts in the prognostic run, though we've had this in the past. This PR restores those, and cleans up the timing logging as a json. 

Significant internal changes:
- Python wrapper timings logged to json
- Intermediate restarts now written out if specified in namelist

Resolves #1785 

- test_unit: [83%](https://output.circle-artifacts.com/output/job/e6b86cc2-9fbd-4ca1-8e94-3cbb26e9b099/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -374,17 +374,10 @@ class TimeLoop(
             "total_precip_after_physics": self._state[TOTAL_PRECIP],
         }
 
-    def _print_timing(self, name, min_val, max_val, mean_val):
-        self._print(f"{name:<30}{min_val:15.4f}{max_val:15.4f}{mean_val:15.4f}")
-
-    def print_global_timings(self, root=0):
+    def log_global_timings(self, root=0):
         is_root = self.rank == root
         recvbuf = np.array(0.0)
         reduced = {}
-        self._print("-----------------------------------------------------------------")
-        self._print("         Reporting clock statistics from python                  ")
-        self._print("-----------------------------------------------------------------")
-        self._print(f"{' ':<30}{'min (s)':>15}{'max (s)':>15}{'mean (s)':>15}")
         for name, value in self._timer.times.items():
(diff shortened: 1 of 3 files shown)


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -378,28 +374,19 @@ def _apply_physics(self) -> Diagnostics:
...
 
-    def _print_timing(self, name, min_val, max_val, mean_val):
-        self._print(f"{name:<30}{min_val:15.4f}{max_val:15.4f}{mean_val:15.4f}")
-
-    def _print_global_timings(self, root=0):
+    def log_global_timings(self, root=0):
         is_root = self.rank == root
         recvbuf = np.array(0.0)
         reduced = {}
-        self._print("-----------------------------------------------------------------")
```
oliverwm1, line 388: why not output both human-readable and machine-readable timings as is currently done?

oliverwm1 (approved):
I'm approving, although I think it would be better to maintain the human-readable timings as currently implemented. No reason we can't output these in addition to the serialized json.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Sure, that's fine

PR #1995, opened 2022-08-10
remove hard-coded-solar-filename

YakelynRJ:
This PR removes the hard-coded FORCING_DIR (imported from config). Instead of using FORCING_DIR to check the if statements, a full path file_name is used. getdata.py was cleanup a bit but is essentially the same. 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/67987b43-781e-48c6-90c1-f7aa7e81d25a\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/radiation/python/radiation_astronomy.py
```diff
@@ -50,7 +49,7 @@ def __init__(self, me, isolar, solar_filename):
                     " with cycle approximation (old values)!",
                 )
 
-            file_exist = os.path.isfile(os.path.join(FORCING_DIR, self.solar_fname))
```
brianhenn, line 53: So is this needed because the `solar_fname` being provided already has the forcing dir now?

brianhenn (commented):
Happy to approve but maybe you can explain why this was needed a little more clearly?

external/radiation/python/radiation_astronomy.py
YakelynRJ: FORCING_DIR is imported from config so there were conflicts when I used a different path.

nbren12 (commented):
Sorry I forgot to hit "submit"....

external/radiation/python/getdata.py
```diff
@@ -334,40 +334,43 @@ def astronomy(forcing_dir, isolar, tile_number):
...
-                data = xr.open_dataset(os.path.join(forcing_dir, solar_file))
+            file_name = "solarconstant_noaa_an.nc"
+            solar_file = os.path.join(forcing_dir, file_name)
+            if os.path.isfile(solar_file):
+                data = xr.open_dataset(solar_file)
             else:
                 warnings.warn(f'Requested solar data file "{solar_file}" not found!',)
                 raise FileNotFoundError(
                     " !!! ERROR! Can not find solar constant file!!!"
                 )
```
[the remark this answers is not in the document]
YakelynRJ: I just modified to read at the end of the if statements

brianhenn (approved):
@YakelynRJ LGTM

PR #1996, opened 2022-08-10
Improve high-antarctic training performance.

nbren12:
The high-antarctic filter was far too slow to use for training it took 1944 seconds per epoch (even though each epoch is only 1.5 of the data). This was dominated by I/O from the local filesystem.

Pre-saving a filtered dataset reduces the per-epoch time to 16 seconds.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/acf81c7b-144b-4c09-b289-91edfe0bdfeb\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/train_microphysics.py
+++ b/external/fv3fit/fv3fit/train_microphysics.py
@@ -184,6 +184,16 @@ class TransformedParameters(Hyperparameters):
 
         return out
 
+    def prepare_flat_data(self, batched: tf.data.Dataset) -> tf.data.Dataset:
+        unbatched = (
+            batched.map(tfdataset.apply_to_mapping(tfdataset.float64_to_float32))
+            .map(expand_single_dim_data)
+            .unbatch()
+        )
+        for filter in self.filters:
+            unbatched = unbatched.filter(filter)
+        return unbatched
+
     @property
     def transform_factory(self) -> ComposedTransformFactory:
         return ComposedTransformFactory(self.tensor_transform)
@@ -406,11 +416,11 @@ class TrainConfig(TransformedParameters):
             raise NotImplementedError(self.data_format)
 
     def open_train_test(self):
-        train_ds = self.open_dataset(self.train_url, self.nfiles, self.model_variables)
-        test_ds = self.open_dataset(
-            self.test_url, self.nfiles_valid, self.model_variables
-        )
-        return unbatch_datasets(train_ds, test_ds, self.filters)
+        train = self.open_dataset(self.train_url, self.nfiles, self.model_variables)
+        test = self.open_dataset(self.test_url, self.nfiles_valid, self.model_variables)
+        train_ds = self.prepare_flat_data(train)
+        test_ds = self.prepare_flat_data(test) if test else None
+        return train_ds, test_ds
 
 
 def save_jacobians(std_jacobians, dir_, filename="jacobians.npz"):
@@ -423,32 +433,19 @@ def save_jacobians(std_jacobians, dir_, filename="jacobians.npz"):
         np.savez(os.path.join(tmpdir, filename), **dumpable)
 
 
-def unbatch_datasets(train, test, filters):
-    def _prepare(ds):
-        unbatched = (
-            ds.map(tfdataset.apply_to_mapping(tfdataset.float64_to_float32))
-            .map(expand_single_dim_data)
-            .unbatch()
-        )
-        for filter in filters:
-            unbatched = unbatched.filter(filter)
-        return unbatched
-
-    return _prepare(train), _prepare(test) if test else None
-
-
 @register_training_function("transformed", TransformedParameters)
 def train_function(
     hyperparameters: TransformedParameters,
     train_batches: tf.data.Dataset,
     validation_batches: Optional[tf.data.Dataset],
 ) -> PureKerasDictPredictor:
-
-    train_batches, validation_batches = unbatch_datasets(
-        train_batches, validation_batches, hyperparameters.filters
+    train = hyperparameters.prepare_flat_data(train_batches)
+    validation = (
+        hyperparameters.prepare_flat_data(validation_batches)
+        if validation_batches
+        else None
     )
-
-    return _train_function_unbatched(hyperparameters, train_batches, validation_batches)
+    return _train_function_unbatched(hyperparameters, train, validation)
 
 
 def _train_function_unbatched(
--- a/external/fv3fit/fv3fit/train_microphysics.py
+++ b/external/fv3fit/fv3fit/train_microphysics.py
@@ -406,9 +406,7 @@ class TrainConfig(TransformedParameters):
         self, url: str, nfiles: Optional[int], required_variables: Set[str],
     ) -> tf.data.Dataset:
         if self.data_format == "nc":
-            self._open_dataset(
-                url, nfiles, required_variables,
-            )
+            return self._open_dataset(url, nfiles, required_variables,)
         elif self.data_format == "tf":
             # needs to be batched
             return tf.data.experimental.load(url).batch(1)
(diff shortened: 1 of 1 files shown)


external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -533,6 +560,29 @@ def main(config: TrainConfig, seed: int = 0):
...
+    tf.data.experimental.save(train, f"{path}/train")
+    tf.data.experimental.save(test, f"{path}/test")
+
+    # upload config
+    with put_dir(config.out_url) as tmpdir:
+        with open(os.path.join(tmpdir, "config.yaml"), "w") as f:
+            f.write(config.to_yaml())
+
+
+def main(config: TrainConfig, seed: int = 0):
```
mcgibbon, line 577: Would you consider making the "save only" path a new executable module (e.g. `save_microphysics`), instead of altering the API of train_microphysics to have a mode which behaves completely differently? If there are a bunch of kubernetes workflows that would need to be modified for that I won't insist, but typically we try to make different functions/scripts for different tools, rather than writing one script with multiple tools (unless it's clearly an entrypoint, like `vcmtools save` vs `vcmtools train`).

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -282,6 +282,9 @@ class TrainConfig(TransformedParameters):
         nfiles: Number of files to use from train_url
         nfiles_valid: Number of files to use from test_url
         log_level: what logging level to use
+        save_only: If true, don't train, but save the train and test data with
+            tf.data.experimental.save to ``out_url``
+        data_format: one of ['nc', 'tf']
```
mcgibbon, line 287: Can you describe what this option does, and remove "netcdfs" from the train/test_url descriptions above?

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -397,28 +422,33 @@ def save_jacobians(std_jacobians, dir_, filename="jacobians.npz"):
         np.savez(os.path.join(tmpdir, filename), **dumpable)
 
 
-@register_training_function("transformed", TransformedParameters)
-def train_function(
-    hyperparameters: TransformedParameters,
-    train_batches: tf.data.Dataset,
-    validation_batches: Optional[tf.data.Dataset],
-) -> PureKerasDictPredictor:
+def unbatch_datasets(train, test, filters):
```
mcgibbon, line 425: Is there a way to capture that this also applies the filters to the data in the function name? If not, can you add type hints and a docstring?

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -386,6 +391,26 @@ def open_dataset(
...
+            self._open_dataset(
+                url, nfiles, required_variables,
+            )
+        elif self.data_format == "tf":
+            # needs to be batched
+            return tf.data.experimental.load(url).batch(1)
+        else:
+            raise NotImplementedError(self.data_format)
+
+    def open_train_test(self):
```
mcgibbon, line 407: I think there is currently a bug where if you don't call the `save` workflow and train on its output, the `train` workflow never applies the filters even though it says it does. It would also easy to accidentally introduce a bug in the save-then-train workflow by calling this function in `train_main`, since that would mean filters get applied twice.

If you separate "save" into its own script you could remove the filters entirely from the train workflow, which I think solves the bug.

There are different ways to make it clear open_train_test should only be called when you want to filter, one option is to move this into the file for `save` and made a private function.

mcgibbon: This is the only really important comment in this review.

mcgibbon (changes requested):

external/fv3fit/fv3fit/train_microphysics.py
nbren12: I did consider a separate subcommand, but we do indeed have k8s tooling around this end point already...and it's nice to have everything in the yaml.

mcgibbon: Do you have a way to avoid the ignored-filters bug with one endpoint? It seems like if you do a save step and a train step and for both give it a configuration that says to apply these filters, you necessarily have to apply the filters twice, or have one of these endpoints ignore the configuration option.

nbren12: applying filters twice is ok IMO, since the filter isn't slow and filtering is idempotent.

TBH. I see where you are coming from, but I felt this was a relatively unobstrusive change that didn't break anything and has the benefit of keeping the config declarative, and I am fine taking on the tech debt since I'd like to switch to fv3fit.train anyway after this project is over anyway.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: > since that would mean filters get applied twice.

What is wrong with applying filters twice? The result is the same as applying once.

>  think there is currently a bug where if you don't call the save workflow and train on its output, the train workflow never applies the filters even though it says it does

Perhaps I am missing something. Could you say more? As far as I can tell all the CLI entrypoints will apply the filter.

1. fv3fit.train_microphysics -> train_main -> train_function (which calls `prepare_flat_data`) -> _train_function_unbatched
1. fv3fit.train -> train_function (which calls prepare_flat_data)
1. save -> open_train_test (which calls prepare_flat_data).

mcgibbon: I hadn't realized what the filters do, but given it's filtering on variables it should be fine to apply them twice.

I missed the call to `prepare_flat_data` in `train_function`.

Is there any risk to the other parts of `prepare_flat_data` being called twice? Particularly `.unbatch()`? If not the PR should be bug-free, though it could be easy for someone editing this to introduce a bug.

external/fv3fit/fv3fit/train_microphysics.py
mcgibbon: I hadn't realized that filters only select data, thought it was an abstract term for "transforms". Should be fine, then. Thanks for walking me through it!

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -276,12 +286,16 @@ class TrainConfig(TransformedParameters):
 
     Attributes:
 
-        train_url: Path to training netcdfs (already in [sample x feature] format)
-        test_url: Path to validation netcdfs (already in [sample x feature] format)
+        train_url: Path to training data (already in [sample x feature] format)
```
mcgibbon, line 289: Thanks for the doc update!

external/fv3fit/fv3fit/train_microphysics.py
```diff
@@ -184,6 +184,16 @@ def filter_variables(self) -> Set[str]:
 
         return out
 
+    def prepare_flat_data(self, batched: tf.data.Dataset) -> tf.data.Dataset:
```
mcgibbon, line 187: Much better name.

mcgibbon (approved):
LGTM, one question on whether two `.unbatch()` can cause issues but assuming it's not everything should be good.

external/fv3fit/fv3fit/train_microphysics.py
nbren12: I agree it's a concern, but I feel like the signature and name of `prepare_flat_data` makes it clear that the output data is unbatched while the input is batched.

mcgibbon: Looking longer, I was missing the extra `.batch()` call that happens when loading from tfdataset which makes it fine to call `.unbatch()` twice.

mcgibbon: I wish we could strongly type dataset dimensions.

nbren12:
Agreed. I think numpy arrays have some typing support now.

PR #2003, opened 2022-08-12
Update graph network training to use tile, x, y data

mcgibbon:
The existing graph network code operates on [node, z] data. When we use this function with the `fv3fit.train` entrypoint, it will need to operate on [sample, time, tile, x, y, z] data. This PR updates the routine to accept this form of data.

Also in this PR, the `dump` and `load` methods are implemented on PytorchModel.

Added public API:
- GraphNetworkConfig and GraphHyperparameters are added to the top level of fv3fit, GraphNetworkConfig has been modified but was previously private
- Made `sequence_size` a public method in `fv3fit.tfdataset`
- Implemented `__eq__` for StandardScaler

Refactored public API:
- Updated StandardScaler so it can operate on higher-dimensional data

Significant internal changes:
- Added CubedSphereGraphOperation, which can apply arbitrary graph operations from the `dgl` package on cubed-sphere data
- `dump` and `load` on PytorchModel are now implemented
- Added new `Dumpable` base class to use in `fv3fit.train` instead of Predictor, to allow training dumpable artifacts other than Predictor. This was needed because the use case for the graph network does not include the prognostic run, which Predictor is tailored to.
- Updated training function registration so it retains the call signature of the registered function instead of replacing it with a generic call signature
- Simplified test function used for graph network
- Added `DEVICE` constant in `fv3fit.pytorch` to avoid duplicating logic to determine the device for tensors
- Consolidated timestepping logic for graph network training into the TrainingLoop class under an `evaluate_model` helper function

- [x] Tests added

- test_unit: [84%](https://output.circle-artifacts.com/output/job/79c82828-c857-45e3-9c6b-eb2eeee9d171/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/pytorch/graph/network.py
```diff
@@ -0,0 +1,100 @@
...
+            SAGEConv(
+                int(config.n_hidden / 2), int(config.n_hidden / 2), config.aggregator
+            )
+        )
+        self.conv6 = CubedSphereGraphOperation(
+            SAGEConv(config.n_hidden, n_features_out, config.aggregator)
+        )
+        self.config = config
+
+    def forward(self, inputs):
```
mcgibbon, line 86: This function is unchanged.

external/fv3fit/fv3fit/pytorch/predict.py
```diff
@@ -1,76 +1,128 @@
...
-    stack,
-)
-from typing import Any, Dict, Hashable, Iterable, Sequence
+from fv3fit._shared import Dumpable
+from typing import Any, Dict, Hashable, Iterable, Mapping, Tuple
+from fv3fit.pytorch.system import DEVICE
 
 
-class PytorchModel(Predictor):
+class PytorchModel(Dumpable):
```
mcgibbon, line 11: This is no longer a Predictor.

mcgibbon:
I don't know why the changelog says fv3gfs-fortran is being modified, the updated version is currently the version on master.

external/fv3fit/fv3fit/_shared/training_config.py
```diff
@@ -34,9 +35,11 @@
 
 
 TrainingFunction = Callable[
-    [Dataclass, Sequence[xr.Dataset], Sequence[xr.Dataset]], Predictor
+    [Dataclass, Sequence[xr.Dataset], Sequence[xr.Dataset]], Dumpable
 ]
 
+TF = TypeVar("TF", bound=TrainingFunction)
```
mcgibbon, line 41: This makes it so decorating a function does not change it's type signature.

external/fv3fit/fv3fit/pytorch/predict.py
```diff
@@ -1,76 +1,128 @@
...
-            if scalar_singleton_dim:  # remove singleton dimension
-                output = output[..., 0]
-                dims = dims[:-1]
-            da = xr.DataArray(
-                data=output, dims=dims, coords={SAMPLE_DIM_NAME: stacked_coords},
-            ).unstack(SAMPLE_DIM_NAME)
-            dim_order = [dim for dim in self._unstacked_dims if dim in da.dims]
-            ds[name] = da.transpose(*dim_order, ...)
-        return ds
+    def pack_to_tensor(self, ds: xr.Dataset, times_per_window: int) -> torch.Tensor:
```
mcgibbon, line 33: The way I wrote the packing, the times in the window don't overlap, which means if you divide the dataset evenly into 1-day windows there will be one time of day where we never predict (the end of one window, which would evolve to the start of the next window). Is this acceptable or should I refactor it to pack into windows whose start and end times overlap with one another?

external/fv3fit/tests/training/test_graph.py
```diff
@@ -1,136 +1,73 @@
...
-
-
-@dataclasses.dataclass
-class TrainingResult:
-    model: PytorchModel
-    output_variables: Sequence[str]
-    test_dataset: tf.data.Dataset
-    hyperparameters: GraphHyperparameters
-
+from typing import Sequence
```
frodre, line 3: Nice simplification to cover this part of the code.

external/fv3fit/fv3fit/tfdataset.py
```diff
@@ -29,6 +29,13 @@ def apply_to_tuple(
     return tuple(tensor_func(tensor) for tensor in data)
 
 
+def sequence_size(seq):
```
frodre, line 32: Maybe an expensive function if loading is triggered

external/fv3fit/fv3fit/pytorch/training_loop.py
```diff
@@ -32,42 +24,78 @@ class TrainingLoopConfig:
...
+            train_data.unbatch()
+            .shuffle(buffer_size=self.buffer_size)
+            .batch(self.samples_per_batch)
+        )
         train_data = tfds.as_numpy(train_data)
-        validation_data = tfds.as_numpy(validation_data)
-        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        validation_data = validation_data.unbatch()
+        n_validation = sequence_size(validation_data)
+        validation_state = (
```
frodre, line 57: Probably worth adding to the docstring that the entirety of the validation data is loaded into memory.

external/fv3fit/fv3fit/pytorch/graph/network.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class GraphNetwork(nn.Module):
+    def __init__(self, config, n_features_in: int, n_features_out: int):
+        super(GraphNetwork, self).__init__()
+        self.conv1 = CubedSphereGraphOperation(
+            SAGEConv(n_features_in, config.n_hidden, config.aggregator)
+        )
+        self.conv2 = CubedSphereGraphOperation(
+            SAGEConv(config.n_hidden, int(config.n_hidden / 2), config.aggregator)
```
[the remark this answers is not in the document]
frodre: I guess this protects against either value being a float, but that would violate the config type.  Either works.

external/fv3fit/fv3fit/pytorch/predict.py
frodre: Does this drastically reduce the available training data?  As long as we don't have a systematic time of day that is missing from optimization, I don't necessarily see a problem w/ non-overlapping windows.

frodre (commented):
Thanks @mcgibbon!  Just wanted to touch base on the cube sphere operation wrapping when building the network, otherwise looks good.  I'll leave it up to you on the `int( a / b)` vs `a // b`

external/fv3fit/fv3fit/pytorch/graph/network.py
```diff
@@ -0,0 +1,111 @@
...
+        convolved = self.graph_op(graph, reshaped).transpose(1, 0)
+        return convolved.reshape(
+            inputs.shape[0], inputs.shape[1], inputs.shape[2], inputs.shape[3], -1
+        )
+
+
+class GraphNetwork(nn.Module):
+    def __init__(self, config, n_features_in: int, n_features_out: int):
+        super(GraphNetwork, self).__init__()
+        self.conv1 = CubedSphereGraphOperation(
```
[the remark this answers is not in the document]
mcgibbon: For the model implemented here, it would probably make more sense to wrap the entire GraphNetwork in a CubedSphereGraphOperation, since it's only doing graph operations. However this model is outdated, we plan to update it to the U-net implementation which will require doing a pooling or upsampling operation in cube-space between each convolutional block. This will be a better starting point for that update.

The operations it does are also very cheap - reshaping an index twice, neither of which copy memory, and retrieving a graph from LRU cache. Performance-wise, it shouldn't really matter which one of these we do.

external/fv3fit/fv3fit/pytorch/graph/network.py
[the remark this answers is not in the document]
mcgibbon: I'll update it to use `//` for clarity that these should be integers, but I'll keep the `int` cast for safety.

external/fv3fit/fv3fit/pytorch/graph/train.py
```diff
@@ -82,97 +79,88 @@ def train_graph_model(
...
-            processed_train_batches.unbatch().batch(
-                hyperparameters.normalization_fit_samples
-            )
-        )
+        iter(train_batches.unbatch().batch(hyperparameters.normalization_fit_samples))
     )
 
-    normalizer = get_normalizer(sample)
+    scalers = get_scalers(sample)
+    mapping_scaler = get_mapping_scaler(scalers)
```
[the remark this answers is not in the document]
mcgibbon: It's hard because we can't really type hint on the sample type of tfdatasets. This is a scaler that works on tfdatasets with mapping samples as opposed to tensor samples. I'll see what I can do as far as renaming and adding comments/type hints to make it clearer.

external/fv3fit/fv3fit/pytorch/predict.py
mcgibbon: This is only meant to be used for evaluation. The way I made the training data loader, it will randomly select windows like bootstrapping, so each epoch some samples may be over-sampled or not sampled.

external/fv3fit/fv3fit/pytorch/training_loop.py
mcgibbon: If you think this needs to be documented, I'd rather modify the code so it works out-of-memory on the tfdataset. I'll make that change.

external/fv3fit/fv3fit/tfdataset.py
mcgibbon: Unfortunately there's no other way to count tfdataset members. We already do this in the dense code.

external/fv3fit/fv3fit/pytorch/graph/train.py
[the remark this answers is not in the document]
mcgibbon: OK, I think I was able to make this clearer.

EbrahimNabizadeh (approved):
I just had a few minor comments. Otherwise, it looks good to me.

external/fv3fit/fv3fit/pytorch/graph/network.py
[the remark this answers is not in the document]
frodre: Cool, thanks for the information.

frodre (approved):
LGTM

PR #2004, opened 2022-08-12
Add dump and load to PytorchModel

mcgibbon:
PytorchModel currently cannot be saved to or loaded from disk. This PR implements `dump` and `load` for PytorchModel.

Added public API:
- Implemented `__eq__` for StandardScaler

Refactored public API:
- `dump` and `load` on PytorchModel are now implemented

- [x] Tests added

- test_unit: [84%](https://output.circle-artifacts.com/output/job/b3cefdb6-5d62-480b-8833-0d13eb3b4dea/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/_shared/predictor.py
```diff
@@ -20,6 +22,17 @@ def dump(self, path: str) -> None:
         pass
 
 
+class Loadable(abc.ABC):
+    """
+    Abstract base class for objects that can be dumped.
```
frodre, line 27:
```suggestion
    Abstract base class for objects that can be loaded from a directory.
```

external/fv3fit/fv3fit/pytorch/predict.py
```diff
@@ -1,18 +1,54 @@
...
+    """
+    with zipfile.ZipFile(f, "w") as archive:
+        for key, value in mapping.items():
+            with archive.open(str(key), "w") as f_dump:
+                value.dump(f_dump)
+
+
+def load_mapping(cls: Type[L], f: IO[bytes]) -> Mapping[Hashable, L]:
+    """
+    Serialize a mapping to a zip file.
```
frodre, line 41:
```suggestion
    Load a mapping from a zip file.
```

frodre (approved):
LGTM, some light cleanup of the copied docstrings suggested ;)

EbrahimNabizadeh (approved):
This looks good to me! Thanks

external/fv3fit/fv3fit/pytorch/predict.py
```diff
@@ -1,18 +1,54 @@
...
+import os
+import yaml
+import vcm
 
 
-class PytorchModel(Dumpable):
+L = TypeVar("L", bound="BinaryLoadable")
+
+
+class BinaryLoadable(Protocol):
```
nbren12, line 18: Why is Loadable an ABC  and BinaryLoadable a protocol?

Personally I'd prefer the latter for both so that we don't need to inherit from it explicitly.

nbren12: Quoting from https://peps.python.org/pep-0544/

> The same problem appears with user-defined ABCs: they must be explicitly subclassed or registered. This is particularly difficult to do with library types as the type objects may be hidden deep in the implementation of the library. Also, extensive use of ABCs might impose additional runtime costs.

> The intention of this PEP is to solve all these problems by allowing users to write the above code without explicit base classes in the class definition, allowing Bucket to be implicitly considered a subtype of both Sized and Iterable[int] by static type checkers using structural [[wiki-structural]](https://peps.python.org/pep-0544/#wiki-structural) subtyping:

mcgibbon: Because `Dumpable` and `Loadable` are part of the definition of `Predictor`, which we've always had as an ABC for a little extra safety in ensuring we implement the required methods when they're made. The drawback isn't much of one because of the level of integration we already need for code that produces these classes (e.g. registering the functions which produce them).

On the other hand, `BinaryLoadable` is only written to serve as part of the type hint for `load_mapping`.

PR #2006, opened 2022-08-15
Emulation init across months

frodre:
This PR includes the experiment script for running the emulation across all restarts in our generated "v4" training data.

The PR also removes the `climate-sim-pool` from the toleration list in the argo workflow.  We currently get OOM errors on that cluster when running online emulation experiments.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/9ca8365f-25a8-463d-9455-e02cadb5f4b8\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


projects/microphysics/argo/argo.yaml
```diff
@@ -150,7 +146,7 @@ spec:
       value: "med-sim-pool"
       effect: "NoSchedule"
     container:
-      image: "us.gcr.io/vcm-ml/fv3net:latest"
+      image: "us.gcr.io/vcm-ml/fv3net:{{workflow.parameters.image_tag}}"
```
frodre, line 149: Changed when debugging the performance regression (model overwrite).  I left because it seems more consistent, but I understand that it gets annoying when testing new prog-run or training images and am fine reverting.

projects/microphysics/configs/gscond-only.yaml
```diff
@@ -1,13 +1,13 @@
 # submit with
-# python3 argo/end_to_end.py train/precpd-only.yaml
+# python3 argo/end_to_end.py configs/gscond-only.yaml
 kind: prognostic
-experiment: pscaling-test
-name: gscond-only-tscale-zcloud-retrained-v1
+experiment: gscond-precpd
```
frodre, line 4: Setting this script back to the better performing old model vs. the more recent retrained version.

projects/microphysics/configs/gscond-only.yaml
```diff
@@ -267,18 +267,21 @@ config:
...
       classifier_path: gs://vcm-ml-experiments/microphysics-emulation/2022-06-09/gscond-classifier-v1/model.tf
-      path: gs://vcm-ml-experiments/microphysics-emulation/2022-07-26/gscond-only-dense-local-tscale-retrain-v1/model.tf
+      path: gs://vcm-ml-experiments/microphysics-emulation/2022-05-12/gscond-only-tscale-dense-local-41b1c1-v1/model.tf
       enforce_conservative: true
       gscond_cloud_conservative: true
       mask_emulator_levels:
         air_temperature_after_gscond:
           start: 74
           stop: null
+          fill_value: air_temperature_input
```
frodre, line 277: Added the upper zero in since it performs similarly to the fortran masking.

projects/microphysics/argo/argo.yaml
nbren12: I often push prog run images from my VM and submit experiments, so would prefer to revert this change...Maybe 95% of my experiments lately have been like this. I'm fine with adding more argo parameters as long as the default is "latest".

projects/microphysics/experiments/prog-run-all-months.py
```diff
@@ -0,0 +1,25 @@
+import sys
```
nbren12, line 1: It would be nice to move this logic into argo (and//or add a MultiInitJob) too to keep the front-end "declarative". This is not needed now, but if we start doing this kind of experiments a few more times, I can see us running into trouble.

I've found it hard to reproduce results submitted by these .py scripts... which is why I moved to the `query-wandb.py reproduce` and `end_to_end.py` workflow. It's  hard to map runs in wandb to scripts in this folder. 

I've also thought about adding a templating layer to end_to_end.py to enable this kind of thing.

nbren12 (approved):
LGTM. But I think we should move away from custom submission scripts to make the configuration more declarative.

projects/microphysics/experiments/prog-run-all-months.py
frodre: That makes sense, I've mostly been using the configs as templates and then checking reproduce or git file history to do reruns.

PR #2007, opened 2022-08-15
Prognostic run: radiation port available in version of image

brianhenn:
Packages the python radiation port in `external/radiation` and installs it prognostic run image.

Added public API:
- creates a package `radiation` that can be pip installed from `external/radiation`
- makes it available in the  `prognostic_run_radiation` image

Significant internal changes:
- Makes the radiation port a package with a `setup.py`, and separates the validation test from the radiation port code. 

- [X] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/74471129-dc7b-451f-a62b-ecf4d45cf81d/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/radiation/radiation/radiation_gases.py
```diff
@@ -1,7 +1,8 @@
 import numpy as np
 import sys
-from phys_const import con_pi
+from .phys_const import con_pi
 
+# TODO: can we delete this?
 sys.path.insert(0, "..")
```
YakelynRJ, line 6: Yup. I think we can remove that.

YakelynRJ (commented):
LGTM. 

Could you please remove the print statement: print(f"Using file {aeros_file}")
inside getdata.py (aerosol(forcing_dir: str) ). Is not needed would be nice if you can remove that.

external/radiation/tests/test_driver.py
```diff
@@ -1,13 +1,18 @@
-from config import FORCING_DIR, LOOKUP_DIR, nlay
```
nbren12, line 1: Does pytest end up calling this when it collects tests? would be nice to wrap in a function so it acts like a normal pytest test.

external/radiation/setup.py
```diff
@@ -0,0 +1,14 @@
...
+setup(
+    description="A python port of the GFS radiation scheme",
+    name="radiation",
+    author="The Allen Institute for Artificial Intelligence",
+    author_email="noahb@allenai.org",
+    license="BSD license",
+    version="0.1.0",
+    python_requires=">=3.8.0",
+    packages=find_packages(),
+    install_requires=["numba>=0.54.1", "xarray", "numpy", "netcdf4"],
```
nbren12, line 13: is numba already in `constraints.txt`? if not then you should run make lock_pip.

docker/prognostic_run/Dockerfile
```diff
@@ -79,4 +79,13 @@ FROM bld AS prognostic-run
...
 ARG COMMIT_SHA_ARG
-ENV COMMIT_SHA=$COMMIT_SHA_ARG
\ No newline at end of file
+ENV COMMIT_SHA=$COMMIT_SHA_ARG
+
+
+###############################################################################
+#  additionally, with radiation port available
+###############################################################################
+FROM prognostic-run AS prognostic-run-radiation
```
nbren12, line 88: What is the downside of including it in the normal image? It could be confusing to have too many prognostic_run images and related makefile targets.

nbren12 (approved):
Thanks! I didn't review too closely...and mostly focused on the dockerfile and setup.py.

I would just install it in the main prognostic run image and revert your makefile changes tbh....there's a fair amount of complexity there.

for bonus points, you could consider wrapping test_driver.py to behave like a traditional pytest test.

docker/prognostic_run/Dockerfile
brianhenn: Not much, since there aren't really any requirements differences. I'll make them one and the same.

external/radiation/tests/test_driver.py
brianhenn: It does, assuming we add pytest to that nix environment. We can clean up the test interface.

external/radiation/setup.py
brianhenn: It is already there, set at `==0.53.1`. The radiation nix env is using `0.54.1`. I was tempted to just check that the port runs OK at `0.53.1` and leave the constraints alone, but I guess I'll add the constraint from nix into the fv3net and prognostic run requirements.

nbren12: Oh. I think it's probably ok. One trick to enforce a constraint without explicitly adding it someplace is to set it in constraints.txt. `make lock_pip` will only change it if it needs to.

PR #2010, opened 2022-08-18
improve query-wandb.py run ergonomics

nbren12:
the query-wandb.py tool is useful to avoid button clicking in the wandb webapp.

A common use-case is get the link for a given run or list the runs
within a wandb group. This was laborious to do with the json outputs of
query-wandb.py runs

* output plaintext (not json) by default (-o json) will still do json
this makes it easier to get a link from a run
* add --group flag to filter for matching groups
* add --job-type flag

Example:
```
$ query-wandb.py runs --group combined-gscond-dense-local-v2-v1 --job-type piggy
combined-gscond-dense-local-v2-v1-online piggy-back cool-shape-3938 https://wandb.ai/ai2cm/microphysics-emulation/runs/3t7bhsyx
combined-gscond-dense-local-v2-v1-offline piggy-back azure-galaxy-3936 https://wandb.ai/ai2cm/microphysics-emulation/runs/1sgufb9k
```

### DIFF
--- a/projects/microphysics/scripts/query-wandb.py
+++ b/projects/microphysics/scripts/query-wandb.py
@@ -11,6 +11,21 @@ app = typer.Typer()
 PROJECT = "ai2cm/microphysics-emulation"
 
 
+def _get_runs(api, experiment, group, job_type):
+    if experiment:
+        filters = filters = {"tags": experiment}
+    else:
+        filters = {}
+
+    if group:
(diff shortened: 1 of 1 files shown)


AnnaKwa (approved):
LGTM! In the future it could be nice to have some of the more general query commands in a more visible place in fv3net.

PR #2011, opened 2022-08-19
Add encoder-decoder architecture for CycleGAN training

mcgibbon:
This PR adds an "autoencoder" training function which trains the generator portion of a CycleGAN as an autoencoder. It is tested on synthetic data consisting of sine waves of varying period and amplitude.

Added public API:
- added `CycleGANLoader` as a `TFDatasetLoader`, will load multiple other `TFDatasetLoader`s and include them as tuples within its entries
- added "autoencoder" training type with AutoencoderHyperparameters
- added PytorchPredictor class for pytorch-based non-autoregressive models

Refactored public API:
- `fv3fit.set_random_seed` will now also set the random seed for pytorch
- the zarr tfdataset loader will no longer attempt to stack dimensions, meaning it only works if no dimensions other than those listed in `unstacked_dims` are present
- `WindowedZarrLoader` now returns batched data, with a configurable batch size
- Fixed bug where WindowedZarrLoader could only load data once, and then had no entries remaining on subsequent epochs
- Renamed PytorchModel to PytorchAutoregressor
- Registered PytorchAutoregressor with fv3fit's IO routines
- Renamed fv3fit's IO registry routine `get_name` to `get_dumpable_name`

Significant internal changes:
- Added model building code for CycleGAN in fv3fit

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/443c4700-be93-4297-b41e-97f40692a177/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -27,16 +27,22 @@ def __post_init__(self):
...
         return data
 
 
 def open_zarr_using_filecache(url: str):
     cachedir = tempfile.mkdtemp()
     return xr.open_zarr(
-        "filecache::" + url, storage_options={"filecache": {"cache_storage": cachedir}}
+        "filecache::" + url,
+        storage_options={"filecache": {"cache_storage": cachedir}},
+        decode_times=False,
```
mcgibbon, line 45: Time can't be retrieved as a variable anyways, needed this to load 5-year long data.

external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -56,6 +62,45 @@ def get_n_windows(n_times: int, window_size: int) -> int:
...
+                for domain_config in d["domain_configs"]
+            ]
+        return CycleGANLoader(domain_configs=domain_configs)
+
+
+RECURSING = False
+
+
+@contextlib.contextmanager
+def prevent_recursion():
```
mcgibbon, line 94: Needed so when checking if something is a CycleGANLoader we don't check if it's a CycleGANLoader and in order to do so, check that it's a CycleGANLoader, etc.

external/fv3fit/fv3fit/pytorch/training_loop.py
```diff
@@ -52,14 +52,17 @@ def fit_loop(
...
         train_data = tfds.as_numpy(train_data)
-        validation_data = validation_data.unbatch()
-        n_validation = sequence_size(validation_data)
-        validation_state = (
-            torch.as_tensor(next(iter(validation_data.batch(n_validation))).numpy())
-            .float()
-            .to(DEVICE)
-        )
-        min_val_loss = np.inf
+        if validation_data is not None:
```
mcgibbon, line 55: This was not previously tested without validation data (still isn't, but it's not within the scope of this PR).

external/fv3fit/fv3fit/train.py
```diff
@@ -181,7 +181,7 @@ def main(args, unknown_args=None):
     os.makedirs("artifacts", exist_ok=True)
     logging.basicConfig(
         level=logging.INFO,
-        format="%(asctime)s [%(levelname)s] %(message)s",
+        format="%(asctime)s [%(levelname)s] %(filename)s::L%(lineno)d : %(message)s",
```
mcgibbon, line 184: Is this OK to change? Should I revert it? I really like being able to see where log lines are coming from.

brianhenn (commented):
Leaving my comments so far, have really only looked at the autoencoder architecture part so far.

external/fv3fit/fv3fit/_shared/io.py
```diff
@@ -18,17 +18,20 @@ class _Register:
 
     def __init__(self) -> None:
         self._model_types: MutableMapping[str, Type[Predictor]] = {}
+        self._dump_types: MutableMapping[str, Type[Dumpable]] = {}
 
-    def __call__(self, name: str) -> Callable[[Type[Predictor]], Type[Predictor]]:
-        if name in self._model_types:
+    def __call__(self, name: str) -> Callable[[Type[Dumpable]], Type[Dumpable]]:
```
brianhenn, line 23: Is this added complexity needed? Given the current definitions of `Predictor`, `Loadable`, and `Dumpable`, it seems like the intuitive thing to be registering is `Predictor`. What's the value of registering something that can only be dumped but not loaded?

external/fv3fit/tests/training/test_autoencoder.py
```diff
@@ -0,0 +1,174 @@
...
+from fv3fit.pytorch.cyclegan import AutoencoderHyperparameters, train_autoencoder
+from fv3fit.pytorch.cyclegan.train import TrainingConfig
+from fv3fit.tfdataset import iterable_to_tfdataset
+import collections
+import os
+import fv3fit.pytorch
+import fv3fit
+
+
+def get_tfdataset(nsamples, nbatch, ntime, nx, ny, nz):
```
brianhenn, line 13: Can this get a docstring with maybe some equations or visual description of what's happening here? (You do this in the test itself below but more would be helpful.) 

The setup of the test data is pretty complicated. Does the test actually need to be this complex to demonstrate that the autoencoder works?

external/fv3fit/tests/training/test_autoencoder.py
```diff
@@ -0,0 +1,174 @@
...
+    grid_x = grid_x[None, None, None, :, :, None]
+    grid_y = grid_y[None, None, None, :, :, None]
+
+    def sample_iterator():
+        # creates a timeseries where each time is the negation of time before it
+        for _ in range(nsamples):
+            ax = np.random.uniform(0.1, 1.5, size=(nbatch, 1, ntile, nz))[
+                :, :, :, None, None, :
+            ]
+            bx = np.random.uniform(8, 16, size=(nbatch, 1, ntile, nz))[
```
brianhenn, line 28: Seems like the period should depend on the grid size instead of being hardcoded?

external/fv3fit/tests/training/test_autoencoder.py
```diff
@@ -0,0 +1,174 @@
...
+    bias = predicted.isel(time=1) - reference.isel(time=1)
+    mean_bias: xr.Dataset = bias.mean()
+    mse: xr.Dataset = (bias ** 2).mean() ** 0.5
+    for varname in state_variables:
+        assert np.abs(mean_bias[varname]) < 0.1
+        assert mse[varname] < 0.1
+
+
+def test_autoencoder_overfit(tmpdir):
+    fv3fit.set_random_seed(0)
```
brianhenn, line 132: Similar request re docstrings here, particularly highlighting how this test is different from the one above it.

external/fv3fit/fv3fit/pytorch/cyclegan/train.py
```diff
@@ -0,0 +1,145 @@
...
+    loss: LossConfig = LossConfig(loss_type="mse")
+    noise_amount: float = 0.5
+
+    @property
+    def variables(self):
+        return tuple(self.state_variables)
+
+
+@curry
+def define_noisy_input(data: tf.Tensor, stdev=0.1) -> Tuple[tf.Tensor, tf.Tensor]:
```
brianhenn, line 58: Is the noisy input really a feature of the autoencoder, or is this more part of the test framework you've chosen? Feels like the latter, in which case the code related to it should move there.

external/fv3fit/tests/training/test_autoencoder.py
mcgibbon: Sort of, what happened was I chose periods that are only as large as they need to be in order to be "resolved enough", and then I chose grid sizes to suit those periods. If the grid size is too small to hold a whole wave, that can still be fine, possibly. But I really wouldn't want someone to put `nx=4` and then have the code say the period is `nx/2` (which would give constant data).

external/fv3fit/tests/training/test_autoencoder.py
mcgibbon: Will add.

external/fv3fit/fv3fit/pytorch/cyclegan/network.py
```diff
@@ -0,0 +1,339 @@
...
+def no_activation():
+    return nn.Identity()
+
+
+class ConvolutionFactory(Protocol):
+    def __call__(self, in_channels: int, out_channels: int) -> nn.Module:
+        ...
+
+
+class ConvolutionFactoryFactory(Protocol):
```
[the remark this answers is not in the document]
mcgibbon: OK I was able to remove one level of factory by adding curry, at the expense of some type safety.

external/fv3fit/fv3fit/pytorch/cyclegan/train.py
```diff
@@ -0,0 +1,145 @@
...
+
+from fv3fit._shared import register_training_function
+from typing import (
+    List,
+    Optional,
+    Tuple,
+)
+from fv3fit.tfdataset import ensure_nd, apply_to_mapping
+from .network import Generator
+from fv3fit.pytorch.graph.train import (
```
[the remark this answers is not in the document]
mcgibbon: I moved the first two, I'd like to leave `get_Xy_dataset` for now because I've made significant changes to it in the follow-on PR and want to avoid merge conflicts. I will clean this up in that PR.

external/fv3fit/fv3fit/pytorch/cyclegan/train.py
```diff
@@ -0,0 +1,145 @@
...
+    Train a denoising autoencoder for cubed sphere data.
+
+    Args:
+        hyperparameters: configuration for training
+        train_batches: training data, as a dataset of Mapping[str, tf.Tensor]
+            where each tensor has dimensions [sample, time, tile, x, y(, z)]
+        validation_batches: validation data, as a dataset of Mapping[str, tf.Tensor]
+            where each tensor has dimensions [sample, time, tile, x, y(, z)]
+    """
+    train_batches = train_batches.map(apply_to_mapping(ensure_nd(6)))
```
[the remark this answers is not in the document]
mcgibbon: Done.

external/fv3fit/fv3fit/pytorch/cyclegan/train.py
```diff
@@ -0,0 +1,145 @@
...
+        val_state = get_state(data=validation_batches)
+    else:
+        val_state = None
+
+    train_state = get_state(data=train_batches)
+
+    train_model = build_model(
+        hyperparameters.generator, n_state=next(iter(train_state)).shape[-1]
+    )
+    print(train_model)
```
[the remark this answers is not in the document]
mcgibbon: This is very useful to have (pytorch model prints are really descriptive), so I updated it to a `logging.debug` call.

external/fv3fit/fv3fit/tfdataset.py
```diff
@@ -141,6 +142,44 @@ def process_shape(shape):
     )
 
 
+def generator_to_tfdataset(
```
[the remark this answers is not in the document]
mcgibbon: Done.

external/fv3fit/tests/training/test_autoencoder.py
mcgibbon: Added a description to the function.

external/fv3fit/fv3fit/_shared/io.py
mcgibbon: Reverted changes to this file and instead set it to dump/load a new Reloadable ABC inherited by Predictor.

brianhenn: Yeah, as discussed a `Predictor` is something that can be dumped/loaded and that can `.predict()`. Beyond that I personally didn't expect specific behavior of the predict method.

external/fv3fit/fv3fit/pytorch/cyclegan/train.py
brianhenn: OK, that's fine then.

external/fv3fit/tests/training/test_autoencoder.py
brianhenn: OK that sounds fine for now then.

EbrahimNabizadeh (approved):
I mainly looked at the training and encoder-decoder parts. Overall it looks good to me, and I just had a few comments.

external/fv3fit/fv3fit/_shared/predictor.py
```diff
@@ -33,7 +33,15 @@ def load(cls: Type[L], path: str) -> L:
         ...
 
 
-class Predictor(Dumpable):
+class Reloadable(Dumpable, Loadable):
```
brianhenn, line 36: This is an improvement, though maybe an even easier way is to modify the `predict` method of the `Predictor` so that it can include the graph network? Even if that model's method won't work in the prognostic run. Up to you.

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: OK so it turns out I don't actually need this, so I deleted it.

It turns out we don't check that the config itself is a CycleGANLoader to check that it's a CycleGANLoader, we check whether it's sub-members are CycleGANLoaders. Eventually we'll run out of sub-members so this can never be an infinite loop.

brianhenn (approved):
Approving this subject to Ibrahim's comments and the optional below.

external/fv3fit/fv3fit/_shared/predictor.py
mcgibbon: I think this way makes it more clear that we're assuming the loaded thing is a Predictor without type-safety. Autoregressive and "predictive" models are different enough to warrant their own concept, and then I'm also probably going to want to have a different API for CycleGAN. Something that does all three of these is probably doing too much.

external/fv3fit/fv3fit/pytorch/cyclegan/network.py
```diff
@@ -0,0 +1,339 @@
...
+            in_channels=min_filters,
+        )
+
+        self._out_conv = flat_convolution(kernel_size=3)(
+            in_channels=2 * min_filters, out_channels=channels
+        )
+
+    def forward(self, inputs):
+        # data will have channels last, model requires channels first
+        inputs = inputs.permute(0, 3, 1, 2)
```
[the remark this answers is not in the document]
mcgibbon: Done.

PR #2012, opened 2022-08-19
Prognostic run: radiation piggyback

brianhenn:
Allows for running the python radiation scheme port in the prognostic run. See [notebook](https://github.com/ai2cm/explore/blob/master/brianh/2022-08-26-validate-rad-port/2022-08-26-validate-rad-port.ipynb) showing that the port's radiative fluxes closely match the Fortran radiative fluxes. 

Added public API:
- Adds an optional `radiation_scheme` entry to the prognostic run user configuration, typed by a `RadiationStepperConfig` class, which instantiates a `RadiationStepper` for piggyback computations in the prognostic run. The config also allows for specifying an optional ML model or prescribe to provide the input states for the radiation scheme. 
- Adds `Radiation` wrapper and `RadiationConfig` config classes to  `radiation` to handle generating the needed flags and inputs, updating, and running the radiation scheme. 

Refactored public API:
- Renames the `getdata` module of `radiation` to `io` and gives the io functions clearer names. Makes some minor additions and changes to the `radiation` package public API to support getting data needed to run the port.

Significant internal changes:
- Physics calls in the runfile loop are split into pre-radiation, radiation, and post-radiation steps.
- Bumps `fv3gfs-fortran` submodule to use a version of the wrapper that can call radiation physics separately.
- `radiation` adds an internal `preprocessing` module for generating necessary inputs for running the radiation driver

[x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/430351fc-ab15-42a3-81ca-4d6892fc601d/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -133,6 +136,41 @@ def add_tendency(state: Any, tendency: State, dt: float) -> Tuple[State, State]:
...
+    ),
+    forcing_data_path: str = "gs://vcm-fv3gfs-serialized-regression-data/physics/forcing/*",  # noqa: 501
+    lookup_local_dir: str = "./data/lookup",
+    forcing_local_dir: str = "./data/forcing",
+) -> None:
+    """Gets lookup tables and forcing needed for the radiation scheme. TODO: Make lookup
+    data part of writing a run directory; make scheme able to read existing forcing.
+    """
+    os.makedirs(lookup_local_dir)
+    copy_cmd = ["gsutil", "cp", lookup_data_path, lookup_local_dir]
```
nbren12, line 151: Can you avoid shelling out to system utilities in this script? `tarfile` is a python built in library, and you can use fsspec for the downloading.

external/radiation/radiation/radiation_astronomy.py
```diff
@@ -317,8 +317,8 @@ def sol_update(self, jdate, kyear, deltsw, deltim, lsol_chg, me, solar_data):
         if me == 0:
             print(
                 "for cosz calculations: nswr,deltim,deltsw,dtswh =",
-                f"{nswr[0]}, {deltim[0]}, {deltsw[0]}, {dtswh[0]}, anginc, nstp =",
```
nbren12, line 320: Should this debugging change be reverted?

external/radiation/radiation/getdata.py
```diff
@@ -292,6 +293,20 @@ def sw(lookup_dir: str):
     return sw_dict
 
 
+def sigma(restart_dir, p_ref=101325.0):
```
nbren12, line 296: Rename to highlight I/O?

```suggestion
def load_hybrid_coordinate(restart_dir, p_ref=101325.0):
```

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -217,6 +255,7 @@ def __init__(
         self._log_debug(f"States to output: {self._states_to_output}")
         self._prephysics_stepper = self._get_prephysics_stepper(config, hydrostatic)
         self._postphysics_stepper = self._get_postphysics_stepper(config, hydrostatic)
+        self._radiation_wrapper = self._get_radiation_wrapper(config, namelist)
```
nbren12, line 258: Why is it a "wrapper" rather than a "stepper"? Can you make it a stepper?

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -334,6 +373,19 @@ def _open_model(self, ml_config: MachineLearningConfig, step: str):
         self._log_info("Model Loaded")
         return model
 
+    def _get_radiation_wrapper(self, config: UserConfig, namelist: f90nml.Namelist):
+        if config.radiation_scheme and config.radiation_scheme.kind == "python":
+            if self.rank == 0:
+                _download_radiation_assets()
+            MPI.COMM_WORLD.barrier()
+            wrapper: RadiationWrapper = RadiationWrapper.from_config(
```
nbren12, line 381: Should `from_config` call `rad_init`? Would make the API tidier.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -344,10 +396,36 @@ def _step_dynamics(self) -> Diagnostics:
...
-        self._log_debug(f"Physics Step (compute)")
-        self._fv3gfs.compute_physics()
-        # no diagnostics are computed by default
+    def _step_pre_radiation_physics(self) -> Diagnostics:
+        self._log_debug(f"Pre-radiation Physics Step")
+        self._fv3gfs.step_pre_radiation()
+        if self._radiation_wrapper is not None:
+            self._radiation_wrapper.rad_update(self.time, self._timestep)
+        return {
+            f"{name}_pre_radiation": self._state[name]
```
nbren12, line 405: magic string here...could be tricky to track down this later on.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -344,10 +396,36 @@ def _step_dynamics(self) -> Diagnostics:
...
+        if self._radiation_wrapper is not None:
+            diagnostics: Diagnostics = self._radiation_wrapper.rad_compute(
+                self._state,
+                self._fv3gfs.get_tracer_metadata(),
+                self.time,
+                self._timestep,
+                self.rank,
+            )
+            for name, diag in diagnostics.items():
+                self._log_info(f"{name}: {diag}")
```
nbren12, line 420: Do you need this ad-hoc logging? Seems a bit messy for all the steps to print things to the console in potentially inconsistent ways.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+
+@dataclasses.dataclass
+class RadiationWrapperConfig:
+    """"""
+
+    kind: Literal["python"]
+    input_model: Optional[MachineLearningConfig] = None
+
+
+class RadiationWrapper:
```
nbren12, line 80: Can you make this a Stepper and hide some of the idiosyncratic and redundantly named methods e.g. `rad_init`, `rad_update` and `rad_compute`? Do you really need separate concepts for `compute` and `update`? That is the main departure from the Stepper API.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+    @classmethod
+    def from_config(cls, config: RadiationWrapperConfig) -> "RadiationWrapper":
+        if config.input_model:
+            model: Optional[MultiModelAdapter] = open_model(config.input_model)
+        else:
+            model = None
+        driver = RadiationDriver()
+        return cls(model, driver)
+
+    def rad_init(
```
nbren12, line 97: Maybe call this in the `__init__` and make it a private method.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+        tracer_metadata: Mapping[str, Any],
+        time: cftime.DatetimeJulian,
+        dt_atmos: float,
+        rank: int,
+        lookup_dir: str = "./data/lookup",
+    ):
+        if self._rad_config is None:
+            raise ValueError(
+                "Radiation driver not initialized. `.rad_init` must be called "
+                "before `.rad_compute`."
```
nbren12, line 246: Maybe just call `rad_init` from the constructor and remove this error message. Is there some use case for the lazy loading?
```suggestion
```

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+]
+
+
+def _rename_out(
+    out: Tuple[Mapping[str, np.ndarray], ...], out_names: Sequence[str] = OUT_NAMES
+) -> Mapping[str, np.ndarray]:
+    radtendout, diagout, _ = out
+    renamed = {}
+    for name in out_names:
+        obj = radtendout if "surface" in name else diagout
```
nbren12, line 478: This parsing logic seems a bit brittle...maybe just write an explicit list/mapping someplace. Bonus points if it is something that can be static typed.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+        random_numbers = getdata.random(ncolumns, nz, NGPTSW, NGPTLW)
+        lw_lookup = getdata.lw(lookup_dir)
+        sw_lookup = getdata.sw(lookup_dir)
+        out = self._driver._GFS_radiation_driver(
+            model, statein, sfcprop, grid, random_numbers, lw_lookup, sw_lookup
+        )
+        return _unstack(_rename_out(out), coords)
+
+
+class UpdatedState(DerivedFV3State):
```
nbren12, line 265: This is just a lazily merged dict object, but the name makes me a bit scared since it implies some other type of state is being added to the fv3net prognostic run. Do you need this object? If so, does it need to be a "DerivedFV3State" or can it be a read-only `Mapping[str, xr.DataArray]`?

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
```diff
@@ -0,0 +1,484 @@
...
+    def __init__(self, state: State, predictions: State):
+        self._state = state
+        self._predictions = predictions
+
+    def __getitem__(self, key: Hashable) -> xr.DataArray:
+        return self._predictions.get(key, self._state[key])
```
nbren12, line 271: This will fail if `key` is not in `self._state` even if it is in `self._predictions`. Is that what you want?

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -334,6 +373,19 @@ def _open_model(self, ml_config: MachineLearningConfig, step: str):
         self._log_info("Model Loaded")
         return model
 
+    def _get_radiation_wrapper(self, config: UserConfig, namelist: f90nml.Namelist):
+        if config.radiation_scheme and config.radiation_scheme.kind == "python":
+            if self.rank == 0:
+                _download_radiation_assets()
```
nbren12, line 379: Should this be in RadiationWrapper.from_config? The TimeLoop seems to have to know a lot about RadiationWrapper (methods, input data,etc). I bet you could decouple them much more.

nbren12 (approved):
Thanks for putting this together @brianhenn. Most of my comments are on high level APIs/names. It would be great if you could introduce fewer new concepts to the prognostic run.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -133,6 +136,41 @@ def add_tendency(state: Any, tendency: State, dt: float) -> Tuple[State, State]:
     return updated, tendency_filled_frac  # type: ignore
 
 
+def _download_radiation_assets(
+    lookup_data_path: str = (
+        "gs://vcm-fv3gfs-serialized-regression-data/physics/lookupdata/lookup.tar.gz"  # noqa: E501
+    ),
+    forcing_data_path: str = "gs://vcm-fv3gfs-serialized-regression-data/physics/forcing/*",  # noqa: 501
```
nbren12, line 143: Are you concerend about this being inconsistent with the prog run? Would be nice to get it from the wrapper.

nbren12:
I'd also consider writing an integration test for the RadiationWrapper. I realize it is pretty data hungry, but I think it will be a good heuristic for some of the API design issues.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Sure

external/radiation/radiation/radiation_astronomy.py
brianhenn: It's not debugging ... the print statement (and apparently only the print statement) expects that these scalars actually have a singleton dimension, such that it fails if they don't. This is odd behavior that doesn't affect the port itself so seems easiest to just get rid of it.

external/radiation/radiation/getdata.py
brianhenn: Hmm, my thinking was that the functions in `getdata` would be called in practice by `getdata.<func>` which accomplishes what you're suggesting. Alternatively we could rename all of the functions in the module and call them without the module name.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Thoughts on how to get around this? We're limited by the fact that all diags must be pooled in the `diagnostics` dict over the course of the substeps, so that if we want to get the model states at different substeps separately (the use case here) there's a name clash. It seems like we need handling of a `substep` xarray dimension on the diagnostics which can then be parsed by the diagnostics manager to put it in the right zarr at the end of the model timestep. That's a lot of new diagnostics handling capabilities though, so I temporarily punted and just suffixed the diagnostics, unsatisfying as that is.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Will delete this.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
brianhenn: Probably not. I think they could be combined into the stepper `__call__` as you suggest.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Yeah, it could be moved there. 

Some of this asset downloading could be avoided entirely if the radiation port could read the forcing files as the text files that are written into the run directory anyway for FV3GFS. Unfortunately the port changed its i/o to read them from netcdfs instead, so now we're effectively putting them into the run directory twice. This is an obvious todo for the port itself.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Yeah, as mentioned above the port i/o would need to be rewritten to allow this, but it would be nice.

brianhenn:
> I'd also consider writing an integration test for the RadiationWrapper. I realize it is pretty data hungry, but I think the ability to write such a test in <10 lines would be a good heuristic for some of the API design issues.

Agreed. I was originally hoping that this could come in the form of a validation test against the Fortran fluxes (such that we could delete the old validation test against serialbox data), but I think the random numbers need to be resolved before we can reliably do that. I think instead an integration/regression test of the radiation wrapper/stepper is doable, much like the regression tests we have already for ML and nudging.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
brianhenn: I'm tempted to take the capability of running the port with ML inputs out of this PR since it was a stub anyway and we're not emphasizing that route right now. If so will delete this.

workflows/prognostic_c48_run/runtime/loop.py
brianhenn: Yeah, it can be a `Stepper` -- will do

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
brianhenn: Ended up keeping this capability and adopting your suggestion.

external/radiation/radiation/radiation_astronomy.py
nbren12: I don't see any changes responding to this comment, but the tests are passing, so I'll resolve it.

external/radiation/radiation/getdata.py
nbren12: That is a good naming approach.

external/radiation/tests/test_preprocessing.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.mark.parametrize(
+    ["in_dict", "flattened_dict", "raises_type_error"],
+    [
+        pytest.param({"a": ARR}, {"a": ARR}, False, id="flat"),
+        pytest.param({"a": {"b": ARR}}, {"a_b": ARR}, False, id="nested"),
+        pytest.param({"a": 3}, None, True, id="type_error"),
+    ],
+)
+def test__flatten(in_dict, flattened_dict, raises_type_error):
```
nbren12, line 16: Nit. It would be a bit less complex and readable to split the error raising test into a new test function.

workflows/prognostic_c48_run/runtime/steppers/radiation.py
```diff
@@ -0,0 +1,221 @@
...
+        # idat is supposed to be model initalization time but is unused w/ current flags
+        idat = np.array(
+            [time.year, time.month, time.day, 0, time.hour, time.minute, time.second, 0]
+        )
+        jdat = np.array(
+            [time.year, time.month, time.day, 0, time.hour, time.minute, time.second, 0]
+        )
+        fhswr = np.array(float(self._rad_config["fhswr"]))
+        dt_atmos = np.array(float(dt_atmos))
+        aerosol_data = io.load_aerosol(self._forcing_local_dir)
```
nbren12, line 134: Looks like this I/O happens every time step? Should you move this to the initialization?

workflows/prognostic_c48_run/runtime/steppers/radiation.py
```diff
@@ -0,0 +1,221 @@
...
+        sw_lookup = io.load_sw(self._lookup_local_dir)
+        out = self._driver._GFS_radiation_driver(
+            model, statein, sfcprop, grid, random_numbers, lw_lookup, sw_lookup
+        )
+        out = preprocessing.postprocess_out(out)
+        return preprocessing.unstack(out, coords)
+
+    def _generate_inputs(self, state: State, time: cftime.DatetimeJulian) -> State:
+        if self._input_generator is not None:
+            _, _, state_updates = self._input_generator(time, state)
```
nbren12, line 183: I'm finding it hard to follow the indirection here. First, I'd recommend inlining `_generate_inputs` where it is called and deleting this function. Also, what is the _input_generator and is it really necessary to have the  complex/untested`MergedState` class? 

The pattern seems roughly similar to this https://github.com/ai2cm/fv3net/blob/c4e7f17d25bd97c83ba291edd8a6ef7afc451510/workflows/prognostic_c48_run/runtime/transformers/core.py#L35 and perhaps more elegantly handled as function decorator.

workflows/prognostic_c48_run/runtime/loop.py
nbren12: I'd probably try to move this kind of renaming up the call-stack they way we do with `.e.g self.monitor` in `__iter__`. just feels odd for a function to know its own name sometimes.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
nbren12: I'm a bit nervous about the doubling down on this class. Let's VC about why it is needed?

external/radiation/radiation/preprocessing.py
```diff
@@ -0,0 +1,257 @@
...
+    "htrsw": "total_sky_shortwave_heating_rate_python",
+    "swhc": "clear_sky_shortwave_heating_rate_python",
+}
+
+State = MutableMapping[Hashable, xr.DataArray]
+Diagnostics = MutableMapping[Hashable, xr.DataArray]
+
+
+def model(
+    rad_config: MutableMapping[Hashable, Any],
```
nbren12, line 70: Does it need to mutable? Can you make it a TypeDict or dataclass...which would be valuable documentation.

nbren12 (commented):
Thanks for the changes Brian. The interactions with loop.py are much improved, though I think more logic should be moved to external/radiation to facilitate reuse outside the prog run and would like to explore if the PR can introduce fewer new concepts and surface area for future bugs to grow (e.g. MergedState).

external/radiation/radiation/preprocessing.py
```diff
@@ -0,0 +1,257 @@
...
+    state: State, sfc_var_mapping: Mapping[str, str] = SFC_NAMES_IN_MAPPING
+) -> Mapping[str, np.ndarray]:
+    sfc = xr.Dataset({v: state[k] for k, v in list(sfc_var_mapping.items())})
+    # we only want the first of "hprime" variables
+    sfc["hprime"] = sfc["hprime"].isel({"orographic_variable": 0})
+    stacked_sfc = _stack(sfc)
+    return {name: stacked_sfc[name].values for name in stacked_sfc.data_vars}
+
+
+def postprocess_out(
```
nbren12, line 216: I'm not sure what is meant by flattening, post_processing etc. Maybe just flatten the call stack here and inline it in  the calling context. Do we really parsing/interpretation to do this? I think being verbose/explicit here would be advantageous.

workflows/prognostic_c48_run/tests/test_radiation.py
```diff
@@ -1,3 +1,124 @@
...
+        }
+    ]
+    return config
+
+
+def run_native(config, rundir):
+    with tempfile.NamedTemporaryFile("w") as f:
+        yaml.safe_dump(config, f)
+        fv3_script = "runfv3"
+        subprocess.check_call([fv3_script, "create", rundir, f.name])
```
nbren12, line 108: There is a python api for `runfv3 create`: https://github.com/ai2cm/fv3net/blob/master/workflows/prognostic_c48_run/runtime/segmented_run/api.py#L9.

And you could quickly add one for `run_native` which simply calls the `run_segment` command defined here: https://github.com/ai2cm/fv3net/blob/1d7d57867449c8fd68ced9324468fe3ebaa8edf5/workflows/prognostic_c48_run/runtime/segmented_run/run.py#L26

Also is the `create` step necessary. `run_native` is meant as an alternative to the `create/append` subcommands.

external/radiation/radiation/preprocessing.py
nbren12: this is optional...maybe for a future pr.

workflows/prognostic_c48_run/runtime/radiation_wrapper.py
nbren12: We pair programmed a path to delete this by implementing a `radiation.Radiation.input_variables` property and using it like this:
```


required_names = self._radiation.input_variables
inputs = {name: state[name] for name in required_names}
return {**inputs, **state_updates}
```

workflows/prognostic_c48_run/runtime/loop.py
nbren12: punting on this. It's a small amount of tech debt for now.

workflows/prognostic_c48_run/runtime/steppers/radiation.py
```diff
@@ -0,0 +1,221 @@
...
+        self,
+        driver: radiation.RadiationDriver,
+        rad_config: MutableMapping[Hashable, Any],
+        comm: MPI.COMM_WORLD,
+        timestep: float,
+        tracer_inds: Mapping[str, int],
+        input_generator: Optional[Union[PureMLStepper, Prescriber]],
+    ):
+        self._driver: radiation.RadiationDriver = driver
+        self._rad_config: MutableMapping[Hashable, Any] = rad_config
```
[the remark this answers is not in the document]
nbren12: in the interest of time...maybe just punt on this too.

workflows/prognostic_c48_run/runtime/steppers/radiation.py
brianhenn: As discussed will move this

workflows/prognostic_c48_run/runtime/steppers/radiation.py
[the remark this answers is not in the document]
brianhenn: I think the documentation needs to get written or it will be forgotten, so will include.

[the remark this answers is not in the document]
brianhenn: Done

workflows/prognostic_c48_run/runtime/steppers/radiation.py
brianhenn: Done

external/radiation/radiation/preprocessing.py
brianhenn: Dpme

external/radiation/radiation/preprocessing.py
brianhenn: The radiation driver returns a tortuously nested set of dicts. A future refactor could clean up its API and eliminate the need for this.

workflows/prognostic_c48_run/tests/test_radiation.py
brianhenn: Done

PR #2015, opened 2022-08-22
Mask extrapolated areas in PIRE pressure level diagnostics

AnnaKwa:
Areas in higher elevation (e.g. Psfc < 850 hPa) are extrapolated in the fortran diagnostics. The extrapolation is very different for the case where the model is run at coarse resolution versus if run at fine res and then coarsened. This update masks out extrapolated areas in the verification so that this issue does not affect the calculated biases.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/b6b557dc-242d-4f10-885f-a83c512cb7dc\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


oliverwm1 (approved):
LGTM

PR #2016, opened 2022-08-24
Limit net condensation for gscond emulator

frodre:
This PR limits the gscond emulator (when `enforce_conservative = True`) to obey available liquid/vapor quantities for net_condensation.

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/f6478d1a-ab0b-4fb4-bb91-5afac2bac19f/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/emulation/tests/test_zhao_carr.py
```diff
@@ -1,23 +1,42 @@
...
 import numpy as np
+from emulation.zhao_carr import (
+    _limit_net_condensation_conserving,
+    Input,
+    ice_water_flag,
+    latent_heat_phase_dependent,
+)
+
+
+def test__limit_net_condenstation():
```
AnnaKwa, line 10:
```suggestion
def test__limit_net_condensation():
```

AnnaKwa (approved):
LGTM! What is the difference between `enforce_conservative` (this limiter) versus `gscond_cloud_conservative`? is this limiter applied to the combined change after both gscond and prcp changes are predicted?

frodre:
Thanks!

>What is the difference between enforce_conservative (this limiter) versus gscond_cloud_conservative?

It's subtle, but `gscond_cloud_conservative` means the gscond clouds are predicted "conservatively" by using the inverse of the specific humidity prediction, instead of predicted by the ML.  We eventually moved into using only the specific humidity prediction for gscond and making sure T and qc were conservative, by updating clouds and temperature based on the humidity prediction (`enforce_conservative`).  `enforce_conservative` also makes sure if we do any masking adjustments that the updated states are all consistent.

In current practice they are always set to `True`.  It's just a byproduct of the very step-by-step trial and error nature of reaching our current scheme 😅

PR #2022, opened 2022-08-29
Add CycleGAN training routine

mcgibbon:
This PR adds a "cyclegan" training function which trains a CycleGAN model.

Added public API:
- added registered "cyclegan" training function to fv3fit with `fv3fit.pytorch.CycleGANHyperparameters` configuration
- added fv3fit.pytorch.CycleGAN Reloadable

Refactored public API:
- PytorchPredictor is now initialized with and contains as attribute a scalars mapping that uses str instead of Hashable for keys

Significant internal changes:
- Pytorch autoencoder architecture has changed, and is no longer a U-Net

- [x] Tests added


external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _init_targets(self, shape: Tuple[int, ...]):
+        self.target_real = torch.autograd.Variable(
+            torch.Tensor(shape).fill_(1.0).to(DEVICE), requires_grad=False
+        )
+        self.target_fake = torch.autograd.Variable(
+            torch.Tensor(shape).fill_(0.0).to(DEVICE), requires_grad=False
+        )
+
+    def evaluate_on_dataset(
```
mcgibbon, line 250: This function is a work in progress, it's unclear what metrics will be appropriate for automatically evaluating the quality of CycleGAN outputs given that there exists no paired output. I may entirely replace this function when I start using it on real data so I'd like to avoid iterating much at this stage.

external/fv3fit/fv3fit/pytorch/cyclegan/network.py
```diff
@@ -1,280 +0,0 @@
-from typing import Callable, Literal, Protocol, Union
```
mcgibbon, line 1: Given the code in this file didn't match the CycleGAN paper as accurately as the code now does, I would suggest not to compare the new code against this previous code but instead to compare it to the paper reference and [reference implementation by the lead author](https://github.com/junyanz/pytorch-CycleGAN-and-pix2pix/blob/master/models/cycle_gan_model.py).

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @property
+    def std(self) -> np.ndarray:
+        """
+        Standard deviation of the observed samples.
+        """
+        return np.sqrt(self._sum_squared / self._count - self.mean ** 2)
+
+
+def get_r2(predicted, target) -> float:
```
mcgibbon, line 183: Do we have or want a vcm shared implementation of this?

external/fv3fit/fv3fit/pytorch/graph/train.py
```diff
@@ -117,7 +114,7 @@ def train_graph_model(
     predictor = PytorchAutoregressor(
         state_variables=hyperparameters.state_variables,
         model=train_model,
-        scalers=cast(Mapping[Hashable, StandardScaler], scalers),
+        scalers=scalers,
```
mcgibbon, line 117: It was necessary to switch to str for variable names so that I can store separate "a_" and "b_" scalers for the input and outpoint domain in one scalers mapping. Technically this isn't needed by PytorchAutoregressor, but it shares dumping/loading code with the CycleGAN reloadable to reduce code duplication. There doesn't appear to be any loss of interoperability with xarray by doing so, all predict methods still use xr.Dataset.

external/fv3fit/fv3fit/pytorch/predict.py
```diff
@@ -382,7 +385,7 @@ def _unpack_tensor(
             else:
                 n_features = 1
                 var_data = data[..., i_feature]
-            var_data = scalers[varname].denormalize(var_data)
+            var_data = scalers[varname].denormalize(var_data.to("cpu").numpy())
```
mcgibbon, line 382: This was a bug before, but untested because it was never run on a GPU.

external/fv3fit/fv3fit/pytorch/system.py
```diff
@@ -1,3 +1,14 @@
...
 
-DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+if os.environ.get("TORCH_CPU_ONLY", False):
+    DEVICE = torch.device("cpu")
+else:
+    DEVICE = torch.device(
+        "cuda:0"
+        if torch.cuda.is_available()
+        else "mps"
+        if torch.backends.mps.is_available()
```
mcgibbon, line 12: Gotta go fast

external/fv3fit/fv3fit/tfdataset.py
```diff
@@ -27,11 +27,13 @@ def apply_to_mapping(
     return {name: tensor_func(tensor) for name, tensor in data.items()}
 
 
-@curry
 def apply_to_tuple(
```
mcgibbon, line 30: This was unused and broken before, I've added an external package test to confirm tuple entries are passed as *args to map functions.

mcgibbon:
Note there is no automated unit test of the full CycleGAN itself, I did this intentionally because a) CycleGAN cannot overfit on one sample, and b) true training of the CycleGAN is too slow to run on PR in our test suite.

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -0,0 +1,235 @@
...
+        super(FoldTileDimension, self).__init__()
+        self._wrapped = wrapped
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        x = inputs.reshape(-1, *inputs.shape[2:])
+        x = self._wrapped(x)
+        return x.reshape(inputs.shape[0], inputs.shape[1], *x.shape[1:])
+
+
+def single_tile_convolution(
```
mcgibbon, line 91: Abstracting convolution with a factory like this will allow me to swap single tile convolution for graph-based analogous implementations through configuration only, while leaving the model building code unchanged.

mcgibbon:
Superceded by #2032

PR #2024, opened 2022-08-30
Add synthetic data loaders to fv3fit

mcgibbon:
This PR adds synthetic data loader configurations for sin waves, square waves, and gaussian noise to fv3fit.

Refactored public API:
- fv3fit data loader configurations can now match new SyntheticWaves or SyntheticNoise APIs.

Significant internal changes:
- Fixed a bug in CycleGANLoader's `from_dict` initializer where it would ignore non-domain configuration (e.g. batch size)

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/c1152fe0-2239-453d-967b-2cc49ce89b79/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/data/synthetic.py
+++ b/external/fv3fit/fv3fit/data/synthetic.py
@@ -15,8 +15,8 @@ class SyntheticNoise(TFDatasetLoader):
     ntime: int
     nx: int
     nz: int
+    noise_amplitude: float
     scalar_names: List[str] = dataclasses.field(default_factory=list)
-    noise_amplitude: float = 1.0
 
     def open_tfdataset(
         self, local_download_path: Optional[str], variable_names: Sequence[str],
@@ -69,7 +69,7 @@ class SyntheticWaves(TFDatasetLoader):
         period_max: maximum period of waves
         phase_range: fraction of 2*pi to use for possible range of
             random phase, should be a value between 0 and 1.
-        type: one of "sinusoidal" or "square"
+        wave_type: one of "sinusoidal" or "square"
     """
 
     nsamples: int
@@ -77,13 +77,13 @@ class SyntheticWaves(TFDatasetLoader):
     ntime: int
     nx: int
     nz: int
+    wave_type: str
     scalar_names: List[str] = dataclasses.field(default_factory=list)
     scale_min: float = 0.0
     scale_max: float = 1.0
     period_min: float = 8.0
     period_max: float = 16.0
     phase_range: float = 1.0
-    type: str = "sinusoidal"
 
     def open_tfdataset(
         self, local_download_path: Optional[str], variable_names: Sequence[str],
@@ -97,12 +97,15 @@ class SyntheticWaves(TFDatasetLoader):
                 variable name to variable value, and each value is a tensor whose
                 first dimension is the batch dimension
         """
-        if self.type == "sinusoidal":
+        if self.wave_type == "sinusoidal":
             func = np.sin
-        elif self.type == "square":
+        elif self.wave_type == "square":
 
             def func(x):
                 return np.sign(np.sin(x))
+
+        else:
+            raise ValueError(f"Invalid wave_type {self.wave_type}")
 
         dataset = get_waves_tfdataset(
             variable_names,
--- a/external/fv3fit/fv3fit/data/synthetic.py
+++ b/external/fv3fit/fv3fit/data/synthetic.py
@@ -1,0 +1,235 @@
+from .base import TFDatasetLoader, register_tfdataset_loader
+import dataclasses
+from typing import Optional, Sequence, List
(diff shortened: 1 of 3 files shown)


external/fv3fit/fv3fit/data/synthetic.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        Args:
+            local_download_path: if provided, cache data locally at this path
+            variable_names: names of variables to include when loading data
+        Returns:
+            dataset containing requested variables, each record is a mapping from
+                variable name to variable value, and each value is a tensor whose
+                first dimension is the batch dimension
+        """
+        if self.wave_type == "sinusoidal":
```
brianhenn, line 100: Nit: push this argument handling into the  `get_waves_tfdataset` function since it is already does post-processing steps with all the arguments it receives. 

Further nit: Instead of adding two new classes of synthetic data loaders, make a single class (e.g., `SyntheticData`) that has required `synthtic_data_kind` argument of type `Literal['waves','noise']`, and whatever other arguments are common to these approaches, and then also takes a bunch of kwargs appropriate to the choice of `synthtic_data_kind`

external/fv3fit/fv3fit/data/synthetic.py
```diff
@@ -0,0 +1,238 @@
...
+    @classmethod
+    def from_dict(cls, d: dict) -> "TFDatasetLoader":
+        return dacite.from_dict(
+            data_class=cls, data=d, config=dacite.Config(strict=True)
+        )
+
+
+def get_waves_tfdataset(
+    variable_names,
+    *,
```
brianhenn, line 139: Just curious what's the reason for needing only the variables names list as a positional?

external/fv3fit/fv3fit/data/synthetic.py
```diff
@@ -0,0 +1,238 @@
...
+from typing import Optional, Sequence, List
+import tensorflow as tf
+import numpy as np
+from ..tfdataset import generator_to_tfdataset
+import dacite
+
+
+@register_tfdataset_loader
+@dataclasses.dataclass
+class SyntheticNoise(TFDatasetLoader):
```
brianhenn, line 12: Curious what's the intended use of the noise dataset?

external/fv3fit/tests/training/test_autoencoder.py
```diff
@@ -99,8 +58,8 @@ def test_autoencoder(tmpdir):
     os.chdir(tmpdir)
     # need a larger nx, ny for the sample data here since we're training
```
brianhenn, line 59: Similar to previous request, I find this test lacking an overarching description of the data problem it's trying to capture with an autoencoder, i.e., it's necessary to get into the weeds of the code, or to take the time to run the test locally with the plotting uncommented, to understand what it's trying to do. Given the tricky nature of this test to write I think it's fair to have a verbose docstring describing where it landed.

brianhenn (approved):
Thanks @mcgibbon, approving subject to the docstring request below along with a couple optional suggestions and requests for clarification.

external/fv3fit/fv3fit/data/synthetic.py
mcgibbon: It's handy for single-sample overfitting tests. I wrote it when I was concerned that waves were particularly difficult for CycleGAN to overfit on, but it turned out CycleGAN is just not good at overfitting.

external/fv3fit/fv3fit/data/synthetic.py
mcgibbon: Personal preference honestly. Whenever I call a function with many arguments, whether it's required to or not I try to put the first argument as positional and the rest of the arguments as keyword. Usually the first argument is obvious enough from the function and feels like the "object" while the other arguments feel like "descriptors" or settings (which I'd say is true here).

external/fv3fit/tests/training/test_autoencoder.py
mcgibbon: Will add some docstrings/comments.

With these settings the autoencoder isn't too bad. It's only a little slower than the other training tests.

mcgibbon: Updated. It didn't help that some of the previous docstrings/comments were false.

external/fv3fit/fv3fit/data/synthetic.py
mcgibbon: I don't know that I can actually do this - if I push the handling into the function, it's basically going to be an if/elif statement with the existing functions inside the blocks. It's also a bit uncomfortable merging the configuration classes because there are configuration options that aren't shared between the noise types, which would get ignored depending on the settings. We can voice chat about it.

PR #2027, opened 2022-09-01
Configure max latitude for high-antarctic filter

nbren12:
The variance of condensation is about 7x smaller south of 75S than between 60S and 75S.  These code changes allowed me to train a model with data only from south of 75S and to launch a non-crashing prognostic run testing this model.

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/327392f2-b9ec-4d37-84ae-ed6c2dc04382\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/emulation/emulation/_emulate/microphysics.py
+++ b/external/emulation/emulation/_emulate/microphysics.py
@@ -3,6 +3,7 @@ import datetime
 from typing import Callable, Optional, List
 import cftime
 import gc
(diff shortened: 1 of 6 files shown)


oliverwm1 (approved):
LGTM. Hopefully the batching helps with memory usage.

PR #2029, opened 2022-09-02
Remove iteration over batches from offline diagnostics compute step

AnnaKwa:
Iterating over batches adds complication to the computation of time mean metrics and also required extra helper functions to deal with averaging over batches. Model prediction is done over batches, which are saved locally to a temporary directory and read in as a single merged dataset of dask arrays, which is then used in the rest of the computations.

The main change is the addition of the `get_prediction` function, which takes the batches sequence, applies the prediction and optional coarsening transforms, and returns the data as a single dataset of dask arrays.

Significant internal changes:
- Removed loop over batches in diagnostics computation

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/bf973bc1-0372-4578-8dbb-2fd575788394\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
@@ -260,6 +260,7 @@ def get_prediction(
 ) -> xr.Dataset:
 
     model_variables = _variables_to_load(model)
+    config.timesteps = sorted(config.timesteps)
     batches = config.load_batches(model_variables)
 
     transforms = [_get_predict_function(model, model_variables)]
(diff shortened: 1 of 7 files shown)


brianhenn (commented):
Thanks @AnnaKwa I like the solution of dumping the predictions locally to make a lazy dataset.

I think we should ensure that the >C48 use cases still work. Ideally we'd have a test for them but I couldn't think of an easy way of doing that that wasn't very slow.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -250,16 +222,77 @@ def _get_data_mapper_if_exists(config):
...
+        )
+    if coarsening_factor > 1:
+        prediction_grid = load_grid_info(f"c{prediction_resolution}")
+        logger.info(f"Coarsening predictions by factor of {coarsening_factor}.")
+        return coarsen_cell_centered(
+            weights=prediction_grid.area, coarsening_factor=coarsening_factor
+        )
+
+
+def get_prediction(
```
brianhenn, line 257: Could this function do some logging? When testing this locally with C384 data the console goes silent for awhile (I think here).

brianhenn (approved):
Thanks @AnnaKwa . Approving subject to the minor requests about logging and .values, and assuming that the failure of the prognostic run report test is unrelated.

PR #2031, opened 2022-09-07
Small updates in the course of CycleGAN work

mcgibbon:
This PR contains small fixes and changes needed to support adding the CycleGAN model.

Added public API:
- Added missing symbols `CycleGANLoader`, `SyntheticNoise`, and `SyntheticWaves` to fv3fit.data
- updated typing on PytorchAutoregressor to use str keys for scalers
- added MPS support to fv3fit.pytorch.system.DEVICE for mac M1s
- fixed bug in apply_to_tuple where it did not follow the *args convention used by tensorflow

Refactored public API:
- fixed bug in CycleGanLoader where batch_size option is ignored

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/b42cca04-75bb-4b97-b34f-4a49981a017c/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/system.py
+++ b/external/fv3fit/fv3fit/pytorch/system.py
@@ -1,3 +1,14 @@
+import torch.backends
 import torch
+import os
 
-DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+if os.environ.get("TORCH_CPU_ONLY", False):
+    DEVICE = torch.device("cpu")
+else:
+    DEVICE = torch.device(
+        "cuda:0"
+        if torch.cuda.is_available()
+        else "mps"
+        if torch.backends.mps.is_available()
+        else "cpu"
+    )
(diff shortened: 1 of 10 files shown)


external/fv3fit/fv3fit/pytorch/system.py
```diff
@@ -1,3 +1,14 @@
+import torch.backends
 import torch
+import os
 
-DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+if os.environ.get("TORCH_CPU_ONLY", False):
```
nbren12, line 5: Any plans for documenting this env var?

nbren12 (approved):
Thanks for extracting this!

external/fv3fit/fv3fit/pytorch/system.py
mcgibbon: Presently no, still too much in flux. I added this only because on the MPS backend it sometimes segfaults instead of giving helpful exceptions. It should only affect myself for now, since nobody else has an M1.

PR #2032, opened 2022-09-07
Add CycleGAN training function to fv3fit

mcgibbon:
This PR adds a "cyclegan" training function which trains a CycleGAN model.

Added public API:
- added registered "cyclegan" training function to fv3fit with `fv3fit.pytorch.CycleGANHyperparameters` configuration
- added fv3fit.pytorch.CycleGAN Reloadable

Refactored public API:
- PytorchPredictor is now initialized with and contains as attribute a scalars mapping that uses str instead of Hashable for keys

Significant internal changes:
- Pytorch autoencoder architecture has changed, and is no longer a U-Net

- [x] Tests added

- test_unit: [82%](https://output.circle-artifacts.com/output/job/da6503d2-f493-48f7-8744-0a46dfd93bd1/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


nbren12 (commented):
We discussed @brianhenn  will be taking the lead on reviewing this. I think one potential was that the code be merged but to a more experimental location e.g. projects/cyclegan. What is your main motivation for bundling the code in fv3fit?

On the other hand, the code looks clean from the 9000 ft view though perhaps could be better unit tested. I'm a bit concerned the end-to-end traing test `test_cyclegan` will be brittle and further slow down the fv3fit tests like the never-since-used BPTT tests.

nbren12:
Overall, I see a lot of value to merging code like this early without a super rigorous review process, but it would be nice to do this have a better home for this kind of code that 1) clearly indicates it is experimental, but 2) allows for integrations with fv3net tooling like tests, argo workflows, etc.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+    scaler_tuple: Tuple[Mapping[str, StandardScaler], Mapping[str, StandardScaler]]
+) -> Mapping[str, StandardScaler]:
+    scalers = {}
+    for prefix, scaler_map in zip(("a_", "b_"), scaler_tuple):
+        for key, scaler in scaler_map.items():
+            scalers[prefix + key] = scaler
+    return scalers
+
+
+class ReplayBuffer:
```
brianhenn, line 112: Did you try training without this, ie, do we really need it? If so it seems like it'd be nice to have a generic random buffer class that takes in a sequence of samples and returns a random subset of the buffer, without the bespoke types and methods here that make this code unlikely to be reused anywhere else.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+                self.data.append(element)
+                to_return.append(element)
+            else:
+                if random.uniform(0, 1) > 0.5:
+                    i = random.randint(0, self.max_size - 1)
+                    to_return.append(self.data[i].clone())
+                    self.data[i] = element
+                else:
+                    to_return.append(element)
+        return torch.autograd.Variable(torch.cat(to_return))
```
brianhenn, line 144: Is this the right API to be using? https://pytorch.org/docs/stable/autograd.html#variable-deprecated

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+            gen_b: torch.Tensor = self.generator_a_to_b(
+                torch.as_tensor(real_a).float().to(DEVICE)
+            )
+            gen_a: torch.Tensor = self.generator_b_to_a(
+                torch.as_tensor(real_b).float().to(DEVICE)
+            )
+            stats_gen_a.observe(gen_a.detach().cpu().numpy())
+            stats_gen_b.observe(gen_b.detach().cpu().numpy())
+        metrics = {
+            # "r2_mean_b_against_real_a": get_r2(stats_real_a.mean, stats_gen_b.mean),
```
brianhenn, line 279: Commented code.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+                if random.uniform(0, 1) > 0.5:
+                    i = random.randint(0, self.max_size - 1)
+                    to_return.append(self.data[i].clone())
+                    self.data[i] = element
+                else:
+                    to_return.append(element)
+        return torch.autograd.Variable(torch.cat(to_return))
+
+
+class StatsCollector:
```
brianhenn, line 147: Do we need the untested running mean/std logic when `mean` and `std` are only ever called once? Seems simpler/less error prone to just make a sequence of predictions and then compute its mean/std over the desired dimensions.

brianhenn (commented):
I did my best to do a comprehensive review but certainly didn't look closely at everything. I did find the separation of concerns to be very helpful in understanding the complex cyclegan workings. I think there are some objects added that a) aren't clear if they are part of the standard cookbook and if we need them and b) aren't tested, that caught my eye. 

I'm agnostic about where new training code should live. It's certainly getting cumbersome to try to sift through fv3fit at this point, but I don't feel like this PR is a place to implement a plug-in type system that was suggested.

nbren12:
> No need to worry about that for now, it's not going to run on PR as it's too slow to train. I may add an on-merge-to-main test if we become concerned it will break.

Will the test run on when manually invoking the fv3fit tests: `pytest external/fv3fit`? I do that frequently and it would be nice to keep the default execution time low enough to provide a useful iteration loop.

> Having N projects in different locations would mean N more python packages we need to have boilerplate for and install in our prognostic run/fv3fit environment, given we want to be able to train them with argo. It's been much easier for us having a package to throw all our experimental model code into instead of setting up new infrastructure each time.

I agree it's annoying to make setup.pys, modify dockerfiles ,etc, but it seems like the latter is still an issue e.g. with the slow test that you don't want to invoke for every commit/PR.  Doesn't testing adjacent subtrees of fv3fit on different cycles in ad-hoc ways seem problematic? It would be nice to capture in the directory structure somehow. Maybe an `fv3fit.models.experimental`  namespace?

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -0,0 +1,424 @@
...
+        discriminator_b = self.discriminator.build(n_state)
+        optimizer_generator = self.generator_optimizer.instance(
+            itertools.chain(
+                generator_a_to_b.parameters(), generator_b_to_a.parameters()
+            )
+        )
+        optimizer_discriminator = self.discriminator_optimizer.instance(
+            itertools.chain(discriminator_a.parameters(), discriminator_b.parameters())
+        )
+        return CycleGANTrainer(
```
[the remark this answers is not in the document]
mcgibbon: CycleGANTrainer is "module public" because it's used in `train_cyclegan.py`, but it's package-private because it's not imported to `fv3fit.pytorch`. I can rename it to `Trainer`, it's just that we have a lot of different objects named things like "train" (e.g. multiple TrainingLoopConfig), and this object's motivation for existing is that cyclegan training is complicated. I'm not sure which name is better.

CycleGAN (and by extension CycleGANModule) are package-public because I will be importing them as symbols in code that evaluates CycleGAN output (which will definitely need to be bespoke compared to evaluating other models).

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: This code is pulled directly from the reference implementation of the CycleGAN. They reported in their paper that they get oscillations in model behavior during training without it (which is referenced in the comment below).

I don't know that we have a need for this anywhere other than the CycleGAN training, `.shuffle()` mostly does the job. I'd prefer to keep it private here. I can rename it to be module-private with `_`.

mcgibbon: I'll switch out the implementation to the one used in the [reference implementation of the author](https://github.com/junyanz/pytorch-CycleGAN-and-pix2pix/blob/eb6ae80412e23c09b4317b04d889f1af27526d2d/util/image_pool.py), with a url reference.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: Will be swapping this code out.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: The issue is we need to calculate the mean over the entire tfdataset, which by definition is out-of-memory. I'll add a comment to this effect in the docstring.

This logic is unit tested in `tests/test_stats_collector.py`. The final mean is the "running mean".

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -0,0 +1,235 @@
+import logging
```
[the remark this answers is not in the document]
mcgibbon: FoldTileDimension is added and the objects are refactored to work with 5-D [sample, tile, z, x, y] data (channels first, previously was channels last). Otherwise it's the same.

mcgibbon:
> > No need to worry about that for now, it's not going to run on PR as it's too slow to train. I may add an on-merge-to-main test if we become concerned it will break.
> 
> Will the test run on when manually invoking the fv3fit tests: `pytest external/fv3fit`? I do that frequently and it would be nice to keep the default execution time low enough to provide a useful iteration loop.
> 
> > Having N projects in different locations would mean N more python packages we need to have boilerplate for and install in our prognostic run/fv3fit environment, given we want to be able to train them with argo. It's been much easier for us having a package to throw all our experimental model code into instead of setting up new infrastructure each time.
> 
> I agree it's annoying to make setup.pys, modify dockerfiles ,etc, but it seems like the latter is still an issue e.g. with the slow test that you don't want to invoke for every commit/PR. Doesn't testing adjacent subtrees of fv3fit on different cycles in ad-hoc ways seem problematic? It would be nice to capture in the directory structure somehow. Maybe an `fv3fit.models.experimental` namespace?

Currently you have to un-comment the `pytest.skip` decorator on the test in order for it to run, so it will never run unintentionally.

To solve the cadence of testing issue, I'll add in a bit-for-bit regression test. If that doesn't work, I'll add in a "code execution test" that "trains" the model but only checks that it runs without exception, not that the training does anything meaningful. Fingers crossed that this works on different systems.

I'll also move the test files into the fv3fit source tree as we discussed, possibly here or maybe in another PR if I move multiple non-cyclegan files along with it.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: Swapped.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: Done.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: Uncommented.

mcgibbon:
Could not add a bit-for-bit test, but I did add a test that the training and prediction run without raising an exception, and I added the bits needed for local regression testing to help refactor the code.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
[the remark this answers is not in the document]
brianhenn: OK that's probably fine, I agree that it's not obvious that dropping the prefix makes the names all that much clearer.

brianhenn (approved):
Approving, noting here that we landed on moving the visual test to `projects` in the next PR.

PR #2034, opened 2022-09-08
Use existing column dQv diagnostic when available

### LINKED ISSUE #1861: <dQv> prognostic diags show as zero in report
ex. https://storage.googleapis.com/vcm-ml-public/argo/2022-05-31t162307-n2f-pire-wind-nudging/index.html

`<dQv>` quantities show up as zero in figures (global mean time series, zonal means, etc.). This is not an issue for <dQu>, which is predicted by the same model. The [offline report](https://storage.googleapis.com/vcm-ml-public/offline_ml_diags/n2f-pire-stable-ml/wind-nudging-correct-inputs-l2-1e-2/wind_tendencies/index.html) displays both of these correctly, and the data saved out of the prognostic run also has nonzero values for dQv. Hence I believe this is a plotting issue and not an actual issue with the tendency not being applied during the prog run.

One thing I noticed in the `diags.zarr` column integrated data was that the ML momentum tendencies save as zero in the first time step of every segment since they're added to the diagnostics in the `_apply_postphysics_to_physics_state` step before the first postphysics step calls the model to predict them. If there's some resampling of the diags such that it only uses that timestep to show in the report, I would understand why it's zero. That doesn't explain why `<dQu>` is plotted as nonzero though, it is also saved as zero during the first timestep of each segment.

oliverwm1:
Fixes issue of dQv being set to 0 in reduced prognostic run diagnostics, even when variable is saved out with proper name.

Resolves #1861 

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/55d63587-6864-4963-88a9-fb6bf12f8279\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/derived_variables.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/derived_variables.py
@@ -183,6 +183,8 @@ def _column_dqv(ds: xr.Dataset) -> xr.DataArray:
         )
         # convert from m/s/s to Pa by multiplying by surface pressure divided by gravity
         column_dqv = 100000 / 9.8065 * ds.column_integrated_dQv
+    elif "column_integrated_dQv_stress" in ds:
+        column_dqv = ds.column_integrated_dQv_stress
     else:
         # assume given dataset has no ML prediction of momentum tendencies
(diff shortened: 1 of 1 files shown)


AnnaKwa (approved):
LGTM!

PR #2035, opened 2022-09-09
Generalize scripts for building prognostic run environment

spencerkclark:
This PR generalizes some scripts used in building the prognostic run docker image so that they can also be used in building code on Gaea or other platforms.  I plan on making a followup PR with the code to actually create the environment on Gaea (this will involve updates to the fortran repo as well).  

- test_unit: [83%](https://output.circle-artifacts.com/output/job/011806af-f870-43d4-9339-0b720fa7983b/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


spencerkclark:
I rebuilt and pushed the regular and GPU base images on my VM after [86e0855](https://github.com/ai2cm/fv3net/pull/2035/commits/86e085560a41a4966fed9a917fc013887c95f79b).

docker/prognostic_run/base.Dockerfile
```diff
@@ -47,33 +47,39 @@ RUN apt-get update && \
...
-RUN bash install_nceplibs.sh /opt/NCEPlibs
+COPY .environment-scripts/install_nceplibs.sh .
+RUN bash install_nceplibs.sh /NCEPlibs /opt/NCEPlibs linux gnu
 
 
 ENV ESMF_DIR=/usr/local/esmf
 ENV CALLPY_DIR=/usr/local
 ENV FMS_DIR=/FMS
 ENV FV3GFS_FORTRAN_DIR=/external/fv3gfs-fortran
-ENV ESMF_INC="-I${ESMF_DIR}/include -I${ESMF_DIR}/mod/modO3/Linux.gfortran.64.mpiuni.default/"
```
spencerkclark, line 65: This second include directory did not exist in the existing build, so it was always just ignored.  Therefore it is safe to delete here.

docker/prognostic_run/Dockerfile
```diff
@@ -4,44 +4,29 @@ FROM ${BASE_IMAGE} as bld
...
-RUN pip3 install --no-dependencies -e /fv3net/external/fv3kube
-
 COPY workflows/post_process_run /fv3net/workflows/post_process_run
-RUN pip3 install --no-dependencies -e /fv3net/workflows/post_process_run
+COPY workflows/prognostic_c48_run/ /fv3net/workflows/prognostic_c48_run
+COPY external/emulation /fv3net/external/emulation
+COPY projects/microphysics/scripts /fv3net/projects/microphysics/scripts
+COPY external/radiation /fv3net/external/radiation
+COPY .environment-scripts/install_fv3net_python_dependencies.sh .
+RUN bash install_fv3net_python_dependencies.sh /fv3net
```
oliverwm1, line 23: It's now not obvious that two things need to be modified for adding 'local' python dependencies (i.e. add COPY to Dockerfile and pip install to shell script).

Could we have an variable that is the list of fv3net packages to be pip installed which gets passed to the script?

.environment-scripts/install_nceplibs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+CLONE_PREFIX=$1
+INSTALL_PREFIX=$2
+PLATFORM=$3
+COMPILER=$4
+
+URL=https://github.com/NCAR/NCEPlibs.git
```
oliverwm1, line 9: no change needed for this PR, but this repo has been archived so we should probably use the NOAA version. https://github.com/NCAR/NCEPlibs. Could you make an issue to that effect?

oliverwm1 (approved):
Looks good to me. Easier to read Dockerfile and useful for installing on other platforms.

Only suggested change is to parametrize the list of internal python dependencies.

docker/prognostic_run/Dockerfile
```diff
@@ -17,10 +17,20 @@ COPY external/fv3kube /fv3net/external/fv3kube
 COPY workflows/post_process_run /fv3net/workflows/post_process_run
 COPY workflows/prognostic_c48_run/ /fv3net/workflows/prognostic_c48_run
 COPY external/emulation /fv3net/external/emulation
-COPY projects/microphysics/scripts /fv3net/projects/microphysics/scripts
```
spencerkclark: This is not actually a package and was accidentally duplicated from a copy statement later on in the Dockerfile.

.environment-scripts/install_nceplibs.sh
spencerkclark: Sure thing -- see #2038.

docker/prognostic_run/Dockerfile
spencerkclark: Implemented in [da8fb0b](https://github.com/ai2cm/fv3net/pull/2035/commits/da8fb0bf2ba5e7a99ac10b0f5002cdab1b9c218c).

PR #2036, opened 2022-09-09
Add artifacts offline-report search CLI

oliverwm1:
It is difficult to find offline ML reports corresponding to particular models. This adds a search feature for offline reports very similar to the prognostic report searching functionality.

Added public API:
- `artifacts offline-report` CLI

Significant internal changes:
- Generalized the `ReportIndex` class to allow it to work as an index for either offline or prognostic run reports

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/5b3202af-4366-4c71-9e1c-8f6bca0bea24/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/artifacts/fv3net/artifacts/report_search.py
+++ b/external/artifacts/fv3net/artifacts/report_search.py
@@ -37,7 +37,7 @@ class ReportIndex:
         return set(itertools.chain.from_iterable(_reports))
 
     def compute(self, url, filename="index.html"):
-        """Compute reports_by_run index from all reports found at url.
+        """Compute reports_by_id index from all reports found at url.
 
         Args:
             url: path to directory containing report subdirectories.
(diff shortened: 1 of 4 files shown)


external/artifacts/fv3net/artifacts/report_search.py
```diff
@@ -3,45 +3,57 @@
...
-from typing import Mapping, Optional, Sequence, Set
+from typing import Callable, Mapping, Optional, Sequence, Set
 
 import gcsfs
 import fsspec
 
 from .utils import _list, _cat_file, _close_session
 
 
+def _get_run_url(line: str) -> Optional[str]:
```
oliverwm1, line 14: this function is unchanged, just moved up from lower in the module

AnnaKwa (approved):
Thanks, this is really useful! LGTM.

PR #2037, opened 2022-09-12
Delete BPTT code from fv3fit

mcgibbon:
A while back we added a backpropagation-through-time training function, which has not been used since. We would likely re-write its logic if we revisited multi-timestep optimization. This PR deletes that code to simplify fv3fit.

Refactored public API:
- removed fv3fit.train_bptt entrypoint and related tests

- [ ] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/d6b2502a-8fc6-4cf7-8672-6c68e08f51b8/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


nbren12 (approved):
LGTM! Thanks.

PR #2044, opened 2022-09-14
Use existing state where prediction is NaN

oliverwm1:
This allows ignoring certain ML predictions and using the wrapped step's predictions instead. Specifically will allow doing fine-only prognostic runs with out-of-sample detection and using baseline physics in place.

(Delete unused sections)
Added public API:
- `taper_nan_mask` to fv3fit to allow masking with NaNs (instead of 0s)

Significant internal changes:
- the `apply` function for the `fv3fit` step transformer now ignores predictions that are equal to `np.nan` and using the existing state (i.e. wrapped function's predictions) for those cells.

- [x] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/b8e5fefc-2690-4085-8928-b922bdec6e35/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
+++ b/workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 import dataclasses
 from typing import Mapping, MutableMapping, Iterable, Hashable, Sequence, Tuple
 
+import numpy as np
 import xarray as xr
 import fv3fit
 from runtime.steppers.machine_learning import (
@@ -119,7 +120,9 @@ class Adapter:
         q2_new, q1_new = non_negative_sphum_mse_conserving(
             inputs[SPHUM], tendencies[SPHUM], self.timestep, q1=tendencies.get(TEMP),
         )
+        q2_new = q2_new.where(tendencies[SPHUM].notnull(), other=np.nan)
         limited_tendencies[SPHUM] = q2_new
         if q1_new is not None:
+            q1_new = q1_new.where(tendencies[TEMP].notnull(), other=np.nan)
             limited_tendencies[TEMP] = q1_new
         return limited_tendencies


workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -84,7 +84,15 @@ def predict(self, inputs: State) -> State:
 
     def apply(self, prediction: State, state: State):
         if self.config.online:
-            state.update(prediction)
+            prediction_filled_nans = {}
+            for k, v in prediction.items():
+                try:
+                    existing_value = state[k]
+                except KeyError:
+                    prediction_filled_nans[k] = v
```
oliverwm1, line 92: This try/except is here because in the units tests there was a case of a predicting a quantity (`surface_precipitation_rate`) that was not in the state.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
```diff
@@ -117,3 +117,37 @@ def add_one_to_temperature():
     assert "surface_precipitation_rate" not in state
     transform(add_one_to_temperature)()
     assert "surface_precipitation_rate" in state
+
+
+def test_adapter_with_nans(state, tmpdir_factory):
```
nbren12, line 122: This is a complex test for the functionality added...mock predictors, temporary files etc. You can make a simpler unit test by refactoring the additions in `apply` to a function `merge_masked_states(left, right)`.

If you worry that function won't be used, then you could add an NaN case to an existing regression test and update the regression data.

external/fv3fit/fv3fit/_shared/taper_function.py
```diff
@@ -13,6 +13,13 @@ def taper_mask(
     return xr.where(novelty_score > cutoff, 0, 1)
 
 
+def taper_nan_mask(
```
nbren12, line 16: Where is this used? I assume this is listed in a yaml someplace.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
```diff
@@ -84,7 +84,15 @@ def predict(self, inputs: State) -> State:
...
+                else:
+                    prediction_filled_nans[k] = v.where(v.notnull(), existing_value)
```
nbren12, line 94: I think the implementation is correct, but it's hard to know if the broader integration with taper_nan_mask is working as you intended. You said there was some issue with the diagnostics that prevented you from saying for sure?

nbren12 (approved):
Strictly speaking I think this is correct and worth merging, but probably worth documenting a bit better (not sure where). I also think the code could be refactored to a more testable design.

workflows/prognostic_c48_run/tests/test_fv3fit_transformer.py
oliverwm1: Good idea.

external/fv3fit/fv3fit/_shared/taper_function.py
oliverwm1: Yes, there is a configuration option for what "taper" function to use.

workflows/prognostic_c48_run/runtime/transformers/fv3fit.py
oliverwm1: Agreed the integration is trickier to test. I did verify offline that an 'out of sample'-wrapped tendency predictor give NaNs where expected. 

The monitored tendencies due to `fv3_physics`, `machine_learning` and `applied_physics` were not what I expected when plotted in a notebook. I'll make something more coherent to share.

PR #2045, opened 2022-09-14
Prognostic run: factory method for radiation stepper inputs

brianhenn:
Adds a cleaner factory method for the radiation stepper, separating out the construction of the prescriber/ML stepper that generates inputs to the radiation stepper, from the construction of the stepper itself. This new prescriber/ML stepper constructor is then shared across the radiation and prephysics stepper construction. Previously running the radiation stepper with prescribed inputs was blocked due to a bug about its factory method's MPI comm input arg, but that was hard to see because the factory method was doing several things.

Significant internal changes:
- in the prognostic run loop, the prephysics and radiation input steppers are now generated by a common method, `_get_prescriber_or_ml_stepper`
- Adds an `.output_variables` property to the `Renaming` and `MultiModel` adapters in `runtime.steppers.machine_learning` to aid in logging what steppers are doing. 
- Also moves some of the model opening helpers out of `loop` and into `runtime.steppers.machine_learning`

[X] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/c12f77b2-7134-4b9e-a097-d35ef3044b69/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/workflows/prognostic_c48_run/runtime/factories.py
+++ b/workflows/prognostic_c48_run/runtime/factories.py
@@ -111,17 +111,22 @@ def get_tendency_prescriber(
 
 def _get_time_lookup_function(
     mapper: Mapping[str, xr.Dataset],
-    variables: Sequence[str],
+    variables: Union[Sequence[str], Mapping[str, str]],
     initial_time: Optional[str] = None,
     frequency_seconds: float = 900.0,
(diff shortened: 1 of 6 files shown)


brianhenn:
Duplicated by #2063

PR #2049, opened 2022-09-16
Use xarray weighting feature in vcm.weighted_average

### LINKED ISSUE #2048: vcm.weighted_average treats NaNs as zeros
Because `vcm.weighted_average` uses `np.nansum` for computing the numerator and denominator, it is possible to get nonsensical results if the quantity being averaged is masked (i.e. has NaNs) but the weighting function is not. This use case happens, for example, when doing interpolation to pressure levels, in which case some data will be missing for lower levels over topography.

For example, I expect this average to be equal to 1.5 (i.e. ignore the Nan) but vcm.weighted_average gives a value of 1:
```
import numpy as np
import xarray as xr
import vcm
mean = vcm.weighted_average(xr.DataArray([np.nan, 1, 2]), xr.DataArray([1,1,1]), dims=['dim_0'])
mean.values
> array(1.)
```

Seems like this could be easily addressed by just using the built-in xarray `weighted` functionality which should handle the missing values for us.

oliverwm1:
Using the xarray weighting feature simplifies our code and also gives us more reasonable results in the case of there being NaNs in the quantity-to-be-averaged.

Significant internal changes:
- vcm.weighted_average now uses the xarray `.weighted` accessor
- one breaking change: if an array consists entirely of NaNs, the resulting weighted average will now also be a NaN instead of 0 (the current behavior)

- [x] Tests added

Resolves #2048 

- test_unit: [83%](https://output.circle-artifacts.com/output/job/269f47a5-909a-49c1-ae3b-691f4cca43ae/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/vcm/tests/test_calc.py
+++ b/external/vcm/tests/test_calc.py
@@ -209,7 +209,7 @@ weights_nans = xr.DataArray(np.full((4,), np.nan), dims=["z"])
     [
         (da, weights, "z", xr.DataArray(17.0 / 6.0)),
         (ds, weights, "z", xr.Dataset({"a": xr.DataArray(17.0 / 6.0)})),
-        (da_nans, weights, "z", xr.DataArray(0.0)),
+        (da_nans, weights, "z", xr.DataArray(np.nan)),
         (da, weights_nans, "z", xr.DataArray(np.nan)),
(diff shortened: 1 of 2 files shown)


AnnaKwa (approved):
LGTM. I can't think of a usage in the diagnostics where this would break anything, but if you want to check you could run the integration test and check that the reports are ok before merging.

PR #2051, opened 2022-09-16
Improvements to fv3fit CycleGAN

mcgibbon:
This PR includes improvements to the CycleGAN code made in the course of training on realistic data.

Refactored public API:
- Added .to(device) method to CycleGAN
- PytorchPredictor can now predict on larger datasets without running out of memory, thanks to batching
- Fixed bug in PytorchPredictor.get_config
- added disable_convolution debugging option to Generator for cyclegan
- PytorchPredictor will now transpose out-of-order dimensions to the expected order when predicting
- PytorchPredictor will now transfer itself to the default device when loading
- Added wandb matplotlib integration to cyclegan training

Significant internal changes:
- Cyclegan initialization now matches the reference paper
- reflection padding and instance normalization were added to CycleGAN training to match reference paper
- Added a GeographicBias layer to CycleGAN which allows adding a position-dependent bias to inputs and outputs
- updated fv3fit io typing to retain class type definition when decorating
- added logging of system device to fv3fit.pytorch.system

- [ ] Tests added

- test_unit: [83%](https://output.circle-artifacts.com/output/job/e64d18b8-d0df-4294-84a8-85ecc949c7df/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -61,7 +64,7 @@ class CycleGANNetworkConfig:
     discriminator_weight: float = 1.0
 
     def build(
-        self, n_state: int, n_batch: int, state_variables, scalers
+        self, n_state: int, nx: int, ny: int, n_batch: int, state_variables, scalers
```
mcgibbon, line 76: Sorry, this change belongs to the "[add GeographicBias layer to CycleGAN](https://github.com/ai2cm/fv3net/pull/2051/commits/86d9c8db89a2cffe5c340f98914d870a86cf80d2)" commit

external/fv3fit/tests/training/test_autoencoder.py
```diff
@@ -29,15 +29,35 @@ def get_synthetic_waves_tfdataset(nsamples, nbatch, ntime, nx, nz):
         nz=nz,
         wave_type="sinusoidal",
         scalar_names=["b"],
-        scale_min=0.5,
-        scale_max=1.5,
+        scale_min=1.0,
+        scale_max=1.0,
```
mcgibbon, line 33: With the architecture matching the paper, the NN can't directly store the amplitude of the input waves, and because those amplitudes aren't correlated with anything, there's no way to infer them. I'll keep an eye out for this affecting scientific results, e.g. if amplitudes of generated patterns don't make sense. But it should stand a better chance on non-synthetic data, where amplitudes are correlated with patterns.

external/fv3fit/fv3fit/pytorch/cyclegan/generator.py
```diff
@@ -56,22 +63,39 @@ def build(
             kernel_size=self.kernel_size,
             max_filters=self.max_filters,
             convolution=convolution,
+            nx=nx,
+            ny=ny,
+            use_geographic_bias=self.use_geographic_bias,
         )
 
 
+class GeographicBias(nn.Module):
```
brianhenn, line 77: Assume that later we'd want to have the tile dimension in the bias layer?

external/fv3fit/fv3fit/pytorch/cyclegan/reloadable.py
```diff
@@ -85,11 +85,15 @@ def load(cls, path: str) -> "CycleGAN":
         """Load a serialized model from a directory."""
         return _load_pytorch(cls, path)
 
+    def to(self, device) -> "CycleGAN":
```
brianhenn, line 88: I know this is torch nomeclature but really wish they used a more informative function name given that it can do a lot of things. At least whatever is passed to `device` should make the intent clearer.

brianhenn (approved):
LGTM. My questions suggest areas where it might be good to add some comments, especially in the case where things are going to change later eg for halos.

external/fv3fit/tests/training/test_autoencoder.py
mcgibbon: I added the instance normalization they apply to the first convolutional block's output, which makes it harder to capture the variance of the input.

external/fv3fit/fv3fit/pytorch/cyclegan/generator.py
mcgibbon: Oh wow, not just later, but now. This is a bug, will fix.

mcgibbon: I think this only executes without an error because channels happens to be 1, so it can broadcast over the tile dimension.

external/fv3fit/fv3fit/pytorch/cyclegan/generator.py
```diff
@@ -128,30 +129,36 @@ def up(in_channels: int, out_channels: int):
...
             in_channels=min_filters,
         )
 
-        self._out_conv = convolution(
-            kernel_size=7,
-            in_channels=min_filters,
-            out_channels=channels,
-            padding="same",
+        out_conv = nn.Sequential(
+            FoldTileDimension(nn.ReflectionPad2d(3)),
```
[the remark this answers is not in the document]
mcgibbon: I'm not sure yet. One possibility is that this module becomes some kind of HaloUpdate module, and we still use normal convolution. Another is that we use graph convolution or message passing, and remove this module. I'll likely try a few things and see what works best.

I am a little surprised by the thought that this only needs 3 halo (reflection padded) points, when the receptive field is definitely more than 3 points. I guess it's that the input/output layers need to have padded input because the definition of their input is static (it's real data), whereas the latent variables can be defined in a way that zero-padding (which is what it does) works just fine. This might mean a halo update based approach isn't a good idea.

PR #2052, opened 2022-09-16
Move fv3fit training tests into source tree

mcgibbon:
After discussion with @nbren12, we agreed to move these fv3fit training tests into the fv3fit source tree. The concern is that it would be easier to delete individual training functions if tests specific to those functions live with their source code.

Significant internal changes:
- moved test_autoencoder.py, test_cyclegan.py, and test_graph.py into the fv3fit source tree where their training functions live

- test_unit: [83%](https://output.circle-artifacts.com/output/job/65b9050e-611d-480f-9d1f-2d6e5432da41/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


nbren12 (approved):
lgtm.

Issue #88, opened 2019-12-18
Clean up I/O routines


nbren12:
Currently, I/O is strewn around this code, base and we keep reinventing the wheel as a result. We are using many different ways of interacting with the filesystem and GCS. These include

- gcsfs/fsspec
- `vcm.cloud.gcs` (wrapper routines around the basic google python api)
- `vcm.cloud.gsutil` (wrappers around `gsutil`) 
- `vcm.cloud.remote_data` (some gcsfs and intake). 
- `vcm.cubedsphere.io` (manual `open_mfdataset` and `ds.to_netcdf` calls). Moreover, `combine_subtiles` should probably be moved to `vcm.combining`.
- `vcm.convenience` many I/O functions using a random assortment of utilties.  Some unused and some not.
- some of the DataFlow pipelines use Apache Beams `fileio` API, which is similar in flavor to `fsspec`.
- `vcm.fv3_restarts` uses `fsspec`. 

Obviously this situation is much more complicated than the functionality we need. I think we should develop a coherent API for I/O operations. Some of these are specific to GFDL data (e.g. tiles/subtiles) others are more general purposes (e.g. opening a remote netCDF file lazily).  I think [intake](https://intake.readthedocs.io/en/latest/quickstart.html)  is designed to solve this problem, but we can always write our own custom solutions.

nbren12:
I think `vcm.convenience.get_root` should be deleted.

nbren12:
Still a bit of an issue imo.

PR #2056, opened 2022-09-19
Delete unused vcm.cloud.gcs module

oliverwm1:
This code is almost unused in fv3net and not used externally (e.g. in notebooks) as far as I know.

- test_unit: [82%](https://output.circle-artifacts.com/output/job/3a70f8eb-1559-41b0-8186-0b957ae284a0/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

Partially addresses #88


nbren12 (approved):
Nice. fsspec is somewhat more reliable now so we probably don't get much from the GA google product.

PR #2067, opened 2022-09-23
Radiation: Updating solar constant

brianhenn:
Previously the radiation port effectively hardcoded the solar constant which prevented validation outside of the original validation timestep. This PR makes the updated solar constant available to the radiation wrapper. It also refactors and renames a preprocessing function (`get_model`) that uses the solar constant to be a wrapper method instead ( (`_get_gfs_physics_control), since the `model` data structure is more about internal configuration of the radiation driver than preprocessing external state. 

See [notebook](https://github.com/ai2cm/explore/blob/master/brianh/2022-09-14-fine-cloud-fluxes/2022-09-23-validate-clearsky.ipynb) showing TOA downward shortwave errors after this fix (about 0.05 W/m^2 RMSE, still not zero due to radiation port vs Fortran scheme time difference to be fixed in another PR) versus [before](https://github.com/ai2cm/explore/blob/master/brianh/2022-09-14-fine-cloud-fluxes/2022-09-14-validate-rad-port.ipynb) (about 0.4 W/m^2). 

Significant internal changes:
- `radiation.Radiation` now has a `_get_gfs_physics_control` method to return flags/scalar configuration parameters needed to call the driver; `gfs_physics_control` is now a typed object nested with the `RadiationConfig` class.
- the time-varying solar constant and solar hour are now passed separately to the radiation driver call
- various renaming of radiation driver variables to improve clarity

- test_unit: [82%](https://output.circle-artifacts.com/output/job/6eafc1d2-4178-444a-aa5d-0de5454ca36a/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/radiation/radiation/radiation_driver.py
+++ b/external/radiation/radiation/radiation_driver.py
@@ -333,7 +333,9 @@ class RadiationDriver:
 
     def GFS_radiation_driver(
         self,
-        Model,
+        GFSPhysicsControl,
+        solcon,
+        solhr,
         Statein,
         Sfcprop,
         Coupling,
@@ -347,36 +349,53 @@ class RadiationDriver:
     ):
 
         return self._GFS_radiation_driver(
-            Model, Statein, Sfcprop, Grid, randomdict, lwdict, swdict,
+            GFSPhysicsControl,
+            solcon,
+            solhr,
+            Statein,
+            Sfcprop,
+            Grid,
+            randomdict,
+            lwdict,
+            swdict,
         )
 
     def _GFS_radiation_driver(
-        self, Model, Statein, Sfcprop, Grid, randomdict, lwdict, swdict,
+        self,
+        GFSPhysicsControl,
+        solcon,
+        solhr,
+        Statein,
+        Sfcprop,
+        Grid,
+        randomdict,
+        lwdict,
+        swdict,
     ):
-        if Model["uni_cld"]:
+        if GFSPhysicsControl["uni_cld"]:
             raise FileNotFoundError(f"uni_cld = True Not implemented")
 
-        if Model["effr_in"]:
+        if GFSPhysicsControl["effr_in"]:
             raise FileNotFoundError(f"effr_in = True Not implemented")
 
-        if not (Model["lsswr"] or Model["lslwr"]):
+        if not (GFSPhysicsControl["lsswr"] or GFSPhysicsControl["lslwr"]):
             return
 
         # --- set commonly used integers
-        LM = Model["levr"]
-        LEVS = Model["levs"]
+        LM = GFSPhysicsControl["levr"]
+        LEVS = GFSPhysicsControl["levs"]
         IM = Grid["xlon"].shape[0]
-        # NFXR = Model["nfxr"] # never used according to lint
-        NTRAC = Model[
+        # NFXR = GFSPhysicsControl["nfxr"] # never used according to lint
+        NTRAC = GFSPhysicsControl[
             "ntrac"
         ]  # tracers in grrad strip off sphum - start tracer1(2:NTRAC)
-        ntcw = Model["ntcw"]
-        ntiw = Model["ntiw"]
-        # ncld = Model["ncld"] # never used according to lint
-        ntrw = Model["ntrw"]
-        ntsw = Model["ntsw"]
-        ntgl = Model["ntgl"]
-        ncndl = min(Model["ncnd"], 4)
+        ntcw = GFSPhysicsControl["ntcw"]
+        ntiw = GFSPhysicsControl["ntiw"]
+        # ncld = GFSPhysicsControl["ncld"] # never used according to lint
+        ntrw = GFSPhysicsControl["ntrw"]
+        ntsw = GFSPhysicsControl["ntsw"]
+        ntgl = GFSPhysicsControl["ntgl"]
+        ncndl = min(GFSPhysicsControl["ncnd"], 4)
 
         LP1 = LM + 1  # num of in/out levels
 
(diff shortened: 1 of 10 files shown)


external/radiation/radiation/radiation_driver.py
```diff
@@ -196,6 +200,7 @@ def radupdate(
         cline,
         solar_data,
         gas_data,
+        me,
```
oliverwm1, line 203: me! (I assume this name is copied from Fortran?)

external/radiation/radiation/radiation_driver.py
```diff
@@ -328,9 +333,6 @@ def radupdate(
 
             return soldict, aerdict, gasdict
 
-        else:
-            return slag, sdec, cdec, solcon
```
oliverwm1, line 332: these outputs are not used anywhere?

external/radiation/radiation/wrapper_api.py
```diff
@@ -162,14 +175,7 @@ def _rad_compute(self, state: State, time: cftime.DatetimeJulian,) -> Diagnostic
...
         ncolumns, nz = statein["tgrs"].shape[0], statein["tgrs"].shape[1]
-        model = get_model(
-            self._rad_config,
-            self._tracer_inds,
-            time,
-            self._timestep,
-            nz,
-            self._comm.rank,
-        )
+        model = self._get_model(time, nz,)
```
oliverwm1, line 178: would `_get_model_config` be accurate? "Model" is already very overloaded for us...

external/radiation/radiation/wrapper_api.py
```diff
@@ -182,3 +188,57 @@ def _rad_compute(self, state: State, time: cftime.DatetimeJulian,) -> Diagnostic
...
+            "me": self._comm.rank,
+            "levs": nz,
+            "levr": nz,
+            "nfxr": self._rad_config.nfxr,
+            "ncld": self._rad_config.ncld,
+            "ncnd": self._rad_config.ncnd,
+            "fhswr": self._rad_config.fhswr,
+            "fhlwr": self._rad_config.fhlwr,
+            # todo: why does solar hour need to be one timestep behind time to validate?
+            "solhr": _solar_hour(time - timedelta(seconds=self._timestep)),
```
oliverwm1, line 208: this is the other issue you were referring to?

external/radiation/radiation/wrapper_api.py
```diff
@@ -182,3 +188,57 @@ def _rad_compute(self, state: State, time: cftime.DatetimeJulian,) -> Diagnostic
...
+            "ncnvcld3d": self._rad_config.ncnvcld3d,
+            "lmfdeep2": self._rad_config.lmfdeep2,
+            "lmfshal": self._rad_config.lmfshal,
+            "sup": self._rad_config.sup,
+            "kdt": self._rad_config.kdt,
+            "do_sfcperts": self._rad_config.do_sfcperts,
+            "pertalb": self._rad_config.pertalb,
+            "do_only_clearsky_rad": self._rad_config.do_only_clearsky_rad,
+            "swhtr": self._rad_config.swhtr,
+            "solcon": self._driver.solar_constant,
```
oliverwm1, line 227: Feels a bit odd that the rest of these are actually config options, while this item is something computed by the code (if I understand correctly). Do we need to include `solcon` in the `_get_model` output?

oliverwm1: Put another way, can/should we separate out the retrieval of config options that remain the same from the retrieval of things like `solhr` and `solcon` that actually depend on time?

external/radiation/radiation/radiation_driver.py
brianhenn: Ha yeah we can call this `rank` if it's clearer.

external/radiation/radiation/radiation_driver.py
brianhenn: They aren't; the Fortran code uses this for a print statement and so it got ported that way. But the real way these values are used by having the `driver.sol` class store them internally.

external/radiation/radiation/wrapper_api.py
brianhenn: See comment below.

external/radiation/radiation/wrapper_api.py
brianhenn: Yes --  I'll make an issue in fv3net to refer to here.

external/radiation/radiation/wrapper_api.py
brianhenn: We are somewhat constrained by the fact that this is port of the Fortran radiation driver, which when called takes a data structure called `Model` of the `GFS_control_type` that holds a ton of different flags/scalars. Two of those things that `GFS_control_type` contains are `solcon` and `solhr`.

The python port stays true to this, and in general I think we want to have the port and Fortran schemes avoid diverging too much. OTOH we could separate them out in their own arguments since the current way of doing things led to bugs. In terms of the name of this structure, I agree that `model` is not very informative. Maybe `GFS_physics_control` instead?

brianhenn: I did make `GFS_physics_control` a typed dataclass that is part of the radiation config, and separated out the time-varying solar arguments from it. I think it's reasonable to try to clean up some of the port's unfortunate features (dicts of untyped objects getting passed around) on a piecemeal basis.

external/radiation/tests/test_driver.py
```diff
@@ -193,9 +201,15 @@ def test_radiation_valiation():
         randomdict = io.load_random_numbers(LOOKUP_DIR, rank)
         lwdict = io.load_lw(LOOKUP_DIR)
         swdict = io.load_sw(LOOKUP_DIR)
+        solcon = Model.pop("solcon")
+        solhr = Model.pop("solhr")
+
+        gfs_physics_control = get_gfs_physics_control(Model)
```
brianhenn, line 207: It'd be nice to abandon this test that uses serialized data instead of having to keep modifying it to work with the port as we change it, but at the moment we don't have another validation test of the port. At some point when it validates "perfectly" or close to it in the prognostic run we can use that, but we're not there quite yet.

PR #2068, opened 2022-09-26
Add halo updates to CycleGAN

mcgibbon:
This PR makes convolution type parameterizable for CycleGAN training, an adds a new "halo_conv2d" convolution option. The previous "conv2d" option is kept only for comparison or debugging purposes.

Added public API:

Refactored public API:
- Added convolution_type option to CycleGANNetworkConfig
- Added strided_kernel_size option to GeneratorConfig, so that even kernel sizes can be used in striding layers and odd kernel sizes can be used in non-striding layers, which ensures we have an even distribution of halo points and makes it more straightforward to align outputs to inputs
- Added in_memory option to CycleGANTrainingConfig, which will convert all training data to pytorch tensors at the start of training.

Significant internal changes:
- Removed "padding" argument from convolution factories, as we always want "same"-like padding behavior
- Removed reflection padding from input and output convolutional layers, since the halo update convolution type does its own padding automatically. This may degrade "conv2d" results, but that configuration only exists for comparison/debugging.

- [x] Tests added

- test_unit: [82%](https://output.circle-artifacts.com/output/job/dd5a93c7-71cf-41cd-ab78-e508b5b644ac/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -144,6 +143,199 @@ def single_tile_convolution(
...
+                # we must make data contiguous after rotating 90 degrees because
+                # the MPS backend doesn't properly manage strides when concatenating
+                # arrays
+                if i_tile % 2 == 0:  # even tile
+                    # south edge
+                    with_halo[i_tile][0][1] = torch.rot90(
+                        inputs[
+                            :, (i_tile - 2) % 6, :, :, -self.n_halo :
+                        ],  # write tile 4 to tile 0
+                        k=-1,  # relative rotation of tile 0 with respect to tile 5
```
brianhenn, line 281: Can you check that these comments are correct? This is pretty hard to follow.

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -144,6 +143,199 @@ def single_tile_convolution(
...
+                with_halo[i_tile][1][1] = inputs[:, i_tile, :, :, :]
+                with_halo[i_tile][0][0] = corner
+                with_halo[i_tile][0][2] = corner
+                with_halo[i_tile][2][0] = corner
+                with_halo[i_tile][2][2] = corner
+                # we must make data contiguous after rotating 90 degrees because
+                # the MPS backend doesn't properly manage strides when concatenating
+                # arrays
+                if i_tile % 2 == 0:  # even tile
+                    # south edge
```
brianhenn, line 276: Can you clarify whether these comments use north/south/east/west in the true sense or relative to each tile's local x/y conventions?

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -144,6 +143,199 @@ def single_tile_convolution(
...
+        self.n_halo = n_halo
+
+    @cpu_only
+    def forward(self, x):
+        return x[
+            ..., self.n_halo : -self.n_halo, self.n_halo : -self.n_halo
+        ].contiguous()
+
+
+class AppendHalos(nn.Module):
```
brianhenn, line 232: I'm glad this passes the test against the other implementation that uses the CubedSphereCommunicator, otherwise this would be a bit of a scary thing to debug.

brianhenn (commented):
Thanks @mcgibbon I'm glad you got this working since it seems relatively complicated. See my questions for areas where the code is hard to follow, which could be improved either with docstrings or pictorial representations.

brianhenn (commented):
Thanks @mcgibbon I appreciate the thoughtful documentation and I think that helped me a lot. But unfortunately it also made it clear I don't understand some of the logic, particularly around padding for transpose convolutions. I think another VC about this would be helpful.

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -161,6 +161,107 @@ def halo_convolution(
...
 
+    Generally we suggest using an even kernel size for strided convolutions,
+    and an odd kernel size for non-strided convolutions. For the non-strided
+    case, an even kernel size gives a symmetric kernel as desired. We've only reasoned
+    this for a stride of 2, but it should hold for any even stride.
+
+    The strided case (which uses a stride of 2) is more complex, but an even
+    kernel size gives a symmetric kernel. This is because a (2, 2) patch of the input
+    domain corresponds to a (1, 1) patch of the output domain. We would need to have a
+    (2, 2) patch in the center of the kernel (which is in input space) in order for
```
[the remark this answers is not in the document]
brianhenn: "symmetric" about the output point?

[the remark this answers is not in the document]
brianhenn: OK I think I understand the padding/kernel_size interaction at this point

brianhenn (approved):
Thanks @mcgibbon. Approving,subject to fixing the tests of course. Still would support a change in the future to remove the Crop object and use ConvTranspose2D's native way of doing this with the `padding` argument.

PR #2076, opened 2022-09-30
prognostic run: flexible radiation port timestep

brianhenn:
Allows the radiation port to run at the same time cadence as the configurable Fortran radiation cadence. 

See [notebook](https://github.com/ai2cm/explore/blob/master/brianh/2022-09-14-fine-cloud-fluxes/2022-09-23-validate-clearsky.ipynb) demonstrating that downward shortwave at TOA exactly validates (<10^-9 W/m^2 abs error at each timestep) over a daylong run. 

Significant internal changes:
- `radiation.Radiation` now tracks the time since initialization in and uses that to decide whether to compute LW and SW based on the `fhswr` and `fhlwr` namelist parameters. 
- If radiation is not computed at a timestep, the port returns radiative fluxes from the last computed timestep, which is the same behavior as the Fortran physics.

[X] Tests added -- port must now validate against Fortran for TOA SW downward flux in C12 case. 

- test_unit: [82%](https://output.circle-artifacts.com/output/job/1e445440-2ae6-4cb3-9c21-969061ec5287/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/radiation/radiation/radiation_astronomy.py
```diff
@@ -31,8 +31,6 @@ def __init__(self, rank, isolar, solar_filename):
         self.isolflg = isolar
         self.solc0 = con_solr
         self.solar_fname = solar_filename
-        self.iyr_sav = 0
```
brianhenn, line 34: Deleted repeated code

external/radiation/radiation/radiation_astronomy.py
```diff
@@ -268,8 +266,6 @@ def sol_update(self, jdate, kyear, deltsw, deltim, lsol_chg, rank, solar_data):
 
                                 else:
                                     i -= 1
-        else:
```
brianhenn, line 271: This is not in Fortran and I'm not sure why it was added to the port. It had the unintended side effect of overriding the solar constant with a fixed value for any radiation calculation after the first one.

external/radiation/radiation/wrapper_api.py
```diff
@@ -223,3 +253,24 @@ def _solar_hour(time: cftime.DatetimeJulian) -> float:
...
+def _get_forecast_time_index(
+    current_time: cftime.DatetimeJulian,
+    init_time: cftime.DatetimeJulian,
+    timestep_seconds: float,
+) -> int:
+    """Get integer index of forecast time, since initial time """
+    forecast_elapsed_seconds = (
+        current_time + timedelta(seconds=timestep_seconds) - init_time
+    ) / timedelta(seconds=1)
+    return int(forecast_elapsed_seconds / timestep_seconds)
```
oliverwm1, line 267:
```suggestion
    forecast_elapsed_seconds = (current_time - init_time).total_seconds()
    return int(forecast_elapsed_seconds / timestep_seconds) + 1
```

IMO a bit more readable.

external/radiation/radiation/config.py
```diff
@@ -86,9 +90,11 @@ class GFSPhysicsControl:
     )
     do_only_clearsky_rad: bool = False
     swhtr: bool = True
-    lprnt: bool = False
     lwhtr: bool = True
+    lprnt: bool = False
     lssav: bool = True
+    nsswr: Optional[int] = None
+    nslwr: Optional[int] = None
```
oliverwm1, line 97: what's the expected behavior if these are not set? Would a default of 1 make more sense?

external/radiation/radiation/radiation_driver.py
```diff
@@ -379,7 +379,7 @@ def _GFS_radiation_driver(
             raise FileNotFoundError(f"effr_in = True Not implemented")
 
         if not (gfs_physics_control.lsswr or gfs_physics_control.lslwr):
-            return
+            return {}
```
oliverwm1, line 382: Would it be cleaner to do this check one level up? Currently there is a somewhat confusing coupling via the check for `len(out) == 0` after the driver call. I'm also fine to leave here if you think it's better.

oliverwm1 (commented):
Great to see the downward SW validating in the notebook. Are there any simple tests we could write that would capture this verification? E.g. something that would detect an off-by-one error for the time? (maybe test value for some grid cell right on the edge of night?)

external/radiation/radiation/config.py
brianhenn: This value has to be set before running the radiation port and is not actually configurable since it will be overwritten as a function of `fhswr/fhlwr` and the physics timestep, so a default value would never be used. I think there's some conflation of immutable namelist flags and mutable control data structure going on here, but I would prefer to deal with that in a future PR that refactors the rad wrapper a bit.

external/radiation/radiation/wrapper_api.py
brianhenn: Done

external/radiation/radiation/radiation_driver.py
brianhenn: Sure. Fortran invokes the driver even when it's not going to run the computations and then promptly exits (which is where this came from), but I don't think there's any reason to need to follow that pattern here.

workflows/prognostic_c48_run/tests/test_radiation.py
```diff
@@ -94,14 +107,25 @@ def get_fv3config():
...
-    config["diagnostics"] = [
+    diagnostics = [
         {
-            "name": "radiative_fluxes.zarr",
+            "name": "state_after_timestep.zarr",
             "chunks": {"time": 2},
             "times": {"frequency": 900, "kind": "interval"},
-            "variables": RADIATIVE_FLUXES,
+            "variables": [
+                flux.name for flux in RADIATIVE_FLUXES if "flux" in flux.name
```
oliverwm1, line 116: Is this because we don't have getters/setters for the heating rates?

workflows/prognostic_c48_run/tests/test_radiation.py
```diff
@@ -7,27 +7,40 @@
 import xarray as xr
 import numpy as np
 import pytest
+import dataclasses
 import os
 
 
+@dataclasses.dataclass
+class PythonRadiativeFlux:
```
oliverwm1, line 15: nit: maybe better named `RadiationDiagnostic` since it includes heating rates. Could also have a `is_flux` property.

oliverwm1 (approved):
Great to have the validation test for TOA SW!

PR #2078, opened 2022-10-03
Update zarr and numcodecs version

mcgibbon:
This PR is updating constraints.txt to get a fix for a bug in zarr. zarr and numcodecs versions are updated.

For some reason other packages changed order within the constraints file, but no other versions were changed.

Requirement changes:
- updated zarr and numcodecs

- [ ] Tests added

- test_unit: [82%](https://output.circle-artifacts.com/output/job/8cb4d23d-76b1-4137-8058-0eb399b527ec/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


oliverwm1:
How did you specify the requirement for zarr to be a more recent version? The typical procedure is to specify this requirement in a specific package if that package is dictating the requirement, or otherwise in the top-level `pip-requirements.txt` file if it is due to a repo-wide requirement (I think fair to say for this case). Then once you've updated the `pip-requirements.txt` (or package-specific setup.py etc), run `make lock_pip`.

I think the order changes comes from different systems having different alphabetical orders for `-` and `=`.

mcgibbon:
Because this isn't technically a lowest version requirement of every package, I did the following:

- Ran `pip install --update zarr`, and noted that zarr and numcodecs got updated
- deleted the `zarr` and `numcodecs` entries from `constraints.txt`
- ran `make constraints.txt`

There would be nothing wrong with also including this in the top level `pip-requirements.txt`, but I don't think we generally worry about finding and specifying the earliest package version that works for our code.

Should I update pip-requirements.txt with a minimum version equal to the new version?

oliverwm1:
> Should I update pip-requirements.txt with a minimum version equal to the new version?

Yes I think so. FYI instead of the procedure you used, you could just update the `pip-requirements.txt` to specify your new requirement and then run `make lock_pip` or `make constraints.txt`. We do use the `pip-requirements.txt` to specify minimum versions we want for particular packages in the fv3net environment (even if they are not strict requirements for a particular package).

mcgibbon:
Added. I re-ran `make constraints.txt` for good measure, nothing changed.

oliverwm1 (approved):
LGTM

PR #2081, opened 2022-10-06
Plot 3-day bias drifts on hovmoller page

oliverwm1:
Add hovmoller bias plots which are restricted to at most 3 days long. This allows viewing of the initial bias spinup.

Example [report](https://storage.googleapis.com/vcm-ml-public/argo/2022-10-06t151239-a67f50858510/hovmoller.html) (see bottom of the page).

Significant internal changes:
- added `trim_duration` method to `RunDiagnostics`
- add section of 3-day hovmoller bias plots

- [x] Tests added

- test_unit: [82%](https://output.circle-artifacts.com/output/job/c74d1132-0af3-4a95-be2e-2dc038d6ff66/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
@@ -282,6 +282,15 @@ def zonal_mean_hovmoller_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
     return plot_2d_matplotlib(diagnostics, "zonal_mean_bias", ["time", "latitude"])
 
 
+@hovmoller_plot_manager.register
+def zonal_mean_short_term_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
+    return plot_2d_matplotlib(
+        diagnostics.trim_duration(np.timedelta64(3, "D")),
+        "zonal_mean_bias",
+        ["time", "latitude"],
+    )
+
+
 def infer_duration_seconds(diagnostics: RunDiagnostics) -> float:
     run, *_ = diagnostics.runs
     time = diagnostics.get_variable(run, "high_frequency_time")
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
@@ -290,7 +290,7 @@ def zonal_mean_short_term_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
         "zonal_mean_bias",
         ["time", "latitude"],
         title="3-day zonal mean bias",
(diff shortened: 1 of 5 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -247,42 +247,53 @@ def diurnal_component_plot(
...
 
 
+@hovmoller_plot_manager.register
+def zonal_mean_short_term_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
+    return plot_2d_matplotlib(
+        diagnostics.trim_duration(np.timedelta64(3, "D")),
+        "zonal_mean_bias",
+        ["time", "latitude"],
+        title="3-day zonal mean bias",
+        plot_vars=[v.lower() for v in GLOBAL_AVERAGE_DYCORE_VARS],
```
oliverwm1, line 293: It's unfortunate we have to do this `.lower()` here (and in other places) to deal with the fact that for some reason [we do a lower-case transformation](https://github.com/ai2cm/fv3net/blob/12d2ed3c259ea669d7e21f29133c7e3d83ca054d/workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py#L74) when saving diagnostics. At some point we should stop doing that, but will be tricky to be backwards compatible.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -247,42 +247,53 @@ def diurnal_component_plot(
...
 @hovmoller_plot_manager.register
-def zonal_mean_hovmoller_bias_plots(diagnostics: Iterable[xr.Dataset]) -> RawHTML:
+def zonal_mean_hovmoller_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
     return plot_2d_matplotlib(diagnostics, "zonal_mean_bias", ["time", "latitude"])
 
 
+@hovmoller_plot_manager.register
+def zonal_mean_short_term_bias_plots(diagnostics: RunDiagnostics) -> RawHTML:
+    return plot_2d_matplotlib(
+        diagnostics.trim_duration(np.timedelta64(3, "D")),
```
brianhenn, line 289: Nit: If `duration` is already required to be a `np.timedelta64` at this point, move the function call inside `trim_duration` and just pass the `*args` to it.

brianhenn (approved):
Looks useful @oliverwm1

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
oliverwm1: I think it's more straightforward to type-hint a `np.timedelta64` duration than to type-hint the args that get passed to `np.timedelta64` (and in principle the duration could be generated some other way, like taking the difference between two times).

PR #2085, opened 2022-10-07
Add reservoir computing recurrence code

AnnaKwa:
This is the first in a set of PRs to add the reservoir computing (RC) code to fv3fit.
Since the reservoir (which is frozen and untrained) and the readout (which is trained) parts of the echo state network can be easily separated, the code reflects this separation of concerns.[ This letter ](https://drive.google.com/file/d/14Eo2kibDIZH_zjd-0l6dxWVjhi4z52rt/view)is a useful reference for the parts of reservoir in the code.

Added public API:
- `ReservoirHyperparameters` 

- [x] Tests added

- test_unit: [82%](https:\/\/output.circle-artifacts.com\/output\/job\/f3612ca9-f5ff-4ec9-ae60-09a76c54b677\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/fv3fit/tests/reservoir/test_reservoir.py
```diff
@@ -0,0 +1,83 @@
...
+        input_dim=2,
+        reservoir_state_dim=2,
+        sparsity=0.0,
+        res_scaling=1.0,
+        input_coupling_sparsity=0,
+    )
+    reservoir = Reservoir(hyperparameters)
+
+    reservoir.W_in = np.ones(reservoir.W_in.shape)
+    reservoir.W_res = np.ones((reservoir.W_res.shape))
```
spencerkclark, line 76:
```suggestion
    reservoir.W_res = np.ones(reservoir.W_res.shape)
```

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,120 @@
...
+    def synchronize(self, synchronization_time_series):
+        self.reset_state()
+        for input in synchronization_time_series:
+            self.increment_reservoir_state(input)
+
+    def _random_uniform_sample_func(self, min=0.0, max=1.0):
+        def _f(d):
+            return np.random.uniform(min, max, size=d)
+
+        return _f
```
spencerkclark, line 85: I think this could be a standalone function instead of being attached to the class.  

You could also consider writing a `random_sparse_matrix(m, n, sparsity, min, max)` function, which would handle converting from `sparsity` to `density`, and constructing the uniform random sample function, both of which are repeated in `_generate_W_in` and `_generate_W_res` separately.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _generate_W_in(self):
+        W_in_rows = []
+        # Generate by row to ensure same number of connections per input element
+        np.random.seed(self.hyperparameters.seed)
+        for k in range(self.hyperparameters.reservoir_state_dim):
+            W_in_rows.append(
+                scipy.sparse.random(
+                    m=self.hyperparameters.input_dim,
+                    n=1,
```
spencerkclark, line 95: I think it might make more sense to swap these and then use `vstack` at the end (this would eliminate the need for the transpose).
```suggestion
                    m=1,
                    n=self.hyperparameters.input_dim,
```

external/fv3fit/tests/reservoir/test_reservoir.py
```diff
@@ -0,0 +1,83 @@
...
+
+    reservoir.reset_state()
+
+    reservoir.increment_state(np.array([0.5, 0.5]))
```
spencerkclark, line 80: Maybe test after two increments, since the first increment only relies on `W_in` (the initial state is zero)?

spencerkclark (commented):
Thanks for breaking this into smaller parts!  I had a few in-line comments, as well as these couple more general ones:

- In Slack you inquired about the naming of classes being potential vague.  Would `ReservoirComputingModel` and `ReservoirComputingModelHyperparameters` be too verbose?  
- I think you have solid test coverage of the `Reservoir` class currently, but perhaps you could also add a test for `synchronize`?

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class ReservoirHyperparameters:
+    """Hyperparameters for reservoir
+
+    input_dim: Size of input vector
```
oliverwm1, line 14: nit: we usually use `dim` to mean the name of a dimension. E.g. see a lot of our VCM code or xarray convention. So I'd suggest renaming to `input_size` and same for `reservoir_state_dim` and `output_dim`.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,120 @@
...
+    reservoir_state_dim: Size of hidden state vector,
```
oliverwm1, line 15: Could also drop the `reservoir` part here and just call this `state_size`, since it's a bit redundant w/ the `ReservoirHyperparameters` class name.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,118 @@
...
+    def _random_uniform_sample_func(self, min=0.0, max=1.0):
+        def _f(d):
+            return np.random.uniform(min, max, size=d)
+
+        return _f
+
+    def _generate_W_in(self):
+        W_in_rows = []
+        # Generate by row to ensure same number of connections per input element
+        np.random.seed(self.hyperparameters.seed)
```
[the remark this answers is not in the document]
AnnaKwa: That makes sense.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,118 @@
...
+        is equally connected into the reservoir. Defaults to 0.
+    input_coupling_scaling: Scaling applied to W_in. Defaults to 1,
+        where all elements are sampled from random uniform distribution
+        [-1, 1]. Changing this affects relative weighting of reservoir memory
+        versus the most recent state.
+    res_scaling: Optional scale value for W_res that can be provided in lieu of
+        spectral radius. This is useful if you know what scaling parameter
+        applied to the uniform distribution [0, 1] for a given reservoir size
+        will lead to the (approximate) desired spectral radius, since eigenvalue
+        calculation for larger reservoirs can be slow.
```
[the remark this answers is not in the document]
AnnaKwa: The eigenvalue computation used to be annoying long for larger reservoirs, but setting the `k` parameter you pointed out makes that part much faster, so I can remove this option.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,120 @@
...
+        def _f(d):
+            return np.random.uniform(min, max, size=d)
+
+        return _f
+
+    def _generate_W_in(self):
+        W_in_rows = []
+        # Generate by row to ensure same number of connections per input element
+        np.random.seed(self.hyperparameters.seed)
+        for k in range(self.hyperparameters.reservoir_state_dim):
```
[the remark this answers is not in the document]
AnnaKwa: Most of the papers don't mention this (I assume they used a density of 1) but the [Wikner 2020 paper](https://drive.google.com/file/d/1uwROhCNJviwJdY5SP3S1u7JK_f0eeykL/view) does this:

> We choose the matrix Win to contain an equal number of nonzero elements in each column, which corresponds to coupling each element of the reservoir input to an equal number of reservoir node states.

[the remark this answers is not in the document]
spencerkclark: Nice thanks!  The code here ensures the number of nonzero elements per row remains constant.  Should it be the other way around?

AnnaKwa:
Thanks for the review!

> Would ReservoirComputingModel and ReservoirComputingModelHyperparameters be too verbose?
I think those sound good. The latter is long but only needs to appear in a few places.

I didn't end up adding a test for synchronization as I wasn't sure how to best test this- I'm not sure how to write a deterministic test case where after N synchronization steps the state is sufficiently uncorrelated with the starting state. The addition of the second `increment_state` call in the last test at least tests that subsequent increments are changing the state as expected (which is basically all the `synchronize` method does).

external/fv3fit/fv3fit/reservoir/reservoir.py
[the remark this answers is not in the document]
AnnaKwa: Thanks,f ixed.

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,115 @@
...
+        logger.info("Resetting reservoir state.")
+        self.state = np.zeros(self.hyperparameters.state_size)
+
+    def synchronize(self, synchronization_time_series):
+        self.reset_state()
+        for input in synchronization_time_series:
+            self.increment_reservoir_state(input)
+
+    def _generate_W_in(self):
+        W_in_rows = []
```
spencerkclark, line 87:
```suggestion
        W_in_columns = []
```

external/fv3fit/tests/reservoir/test_reservoir.py
```diff
@@ -0,0 +1,98 @@
...
+    hyperparameters = ReservoirHyperparameters(
+        input_size=100,
+        state_size=1000,
+        adjacency_matrix_sparsity=0.8,
+        spectral_radius=0.6,
+        input_coupling_sparsity=input_coupling_sparsity,
+    )
+    reservoir = Reservoir(hyperparameters)
+    nonzero_per_col = [
+        reservoir.W_in.T.getrow(i).count_nonzero()
```
spencerkclark, line 66:
```suggestion
        reservoir.W_in.getcol(i).count_nonzero()
```

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,115 @@
...
+        # Generate by column to ensure same number of connections per input element
```
spencerkclark, line 88:
```suggestion
        # Generate by column to ensure same number of connections per input element.
        # See Wikner et al. (2020), https://doi.org/10.1063/5.0005541, for more
        # details.
```

external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -0,0 +1,115 @@
...
+    input_coupling_scaling: float = 1.0
+
+    def __post_init__(self):
+        if not self.output_size:
+            self.output_size = self.input_size
+
+
+class Reservoir:
+    def __init__(
+        self, hyperparameters: ReservoirHyperparameters,
```
spencerkclark, line 66: Did you want to rename these?

spencerkclark (approved):
I had a few more minor suggestions; otherwise it looks good!

external/fv3fit/fv3fit/reservoir/reservoir.py
AnnaKwa: I'll keep these names the same since the "reservoir" is just this part of the model.

PR #2093, opened 2022-10-13
Override volume mount for `external/fv3gfs-fortran` in `make enter` rule

### LINKED ISSUE #2080: `make enter_prognostic_run` results in environment that can no longer import the Python wrapper
#2047 resulted in FMS and FV3GFS being compiled directly in the `external/fv3gfs-fortran` directory.  Previously both of these steps were done in directories outside `external`.  This enabled bind-mounting the user's full local fv3net `external` directory into the image in the `enter_prognostic_run` rule of the fv3net Makefile, without interfering with linking to libraries that were built within the image (e.g. FMS).  The current behavior causes problems, because one will not have compiled FMS or FV3GFS within their local fv3net clone.  Therefore importing the wrapper fails:
```
root@14fd0a7294de:/fv3net/workflows/prognostic_c48_run# python3 -c "from fv3gfs import wrapper"
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/fv3gfs/wrapper/__init__.py", line 2, in <module>
    from ._wrapper import (
ImportError: libFMS.so.2: cannot open shared object file: No such file or directory
```
A simple fix would be to restore the behavior of compiling FMS and FV3GFS in a location outside of `external`.

spencerkclark:
This is a simple way to address #2080.  If, in addition to the general mount of the user's full `external` directory, we add a specific mount of `/fv3net/external/fv3gfs-fortran`, docker will use the image's `fv3gfs-fortran` directory, but the user's `external` directory for everything else.

In images that don't use fv3gfs-fortran, this will add an empty directory (but previously, in those images we were adding the full user's directory of fv3gfs-fortran, which wasn't really necessary either, so I don't think this is necessarily a step back).  

Resolves #2080 

- test_unit: [82%](https://output.circle-artifacts.com/output/job/ff61922f-e4a1-45d7-ae2f-54b676a1da17/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ BEAM_VERSION = 2.37.0
 UBUNTU_IMAGE = ubuntu@sha256:9101220a875cee98b016668342c489ff0674f247f6ca20dfc91b91c0f28581ae
 # prognostic base image is updated manually, not on every commit
 PROGNOSTIC_BASE_VERSION = 1.1.0
+
+# Explicitly mount the /fv3net/external/fv3gfs-fortran directory to prevent
+# it from being overriden by the user's.  Historically we only interactively
+# develop the Python dependencies of the images in fv3net.
 DOCKER_INTERACTIVE_ARGS = \
 	--tty \
 	--interactive \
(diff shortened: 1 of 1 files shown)


brianhenn (approved):
Thanks @spencerkclark

PR #2096, opened 2022-10-18
Fix RC model prediction bugs

AnnaKwa:
- The readout component's prediction always has 2 dimensions and should be reshaped before being used to increment the state. The test of `ReservoirComputingModel.predict` was erroneously passing because the mock regressor class didn't replicate the output dimensions of the sklearn regressors.

- Fixed egregious error where `add_input_noise` returned the input noise array instead of adding it to the input data and returning the noised data
 
Also some changes for convenience in `train_ks.py`:
- added a dump method to the training config to help with experiment tracking
- moved predictor training into a `train()` function that can be imported in other scripts, notebooks, etc

- test_unit: [82%](https:\/\/output.circle-artifacts.com\/output\/job\/d5aea92a-1a1f-43d6-9ea7-c2e9bad85538\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
+++ b/external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
@@ -18,7 +18,7 @@ class MultiOutputMeanRegressor:
     def predict(self, input):
         # returns vector of size n_outputs, with each element
         # the mean of the input vector elements
-        return np.full(self.n_outputs, np.mean(input))
+        return np.full(self.n_outputs, np.mean(input)).reshape(1, -1)
 
 
 def _sparse_allclose(A, B, atol=1e-8):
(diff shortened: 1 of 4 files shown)


external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, n_outputs: int):
     def predict(self, input):
         # returns vector of size n_outputs, with each element
         # the mean of the input vector elements
-        return np.full(self.n_outputs, np.mean(input))
+        return np.full(self.n_outputs, np.mean(input)).reshape(1, -1)
```
AnnaKwa, line 21: This change is to replicate the behavior of the sklearn regressors, which predict 2d arrays.

spencerkclark (approved):
Thanks, yeah, the test you added is exactly what I had in mind.  The other changes look good too.

PR #2099, opened 2022-10-22
Radiation: wrapper solar hour validates

brianhenn:
Allows the python radiation port's TOA SW downward flux to validate over runs started in a wider set of ways. Accurately mimics GFS physics computation of the solar hour based on the model "initialization time" instead of the wrapper segment's first timestep. Before runs started off-the-hour or segmented runs with non-even-hour segment lengths would not validate.

Significant internal changes:
- `radiation.Radiation._solar_hour` now computes solar hour as in Fortran using the initialization time

Requirement changes:
- bumps `fv3gfs-fortran` submodule to use version with initialization time getter

[X] Tests added: The port TOA SW downward flux validation test now passes for a case where the physics initialization time is different from the model's first timestep (before it passed, but only because it wasn't mimicking Fortran)

- test_unit: [82%](https://output.circle-artifacts.com/output/job/f2493f6f-1e47-46db-bf0d-e91f0690a40f/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/radiation/radiation/wrapper_api.py
```diff
@@ -247,13 +246,13 @@ def _rad_compute(self, state: State, time: cftime.DatetimeJulian,) -> Diagnostic
...
-        + time.minute / MINUTES_PER_HOUR
-        + time.second / (MINUTES_PER_HOUR * SECONDS_PER_MINUTE)
-    )
+    def _solar_hour(self, time: cftime.DatetimeJulian) -> float:
+        """This follows the Fortran computation that rounds initial time to the
+        hour. See https://github.com/NOAA-GFDL/SHiELD_physics/issues/17, but for
+        now we want the port to validate against Fortran."""
+        seconds_elapsed = (time - self._init_time).total_seconds()
+        hours_elapsed = seconds_elapsed / SECONDS_PER_HOUR
+        return hours_elapsed + self._init_time.hour
```
brianhenn, line 255: Whoops this should be:
```suggestion
return (hours_elapsed + self._init_time.hour) % 24
```
Not sure if that would have caused an error or not with periodic cosine zenith angle computations, but will fix.

external/radiation/radiation/wrapper_api.py
```diff
@@ -247,13 +247,13 @@ def _rad_compute(self, state: State, time: cftime.DatetimeJulian,) -> Diagnostic
...
-def _solar_hour(time: cftime.DatetimeJulian) -> float:
-    return (
-        time.hour
-        + time.minute / MINUTES_PER_HOUR
-        + time.second / (MINUTES_PER_HOUR * SECONDS_PER_MINUTE)
-    )
+    def _solar_hour(self, time: cftime.DatetimeJulian) -> float:
+        """This follows the Fortran computation that rounds initial time to the
+        hour. See https://github.com/NOAA-GFDL/SHiELD_physics/issues/17, but for
+        now we want the port to validate against Fortran."""
```
oliverwm1, line 253: I don't see any rounding of the time to the nearest hour in this function.

workflows/prognostic_c48_run/runtime/derived_state.py
```diff
@@ -31,6 +31,9 @@ def __getitem__(self, key: str) -> xr.DataArray:
         if key == "time":
             time = self._getter.get_state(["time"])["time"]
             return xr.DataArray(time, name="time")
+        elif key == "initialization_time":
+            initialization_time = self._getter.get_state([key])[key]
+            return round_time(cftime.DatetimeJulian(*initialization_time.timetuple()))
```
oliverwm1, line 36: I think this should basically be the same as the `key=='time'` clause. It's a bit confusing to be doing rounding (to the nearest second?) and also this function is supposed to return a `xr.DataArray` but it looks like this case would return a `cftime.DatetimeJulian`.

external/radiation/radiation/wrapper_api.py
brianhenn: Here is the line:
```
Model%solhr  = mod(Model%phour+Model%idate(1),con_24)
```
This takes the fractional hours since initialization (here "initialization" includes hours+minutes+seconds), and adds it to *only the hour* (`idate(1)`) of the initialization, which effectively subtracts any minutes+seconds in the initialization time from the resulting solar hour.

workflows/prognostic_c48_run/runtime/derived_state.py
brianhenn: Sure, changed it to returning a data array. Our setup has multiple ways of getting the time in the wrapper (this was the other way), which is potentially confusing.

oliverwm1 (approved):

external/radiation/radiation/wrapper_api.py
oliverwm1: I see, thanks for explanation. I think either description works—although maybe a unit test would be good to ensure the behavior stays as expected.

PR #2101, opened 2022-10-28
Add hybrid reservoir computing model

AnnaKwa:
This adds the `HybridReservoirComputingModel` class and some code to `projects/reservoir` for training a hybrid model for the 1D KS equation. The 'imperfect' model in this case is a KS solver with an error added in the solution step (pretty rudimentary, but it does the job). 

I didn't make a dataclass for configuring the `ImperfectModel` object since it seemed a bit overkill to write the I/O for it when we're just playing with a toy model for now. If we decide to move on to applying this in an atmospheric model it should be revisited.

I'm open to alternative suggestions for the imperfect model related objects which have pretty cumbersome names. 'knowledge-based' is a nicer descriptor but maybe wordier.

Examples in [this notebook](https://github.com/ai2cm/explore/blob/master/annak/2022-09-14-reservoir/2022-10-21-hybrid.ipynb)
image

Refactored public API:
- `KuramotoSivashinskyConfig.timestep` is changed to be the actual timestep of the output time series, not the solver timestep. The solver timestep is now timestep/time_downsampling_factor. It was less confusing to configure the imperfect model this way, since its timestep is smaller than the reservoir model's timestep.
 

- [x] Tests added

- test_unit: [83%](https:\/\/output.circle-artifacts.com\/output\/job\/b91503ca-4133-47d6-9478-e19855d286b8\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


projects/reservoir/configs/train_config.yaml
```diff
@@ -15,3 +15,4 @@ n_burn: 250
 input_noise: 0.01
 seed: 0
 n_samples: 6000
+timestep: 0.3
```
spencerkclark, line 18:
```suggestion
timestep: 0.3

```

projects/reservoir/ks.py
```diff
@@ -116,51 +119,85 @@ def _generate_ic(input_dim, seed=0):
...
+    def predict(self, n_steps: int, initial_condition: np.ndarray):
+        if initial_condition.size != self.spatial_downsampling_factor * self.N:
+            # upsample initial condition array by spatial_downsampling_factor
+            ic_x = range(initial_condition.size)
+            interp_ic = interp1d(ic_x, initial_condition)
+            initial_condition = interp_ic(
+                np.linspace(
+                    ic_x[0], ic_x[-1], self.N * self.spatial_downsampling_factor
+                )
+            )
```
spencerkclark, line 172: I might handle this interpolation explicitly within `create_imperfect_prediction_train_data` and maybe instead raise an error here if the initial condition size does not match the appropriate size based on the KS parameters.  Handling interpolation automatically is convenient, but I think it could also mask potentially unintended bugs.

spencerkclark (commented):
Thanks for the changes based on our discussion on Friday.  I had a few more comments in addition to the spatial downsampling question I had on Slack.

spencerkclark (approved):
This looks good to me now -- thanks for adding a test of the interpolation.  I just had a couple more minor comments.

PR #2111, opened 2022-11-10
Add reload_path option to CycleGANNetworkConfig

mcgibbon:
This PR adds the ability to continue training CycleGAN from a previously saved model.

Added public API:
- added reload_path option to CycleGANNetworkConfig

Significant internal changes:
- Updated CycleGAN project files
- fv3fit_torch image is now automatically built on CircleCI

- [ ] Tests added 

- test_unit: [82%](https://output.circle-artifacts.com/output/job/98db66e2-7874-403b-84c9-995bd3228d52/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,7 @@ jobs:
           - post_process_run
           - fv3net
           - fv3fit
+          - fv3fit_torch
           - dataflow
           - artifacts
     machine:
@@ -286,6 +287,11 @@ workflows:
       - build_default:
           name: "build/push fv3fit image"
           image: fv3fit
+          requires:
+            - pytest
+      - build_default:
+          name: "build/push fv3fit_torch image"
+          image: fv3fit_torch
           requires:
             - pytest
       # manually trigger on PRs
--- a/projects/cyclegan/prepare_datasets.py
+++ b/projects/cyclegan/prepare_datasets.py
@@ -1,0 +1,18 @@
+import xarray as xr
+from datetime import datetime
+
+if __name__ == "__main__":
+    paths = {
+        "neg4K": "gs://vcm-ml-experiments/spencerc/2021-06-14/n2f-25km-baseline-minus-4k/fv3gfs_run/",  # noqa: E501
+        "0K": "gs://vcm-ml-experiments/spencerc/2021-05-24/n2f-25km-baseline-unperturbed/fv3gfs_run/",  # noqa: E501
+        "4K": "gs://vcm-ml-experiments/spencerc/2021-05-24/n2f-25km-baseline-plus-4k/fv3gfs_run/",  # noqa: E501
+        "8K": "gs://vcm-ml-experiments/spencerc/2021-08-11/n2f-25km-baseline-plus-8k/fv3gfs_run/",  # noqa: E501
+    }
+    date_string = datetime.now().strftime("%Y-%m-%d")
+    for name, path in paths.items():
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -288,6 +289,11 @@ workflows:
           image: fv3fit
           requires:
             - pytest
+      - build_default:
+          name: "build/push fv3fit_torch image"
```
oliverwm1, line 293: What's the difference between the `fv3fit_torch` image and the `fv3fit` image? We already build a `prognostic_run_gpu` image... was it not possible to use this for your training?

oliverwm1: Or alternatively, could we bump the tag for the `nvidia/cuda` base of the `prognostic_run_gpu` image to the version you need? Does this cause issues for @frodre?

projects/cyclegan/prepare_datasets.py
```diff
@@ -0,0 +1,18 @@
...
+        "0K": "gs://vcm-ml-experiments/spencerc/2021-05-24/n2f-25km-baseline-unperturbed/fv3gfs_run/",  # noqa: E501
+        "4K": "gs://vcm-ml-experiments/spencerc/2021-05-24/n2f-25km-baseline-plus-4k/fv3gfs_run/",  # noqa: E501
+        "8K": "gs://vcm-ml-experiments/spencerc/2021-08-11/n2f-25km-baseline-plus-8k/fv3gfs_run/",  # noqa: E501
+    }
+    date_string = datetime.now().strftime("%Y-%m-%d")
+    for name, path in paths.items():
+        ds_atmos = xr.open_zarr(path + "atmos_dt_atmos.zarr")
+        ds_sfc = xr.open_zarr(path + "sfc_dt_atmos.zarr")
+        ds = xr.merge([ds_atmos, ds_sfc])
+        out_path = f"gs://vcm-ml-experiments/mcgibbon/{date_string}/n2f-25km-{name}-merged.zarr"  # noqa: E501
```
AnnaKwa, line 16: This script is only run once, right, and not repeated as part of a workflow? I'm not sure how large the resulting outputs are but that could be a lot of redundant data if it resaves every time it's run on a different day.

.circleci/config.yml
frodre: We exclusively use the the `fv3fit` image now for tensorflow training on GPUs, which is the reason for the separation (tf is not compatible with CUDA that works with pytorch 😟).  

We held the prog_run image in case we were going to try prognostic runs with a gpu, but we haven't used it in a while.  Could probably remove that image from the CI stack.

oliverwm1: I see. Well yeah we should just be mindful of the creep of more and more images. I would support not building the prognostic run gpu image if its unused.

mcgibbon: Yes I also wanted to avoid adding a new image, unfortunately no cuda version works for both.

projects/cyclegan/prepare_datasets.py
mcgibbon: Yes, this is not run automatically.

AnnaKwa (approved):

.circleci/config.yml
AnnaKwa: I don't have much knowledge about running the prognostic run on GPUs, but could that be an efficient way to parallelize the inference using reservoirs over many smaller subdomains @mcgibbon?

mcgibbon: I'm not sure what you mean, we can chat about it on slack.

PR #2112, opened 2022-11-11
Add restart file coarsening regression tests

spencerkclark:
This PR adds some regression tests for coarse-graining restart files using the high-level interfaces provided in `coarsen_restarts.py`.  They already turned up one bug when using the blended approach, but not coarse-graining the A-grid winds, which I went ahead and fixed.  I tried to write things in a way such that these could facilitate refactoring the somewhat repetitive code in `coarsen_restarts.py`.

### Timing information

Since being refactored to use `synth` these tests now take less than 1.5 seconds each.  They are now more on par with the current slowest tests in vcm (`test_grid_builder`).

### Implementation note

Unfortunately we cannot use a traditional checksum based system, because `mappm` does not produce bit-for-bit identical results on my VM when compared with CircleCI, so I resorted to adding reference data to compare against in GCS.

### Requirement changes

- `synth` is needed to run these tests, but these tests will simply be skipped if it is not installed.

- [x] Tests added

- test_unit: [84%](https://output.circle-artifacts.com/output/job/9a37c377-1772-4a79-9bee-287b91dc0314/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


oliverwm1:
I'm concerned about the slowness of these tests. For comparison, the entire test suite for `vcm` (550 tests) is about 2 minutes in a recent CI test, and runs in as little as 15s on my mac. So this would be a very significant increase in test time for `vcm`.

Do you know what makes the test slow? Is it the interaction with GCS, or the actual coarsening? It seems like we should be able to write this test without relying on GCS.

spencerkclark:
It looks like I need to install synth in the CI environment.

external/vcm/vcm/xarray_loaders.py
```diff
@@ -112,3 +113,62 @@ def dump_nc(ds: xr.Dataset, f: BinaryIO):
...
+    """
+    for name, var in ds.variables.items():
+        kind = var.dtype.kind
+        if kind not in "biufSU":
+            raise NotImplementedError(
+                f"It is not currently possible to serialize data of kind "
+                f"{kind!r}, the kind of variable {name}, to JSON in vcm."
+            )
+    with open(filename, mode) as file:
+        json.dump(ds.compute().to_dict(), file)
```
spencerkclark, line 138: Not sure if we want to add `indent=4` here or not (do we prefer single line JSON files or O(1000)-line JSON files?).

external/vcm/tests/test_coarsen_restarts.py
```diff
@@ -0,0 +1,142 @@
...
+    return data
+
+
+def generate_synthetic_data(restart_schemas, grid_schema):
+    restart_data = generate_synthetic_restart_data(restart_schemas)
+    grid_data = generate_synthetic_grid_spec_data(grid_schema)
+    return restart_data, grid_data
+
+
+@pytest.mark.slow
```
spencerkclark, line 102: These tests are now less than 1.5 seconds each in CI (and less than 1 second each on my VM).  I don't think it does much harm to keep it, but we could remove this mark if we considered it to be a bit of a distraction now.

.environment-scripts/install_local_packages.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 CONDA_ENV=$1
-
+which pip
```
oliverwm1: leftover testing code?

external/vcm/tests/_coarsen_restarts_regression_tests/reference/area-weighted-model-level-without-agrid-winds-fv_core.res.json
```diff
@@ -0,0 +1 @@
...], [-419.9194891442152, -636.0835173202973, 824.6031201175159]], [[-175.99505510761955, -534.8269507621551, 67.40147707494769], [38.82355588656718, 155.47861094659277, 279.24037126640496]], [[7.590980745026847, -326.9134175417396, -787.5903055399301], [259.4353440229235, -419.18895740851406, 327.4215759870131]]]], [[[[-518.2347143407021, -852.6099446047097, 217.27979359633045], [-638.9506817572459, 310.52501752565024, -86.87470015188599]], [[625.4154837879888, -30.914293628859713, -249.04896257443045], [179.5868676917671, -116.45402282745297, -332.55953810169524]], [[-872.9038236934767, -311.78154433911305, 97.36243151995416], [167.53132605143867, -34.310124459537874, -416.3241932386889]], [[801.109653938307, -175.32898970790814, 653.031268944695], [-23.441072483616132, -299.5833146847884, 769.7072133118921]], [[-705.0847355777385, -43.97519048972128, -788.824182733945], [494.2930650223648, -148.62445096638015, 476.3101435422064]], [[-271.87365602550966, -236.98392628353164, 570.6687420863033], [-427.55998462223766, -312.5279107546991, -374.9999977807396]], [[406.1765810580566, 262.41067115299006, -643.3988970397526], [87.3452679958637, -198.3149368084531, 530.3419378585622]]]]]}}}
```
oliverwm1, line 1: no change necessary, but if we wanted to make these even smaller, could use lower precision for the floats. Especially given I imagine we're using a relatively large tolerance when doing the test.

external/vcm/tests/test_coarsen_restarts.py
oliverwm1: Yeah, I'd say leave it.

external/vcm/tests/test_coarsen_restarts.py
```diff
@@ -0,0 +1,154 @@
...
+    restart_schemas = open_restart_schemas(SCHEMA_PATH)
+    grid_spec_schema = open_grid_spec_schema(SCHEMA_PATH)
+    restart_data, grid_data = generate_synthetic_data(restart_schemas, grid_spec_schema)
+
+    func, kwargs = REGRESSION_TESTS[tag]
+    result = func(FACTOR, grid_data, restart_data, **kwargs)
+    result = {category: ds for category, ds in result.items()}
+
+    # To reset the reference data, run this module as a script:
+    # python test_coarsen_restarts --overwrite
```
oliverwm1, line 114:
```suggestion
    # python test_coarsen_restarts.py --overwrite
```
?

external/vcm/vcm/xarray_loaders.py
oliverwm1: Maybe could be an option, or you could just pass generic `kwargs` to `json.dump`? Probably for the reference test data better to have it be compact.

oliverwm1 (approved):
Thanks for the changes—I think a big improvement to not require network access and speed up tests.

Should add `synth` to vcm test requirements.

.environment-scripts/install_local_packages.sh
oliverwm1: This may be from an outdated commit.

oliverwm1:
> Should add synth to vcm test requirements.

I know you did add it to the tox config, but for completeness probably good to also include in setup.py.

oliverwm1:
> Should add synth to vcm test requirements.

Or to be as general as possible, you could skip these tests if synth is not installed. Just thinking from perspective of VCM being a useful standalone package.

external/vcm/tests/_coarsen_restarts_regression_tests/reference/area-weighted-model-level-without-agrid-winds-fv_core.res.json
spencerkclark: Interesting thought.  Since JSON is not a binary format there's not a super clean way of doing this, but if I truncate to five decimal points (e.g. through `xarray.Dataset.round`), the total size of the reference files is only reduced from 604K to 508K (fv3net itself is about 359M).  Therefore I've kept things as is.

external/vcm/tests/test_coarsen_restarts.py
spencerkclark: Oops thanks!

external/vcm/vcm/xarray_loaders.py
spencerkclark: Yeah, I agree no sense adding a huge amount of lines to the diff here.  I went ahead and implemented your suggestion.

spencerkclark:
Alright we should be good now (had a bad rebase there for a moment, but I reset things).  I added a module-level `importorskip` to the `test_coarsen_restarts.py` file so that those tests get skipped if `synth` is not installed, making `synth` fully optional.  I'm not sure if we want to add it in the `tests_require` section of `setup.py` or not (technically it's not really required for running the tests, just some of them, but I generally agree it's probably good to have it listed somewhere other than in the tox config).

oliverwm1:
Yeah, leaving it out of setup.py seems fine to me. Having the importorskip is good.

spencerkclark:
Sounds good.  I'll go ahead and merge.

PR #2121, opened 2022-12-02
Emulation precipitation conservation

frodre:
Adds a conservative precipitation options to post-hoc adjustments of the precpd emulation outputs.  

The two flags added to the model config are:
*  `enforce_strict_precpd_conservative`, which uses strict accounting to calculate precip while enforcing no negative downward  fluxes of precip anywhere in the atmosphere, i.e., there's never more evaporation than precip available.  
* `simple_precip_conservative`, which uses total column water change to determine the surface precipitation (no limiting).

- test_unit: [60%](https://output.circle-artifacts.com/output/job/0317243d-4094-472c-b7de-7a4ab855cbe5/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/emulation/emulation/config.py
```diff
@@ -103,6 +103,11 @@ class ModelConfig:
             built.
         batch_size: number of columns to batch the ml prediction by. May reduce
             memory use.
+        enforce_precpd_conservative: Use a conservation to determine total precip
```
AnnaKwa, line 106: These two flags kind of look similar, could they be a bit more specific to the conservation scheme being used? ex. `..._strict_toa, ..._from_column_water_change` or something to that effect?

external/emulation/emulation/zhao_carr.py
```diff
@@ -226,3 +235,109 @@ def enforce_conservative_phase_dependent(state, emulator):
...
+    num_features, num_samples = precip_to_vapor.shape
+
+    total_precip = np.zeros(num_samples)
+
+    # calculate precip and evaporation starting from TOA
+    for k in range(num_features - 1, -1, -1):
+        precip = limited_c_to_p[k]
+        total_precip += precip
+        evaporation = limited_p_to_v[k]
+        limited_evap = np.minimum(total_precip, evaporation)
```
AnnaKwa, line 284: If not too much of a pain to construct, could you add a test case where the evaporation at higher levels is such that this constraint has to take effect?

external/emulation/emulation/zhao_carr.py
```diff
@@ -226,3 +235,109 @@ def enforce_conservative_phase_dependent(state, emulator):
...
+    return x / RHO_WATER
+
+
+def _strict_conservative_precip_from_TOA_to_surface(
+    condensate_to_precip: np.ndarray, precip_to_vapor: np.ndarray
+):
+
+    """
+    Iterates backwards through precip source and evaporation terms to
+    determine surface precip.  Functio limits evaporation to available
```
AnnaKwa, line 261:
```suggestion
    determine surface precip.  Function limits evaporation to available
```

external/emulation/emulation/zhao_carr.py
```diff
@@ -226,3 +235,109 @@ def enforce_conservative_phase_dependent(state, emulator):
...
+        -1 * precip_src_limited, delp
+    )
+    humidity_out = state[GscondOutput.humidity] + limited_evaporation
+    temperature_out = state[GscondOutput.temperature] + evaporative_cooling
+
+    return {
+        **emulator,
+        PrecpdOutput.cloud_water: cloud_out,
+        PrecpdOutput.humidity: humidity_out,
+        PrecpdOutput.temperature: temperature_out,
```
AnnaKwa, line 322: Is it the case then that if this flag is enabled, the original prediction of temperature change from the Precpd model is not used, and is instead overwritten by the gscond model's T prediction + cooling from the precpd-predicted cloud and q? If this might lead to unexpected behavior can this be noted in the docstrings?

external/emulation/emulation/config.py
```diff
@@ -196,6 +203,12 @@ def _build_masks(self) -> Iterable[Mask]:
         elif self.enforce_conservative_phase_dependent:
             yield emulation.zhao_carr.enforce_conservative_phase_dependent
 
+        if self.precpd_precip_conservative:
+            yield emulation.zhao_carr.conservative_precip_simple
+
+        if self.enforce_precpd_conservative:
```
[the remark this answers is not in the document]
frodre: Added as an if/elif, and also raise an error when True/True happens in the config.

external/emulation/emulation/zhao_carr.py
frodre: Added a note to the docstring about temperature

external/emulation/emulation/zhao_carr.py
frodre: This was supposed to be the test case written but I had the directionality wrong.  I updated the test to make sure the limit is being enforced correctly.

AnnaKwa (approved):
Thanks for addressing the changes! LGTM.

PR #2127, opened 2022-12-16
Refactor momentum tendency update to use new A-grid to D-grid wind transform in wrapper

spencerkclark:
This PR refactors the code used to apply tendency updates to the A-grid winds in the prognostic run.  Beyond being somewhat convoluted, the previous approach had two primary drawbacks:
- It required delaying applying wind updates by a timestep.
- It required updating the `"eastward_wind_after_physics"` and `"northward_wind_after_physics"` variables from the wrapper, which use the opposite vertical dimension orientation convention to variables in the dynamical core and diagnostics. The code was previously not accounting for this and was thus applying the wind corrections upside down.

Significant internal changes:
- Removes the `_apply_postphysics_to_physics_state` step in the prognostic run time loop, which is no longer needed.

- [ ] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -154,7 +154,7 @@ def add_tendency(state: Any, tendencies: State, dt: float) -> Tuple[State, State
         filled_fractions = {}
 
         for name, tendency in tendencies.items():
-            filled, filled_fraction = fillna_tendency(tendency[name])
+            filled, filled_fraction = fillna_tendency(tendency)
             filled_tendencies[name] = filled
             filled_fractions[f"{name}_filled_frac"] = filled_fraction
 
@@ -544,7 +544,6 @@ class TimeLoop(
 
         Mostly used for updating the eastward and northward winds.
         """
-        pass
         # self._log_debug(f"Apply postphysics tendencies to physics state")
         # tendency = {k: v for k, v in self._tendencies.items() if k in ["dQu", "dQv"]}
 
@@ -657,7 +656,7 @@ class TimeLoop(
                 self._step_pre_radiation_physics,
                 self._step_radiation_physics,
                 self._step_post_radiation_physics,
-                self._apply_postphysics_to_physics_state,
+                # self._apply_postphysics_to_physics_state,
                 self.monitor(
                     "applied_physics",
                     self.emulate_or_prescribe_tendency(
--- a/workflows/prognostic_c48_run/runtime/loop.py
+++ b/workflows/prognostic_c48_run/runtime/loop.py
@@ -123,22 +123,16 @@ def compute_total_dgrid_wind_tendencies(state, tendencies):
     # tendencies are provided, we don't need to bother passing zeros through
     # to the transformation function.  We don't need to bother adding zeros
     # if no D-grid wind tendencies are provided either.
-    dQx_wind = tendencies.get(
-        "dQx_wind", xr.zeros_like(state["x_wind"]).assign_attrs(units="m/s/s")
-    )
-    dQy_wind = tendencies.get(
-        "dQy_wind", xr.zeros_like(state["y_wind"]).assign_attrs(units="m/s/s")
-    )
-    dQu = tendencies.get(
-        "dQu", xr.zeros_like(state["eastward_wind"]).assign_attrs(units="m/s/s")
-    )
-    dQv = tendencies.get(
-        "dQv", xr.zeros_like(state["northward_wind"]).assign_attrs(units="m/s/s")
-    )
+    dQx_wind = tendencies.get("dQx_wind", xr.zeros_like(state["x_wind"]))
+    dQy_wind = tendencies.get("dQy_wind", xr.zeros_like(state["y_wind"]))
+    dQu = tendencies.get("dQu", xr.zeros_like(state["eastward_wind"]))
+    dQv = tendencies.get("dQv", xr.zeros_like(state["northward_wind"]))
     (
         dQx_wind_from_dQu_and_dQv,
         dQy_wind_from_dQu_and_dQv,
-    ) = transform_agrid_winds_to_dgrid_winds(dQu, dQv)
+    ) = transform_agrid_winds_to_dgrid_winds(
+        dQu.assign_attrs(units="m/s/s"), dQv.assign_attrs(units="m/s/s")
+    )
     return dQx_wind + dQx_wind_from_dQu_and_dQv, dQy_wind + dQy_wind_from_dQu_and_dQv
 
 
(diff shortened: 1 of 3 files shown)


workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
```diff
@@ -310,6 +310,8 @@ def __call__(self, time, state):
 
     def get_diagnostics(self, state, tendency):
         diags = compute_diagnostics(state, tendency, self.label, self.hydrostatic)
+        momentum_diags = compute_ml_momentum_diagnostics(state, tendency)
+        diags.update(momentum_diags)
         return diags, diags[f"net_moistening_due_to_{self.label}"]
 
     def get_momentum_diagnostics(self, state, tendency):
```
oliverwm1, line 317: We should be able to delete this method on all steppers now (which is great!!)

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -106,32 +111,86 @@ def fillna_tendency(tendency: xr.DataArray) -> Tuple[xr.DataArray, xr.DataArray]
...
+    return dQu, dQv
+
+
+def transform_from_agrid_to_dgrid(
+    u: xr.DataArray, v: xr.DataArray
+) -> Tuple[xr.DataArray, xr.DataArray]:
+    """Transform a vector field on the A-grid in latitude-longitude coordinates
+    to the D-grid in cubed-sphere coordinates.
+
+    u and v must have double precision and contain like units attributes.
```
oliverwm1, line 142:
```suggestion
    u and v must have double precision and contain units attributes.
```

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -106,32 +111,86 @@ def fillna_tendency(tendency: xr.DataArray) -> Tuple[xr.DataArray, xr.DataArray]
...
+        # First update x_wind and y_wind based on the A-grid wind tendencies. If
+        # D-grid wind tendencies are provided later, update x_wind and y_wind
+        # within the for loop.  Tendency diagnostics for dQu, dQv, dQx_wind, and
+        # dQy_wind will be recorded separately.
+        if set(filled_tendencies).intersection(A_GRID_WIND_TENDENCIES):
+            dQu, dQv = prepare_agrid_wind_tendencies(filled_tendencies)
+            transformed = transform_from_agrid_to_dgrid(dQu, dQv)
+            dQx_wind_from_agrid, dQy_wind_from_agrid = transformed
+            updated[X_WIND] = state[X_WIND] + dQx_wind_from_agrid * dt
+            updated[Y_WIND] = state[Y_WIND] + dQy_wind_from_agrid * dt
```
oliverwm1, line 178: Maybe better to just not implement case of dQu and dQx_wind tendencies? Feels like a bit of a red flag that there are now multiple `updated = state + tendency * dt` lines in this function.

What about just having this `add_tendency` function assuming all provided tendencies are on the same grid as the corresponding state variable? i.e. do the A->D grid transformation (if necessary) in separate preceding call?

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -482,31 +541,6 @@ def _step_prephysics(self) -> Diagnostics:
 
         return diagnostics
 
-    def _apply_postphysics_to_physics_state(self) -> Diagnostics:
```
oliverwm1, line 485: so great to delete this!

workflows/prognostic_c48_run/runtime/names.py
```diff
@@ -38,6 +40,7 @@
     "ocean_surface_temperature",
     "surface_temperature",
 ]
+A_GRID_WIND_TENDENCIES = {"dQu", "dQv"}
```
oliverwm1, line 43: use previously defined constants here.

workflows/prognostic_c48_run/runtime/names.py
```diff
@@ -19,13 +19,15 @@
 TSFC = "surface_temperature"
 MASK = "land_sea_mask"
 TIME_KEYS = ["time", "initialization_time"]
+X_WIND = "x_wind"
+Y_WIND = "y_wind"
+AGRID_U_TENDENCY = "dQu"
+AGRID_V_TENDENCY = "dQv"
```
oliverwm1, line 25: maybe use `EAST_WIND` and `NORTH_WIND` instead of `U` and `V`? Bit more consistent with usage elsewhere.

oliverwm1 (commented):
Thanks for this PR, it's great to be able to remove the previous complexity (and weird timestep behavior) from applying wind updates through the physics.

I am concerned about the new complexity in the `add_tendency` function though. Can we keep responsibility for coordinate transformation outside of this?

workflows/prognostic_c48_run/runtime/loop.py
spencerkclark: Sure thing -- I agree there's not really a strong reason to continue supporting that.  

I'll refactor the na-filling out of `add_tendency` in that case as well, since it needs to occur prior to the coordinate transformation.  Along similar lines there's maybe an argument that it should have been that way all along.

workflows/prognostic_c48_run/runtime/steppers/machine_learning.py
spencerkclark: Indeed, thanks for pointing that out!

workflows/prognostic_c48_run/tests/test_regression.py
```diff
@@ -598,7 +611,9 @@ def completed_rundir(configuration, tmpdir_factory):
     tendency_dataset_path = tmpdir_factory.mktemp("tendencies")
 
     if configuration == ConfigEnum.predictor:
-        model = get_mock_predictor()
+        model = get_mock_predictor(
+            dQ1_tendency=1 / 86400, mask_lowest_level_outputs=True
```
oliverwm1, line 615: Is this change in behavior intentional? (I think the dQ1 tendency was zero before). I'm wondering whether the checksum changes are from this or from the change in how wind tendencies are applied, or both. I think it would be better to leave the behavior of this test unchanged if possible.

spencerkclark: As discussed in Slack, indeed it was both (I lumped in some changes that also allowed us to verify that the tendencies were being properly in the vertical).  We decided to revert the changes required to test the vertical orientation of tendency updates, punting that responsibility to the Python wrapper.  The checksum changes now are only due to the fact that wind tendency updates are now applied during the first iteration of the time loop (when they were not in the past).

oliverwm1 (approved):
LGTM, thanks for changes! Would be good to note in commit message that this is a breaking change for prognostic runs with wind corrections.

PR #2133, opened 2023-01-06
Combine reservoir models

AnnaKwa:
This PR makes some changes to the existing `Reservoir` and `ReservoirComputingReadout` classes to allow for easier prediction over multiple subdomains in a single call. 

- `Reservoir.increment_state` can operate on a matrix where the each column is the input from a separate subdomain. This saves iterations if we are to use the same reservoir weights across multiple subdomains. Previously, reservoir state updates only worked if the input and reservoir states were 1d arrays.
- `ReservoirComputingReadout` is refactored to save the linear regression coefficients and intercepts as attributes, rather than an sklearn regressor object. This makes combining multiple readouts more straightforward. Instead of being initialized with an sklearn regressor object, the class is initialized with the `ReadoutHyperparameters` and the sklearn regressor is created and fit when `.fit` is called. The class can be initialized with previously-fit values for these, by default they are None and updated when `ReservoirComputingReadout.fit` is called. An exception is raised if `.fit` is called but these attributes are not None.

This also adds a `CombinedReservoirComputingReadout` class that combines multiple trained readouts into a single readout. Let's say there's N subdomains, each with a reservoir state vector of length s. Unlike the method of combining reservoir prediction across subdomains, each subdomain is assumed to use a different readout, so the input to the combined readout `.predict` is a 1D array of length N x s. The coefficient matrices of N individual readouts are combined into a block diagonal matrix. The intercepts are concatenated into a 1D array of length N x s.

- [x] Tests added

- test_unit: [60%](https://output.circle-artifacts.com/output/job/8a6fd204-026b-44c7-8cce-01f6d30b9890/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/config.py
+++ b/external/fv3fit/fv3fit/reservoir/config.py
@@ -21,9 +21,6 @@ class ReservoirHyperparameters:
         W_res has shape state_size x state_size
     adjacency_matrix_sparsity: Fraction of elements in adjacency matrix
         W_res that are zero
-    output_size: Optional: size of output vector. Can be smaller than input
-        dimension if predicting on subdomains with overlapping
-        input regions. Defaults to same as input_size
     spectral_radius: Largest absolute value eigenvalue of W_res.
         Larger values increase the memory of the reservoir.
     seed: Random seed for sampling
@@ -49,8 +46,10 @@ class ReadoutHyperparameters:
     """
     linear_regressor_kwargs: kwargs to provide when initializing the
         sklearn Ridge regressor for ReservoirComputingReadout
-    square_half_hidden_state: if True, square even elements of state vector
-        as described in in Wikner+ 2020 (https://doi.org/10.1063/5.0005541)
+    square_half_hidden_state: if True, square even terms in the reservoir
+        state before it is used as input to the regressor's .fit and
+        .predict methods. This option was found to be important for skillful
+        predictions in Wikner+2020 (https://doi.org/10.1063/5.0005541)
     """
 
     linear_regressor_kwargs: dict
--- a/external/fv3fit/fv3fit/reservoir/config.py
+++ b/external/fv3fit/fv3fit/reservoir/config.py
@@ -17,7 +17,6 @@ class SubdomainConfig:
 class ReservoirHyperparameters:
     """Hyperparameters for reservoir
 
-    input_size: Size of input vector
     state_size: Size of hidden state vector,
         W_res has shape state_size x state_size
     adjacency_matrix_sparsity: Fraction of elements in adjacency matrix
@@ -35,32 +34,21 @@ class ReservoirHyperparameters:
         where all elements are sampled from random uniform distribution
         [-1, 1]. Changing this affects relative weighting of reservoir memory
         versus the most recent state.
-    ncols_state: If state is 2D (each column is a separate subdomain which uses
-        the same W_in and W_res weights) then use this to initialize the state
-        with the corret shape.
-
     """
 
-    input_size: int
     state_size: int
     adjacency_matrix_sparsity: float
     spectral_radius: float
-    output_size: Optional[int] = None
     seed: int = 0
     input_coupling_sparsity: float = 0.0
     input_coupling_scaling: float = 1.0
-    ncols_state: int = 1
-
-    def __post_init__(self):
-        if not self.output_size:
-            self.output_size = self.input_size
 
 
 @dataclass
 class ReadoutHyperparameters:
     """
-    linear_regressor_kwargs: kwargs to provide when initializing the linear
-        regressor for ReservoirComputingReadout
+    linear_regressor_kwargs: kwargs to provide when initializing the
+        sklearn Ridge regressor for ReservoirComputingReadout
     square_half_hidden_state: if True, square even elements of state vector
         as described in in Wikner+ 2020 (https://doi.org/10.1063/5.0005541)
     """
--- a/external/fv3fit/fv3fit/reservoir/config.py
+++ b/external/fv3fit/fv3fit/reservoir/config.py
@@ -35,6 +35,10 @@ class ReservoirHyperparameters:
         where all elements are sampled from random uniform distribution
         [-1, 1]. Changing this affects relative weighting of reservoir memory
         versus the most recent state.
+    ncols_state: If state is 2D (each column is a separate subdomain which uses
+        the same W_in and W_res weights) then use this to initialize the state
+        with the corret shape.
+
     """
 
     input_size: int
@@ -45,6 +49,7 @@ class ReservoirHyperparameters:
     seed: int = 0
     input_coupling_sparsity: float = 0.0
     input_coupling_scaling: float = 1.0
+    ncols_state: int = 1
 
     def __post_init__(self):
         if not self.output_size:
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -24,7 +24,7 @@ class ImperfectModel(abc.ABC):
 
 
 class ReservoirComputingModel:
-    _METADATA_NAME = "metadata.bin"
+    _RESERVOIR_SUBDIR = "reservoir"
 
     def __init__(
         self, reservoir: Reservoir, readout: ReservoirComputingReadout,
@@ -44,19 +44,18 @@ class ReservoirComputingModel:
             path: a URL pointing to a directory
         """
         self.readout.dump(path)
-        self.reservoir.dump(path)
+        self.reservoir.dump(f"{path}/{self._RESERVOIR_SUBDIR}")
 
     @classmethod
     def load(cls, path: str) -> "ReservoirComputingModel":
         """Load a model from a remote path"""
         readout = ReservoirComputingReadout.load(path)
-        reservoir = Reservoir.load(path)
+        reservoir = Reservoir.load(f"{path}/{cls._RESERVOIR_SUBDIR}")
         return cls(reservoir=reservoir, readout=readout,)
 
 
 class HybridReservoirComputingModel:
-    _READOUT_NAME = "readout.bin"
-    _METADATA_NAME = "metadata.bin"
+    _RESERVOIR_SUBDIR = "reservoir"
 
     def __init__(
         self, reservoir: Reservoir, readout: ReservoirComputingReadout,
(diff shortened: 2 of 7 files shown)


external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -20,36 +27,116 @@ def square_even_terms(v: np.ndarray, axis=1) -> np.ndarray:
...
+            )
+        linear_regressor = Ridge(**self.hyperparameters.linear_regressor_kwargs)
+        if self.square_half_hidden_state is True:
             res_states = square_even_terms(res_states, axis=1)
-        self.linear_regressor.fit(res_states, output_states)
+        linear_regressor.fit(res_states, output_states)
+        self.coefficients = linear_regressor.coef_
+        self.intercepts = linear_regressor.intercept_
 
-    def predict(self, input):
```
AnnaKwa, line 41: Dims change here because sklearn regressor assumed prediction is on 2D array, but we don't use it anymore.

external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -20,36 +27,115 @@ def square_even_terms(v: np.ndarray, axis=1) -> np.ndarray:
...
+            )
+        self.square_half_hidden_state = square_state_settings[0]
+
+    def predict(self, input: np.ndarray):
+        if self.square_half_hidden_state:
+            input = square_even_terms(input, axis=0)
+        return self.coefficients * input + self.intercepts
+
+    @classmethod
+    def load(cls, paths: Sequence[str]) -> "CombinedReservoirComputingReadout":
```
mcgibbon, line 122: Is it going to be feasible to configure/pass this a list of paths? Or should the class itself be responsible for organizing and loading sub-paths from one path? I notice there isn't a `save` method on this.

external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -35,6 +35,10 @@ class ReservoirHyperparameters:
         where all elements are sampled from random uniform distribution
         [-1, 1]. Changing this affects relative weighting of reservoir memory
         versus the most recent state.
+    ncols_state: If state is 2D (each column is a separate subdomain which uses
```
mcgibbon, line 38: Is it necessary for this and `state_size` to be passed as configuration/yaml arguments, or can they be determined automatically at train/run time from tensor shapes? This seems fairly difficult to configure and I don't see how it gets ignored for non-2D states.

mcgibbon (commented):
Is there a training function Reservoir or CombinedReservoir is used with?

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -69,7 +69,12 @@ def load(cls, path: str) -> "ReservoirComputingModel":
 
         f = io.BytesIO(mapper[cls._READOUT_NAME])
         readout_components = joblib.load(f)
-        readout = ReservoirComputingReadout(**readout_components)
+        readout_hyperparameters = dacite.from_dict(
+            ReadoutHyperparameters, readout_components.pop("hyperparameters")
+        )
+        readout = ReservoirComputingReadout(
```
mcgibbon, line 75: Can you move the low-level logic on how to load a ReservoirComputingReadout into a `ReservoirComputingReadout.load` class method? Similarly for `Reservoir`. This will avoid e.g. your duplicated change in `HybridReservoirComputingModel` when these classes are modified.

external/fv3fit/tests/reservoir/test_readout.py
```diff
@@ -27,48 +28,113 @@ def test_square_even_terms(arr, axis, expected):
...
+            linear_regressor_kwargs={}, square_half_hidden_state=False
+        ),
+        coefficients=np.ones(shape=(2, 2)),
+        intercepts=np.zeros(2),
+    )
+    with pytest.raises(ValueError):
+        CombinedReservoirComputingReadout(readouts=[readout_1, readout_2])
+
+
+def test_combined_load(tmpdir):
```
mcgibbon, line 115: Can this test be simplified by writing a `CombinedReservoirComputingReadout.dump` method, that automatically determines subdirectory names? Do we expect the ability to combine paths from potentially separate training runs to be used?

external/fv3fit/fv3fit/reservoir/config.py
AnnaKwa: Removed, and instead `reset_state` now takes an arg `input_shape` which will provide this information when it is needed to initialize the reservoir state.

external/fv3fit/tests/reservoir/test_readout.py
AnnaKwa: > Do we expect the ability to combine paths from potentially separate training runs to be used?
No, I don't see that happening.

Since training will be in parallel, I think it will be easier to have the training function handle saving all the separate readouts into subdirectories rather than waiting for all the readouts to train before combining into that class to dump.

external/fv3fit/fv3fit/reservoir/readout.py
AnnaKwa: Good point, configuring a list of >10s of paths would be annoying. Changed to list and read from sorted subdirs in `path`. I didn't include a dump method on this class because I figured the saving of each readout to the specified subdir could be handled by the training function.

mcgibbon: The one thing to keep in mind with that is the responsibility for knowing how sub-models are stored at a low level now exists in two places - your code that writes the models, and this code that reads them. Ideally you'd be able to remove that low level detail from the writing code and keep it on this class. One possible way to do something like this would be to use the same `get_combined_reservoir_paths(path, n_reservoirs)` helper function in both places.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -49,38 +43,19 @@ def dump(self, path: str) -> None:
...
-
-        mapper[self._READOUT_NAME] = self.readout.dumps()
-        metadata = {
-            "reservoir_hyperparameters": dataclasses.asdict(
-                self.reservoir.hyperparameters
-            )
-        }
-        mapper[self._METADATA_NAME] = yaml.safe_dump(metadata).encode("UTF-8")
+        self.readout.dump(path)
+        self.reservoir.dump(path)
```
mcgibbon, line 47: Is it safe to dump these both to the same path? This class doesn't know the low-level details of how these subcomponents are stored, it's very possible for them to re-use the same filenames (especially if someone modifies them after you).

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -49,38 +43,19 @@ def dump(self, path: str) -> None:
...
-        metadata = yaml.safe_load(mapper[cls._METADATA_NAME])
-
-        reservoir_hyperparameters = dacite.from_dict(
-            ReservoirHyperparameters, metadata["reservoir_hyperparameters"]
-        )
-
-        return cls(reservoir=Reservoir(reservoir_hyperparameters), readout=readout,)
+        readout = ReservoirComputingReadout.load(path)
+        reservoir = Reservoir.load(path)
+        return cls(reservoir=reservoir, readout=readout,)
```
mcgibbon, line 54: Thanks!

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -49,38 +43,19 @@ def dump(self, path: str) -> None:
...
 
 
 class HybridReservoirComputingModel:
-    _READOUT_NAME = "readout.pkl"
+    _READOUT_NAME = "readout.bin"
```
mcgibbon, line 58: Is this attribute still used?

external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -17,39 +25,175 @@ def square_even_terms(v: np.ndarray, axis=1) -> np.ndarray:
...
     def __init__(
-        self, linear_regressor: Ridge, square_half_hidden_state: bool = False,
+        self,
+        hyperparameters: ReadoutHyperparameters,
+        coefficients: Optional[np.ndarray] = None,
+        intercepts: Optional[np.ndarray] = None,
     ):
-        self.linear_regressor = linear_regressor
-        self.square_half_hidden_state = square_half_hidden_state
+        self.hyperparameters = hyperparameters
```
mcgibbon, line 71: Task: Class docstring needs to be updated for new/changed attributes.

external/fv3fit/fv3fit/reservoir/model.py
AnnaKwa: Oops, no. Removed.

external/fv3fit/fv3fit/reservoir/readout.py
AnnaKwa: That's a good idea. I just realized this part needs to be redone before merging- right now I don't think it's possible to use `ReservoirComputingModel.dump` for multiple models and save the readouts to the directory structure assumed here. Combined with the other suggestion to not dump readout/reservoir to the same path, the new assumed directory structure when reading and writing multiple subdomain models is
```
dir
|- model_0
   |- readout
       |- weights, etc.
   |- reservoir
       |- weights, etc.
|- model_1
    |- ...
```
`CombinedReservoirComputingReadout.load` will take `dir` as the path arg and search for directories ending in numbers, then open their readout subdirs.

external/fv3fit/fv3fit/reservoir/model.py
AnnaKwa: See above comment- changed to dump to readout/reservoir subdirectories.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -49,39 +43,19 @@ def dump(self, path: str) -> None:
...
-        readout = ReservoirComputingReadout(**readout_components)
-        metadata = yaml.safe_load(mapper[cls._METADATA_NAME])
-
-        reservoir_hyperparameters = dacite.from_dict(
-            ReservoirHyperparameters, metadata["reservoir_hyperparameters"]
-        )
-
-        return cls(reservoir=Reservoir(reservoir_hyperparameters), readout=readout,)
+        readout = ReservoirComputingReadout.load(path)
+        reservoir = Reservoir.load(f"{path}/{cls._RESERVOIR_SUBDIR}")
```
mcgibbon, line 53: nit (non-blocking): leave this for another PR, but it feels odd that one of these is working in a subdirectory and the other is not.

mcgibbon (approved):
LGTM. One comment to think about for later.

PR #2140, opened 2023-01-13
Combine optimizer settings for cyclegan

mcgibbon:
Currently the CycleGAN network configuration has separate configurations for the generator and discriminator optimizers. This is an issue when running hyperparameter sweeps, as there is no way to link the learning rate of these two optimizers when performing sweeps.

This PR merges the configuration for the two optimizers, which will allow hyperparameter sweeps. The relative learning rates of these two models can still be modified by setting different scale factors on the generator and discriminator loss functions.

Refactored public API:
- CycleGANNetworkConfig now has one `optimizer` setting instead of separate `discriminator_optimizer` and `generator_optimizer` settings

- test_unit: [60%](https://output.circle-artifacts.com/output/job/c4ef5656-3ffa-40d1-8307-85ee0aaff222/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
+++ b/external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
@@ -33,10 +33,8 @@ class CycleGANNetworkConfig:
     Configuration for building and training a CycleGAN network.
 
     Attributes:
-        generator_optimizer: configuration for the optimizer used to train the
-            generator
-        discriminator_optimizer: configuration for the optimizer used to train the
-            discriminator
+        optimizer: configuration for the optimizer used to train the
+            generator and discriminator
         generator: configuration for building the generator network
         discriminator: configuration for building the discriminator network
         identity_loss: loss function used to make the generator which outputs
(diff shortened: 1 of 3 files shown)


external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
```diff
@@ -33,10 +33,8 @@ class CycleGANNetworkConfig:
     Configuration for building and training a CycleGAN network.
 
     Attributes:
-        generator_optimizer: configuration for the optimizer used to train the
-            generator
-        discriminator_optimizer: configuration for the optimizer used to train the
-            discriminator
+        optimizer: configuration for the optimizer used to train the
```
AnnaKwa, line 36: Now that these are combined, can you add to this docstring describing how the discriminator and generator loss configs can be used to set different learning rates for them?

AnnaKwa (approved):
LGTM! If you could add a mention of how you can still set different LRs that would be helpful.

external/fv3fit/fv3fit/pytorch/cyclegan/cyclegan_trainer.py
mcgibbon: This should be covered well enough by the description for `discriminator_weight` and `generator_weight`. There is no way to set different learning rates, but the optimizer update is proportional to both the learning rate and the loss itself (which the weight modifies).

PR #2145, opened 2023-01-20
Add postphysics bias correction prescriber option

AnnaKwa:
Currently the prognostic run `Prescriber` option only is available for state updates in the prephysics stepper. 

This PR adds 
- a`bias_correction` top level configuration field, which uses the same config as other prescribers and will be used as a postphysics stepper. I'm open to alternate naming suggestions.
- an optional field `PrescriberConfig.tendency_variables` which is a mapping of tendency field names in the prescriber dataset to the tendency name conventions used in the prog run (e.g. "dQ1"). If this field exists, the `Prescriber` will return these tendencies (as opposed to always returning empty tendencies before)

- test_unit: [60%](https:\/\/output.circle-artifacts.com\/output\/job\/ba68c227-0050-434c-bb03-2bd2a4b97719\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/prognostic_c48_run/runtime/steppers/prescriber.py
+++ b/workflows/prognostic_c48_run/runtime/steppers/prescriber.py
@@ -24,6 +24,8 @@ class PrescriberConfig:
             defaults to True
         reference_initial_time: if time interpolating, time of first point in dataset
         reference_frequency_seconds: time frequency of dataset
+        tendency_variables: optional mapping of tendency variable names in the dataset
+            to the state variables they update
 
     Example::
 
@@ -42,6 +44,7 @@ class PrescriberConfig:
     consolidated: bool = True
     reference_initial_time: Optional[str] = None
     reference_frequency_seconds: float = 900
+    tendency_variables: Optional[Mapping[str, str]] = None
 
 
 class Prescriber:
@@ -53,6 +56,7 @@ class Prescriber:
         communicator: pace.util.CubedSphereCommunicator,
         time_lookup_function: Callable[[cftime.DatetimeJulian], State],
         variables: Mapping[str, str],
+        tendency_variables: Optional[Mapping[str, str]],
     ):
         """Create a Prescriber object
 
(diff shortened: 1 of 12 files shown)


workflows/prognostic_c48_run/runtime/steppers/prescriber.py
```diff
@@ -62,22 +66,26 @@ def __init__(
...
         """
         self._communicator = communicator
         self._time_lookup_function = time_lookup_function
         self._variables = variables
+        self._tendency_variables = tendency_variables or {}
 
     def _open_prescribed_timestep(self, time: cftime.DatetimeJulian) -> xr.Dataset:
         ds = scatter_within_tile(time, self._time_lookup_function, self._communicator)
-        return ds.rename(self._variables)
+        return ds.rename(**self._variables, **self._tendency_variables)
```
brianhenn, line 80: Previously your PR made the rank available to the stepper. Why was that and is it no longer necessary when using the prescriber?

brianhenn (commented):
Thanks, my main request is thinking about the config validation with respect to bias correction vs. ML/nudging. See below.

workflows/prognostic_c48_run/tests/_regtest_outputs/test_prepare_config.test_prepare_ml_config_regression[emulator].out
```diff
@@ -442,7 +443,6 @@ zhao_carr_emulation:
     cloud_squash: null
     enforce_conservative: false
     enforce_conservative_phase_dependent: false
-    enforce_strict_precpd_conservative: false
```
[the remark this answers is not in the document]
Comment (AnnaKwa) [reply workflows/prognostic_c48_run/tests/_regtest_outputs/test_prepare_config.test_prepare_ml_config_regression[emulator].out]: I think it's because I only ran the regtest-reset option on `test_regression.py` but there was an additional regression test in the prepare config tests for the emulation that produced this output. I reran that and I think it fixes the broken test.

workflows/prognostic_c48_run/runtime/steppers/prescriber.py
AnnaKwa: The original version was going to use a "predictor" that looked up the correction from a dataset and applied it within each rank. That needed rank info since all our current usages of ML are column local and don't care about x/y/tile location. After our discussion, I figured it was more straightforward to add a new postphysics stepper for bias correction so that it could use the existing functionality of the prescriber to assigning the local corrections.

brianhenn: Got it, thanks for the clarification

brianhenn (approved):
LGTM thanks!

PR #2149, opened 2023-01-25
Postphysics prescriber fix

AnnaKwa:
I had to fix the prescribed tendency option to add the tendency variables to the list to load in the time lookup function.
I also added the prescribed tendencies to the diagnostics as `{tendency variable}_prescribed`.

- test_unit: [60%](https:\/\/output.circle-artifacts.com\/output\/job\/0d2cc5f3-16ed-4485-8cf0-5b9de36ca489\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/workflows/prognostic_c48_run/runtime/factories.py
+++ b/workflows/prognostic_c48_run/runtime/factories.py
@@ -150,7 +150,7 @@ def get_prescriber(
         mapper = {}
     time_lookup_function = _get_time_lookup_function(
         mapper,
-        list(config.variables),
+        list(config.variables) + list(config.tendency_variables or []),
         config.reference_initial_time,
         config.reference_frequency_seconds,
     )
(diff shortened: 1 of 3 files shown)


workflows/prognostic_c48_run/runtime/factories.py
```diff
@@ -150,7 +150,7 @@ def get_prescriber(
         mapper = {}
     time_lookup_function = _get_time_lookup_function(
         mapper,
-        list(config.variables),
+        list(config.variables) + list(config.tendency_variables or []),
```
brianhenn, line 153: I'm thinking we should update the [prescriber tests](https://github.com/ai2cm/fv3net/blob/850965fe3b40128af1c3335c459cbda804968828/workflows/prognostic_c48_run/tests/test_prescriber.py#L132) to cover this (esp given that this slipped through).

brianhenn (approved):
Thanks @AnnaKwa LGTM

PR #2152, opened 2023-01-27
Always obtain offline diagnostics grid information from catalog

spencerkclark:
It is currently not possible to pass flags to the compute step in the offline diagnostics workflow.  This PR adds an option to do that.  The motivation for me adding this is to enable producing an offline report using C12 test data via the `--evaluation-grid` flag.

Refactored public API:

- The `evaluation-grid` flag associated with setting the evaluation grid resolution of the offline diagnostics workflow no longer accepts a path to a netCDF file.  It instead takes in a resolution string that can be used to read in grid information from the catalog, e.g. `"c12"`.  This makes it more consistent with other workflows (e.g. [in data loading for ML training](https://github.com/ai2cm/fv3net/blob/46afd21253778b4be32c2333c3f53311177ccbe6/external/loaders/loaders/batches/_batch.py#L73)) and more convenient (one no longer needs to construct a specific grid file or look up a path to a previous one to use an evaluation grid other than the default).

- test_unit: [60%](https://output.circle-artifacts.com/output/job/37b83e4e-f466-4829-aff7-caab15b63ff2/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
+++ b/workflows/diagnostics/fv3net/diagnostics/offline/compute.py
@@ -87,14 +87,16 @@ def _get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--evaluation-grid",
         type=str,
-        default=None,
+        default=EVALUATION_RESOLUTION,
         help=(
-            "Optional path to a grid data netcdf that sets the grid resolution at "
-            "which the diagnostics will be computed. If not provided, evaluation grid "
-            "resolution will be C48. If validation data resolution is higher than "
-            "the evaluation grid resolution by an integer multiple, the validation "
-            "data will be coarsened to the evaluation grid resolution. Otherwise the "
-            "evaluation grid resolution must match the validation data resolution."
+            f"Optional arguent that sets the grid resolution at which the diagnostics"
+            f"will be computed. If not provided, evaluation grid resolution will "
+            f"be {EVALUATION_RESOLUTION}. If validation data resolution is higher "
+            f"than the evaluation grid resolution by an integer multiple, the "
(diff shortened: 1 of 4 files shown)


frodre (approved):

brianhenn (commented):
As discussed, we'll add that the evaluation grid argument will be a catalog key.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -87,14 +87,16 @@ def _get_parser() -> argparse.ArgumentParser:
...
-        default=None,
+        default=EVALUATION_RESOLUTION,
         help=(
-            "Optional path to a grid data netcdf that sets the grid resolution at "
-            "which the diagnostics will be computed. If not provided, evaluation grid "
-            "resolution will be C48. If validation data resolution is higher than "
-            "the evaluation grid resolution by an integer multiple, the validation "
-            "data will be coarsened to the evaluation grid resolution. Otherwise the "
-            "evaluation grid resolution must match the validation data resolution."
+            f"Optional arguent that sets the grid resolution at which the diagnostics"
```
brianhenn, line 92: Specify the required string format in this docstring, eg "c12"?

brianhenn (approved):
Thanks @spencerkclark

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
spencerkclark: Thanks -- added.

brianhenn (approved):
Thanks, approving again just to acknowledge the need to modify the catalog data to get the offline test to pass.

PR #2153, opened 2023-02-01
Add combined output model

AnnaKwa:
This PR adds the `CombinedOutputModel` class, which combines multiple models with nonoverlapping output variables into a single model. It replicates functionality in the prognostic run's [`MultipleModelAdapter`](https://github.com/ai2cm/fv3net/blob/46afd21253778b4be32c2333c3f53311177ccbe6/workflows/prognostic_c48_run/runtime/steppers/machine_learning.py#L150), but sometimes it is more efficient to combine models earlier in the workflow. e.g. using the novelty detector to wrap a `CombinedOutputModel` once instead of wrapping individual models helps to cut down on evaluation time.

- [x] Tests added

- test_unit: [60%](https:\/\/output.circle-artifacts.com\/output\/job\/823fa706-f586-4868-bd7a-c0cec945e495\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)

### DIFF
--- a/external/fv3fit/tests/test_combined_output_model.py
+++ b/external/fv3fit/tests/test_combined_output_model.py
@@ -24,12 +24,15 @@ def test_CombinedOutputModel():
     X = xr.Dataset({"in0": da, "in1": da, "in2": da})
 
     combined_model = CombinedOutputModel([model0, model1])
+    assert set(combined_model.input_variables) == {"in0", "in1", "in2"}
+    assert set(combined_model.output_variables) == {"out0a", "out0b", "out1a", "out1b"}
+
     combined_prediction = combined_model.predict(X)
     np.testing.assert_array_equal(
-        combined_prediction["out0a"].values, model0.predict(X)["out0a"]
+        combined_prediction["out0a"], model0.predict(X)["out0a"]
     )
     np.testing.assert_array_equal(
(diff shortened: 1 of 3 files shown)


external/fv3fit/tests/test_combined_output_model.py
```diff
@@ -0,0 +1,78 @@
...
+    X = xr.Dataset({"in0": da, "in1": da, "in2": da})
+
+    combined_model = CombinedOutputModel([model0, model1])
+    combined_prediction = combined_model.predict(X)
+    np.testing.assert_array_equal(
+        combined_prediction["out0a"].values, model0.predict(X)["out0a"]
+    )
+    np.testing.assert_array_equal(
+        combined_prediction["out1a"].values, model1.predict(X)["out1a"]
+    )
```
oliverwm1, line 33: mostly out of curiosity, why do you need `.values` for the combined predictions but not for the individual model predictions?

oliverwm1 (approved):
Thanks, I can see this being a useful feature!

external/fv3fit/tests/test_combined_output_model.py
AnnaKwa: No reason, cleaned it up.

PR #2157, opened 2023-02-08
vcm: in-cloud <-> gridcell-mean condensate mappings

brianhenn:
Adds functions to `vcm` to convert between in the standard grid-cell mean condensate mixing ratios and the "in-cloud" equivalent. This exactly follows the [conversion](https://github.com/ai2cm/fv3gfs-fortran/blob/3238351a062385baae73f91bc121dcd77347b442/FV3/gfsphysics/physics/radiation_clouds.f#L1920) done in the GFS radiation scheme to compute effective optical properties. Adds the conversion to "in-cloud" condensate to `DerivedMapping`, and adds the conversions back grid-cell mean as `DataTransform`s; this is because the former is an unambiguous result from standard wrapper variables, but the latter is not. The use case is training ML from standard wrapper outputs that a) predicts the "in-cloud" mixing ratio and b) produces the grid-cell mean mixing ratio as a transformed output variable, to produce ML cloud fields for the ported radiation scheme.

Added public API:
- `gridcell_to_incloud_condensate`, `incloud_to_gridcell_condensate` added to `vcm`
- `incloud_[water,ice]_mixing_ratio` added to `DerivedMapping`
- `cloud_[water,ice]_mixing_ratio_from_incloud` added as `DataTransform`s

- [X] Tests added

- test_unit: [61%](https://output.circle-artifacts.com/output/job/d8f7207b-6771-4a81-8efb-db185ad6601d/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)

### DIFF
--- a/external/vcm/vcm/derived_mapping.py
+++ b/external/vcm/vcm/derived_mapping.py
@@ -384,6 +384,7 @@ def upward_heat_flux_at_surface(self):
 @DerivedMapping.register(
     "incloud_water_mixing_ratio",
     required_inputs=["cloud_amount", "cloud_water_mixing_ratio"],
+    use_nonderived_if_exists=True,
 )
 def incloud_water_mixing_ratio(self):
     result = vcm.gridcell_to_incloud_condensate(
@@ -395,6 +396,7 @@ def incloud_water_mixing_ratio(self):
 @DerivedMapping.register(
     "incloud_ice_mixing_ratio",
     required_inputs=["cloud_amount", "cloud_ice_mixing_ratio"],
+    use_nonderived_if_exists=True,
 )
 def incloud_ice_mixing_ratio(self):
     result = vcm.gridcell_to_incloud_condensate(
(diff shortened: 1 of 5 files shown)


external/vcm/vcm/derived_mapping.py
```diff
@@ -379,3 +379,55 @@ def upward_heat_flux_at_surface(self):
...
+    )
+    return result.assign_attrs(long_name="in-cloud ice mixing ratio", units="kg/kg")
+
+
+@DerivedMapping.register(
+    "gridcell_cloud_water_mixing_ratio",
+    required_inputs=["cloud_amount", "incloud_water_mixing_ratio"],
+)
+def gridcell_cloud_water_mixing_ratio(self):
+    # needs a nonstandard name to avoid circular reference
```
oliverwm1, line 413: Could you say more about this? Feels like this may be pointing to a design issue with `vcm.DerivedMapping`.

brianhenn: The DerivedMapping doesn't allow a set of conversions where 'a' depends on 'b' and 'b' depends on 'a', since it recurses to find the dependency graph of everything in the mapping (it's a DAG). I think it's fine, and in my case I think it's useful to have a different name to differentiate the output of these conversions from a standard mixing ratio field.

AnnaKwa: In the case of variables that could be obtained in different ways I think the `TransformedPredictor` may be a more suitable approach, e.g. there's a common naming pattern for transforms there that includes how the quantity was derived like `Q1_from_...`.

AnnaKwa (commented):
If I'm understanding the potential cases of converting to/from gridcell / incloud, I think making the conversion a `DataTransform` would clarify the issue that you pointed out in your comment.

external/vcm/vcm/derived_mapping.py
oliverwm1: I'm fine either way. I agree this feels like a use-case that the `TransformedPredictor` is naturally suited for. But it also requires an additional configuration burden which is a downside. If there is something meaningfully different about the differently named diagnostics, then maybe it is okay.

brianhenn: @AnnaKwa and I just discussed this and came to the conclusion that the conversion to "in-cloud" is unambiguous from wrapper variables and is appropriate for the `DerivedMapper`. The conversion back to grid-cell means from in-cloud is more problem-specific and also creates the potential circular reference, so it should be moved to a `DataTransform`. This has the nice result of only needing additional config on the output variables side, not on the training data.

AnnaKwa (approved):
Thanks for the updates, LGTM!

PR #2159, opened 2023-02-15
Add dimension ordering option to NCDirLoader

AnnaKwa:
This adds options to `fv3fit.data.NCDirLoader` to be used for reservoir computing model training.

There are a few requirements specific to this training that I wanted to be able to enforce when the tfdataset is created. The reservoir training data is first preprocessed by saving ranks separately into time chunked netcdfs, and when loaded should be ordered in time and have x, y dimensions in a specific order so that the subdomains are always divided and reconstituted in the same way. Since the data is now always passed to training as a tfdataset as opposed to xarray datasets, it is harder to enforce dimension ordering on the tensors. The dimension order in the tensors can be set in this loader using the `dim_order` field.

I also added an option for `varying_first_dim` to allow the case where batches have different numbers of timesteps.

Added public API:
-  `fv3fit.data.ReservoirTimeSeriesLoader`

Significant internal changes:
- add the optional arg `varying_first_dim` to `nc_dir_to_tfdataset` so it can be passed to the call to `nc_files_to_tf_dataset` (it is already an arg of the second function)

- [x] Tests added

- test_unit: [61%](https://output.circle-artifacts.com/output/job/5ab29059-64e1-49ab-8b2c-0025aeb18535/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -243,3 +244,58 @@ def generator():
             yield record
 
     return generator
+
+
+@register_tfdataset_loader
+@dataclasses.dataclass
+class ReservoirTimeSeriesLoader(TFDatasetLoader):
```
mcgibbon, line 251: Please add a docstring.

Could you name this class according to what it does, without mentioning "Reservoir"? While it's custom made to fit your requirements for reservoir training, it would help to make it clear at the level of implementing the data loading what those requirements are. It's also possible this loader will be re-used for other training functions with similar requirements.

To me this seems like a simple NetCDF directory loader with a small feature of being able to enforce dimension ordering different from what's in the files being loaded, and the code is mostly duplicate against NCDirLoader. Is there a reason you didn't add an `Optional[Sequence[str]]` argument to `NCDirLoader` and use that?

AnnaKwa: Moved functionality into NCDirLoader

external/fv3fit/tests/data_/test_netcdf.py
```diff
@@ -32,3 +36,75 @@ def test_NCDirLoader(tmp_path: Path):
...
+            url=data_path, dim_order=dim_order, shuffle=False, varying_first_dim=True,
+        )
+        dataset = loader.open_tfdataset(
+            local_download_path=None, variable_names=["a_sfc", "b"]
+        )
+        first_batch = next(iter(dataset))
+        assert first_batch["b"].shape == expected_shape_first_batch
+
+
+def test_NCDirLoader_varying_first_dim_allowed():
```
mcgibbon, line 82: optional: consider adding a test for when varying_first_dim is not allowed but the first dimension does vary.

external/fv3fit/tests/data_/test_netcdf.py
```diff
@@ -32,3 +36,75 @@ def test_NCDirLoader(tmp_path: Path):
...
+        with temporary_netcdfs_dir() as data_path:
+            loader = fv3fit.data.NCDirLoader(
+                url=data_path,
+                dim_order=["time", "x", "y"],
+                shuffle=False,
+                varying_first_dim=True,
+            )
+            loader.open_tfdataset(
+                local_download_path=None, variable_names=["a_sfc", "b"]
+            )
```
mcgibbon, line 110: If you isolate the `pytest.raises` to the part you expect to cause an exception, it will prevent the test from passing if we introduce a new bug into e.g. `temporary_netcdfs_dir` or `NCDirLoader`.

```suggestion
    with temporary_netcdfs_dir() as data_path:
        loader = fv3fit.data.NCDirLoader(
            url=data_path,
            dim_order=["time", "x", "y"],
            shuffle=False,
            varying_first_dim=True,
        )
        with pytest.raises(ValueError):
            loader.open_tfdataset(
                local_download_path=None, variable_names=["a_sfc", "b"]
            )
```

mcgibbon (approved):
Please include these minor clarity and safety updates, but then looks ready to merge.

external/fv3fit/fv3fit/data/netcdf/load.py
```diff
@@ -139,7 +153,17 @@ def dtype(self):
...
+            # Operates on individual data arrays to allow for inserting z dim
+            # of size 1 for 2D variables. This gives all tensors the same number
+            # of dimensions for easier concatenation later on in training.
+            tensors = {}
+            for key in variables:
+                data_array = self._ensure_consistent_dims(ds[key])
+                tensors[key] = tf.convert_to_tensor(data_array, dtype=self.dtype)
+            return tensors
+        else:
+            return to_tensor(ds, variables, dtype=self.dtype)
```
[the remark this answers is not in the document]
AnnaKwa: `mypy` forces the check whether `self.dim_order is None` to be in the `_ensure_consistent dims` function, so the whole if statement can just be removed.

PR #2161, opened 2023-02-16
Organize 1D-specific reservoir computing code

AnnaKwa:
No real code changes here, just reorganization to keep the one-dim specific code separate from later additions to `fv3fit.reservoir`. Since the periodic 1D case has slightly different configuration than the cubedsphere case, I figure it is simplest to just move all the older code to a `one_dim` subdir and have the top level of `fv3fit.reservoir` contain code for the cubedsphere problem. 

There is one small addition of a `CubedsphereSubdomainConfig`, see comment.

- test_unit: [61%](https:\/\/output.circle-artifacts.com\/output\/job\/4872b5a8-9a25-4477-a141-3ca00a0063a1\/artifacts\/0\/tmp\/coverage\/htmlcov-test_unit\/index.html)


external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -1,16 +1,16 @@
...
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass, asdict
+from typing import Tuple, Sequence
 import fsspec
 from typing import Optional
 import yaml
 
 
 @dataclass
-class SubdomainConfig:
```
AnnaKwa, line 9: This is the only new code, I left this change in this PR so that the `ReservoirTrainingConfig` used for cubedsphere data can stay in this PR.

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -1,78 +1,10 @@
 import numpy as np
 
 
-def _slice(arr: np.ndarray, inds: slice, axis: int = 0):
+def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
```
spencerkclark, line 4: I guess this is needed for future work?

external/fv3fit/fv3fit/reservoir/one_dim/domain_1d.py
```diff
@@ -0,0 +1,79 @@
...
+from ..domain import slice_along_axis
+
+
+def _slice(arr: np.ndarray, inds: slice, axis: int = 0):
+    # https://stackoverflow.com/a/37729566
+    # For slicing ndarray along a dynamically specified axis
+    # same as np.take() but does not make a copy of the data
+    sl = [slice(None)] * arr.ndim
+    sl[axis] = inds
+    return arr[tuple(sl)]
```
spencerkclark, line 11: Can probably remove this implementation, since it is now imported from `..domain`.

spencerkclark (approved):
Looks good to me.  Thanks for updating to leverage `git mv` where possible!

external/fv3fit/fv3fit/reservoir/domain.py
AnnaKwa: Yup, used in the cubedsphere subdomains.

PR #1218, opened 2021-05-19
add log viewer CLI

nbren12:
This adds a small webapp for helping to view the GFDL logs from a
segmented run. Currently, this only plots the globally averaged
variables.

Added public API:
- CLI: `prognostic_run_diags logs`

Requirement changes:
- Install streamlit (see license information below).

Screenshot: 
Screen Shot 2021-05-19 at 1 42 22 PM

PR #2163, opened 2023-02-20
fv3fit: Gridcell random forest

brianhenn:

(Delete unused sections)
Added public API:

Refactored public API:

Significant internal changes:

Requirement changes:

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/sklearn/_random_forest.py
+++ b/external/fv3fit/fv3fit/sklearn/_random_forest.py
@@ -22,6 +22,7 @@ from .._shared.training_config import RandomForestHyperparameters
 from .. import _shared
 from .._shared import (
     scaler,
+    stack,
     stack_non_vertical,
     match_prediction_to_input_coords,
     SAMPLE_DIM_NAME,
@@ -93,6 +94,7 @@ class RandomForest(Predictor):
             packer_config=hyperparameters.packer_config,
             # pass n_jobs along here to use joblib
             n_jobs=hyperparameters.n_jobs,
+            predict_columns=hyperparameters.predict_columns,
         )
         self.input_variables = self._model_wrapper.input_variables
         self.output_variables = self._model_wrapper.output_variables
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/sklearn/_random_forest.py
```diff
@@ -274,10 +278,12 @@ def _predict_on_stacked_data(self, stacked_data):
         )
 
     def predict(self, data):
-        stacked_data = stack_non_vertical(
-            safe.get_variables(data, self.input_variables)
+        input_data = safe.get_variables(data, self.input_variables)
+        stacked_data = (
+            stack(input_data)
+            if self.predict_columns is False
+            else stack_non_vertical(input_data)
```
AnnaKwa, line 285: I think you could just delete the `stack_non_vertical` function and just call `stack` in either case with different unstacked_dims args. `stack_non_vertical` is pretty old and basically only exists to do some checks in the case that `pfull` is the vertical dimension which was only the case in the very early days of our ML workflow.

external/fv3fit/tests/test_sklearn_wrapper.py
```diff
@@ -164,6 +164,35 @@ def test_predict_returns_unstacked_dims():
...
+    _, wrapper = fit_wrapper_with_gridcell_data()
+    input_data = get_unstacked_data()
+    # works
+    _ = wrapper.predict(input_data)
+    wrapper.predict_columns = True
+    with pytest.raises(ValueError):
+        # wrong number of features
+        _ = wrapper.predict(input_data)
+
+
```
AnnaKwa, line 195: Could you include a test where the trained model output a scalar for each z level point and the wrapper does the correct unstacking so that the predicted dataset has the full column of z predicted?

AnnaKwa (commented):
LGTM, could you just add a couple tests?

external/fv3fit/fv3fit/sklearn/_random_forest.py
brianhenn: Sure. `stack_non_vertical` is called in the novelty detectors etc. but switching it to `stack` isn't hard.

AnnaKwa (approved):
LGTM, thanks!

PR #2164, opened 2023-02-21
Batch ridge regressor training

AnnaKwa:
This PR adds a linear regressor class that can be incrementally trained on batches of data. The matrices $X^T X$ and $X^T Y$ are updated in place after each batch of data, the the equation $(X^T X + \lambda I )W = X^T Y$ is solved for weights W after all the batches are iterated over.
This is useful for cases where there might be a lot of samples in the reservoir computing time series training data, so that the readout does not have to be fit all at once on the entire dataset loaded into memory. 

(Delete unused sections)
Added public API:
- BatchLinearRegressor, BatchLinearRegressorHyperparameters. There is a new optional kwarg `calculate_weights` on the `ReservoirComputingReadout.fit` method, the current default of True solves the equation after each batch to retain the original behavior where calling `fit` once would automatically calculate coefficients and intercepts. This option can be set to False when fitting over batches in the training function, where `BatchLinearRegressor.calculate_weights` is called once after the loop over `update_batch` calls to save computation.

Significant internal changes:
- This regressor replaces the sklearn `Ridge` regressor in the readout class. The readout prediction equation looks slightly different now since sklearn does some rearranging of the weights into a different row/column format than here. 
 
- [x] Tests added

- test_unit: [61%](https://output.circle-artifacts.com/output/job/61c8b588-e28a-40a9-9a79-d42f012a6a8f/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/tests/reservoir/test_readout.py
```diff
@@ -156,3 +165,84 @@ def test_combined_load(tmpdir):
...
+        l2=l2, add_bias_term=True, use_least_squares_solve=True
+    )
+    batch_lr = BatchLinearRegressor(config)
+    for Xb, yb in zip(X_batches, y_batches):
+        batch_lr.batch_update(Xb, yb)
+    coefficients, intercepts = batch_lr.get_weights()
+
+    full_lr = BatchLinearRegressor(config)
+    full_lr.batch_update(X_full, y_full)
+    coefficients_full, intercepts_full = batch_lr.get_weights()
```
spencerkclark, line 217: I think this test may be falsely passing!
```suggestion
    coefficients_full, intercepts_full = full_lr.get_weights()
```

spencerkclark (commented):
Thanks @AnnaKwa -- this looks clever.  Take a look though, I think there may be a bug in including the regularization term that prevents the results of a batched fit from matching those of an unbatched one.

external/fv3fit/tests/reservoir/test_readout.py
AnnaKwa: Whoops. This typo covered up the other bug in the regularization!

external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -41,18 +41,34 @@ class ReservoirHyperparameters:
...
+@dataclass
+class BatchLinearRegressorHyperparameters:
+    """
+    l2: ridge regression coefficient
+    add_bias_term: Use default of True if input samples do not already
+        have a constant term to fit the intercept. Default True value is
+        the same behavior as sklearn regressors.
+    use_least_squares_solve: Can set to True for simple test cases
+        where the system is overdetermined and the default np.linalg.solve
+        encounters errors with nonsingular XT.X
```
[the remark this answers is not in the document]
AnnaKwa: Whoops yes, I wrote this backwards.

external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -50,6 +54,47 @@ def _sort_subdirs_numerically(subdirs: Sequence[str]) -> Sequence[str]:
...
+    def batch_update(self, X, y):
+        if self.hyperparameters.add_bias_term:
+            X = self._add_bias_feature(X)
+
+        if self.A is None and self.B is None:
+            self.A = np.zeros((X.shape[1], X.shape[1]))
+            self.B = np.zeros((X.shape[1], y.shape[1]))
+
+        reg = self.hyperparameters.l2 * np.identity(X.shape[1])
+        self.A = np.add(self.A, (np.dot(X.T, X) + reg))
```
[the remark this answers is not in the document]
AnnaKwa: Good catch, important bug.

external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -50,6 +54,59 @@ def _sort_subdirs_numerically(subdirs: Sequence[str]) -> Sequence[str]:
...
+        self.hyperparameters = hyperparameters
+        self.A = None
+        self.B = None
+
+    def _add_bias_feature(self, X):
+        return np.concatenate([X, np.ones((X.shape[0], 1))], axis=1)
+
+    def _check_X_last_col_constant(self, X):
+        last_col = X[:, -1]
+        if len(np.unique(last_col)) > 1:
```
spencerkclark, line 72: It would seem a little bit more exact to check that the last column is all ones.

spencerkclark (approved):
A couple more minor suggestions, but otherwise looks good!

PR #2165, opened 2023-02-22
`fv3fit.reservoir` cubedsphere and rank dividers

AnnaKwa:
This PR adds two classes to the reservoir code to aid in processing the cubedsphere data into the form used in training.
It also puts a short preprocessing script in `projects/reservoir` that shows how the `CubedsphereDivider` is used to split and save data into rank regions so that at training time only the rank data needed is loaded. 

For context, the training function is not PR'd yet but will take in the path to a rank of data in its training config, then train readouts for the subdomains within that rank and combine them into a single `CombinedOutputReadout`, resulting in one reservoir model per rank.

Added public API:
- `fv3fit.reservoir.CubedsphereDivider` mainly utilizes `pace.util.CubedSpherePartitioner` but with the ability to include halo points around each rank's data. Also adds some constraints on the format of data being partitioned into ranks and adds metadata to the rank datasets returned by `CubedsphereDivider.get_rank_data`. This class operates on xarray datasets as it is intended for use before the training function is called.
-  `RankDivider` partitions a rank of data into subdomains with overlapping regions. This class operates on tensors since it is used in the training function after the the main training script loads data into a `tf.dataset`.

- [x] Tests added
 
- test_unit: [61%](https://output.circle-artifacts.com/output/job/2314a350-2a39-43b2-bc49-86c528a96d09/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


projects/reservoir/fv3/save_ranks.py
```diff
@@ -0,0 +1,133 @@
...
+    data = intake.open_zarr(args.data_path).to_dask()
+
+    tstart = (
+        data.time.values[0]
+        if not args.start_time
+        else vcm.parse_datetime_from_str(args.start_time)
+    )
+    tstop = (
+        data.time.values[-1]
+        if not args.stop_time
```
mcgibbon, line 100: nit: You should generally check for None values with `if args.start_time is None` instead of casting to boolean, both because it's clearer what the intent of the code is/easier to read and because types can have some interesting casting rules (I don't know what date is the "zero" date if any that casts to False, but I wouldn't be surprised if it exists).

```suggestion
        if args.start_time is None
        else vcm.parse_datetime_from_str(args.start_time)
    )
    tstop = (
        data.time.values[-1]
        if args.stop_time is None
```

projects/reservoir/fv3/save_ranks.py
```diff
@@ -0,0 +1,133 @@
...
+    for d in ["tile", "x", "y"]:
+        dims_.append(d)
+        extent_.append(dims.pop(d))
+    for k, v in dims.items():
+        dims_.append(k)
+        extent_.append(v)
+    return dims_, extent_
+
+
+if __name__ == "__main__":
```
mcgibbon, line 84: Non-blocking suggestion: You might pass on this for this time but I'm spending a fair amount of review time manually "testing" that the argument parsing into variables done in this main block is correct. It would be better if you could isolate it into a unit test that takes in the command-line arguments and tested that the command-line args have the expected effect on internal variables like `tstart`, `tstop`, and `time_chunks`.

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +16,213 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
...
+
+
+class CubedsphereDivider:
+    def __init__(
+        self,
+        tile_layout: Layout,
+        global_dims: Sequence[str],
+        global_extent: Sequence[int],
+    ):
+        """ Class for dividing global cubedsphere data into ranks for preprocessing.
```
mcgibbon, line 28:
```suggestion
        """Class for dividing global cubed sphere data into ranks for preprocessing.
```

mcgibbon: Can you elaborate on what "preprocessing" means? Is the division into ranks needed for a preprocessing step (which one, why?) or primarily for training reasons? Is it divided in preparation for preprocessing or as part of preprocessing?

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +16,213 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
     sl = [slice(None)] * arr.ndim
     sl[axis] = inds
     return arr[tuple(sl)]
+
+
+class CubedsphereDivider:
```
mcgibbon, line 21: As long as this class is only used by `save_ranks.py`, I would suggest putting it in that file or at least in that directory. I don't see this needing to be re-used by fv3fit as written, because you preprocess data with this before passing it to fv3fit.

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +16,213 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
...
+        self.global_dims = global_dims
+        self.global_extent = global_extent
+        tile_partitioner = pace.util.TilePartitioner(layout=tile_layout)
+        self.cubedsphere_partitioner = pace.util.CubedSpherePartitioner(
+            tile=tile_partitioner
+        )
+        self.total_ranks = self.cubedsphere_partitioner.total_ranks
+
+    def get_rank_data(self, dataset: xr.Dataset, rank: int, overlap: int) -> xr.Dataset:
+        """ Returns dataset of a single rank of data, including overlap cells on edges
```
mcgibbon, line 47:
```suggestion
        """Returns dataset of a single rank of data, including overlap cells on edges
```

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +16,213 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
...
+        if overlap > 0:
+            dataset_ = append_halos(dataset, overlap)
+            x_slice = subtile_selection["x"]
+            subtile_selection["x"] = slice(
+                x_slice.start, x_slice.stop + 2 * overlap, None
+            )
+            y_slice = subtile_selection["y"]
+            subtile_selection["y"] = slice(
+                y_slice.start, y_slice.stop + 2 * overlap, None
+            )
```
mcgibbon, line 83: Suggestion: I think this code is correct but at a glance it looks like it has a bug where it's extending halos only on the right hand side. Adding a test of the data output values from this function would make it clearer that this code works correctly.

external/fv3fit/fv3fit/reservoir/domain.py
mcgibbon: nit: Elsewhere we've always called it `CubedSphere`

mcgibbon (changes requested):
Main comment is to move this code to where it's used, since it's (currently) really separate from fv3fit.

external/fv3fit/fv3fit/reservoir/domain.py
AnnaKwa: Added more description to the docstring. Effectively the preprocessing chunks it into ranks as well for more efficient loading time and memory usage.

projects/reservoir/fv3/tests/test_save_ranks_cli.py
```diff
@@ -0,0 +1,41 @@
+import pytest
```
mcgibbon, line 1: I think you could delete this test file, unless you want to make sure the CLI stays backwards-compatible and the argument names don't change. It's not necessary to test `argparse`. What I was talking about earlier was the argument handling logic that came after the argparse calls, outside of the new unit-tested routines.

external/fv3fit/fv3fit/keras/_models/shared/halos.py
```diff
@@ -45,7 +45,8 @@ def _quantities_from_dataset(
     ds: xr.Dataset, n_halo: int
 ) -> Mapping[Hashable, pace.util.Quantity]:
     return_dict = {}
-    nx, ny, nz = ds.dims["x"], ds.dims["y"], ds.dims["z"]
+    nx, ny = ds.dims["x"], ds.dims["y"]
+    nz = ds.dims["z"] if "z" in ds.dims else 0
```
mcgibbon, line 49: Setting this to zero makes me nervous because it will result in size-zero arrays if an array with a z-dimension is requested. No z-dimensions should get requested in this routine, so this would be a way to make sure that's the case and error more loudly if something breaks.

```suggestion
    # if nz is not in the dataset, should not get used
    # we set it to -1 to cause errors if used
    nz = ds.dims["z"] if "z" in ds.dims else -1
```

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +14,147 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
     sl = [slice(None)] * arr.ndim
     sl[axis] = inds
     return arr[tuple(sl)]
+
+
+class RankDivider:
```
mcgibbon, line 19: I'm guessing this gets used in your fv3fit training code, which is not in this PR.

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -8,3 +14,147 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
...
+            )
+        self.rank_dims = rank_dims
+        self.overlap = overlap
+        self.rank_extent = rank_extent
+
+        self.x_ind = rank_dims.index("x")
+        self.y_ind = rank_dims.index("y")
+
+        self._partitioner = pace.util.TilePartitioner(subdomain_layout)
+        self.n_subdomains = subdomain_layout[0] * subdomain_layout[1]
```
mcgibbon, line 63: nit (non-blocking): It seems a little random here which variables are private and public. It may make things clearer to make all private variables private.

mcgibbon (approved):
Some suggestions, but looks good!

external/fv3fit/fv3fit/reservoir/domain.py
AnnaKwa: Correct.

projects/reservoir/fv3/tests/test_save_ranks_cli.py
AnnaKwa: Ok I see. I thought you were referring to the possibility of the multiple variable names getting messed up and ending up saving data with just one, or some error like that.

PR #2171, opened 2023-03-07
Update argo jobs' wandb token

AnnaKwa:
This changes the training and offline argo templates to use the secret associated with the wandb GCP service account instead of my personal token.


frodre (approved):
The era of Anna-created jobs on wandb is coming to a close.  LGTM, thanks for the quick update.

PR #2172, opened 2023-03-09
Refactor readout

AnnaKwa:
This PR refactors parts of the reservoir code that are concerned with the fitting of the linear regression readout layer. Previously, the `BatchLinearRegressor` was a component of the `ReservoirComputingReadout` class, and `ReservoirComputingReadout.fit` was called to learn the weights. This PR removes the linear regressor and fitting from the readout, such that the readout is initialized and saved with already-trained weights. This is so that the final saved reservoir model `Predictor` (which contains a readout) does not have any code associated with fitting the model.

Below is a sketch of the part of the training code showing what objects do the fitting and what is actually saved in the Predictor. 
```
for X, Y in training_data_batches:
    batch_linear_regressor.batch_update(X, Y).   # changes the previous training code where readout.fit was called
coefficients, intercepts = batch_linear_regressor.get_weights()
readout = ReservoirComputingReadout(coefficients, intercepts)
model = ReservoirComputingModel(reservoir, readout, square_half_hidden_state)
model.dump(...)
```

The option to square even terms in the readout input is now moved to the higher level `ReservoirComputingModel` class. A subsequent PR adding a data formatter object that is an attribute of the final saved `ReservoirComputingModel` will be responsible for making sure that the data is processed at inference time (division in to subdomains, stacking/unstacking, squaring readout inputs) the same way that is handled in training.

This breaks a lot of the 1D periodic reservoir code, so I removed it and added a deprecation error if a user tries to import this code. If needed in the future I can restore this code and refactor it to use the new readout structure. Right now I don't see any use for the 1D prototype, so it's a lot easier to just remove that code than maintain it to work with refactors.

- [x ] Tests added

- test_unit: [61%](https://output.circle-artifacts.com/output/job/d826b294-945d-47d6-a63c-fc0cae1a0303/artifacts/0/tmp/coverage/htmlcov-test_unit/index.html)


external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -58,20 +58,6 @@ class BatchLinearRegressorHyperparameters:
     use_least_squares_solve: bool = False
 
 
-@dataclass
-class ReadoutHyperparameters:
```
AnnaKwa, line 62: This class is no longer needed, as the readout is not responsible for creating and training a regressor.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -1,18 +1,60 @@
-from .readout import ReservoirComputingReadout
+import fsspec
```
AnnaKwa, line 1: The changes in this module are related to moving the `square_hidden_state_input` option out of the readout and into the model's `.predict` method.

external/fv3fit/fv3fit/reservoir/readout.py
```diff
@@ -1,14 +1,10 @@
-import dacite
```
AnnaKwa, line 1: The changes in this module:
- Simplify the readout class to be initialized with coefficients and intercepts from a trained regressor, and remove the `.fit` method. The fitting of the regressor will be done in the training function by calling `BatchLinearRegressor.batch_update`. 
- Save the readout coefficients as a sparse scipy array to save on space.
- Remove the `CombinedReservoirComputingReadout` class. Instead, a standalone `combine_readouts` function can be used to combine multiple readouts into a single block diagonal coefficients matrix. (Initially when I wrote this class I thought that subdomains' readouts would be trained and saved in parallel and we would then load individual readouts before combining. After writing the training function I realized this was less efficient and all readouts could be solved at once and combined without intermediate saves.)

external/fv3fit/tests/reservoir/test_domain.py
```diff
@@ -1,6 +1,12 @@
 import numpy as np
+import os
 import pytest
-from fv3fit.reservoir.domain import slice_along_axis
```
AnnaKwa, line 3: Nothing is new here, after deleting `test_1d_domain` I consolidated the tests in `test_domain` and `test_cubedsphere_domain` into a single test module.

external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
```diff
@@ -54,19 +59,26 @@ def test_dump_load_preserves_matrices(tmpdir):
...
 
     loaded_predictor = ReservoirComputingModel.load(output_path)
-    assert _sparse_allclose(loaded_predictor.reservoir.W_in, predictor.reservoir.W_in)
-    assert _sparse_allclose(loaded_predictor.reservoir.W_res, predictor.reservoir.W_res)
+    assert _sparse_allclose(
+        loaded_predictor.reservoir.W_in.todense(), predictor.reservoir.W_in.todense()
+    )
+    assert _sparse_allclose(
+        loaded_predictor.reservoir.W_res.todense(), predictor.reservoir.W_res.todense()
+    )
```
spencerkclark, line 78: Is there a need for `_sparse_allclose` if converting the inputs to dense matrices ahead of time?  Or I guess alternatively, what prompted the need to convert the inputs to dense matrices?

spencerkclark (approved):
Impressive cleanup!  Nice to see this code maturing.  Just a minor question in one of the tests, but otherwise this looks good to me.

external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
AnnaKwa: Good catch, I removed the todense() conversion since directly comparing the sparse matrices in the objects is safest.

PR #2173, opened 2023-03-15
Fix unrecognized quotation characters from #2106

spencerkclark:
I guess this is what happens when you merge a (seemingly simple) PR that was open so long you forget whether you tested it...I'm not sure how these different quotation characters made it in here, but this fully restores things back to the way they were [here](https://github.com/ai2cm/fv3net/pull/2047/files/6dbb859f836677df03a118fb702398181a7958d4#diff-5594e73953cd77a2c885c6e61a47201d80bef8f900e034f4d7d5811ad0664489) as #2106 intended to do.  Without using the proper quotation mark characters the build process fails immediately within `install_base_software.sh`.


elynnwu (approved):
lgtm, scary quotation 😨

spencerkclark:
Thanks @elynnwu!

PR #2174, opened 2023-03-15
Move black before flake8 in pre-commit order

mcgibbon:
Currently when pre-commit runs, flake8 will complain about issues before black has a chance to correct them. This PR moves black before flake8 in the pre-commit order, so that it has a chance to fix issues before flake8 checks for them.

- [ ] Tests added


elynnwu (approved):
lgtm

PR #2175, opened 2023-03-15
Delete unused autoencoder model to avoid refactoring

mcgibbon:
An autoencoder model was created early on to help develop the CycleGAN code, as it allows testing the encoder-decoder architecture without testing the GAN loss. The downside of this is refactoring the CycleGAN code requires refactoring two training functions, one of which is no longer used.

This PR deletes the unused autoencoder training function, to make it easier to refactor the CycleGAN code.

Refactored public API:
- Deleted "autoencoder" train function

- [ ] Tests added


AnnaKwa (approved):
LGTM

PR #2176, opened 2023-03-15
add SchedulerConfig to cyclegan training config

mcgibbon:
The CycleGAN model seems to perform better when it is able to train on a high learning rate to get into the neighbourhood of correct results, then decrease to lower learning rates where artifacting behaviours can be learned out of the model.

This PR adds a "scheduler" configuration option with SchedulerConfig to allow the user to configure an arbitrary learning rate scheduler to use in CycleGAN training.

Added public API:
- added SchedulerConfig to fv3fit.pytorch.optimizer

Refactored public API:
- added "scheduler" option to CycleGANTrainingConfig, which defaults to no learning rate scheduling

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/optimizer.py
+++ b/external/fv3fit/fv3fit/pytorch/optimizer.py
@@ -1,8 +1,11 @@
 import torch.optim as optim
+import torch.optim.lr_scheduler as lr_scheduler
 import dataclasses
 from typing import (
     Any,
     Mapping,
+    Optional,
+    Protocol,
 )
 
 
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/pytorch/optimizer.py
```diff
@@ -17,3 +20,27 @@ def instance(self, params):
...
+
+
+@dataclasses.dataclass
+class SchedulerConfig:
+    name: Optional[str] = None
+    kwargs: Mapping[str, Any] = dataclasses.field(default_factory=dict)
+
+    def instance(self, optimizer) -> Scheduler:
+        if self.name is None:
+            return NullScheduler()
```
elynnwu, line 42: [nit] this is technically not of type `Scheduler`, should it be `class NullScheduler(Scheduler)`?

elynnwu (approved):

external/fv3fit/fv3fit/pytorch/optimizer.py
mcgibbon: Scheduler is a protocol type, meaning you don't have to subclass it to be an instance of it. `isinstance(null_scheduler, Scheduler)` would be True for `null_scheduler: NullScheduler`.

PR #2177, opened 2023-03-15
Copy Discriminator and ImagePool into fv3fit.pytorch.recurrent

mcgibbon:
Currently the FMR recurrent model in fv3fit.pytorch.recurrent re-uses Discriminator and ImagePool from the CycleGAN code. This means refactoring the CycleGAN model requires additionally refactoring the recurrent model, which is no longer in development.

Rather than refactor both models, this PR decouples the recurrent FMR model from Discriminator, DiscriminatorConfig, and ImagePool in the CycleGAN code by copying these into the recurrent module.

Significant internal changes:
- fv3fit.pytorch.recurrent no longer relies on CycleGAN's implementations of Discriminator, DiscriminatorConfig, or ImagePool

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/recurrent/discriminator.py
+++ b/external/fv3fit/fv3fit/pytorch/recurrent/discriminator.py
@@ -1,0 +1,126 @@
+import dataclasses
+
+import torch.nn as nn
+from toolz import curry
+import torch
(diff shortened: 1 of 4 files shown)


AnnaKwa (approved):
LGTM

PR #2182, opened 2023-03-15
Remove extra instance normalization from first discriminator layer

mcgibbon:
In comparison to the reference Zhu et al. model, our cyclegan includes an extra instance normalization on the first discriminator layer. This leads to the model producing output distributions that are scaled relative to the true distribution, as the discriminator is not able to reason about means and variability.

This PR removes the extra instance normalization from this first layer.

Significant internal changes:
- Removed extra instance normalization from Discriminator in cyclegan and recurrent.

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
+++ b/external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
@@ -77,18 +77,19 @@ class Discriminator(nn.Module):
             raise ValueError("n_convolutions must be at least 1")
         # max_filters = min_filters * 2 ** (n_convolutions - 1), therefore
         min_filters = int(max_filters / 2 ** (n_convolutions - 1))
-        convs = [
-            ConvBlock(
+        # first convolutional block must not have instance normalization, so that the
+        # discriminator can use information about the mean and standard deviation of
+        # the input data (generated images)
+        convs = []
+        convs.append(
+            convolution(
                 in_channels=in_channels,
                 out_channels=min_filters,
-                convolution_factory=curry(convolution)(
-                    kernel_size=strided_kernel_size, stride=2,
-                ),
-                activation_factory=leakyrelu_activation(
-                    negative_slope=0.2, inplace=True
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
```diff
@@ -77,18 +77,19 @@ def __init__(
...
-                ),
-                activation_factory=leakyrelu_activation(
-                    negative_slope=0.2, inplace=True
-                ),
+                kernel_size=strided_kernel_size,
+                stride=2,
             )
-        ]
+        )
+        convs.append(leakyrelu_activation(negative_slope=0.2, inplace=True)())
```
gideonite, line 92: So it looks like you have separated convs from activation functions? In this case, does the list `convs` still apply? Maybe it is now layers since you have a conv layer followed by a ReLU activation "layer".

Also, curious what the motivation for `in_place` is.

gideonite (approved):
Looks great but I don't have a lot of context for your project (yet!)

external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
mcgibbon: Yes, I've separated it just because in a later PR I add a geographically-resolved bias module between the convolution operation and the activation function.

These could be renamed "modules", but I find it a bit more descriptive to call it convs, because it's still applying a series of convolutional layers. The conv, geographic bias, and activation modules together form a single convolutional layer with a geographically-dependent bias (at least, that's how I think about it).

`inplace` just makes it run a little faster if you know you aren't re-using the input somewhere else.

mcgibbon: It might make it more clear if I combine these 2-3 layers in a `nn.Sequential` container so it is indeed a list of convolutional layers - I'll try to remember to do this in a later PR / on my working branch.

PR #2183, opened 2023-03-15
Fix CycleGAN checkpoint dumping

mcgibbon:
Currently CycleGAN's checkpoints are dumped directly instead of through `fv3fit._shared.io.dump`, which means they cannot be loaded with `fv3fit.load` as they are missing a `name` file. They also can only be saved to local paths, as the code used for path interpolation does not work with gcsfs paths.

This PR fixes both issues.

Refactored public API:
- CycleGAN checkpoint models can now be loaded with `fv3fit.load`
- CycleGAN training will now report checkpoint save paths through its logger
- CycleGAN checkpoints can now be saved to remote paths

- [ ] Tests added


AnnaKwa (approved):
LGTM

PR #2184, opened 2023-03-15
Add use_geographic_embedded_bias options in CycleGAN

mcgibbon:
This PR adds `use_geographic_embedded_bias` options to DiscriminatorConfig and GeneratorConfig used by the CycleGAN training, which when enabled (default disabled) will add geographically-varying bias vectors to the first model layer output. This allows significantly more geographically-resolved features than the previous `use_geographic_bias` option, as bias is added in embedded space, more similar to what's done in FourCastNet.

Refactored public API:
- fv3fit.pytorch.cyclegan.DiscriminatorConfig and fv3fit.pytorch.cyclegan.GeneratorConfig now have `use_geographic_embedded_bias` options

- [ ] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
+++ b/external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
@@ -5,6 +5,7 @@ from toolz import curry
 import torch
 from .modules import (
     ConvolutionFactory,
+    GeographicBias,
     single_tile_convolution,
     leakyrelu_activation,
     ConvBlock,
@@ -24,17 +25,27 @@ class DiscriminatorConfig:
         n_convolutions: number of strided convolutional layers before the
             final convolutional output layer, must be at least 1
         kernel_size: size of convolutional kernels
+        strided_kernel_size: size of convolutional kernels in the
+            strided convolutions
         max_filters: maximum number of filters in any convolutional layer,
             equal to the number of filters in the final strided convolutional layer
+        use_geographic_embedded_bias: if True, include a layer that adds a trainable
+            bias vector after the initial encoding layer. This bias is a
+            function of horizontal coordinates.
     """
 
     n_convolutions: int = 3
     kernel_size: int = 3
     strided_kernel_size: int = 3
     max_filters: int = 256
+    use_geographic_embedded_bias: bool = False
 
     def build(
-        self, channels: int, convolution: ConvolutionFactory = single_tile_convolution,
+        self,
+        channels: int,
+        nx: int,
+        ny: int,
+        convolution: ConvolutionFactory = single_tile_convolution,
     ):
         return Discriminator(
             in_channels=channels,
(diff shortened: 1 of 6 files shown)


external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
```diff
@@ -89,7 +111,12 @@ def __init__(
                 stride=2,
             )
         )
+        if use_geographic_embedded_bias:
+            convs.append(
+                GeographicBias(channels=min_filters, nx=int(nx / 2), ny=int(ny / 2))
```
oliverwm1, line 116: This divide-by-2 for `nx` and `ny` comes from the `stride=2` for the first convolution layer? Any concern about not having the geographic bias on the native grid? (or put another way, could you do embedding without stride=2 convolution?)

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
```diff
@@ -29,6 +29,19 @@ def no_activation():
     return nn.Identity()
 
 
+class GeographicBias(nn.Module):
+    """
+    Adds a trainable bias vector of shape [6, channels, nx, ny] to the layer input.
```
oliverwm1, line 34: Where does this hard-coded value of 6 come from? Should it be parametrized?

oliverwm1 (commented):
Generally looks good but had a couple questions, and wonder whether the asymmetry in dimension size for the generator and discriminator geographic biases might be an issue.

external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
mcgibbon: I'm not worried, unless I see poor performance in training. I'd be more worried about modifying the discriminator architecture by adding a non-strided convolution (if I did nonlinear embedding).

If I did first linearly embed into a `min_filters` space, then adding the bias before or after a strided convolution which follows would be identical, as it's a linear transformation. It's just that the bias would be over-defined and 75% of the information would be discarded.

external/fv3fit/fv3fit/pytorch/cyclegan/modules.py
mcgibbon: 6 is the number of tiles on a cubed sphere. Since this is a module to add a bias which is resolved on the cubed sphere, I thought it should be fine for the implementation to require a tile dimension.

I've added a short statement to the docstring.

external/fv3fit/fv3fit/pytorch/cyclegan/discriminator.py
oliverwm1: Okay, not sure I totally follow, but if performance is as expected, then I agree it's likely okay.

oliverwm1 (approved):

PR #2185, opened 2023-03-15
Fix/cyclegan gen odd kernel size

mcgibbon:
Currently the CycleGAN generator cannot use an odd strided_kernel_size, as it produces incompatible output shapes. This PR supplies an `output_padding` argument in that case which fixes the output shapes, so that an odd strided_kernel_size can be used.

Using a strided kernel size of 3 instead of 4, like was done in the original Zhu et al., appears to give improved performance and skill.

Refactored public API:
- Fixed bug to allow cyclegan generator to use an odd `strided_kernel_size`.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/pytorch/cyclegan/generator.py
+++ b/external/fv3fit/fv3fit/pytorch/cyclegan/generator.py
@@ -149,6 +149,7 @@ class Generator(nn.Module):
                 convolution_factory=curry(convolution)(
                     kernel_size=config.strided_kernel_size,
                     stride=2,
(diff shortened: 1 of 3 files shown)


AnnaKwa (approved):
LGTM. In the `GeneratorConfig` docstring there's a note about even strided kernel size being recommended, is this no longer true after this fix?

mcgibbon:
> LGTM. In the `GeneratorConfig` docstring there's a note about even strided kernel size being recommended, is this no longer true after this fix?

That's correct, I've updated the docstring now.

PR #2186, opened 2023-03-15
Add keras autoencoder

AnnaKwa:
Adds a basic autoencoder class with encode/decode methods. It is an optional component in reservoir training (followup PR) to reduce input dimensionality.
Right now its only usage is for loading an already-trained autoencoder for use in the reservoir training. A training function for a column autoencoder is planned for a future PR, which is where the code for encoder/decoder building and related tests will be added.


external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -0,0 +1,43 @@
+import os
+import tensorflow as tf
+from fv3fit._shared import get_dir, put_dir
+
+
+class Autoencoder(tf.keras.Model):
```
mcgibbon, line 6: Please add type hints to the methods of this function. Some docstrings would also be nice, but type hints are good enough.

For the models, you could define a protocol class:
```from typing import Protocol

class Model(Protocol):

    def predict(self, x: tf.Tensor) -> tf.Tensor:
        ...

    def __call__(self, x: tf.Tensor) -> tf.Tensor:
        ...

    def save(self, path: str) -> None:
        ...
```

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -0,0 +1,43 @@
...
+    _DECODER_NAME = "decoder.tf"
+
+    def __init__(self, encoder, decoder):
+        super(Autoencoder, self).__init__()
+        self.n_latent_dims = encoder.layers[-1].output.shape[-1]
+        self.encoder = encoder
+        self.decoder = decoder
+
+    def call(self, x):
+        encoded = self.encoder(x)
```
mcgibbon, line 17: Is there a reason you call `self.encoder.__call__` in `call` but `self.encoder.predict` in `encode`?

external/fv3fit/fv3fit/reservoir/autoencoder.py
mcgibbon: Scratch that, just use tf.keras.Model for models. I just realized dump and then load doesn't work if it's not a keras model.

external/fv3fit/fv3fit/reservoir/autoencoder.py
AnnaKwa: There's some differences between `Model.call` and `Model.predict` (https://keras.io/getting_started/faq/#whats-the-difference-between-model-methods-predict-and-call) so to be safe I just tried to be consistent. I figured if calling encode/decode the user is always just wanting the output values, versus the other potential uses of `__call__`.

mcgibbon: Optional suggestion: rename this method to `__call__` and add a `.predict` method which uses `.predict` on the underlying encoder and decoder.

mcgibbon (approved):
LGTM, one optional suggestion.

external/fv3fit/fv3fit/reservoir/autoencoder.py
AnnaKwa: As per https://stackoverflow.com/a/59407927/5597448 I opted to keep the current version (define `call`, not `__call__`)

PR #2187, opened 2023-03-16
Add handling for 'time' variable to WindowedZarrLoader

mcgibbon:
This PR adds the ability to load time as a number of seconds since epoch to the WindowedZarrLoader.

Refactored public API:
- If "time" is requested from the WindowedZarrLoader, it will be returned in records as a number of samples since epoch (since 1970-01-01)
- SyntheticZarrLoader can now include time as a record variable
- WindowedZarrLoader now returns variables in the dimension order they were requested. This behavior was unintentionally changed in an earlier commit.
- WindowedZarrLoader can now return variables which are missing any of the requested dimensions, not just the last dimension. This was required in order to return the "time" variable.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/data/tfdataset.py
+++ b/external/fv3fit/fv3fit/data/tfdataset.py
@@ -9,24 +9,18 @@ from ..tfdataset import generator_to_tfdataset
 import tempfile
 import xarray as xr
 import numpy as np
-from vcm import safe
 
 
 SAMPLE_DIM_NAME = "_fv3fit_sample"
 
 
-def stack(ds: xr.Dataset, unstacked_dims: Sequence[str]):
+def stack(ds: xr.DataArray, unstacked_dims: Sequence[str]):
     stack_dims = [dim for dim in ds.dims if dim not in unstacked_dims]
     unstacked_dims = [dim for dim in unstacked_dims if dim in ds.dims]
     if len(stack_dims) == 0:
         ds_stacked = ds.expand_dims(dim=SAMPLE_DIM_NAME, axis=0)
     else:
-        ds_stacked = safe.stack_once(
-            ds,
-            SAMPLE_DIM_NAME,
-            stack_dims,
-            allowed_broadcast_dims=list(unstacked_dims) + ["time", "dataset"],
-        )
+        ds_stacked = ds.stack({SAMPLE_DIM_NAME: stack_dims})
     return ds_stacked.transpose(SAMPLE_DIM_NAME, *unstacked_dims)
 
 
@@ -49,10 +43,7 @@ class VariableConfig:
     def get_record(self, name: str, ds: xr.Dataset, unstacked_dims: Sequence[str]):
         if self.times == "start":
             ds = ds.isel(time=0)
-        # must stack array instead of dataset dataset to avoid attaching dimensions
-        # to it that happen to exist in other variables
         array = stack(ds[name], unstacked_dims)
-        print(name, array.shape, stack(ds[[name]], unstacked_dims)[name].shape)
         return array.values
 
 
--- a/external/fv3fit/fv3fit/data/tfdataset.py
+++ b/external/fv3fit/fv3fit/data/tfdataset.py
@@ -1,5 +1,6 @@
 import dataclasses
 import os
+import cftime
 from typing import List, Mapping, Sequence, Optional
 import tensorflow as tf
 from .base import TFDatasetLoader, register_tfdataset_loader, tfdataset_loader_from_dict
@@ -16,8 +17,7 @@ SAMPLE_DIM_NAME = "_fv3fit_sample"
 
 def stack(ds: xr.Dataset, unstacked_dims: Sequence[str]):
     stack_dims = [dim for dim in ds.dims if dim not in unstacked_dims]
-    unstacked_dims = [dim for dim in ds.dims if dim in unstacked_dims]
-    unstacked_dims.sort()  # needed to always get [x, y, z] dimensions
+    unstacked_dims = [dim for dim in unstacked_dims if dim in ds.dims]
     if len(stack_dims) == 0:
         ds_stacked = ds.expand_dims(dim=SAMPLE_DIM_NAME, axis=0)
     else:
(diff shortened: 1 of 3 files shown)


external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -16,8 +17,7 @@
 
 def stack(ds: xr.Dataset, unstacked_dims: Sequence[str]):
     stack_dims = [dim for dim in ds.dims if dim not in unstacked_dims]
-    unstacked_dims = [dim for dim in ds.dims if dim in unstacked_dims]
-    unstacked_dims.sort()  # needed to always get [x, y, z] dimensions
```
oliverwm1, line 20: was this `.sort()` causing an issue for your feature? Or did you just remove since it's unnecessary?

external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -119,6 +118,10 @@ class WindowedZarrLoader(TFDatasetLoader):
     necessarily appear in each iteration over the tf.data.Dataset. Each sample window
     is independently selected along any stacked (sample) dimensions.
 
+    If the "time" variable itself is loaded, it will be converted to a number of
+    seconds since 1970-01-01. It is also required that the time variable be
+    one-dimensional in "time" (as opposed to varying along some perturbation axis).
```
oliverwm1, line 123: This conversion is pretty specific for the general purpose `WindowedZarrLoader`. (looks like the class mostly just does subselection right now, no conversion of actual values?)

external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -47,21 +47,20 @@ def __post_init__(self):
             raise TypeError("times must be one of 'window' or 'start'")
 
     def get_record(self, name: str, ds: xr.Dataset, unstacked_dims: Sequence[str]):
-        for dim in unstacked_dims[:-1]:
-            if dim not in ds[name].dims:
-                raise ValueError("variable {} has no dimension {}".format(name, dim))
         if self.times == "start":
             ds = ds.isel(time=0)
-        ds = stack(ds, unstacked_dims)
+        # must subset dataset to one variable to avoid attaching dimensions
```
brianhenn, line 52: Couldn't we just make `stack` operate on data arrays instead and not have to worry about this?

external/fv3fit/fv3fit/data/tfdataset.py
brianhenn: Would tend to agree with Oli, could the time conversion be moved into the CycleGAN training function? I know the CycleGAN is the only user of time from the `WindowedZarrLoader` at the moment but maybe it won't be.

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: It was causing issues. The tests don't care, but it broke my models for production because my post-evaluation scripts expect specific dimension orders, and they were no longer followed when I added `.sort` in a previous commit.

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: It is and it isn't - the time variable has to be converted to a data type tensorflow can parse (there's no such thing as a cftime.DatetimeJulian tensor). This was the most unopinionated conversion I could think of.

external/fv3fit/fv3fit/data/tfdataset.py
brianhenn: Seems fine, but can you jog my memory why you had commented in before that a specific order was needed?

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: We probably could. It may even "just work" - I'll try.

external/fv3fit/fv3fit/data/tfdataset.py
oliverwm1: Yes... this code is used is many places, so some care required here IMO.

external/fv3fit/fv3fit/data/tfdataset.py
brianhenn: OK thanks, I wasn't connecting that time could never be in a tf dataset as a datetime object, I'm fine keeping it then. Maybe just clean up the `records` function a bit by moving the specific time handling into its own function, and think about whether this could just be tested via a functional unit test instead of (in addition to?) a test on the loader.

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: It needed some other changes, but they're for the better. Updated.

external/fv3fit/fv3fit/data/tfdataset.py
mcgibbon: I think the code itself for the time handling within `generator` is simple enough that it doesn't need a test - it's the same datetime conversion code people copy paste from stack overflow all the time. When I added the loader test, it caught a fair number of failures to do with making sure the result is aligned correctly with the other data and has the right dimensions, and to do with things I hadn't thought of when loading different types of zarr data (e.g. the data type is different if you save cftime vs datetime). So I think the larger-scale test is needed.

It's actually quite nice having tests at the level of the loader because they're meant to be fairly backwards-compatible and really thought of as a "unit", so changes to implementation details rarely require the tests themselves to be updated/refactored, only added to.

brianhenn (approved):
Thanks @mcgibbon. Approving with a couple requests for continued cleanup of the stacking routine.

external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -49,10 +49,11 @@ def __post_init__(self):
     def get_record(self, name: str, ds: xr.Dataset, unstacked_dims: Sequence[str]):
         if self.times == "start":
             ds = ds.isel(time=0)
-        # must subset dataset to one variable to avoid attaching dimensions
+        # must stack array instead of dataset dataset to avoid attaching dimensions
         # to it that happen to exist in other variables
-        ds = stack(ds[[name]], unstacked_dims)
-        return ds[name].values
+        array = stack(ds[name], unstacked_dims)
```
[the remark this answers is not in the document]
mcgibbon: Good suggestion, updated.

external/fv3fit/fv3fit/data/tfdataset.py
```diff
@@ -49,10 +49,11 @@ def __post_init__(self):
     def get_record(self, name: str, ds: xr.Dataset, unstacked_dims: Sequence[str]):
         if self.times == "start":
             ds = ds.isel(time=0)
-        # must subset dataset to one variable to avoid attaching dimensions
+        # must stack array instead of dataset dataset to avoid attaching dimensions
```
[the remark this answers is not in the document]
mcgibbon: With your suggested updates this is a little more obvious. Removed.

PR #2188, opened 2023-03-17
Make offline compute compatible with scream

elynnwu:
In this PR, we added the capability of taking native scream model output and running it through `fv3fit.train` and `fv3net.diagnostics.offline.compute`.

Added public API:
- `format_output.py`: this is a post processing script for native scream output. Currently, we also include making random tendencies values so that it can go through training and offline diagnostics workflow.
- `open_nudge_to_fine_scream`: similar to `open_nudge_to_fine` but specifically for scream
- `gsrm_name_from_resolution_string`: new function in vcm to determine whether to use fv3 or scream based on resolution string

Refactored public API:
- `BatchesFromMapperConfig` now has `ptop`. Defaults are set to fv3 setting, but can be changed in the input namelist
- `offline/compute.py`: `_get_transect` takes in `ptop` as input
- DiagArg: now has `horizontal_dims` and `vertical_dim`

Significant internal changes:
- `vcm/calc/thermo/constants.py`: changed `TOA_PRESSURE` to public variable so that it can be used in other package
- catalog: add scream ne30 grid and land sea mask
- Changed `HORIZONTAL_DIMS` to `HORIZONTAL_DIMS_FV3`

- [x] Tests added


external/loaders/loaders/batches/_batch.py
```diff
@@ -89,6 +92,8 @@ def __post_init__(self):
                 "Timesteps provided for selection must be unique. "
                 f"Duplicated times were found: {duplicate_times}"
             )
+        if "ne" in self.res:
+            self.gsrm_name = "scream"
```
oliverwm1, line 96: Why provide `gsrm_name` as a config option if we just overwrite it based on `res`? If we can just use the existing `res` parameter to describe the horizontal grid, that would be better.

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -179,10 +182,10 @@ def open_nudge_to_fine(
...
         "eastward_wind_tendency_due_to_nudging": "dQu",
         "northward_wind_tendency_due_to_nudging": "dQv",
-        "tendency_of_air_temperature_due_to_fv3_physics": "pQ1",
-        "tendency_of_specific_humidity_due_to_fv3_physics": "pQ2",
-        "tendency_of_eastward_wind_due_to_fv3_physics": "pQu",
-        "tendency_of_northward_wind_due_to_fv3_physics": "pQv",
+        f"tendency_of_air_temperature_due_to_{gsrm_name}_physics": "pQ1",
+        f"tendency_of_specific_humidity_due_to_{gsrm_name}_physics": "pQ2",
+        f"tendency_of_eastward_wind_due_to_{gsrm_name}_physics": "pQu",
+        f"tendency_of_northward_wind_due_to_{gsrm_name}_physics": "pQv",
```
oliverwm1, line 188: How much control do we have over the SCREAM output names? Do you know what the names of these quantities (physics tendencies) will be? 

And if they're not named `tendency_of_air_temperature_due_to_scream_physics`, probably we should just not make this change. Note that this function does not fail if these variables don't exist in the given dataset.

external/vcm/vcm/catalog.yaml
```diff
@@ -998,3 +998,27 @@ sources:
...
+    description: Lat, lon, and area of NE30 data
+    driver: zarr
+    metadata:
+      grid: ne30
+      variables:
+        - area
+        - lat
+        - lon
+    args:
+      urlpath: "gs://vcm-scream/latLonArea/ne30/ne30.zarr/"
```
oliverwm1, line 1012: I wonder if we should save this `area` data in units of `m^2` instead of the current `rad^2`? High potential for bugs, although I understand a desire to be consistent with SCREAM output.

oliverwm1: Also another question: did you reshape this data to 6 x 90 x 40? Or is that the native output shape?

external/vcm/vcm/catalog.yaml
```diff
@@ -998,3 +998,27 @@ sources:
...
+      consolidated: true
+
+  landseamask/ne30:
+    description: land_sea_mask of NE30 data
+    driver: zarr
+    metadata:
+      grid: ne30
+      variables:
+        - land_sea_mask
```
oliverwm1, line 1021: I don't *think* it will cause issues, but just FYI the FV3 land sea mask includes a sea-ice mask (value=2), whereas it seems the SCREAM one is just 0 (ocean) or land (1).

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -143,11 +150,32 @@ def insert_rmse(ds: xr.Dataset):
 
 
 def load_grid_info(res: str = "c48"):
+    if "ne" in res:
+        return load_grid_info_scream(res)
+    elif "c" in res:
```
oliverwm1, line 155: We should probably define this convention somewhere centralized (in `vcm`?) instead of having similar functions that distinguish between scream and fv3 "res" strings in many places.

I'd also suggest checking `res.startswith("ne")` and `res.startswith("c")` instead of `in`. (Maybe another grid will come along with a `c` in it somewhere!)

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -143,11 +150,32 @@ def insert_rmse(ds: xr.Dataset):
...
+    return safe.get_variables(grid_info, GRID_INFO_VARS_FV3).drop_vars(
+        "tile", errors="ignore"
+    )
+
+
+def load_grid_info_scream(res):
+    grid = catalog[f"grid/{res}"].read()
+    land_sea_mask = catalog[f"landseamask/{res}"].read()
+    r_earth = 6.376e6
+    grid.area.values = grid.area.values * r_earth ** 2  # convert from radian^2 to m^2
```
oliverwm1, line 175: This is a bit scary... do we expect user to remember to do this every time they load the "area" data for SCREAM? In practice this catalog item gets used a lot in notebooks and other places.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -173,7 +173,9 @@ def _consolidate_dimensioned_data(ds):
     return ds_diagnostics, ds_scalar_metrics
 
 
-def _get_transect(ds_snapshot: xr.Dataset, grid: xr.Dataset, variables: Sequence[str]):
+def _get_transect(
+    ds_snapshot: xr.Dataset, grid: xr.Dataset, variables: Sequence[str], ptop: float,
```
oliverwm1, line 177: the added `ptop` argument is not used in this function.

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+    "VerticalLayerInterface",
+    # "pseudo_density",  # this is pressure thickness
+    "area",
+    "lat",
+    "lon",
+    "hyai",
+    "hybi",
+]
+
+NP = 4
```
oliverwm1, line 31: Is this used anywhere?

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+NPG = 2
```
oliverwm1, line 32: I guess this has some standard meaning for SCREAM, but I have no idea what it represents. Either give a more useful variable name (preferred) or add a short comment.

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+            "z": np.arange(0, levs),
+            "z_interface": np.arange(0, levs + 1),
+            "ncol": np.arange(0, Ncols),
+            "tile": np.arange(0, 6),
+            "x": np.arange(0, x_dim),
+            "y": np.arange(0, y_dim),
+        }
+    )
+    ds = ds.stack(ncol=("tile", "x", "y"))
+    ds = ds.unstack("ncol")
```
oliverwm1, line 163: Why stack and then unstack?

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+        data_vars="minimal",
+    )
+    if "horiz_winds" in output_variables:
+        u = ds.horiz_winds.isel(dim2=0).rename({"x_wind"})
+        v = ds.horiz_winds.isel(dim2=1).rename({"y_wind"})
+
+    rename_vars: Mapping[Hashable, Hashable] = {
+        "T_mid": "air_temperature",
+        "qv": "specific_humidity",
+        "omega": "upward_air_velocity",
```
oliverwm1, line 99: The standard name for omega is `lagrangian_tendency_of_air_pressure`. `upward_air_velocity` might get confused with the vertical velocity in units of m/s.

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+        data_path: path to input data directory, remote or local
+        file_name: either a single file name
+            or a wildcard pattern to match multiple names
+        output_path: path to output data directory, remote or local
+        output_file_name: name of output file
+        output_variables: list of variables to output
+    """
+    Ncols = NE ** 2 * 6 * NPG ** 2
+    x_dim = int(Ncols / 6 / NE / 3)
+    y_dim = int(Ncols / 6 / x_dim)
```
oliverwm1, line 85: I don't follow rationale for this definition of `x_dim` and `y_dim`. Let's chat offline.

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
+from typing import Hashable, List, Mapping
```
oliverwm1, line 1: How would this script get called? Are you imagining some automated post-processing, or more of an ad-hoc approach? Will the script require a lot of memory to run for our anticipated datasets?

For now, this script may be better suited for the `fv3net/projects` location, which is a good place for untested code. Everything else in `post_process_run` is quite well unit tested and more modular so I'm not sure this script is currently a good fit.

external/loaders/loaders/batches/_batch.py
```diff
@@ -138,8 +144,9 @@ def batches_from_mapper(
     shuffle_timesteps: bool = True,
     shuffle_samples: bool = False,
     data_transforms: Optional[Sequence[Mapping]] = None,
+    gsrm_name: str = "fv3",
```
oliverwm1, line 147: Same point as above in terms of `res` and `gsrm_name` arguments.

oliverwm1 (changes requested):
Great that we can get model training/offline diagnostics compatible with scream!

Could we get away without adding a new `gsrm_name` config parameter?

elynnwu:
Sample native scream output: `gs://vcm-ml-scratch/elynnw/scream/output.scream.monthly.AVERAGE.nmonths_x1.0001-02-01-00000.nc`

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+    ds = ds.assign_coords(
+        {
+            "z": np.arange(0, levs),
+            "z_interface": np.arange(0, levs + 1),
+            "ncol": np.arange(0, Ncols),
+            "tile": np.arange(0, 6),
+            "x": np.arange(0, x_dim),
+            "y": np.arange(0, y_dim),
+        }
+    )
```
spencerkclark, line 161: Is it necessary to assign these coordinates, or can we get away without it?

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+    ptop = 225.52
+    delp = p_mid.copy()
+    dz = z_int[:, :, 0:-1] - z_int[:, :, 1:]
+    p_int = z_int.copy()
+    p_int[:, :, 0] = ptop
+    for k in range(1, levs):
+        p_int[:, :, k] = (
+            dz[:, :, k - 1] * p_mid[:, :, k] + dz[:, :, k] * p_mid[:, :, k - 1]
+        ) / (dz[:, :, k - 1] + dz[:, :, k])
+    p_int[:, :, -1] = ps / 2.0  # surface pressure bug
```
spencerkclark, line 57: Is this method more accurate than using `hyai`, `hybi`, and `ps` to compute the pressure on the layer interfaces, because SCREAM is a non-hydrostatic model?  It's too bad they don't have this output as a diagnostic directly...

More generally it would probably be good to have this written in more xarray-style code, e.g. without the assumption of array shapes and position of the vertical dimension.

workflows/post_process_run/scream_post/format_output.py
```diff
@@ -0,0 +1,186 @@
...
+    "SW_clrsky_flux_dn@bot",
+    "LW_clrsky_flux_dn@bot",
+    "LW_clrsky_flux_up@bot",
+    "SW_clrsky_flux_up@tom",
+    "LW_clrsky_flux_up@tom",
+    "VerticalLayerInterface",
+    "p_mid",
+    "ps",
+    "VerticalLayerInterface",
+    # "pseudo_density",  # this is pressure thickness
```
spencerkclark, line 23: Is this not included in the dataset?

workflows/post_process_run/scream_post/format_output.py
spencerkclark: Also interested in this just for my own understanding.

spencerkclark (commented):
Thanks @elynnwu -- I think @oliverwm1 covered the important things.  I had a few more minor comments as I looked this over.

external/loaders/loaders/batches/_batch.py
elynnwu: We provide `gsrm_name` inside `mapper_config`. This is because functions like `open_nudge_to_fine` does not know anything about resolution but does need the name to read in the correct physics tendency variables. 
Then in the general config, instead of adding another `gsrm_name`, we use `res` to determine `gsrm_name`.

external/loaders/loaders/mappers/_nudged/_nudged.py
elynnwu: Apparently, it is possible for us to change scream output names (in the future). Currently there isn't any physics tendency variables, so I'm keeping the same naming convention here when making fake data.

external/vcm/vcm/catalog.yaml
elynnwu: Right, we only have 0 and 1 in this land sea mask. I'm waiting to swap this to scream's land sea mask when I have access to it, not sure if that one will have sea ice.

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
elynnwu: It actually needs to be passed to `interpolate_to_pressure_levels`, updated now

workflows/post_process_run/scream_post/format_output.py
elynnwu: Nope, deleted

workflows/post_process_run/scream_post/format_output.py
elynnwu: No longer need these

workflows/post_process_run/scream_post/format_output.py
elynnwu: No reshaping

workflows/post_process_run/scream_post/format_output.py
elynnwu: No renaming for now

workflows/post_process_run/scream_post/format_output.py
elynnwu: These were some magic integers I came up with, but per our discussion, no more reshaping. We use native scream output shape.

workflows/post_process_run/scream_post/format_output.py
elynnwu: Moved to `fv3net/projects/scream`. This script is called once offline to transform scream native output to zarr, it also includes some hacks to work with the current set of scream output. We expect to remove these hacks (e.g., surface pressure is 2x, horizontal winds are written in the same variable but will be split into 2) once we have the new scream output. Wanted to still commit this so that the dataset is reproducible

external/loaders/loaders/batches/_batch.py
elynnwu: Responded above

workflows/post_process_run/scream_post/format_output.py
elynnwu: We can get way without it, removed.

workflows/post_process_run/scream_post/format_output.py
elynnwu: In the future, we should have `delp` directly as part of the output. Changed to using `hyai`, `hybi`, `p0`, and `ps` to calculate the full pressure.

workflows/post_process_run/scream_post/format_output.py
elynnwu: Not currently, but it should be soon.

spencerkclark (commented):
Thanks for the updates @elynnwu!  Just a question / suggestion related to time.  I'll leave it to @oliverwm1 to review the new dimension name code and give the final approval since that was something you both discussed.

external/loaders/loaders/batches/_batch.py
```diff
@@ -192,10 +200,17 @@ def batches_from_mapper(
...
+            transforms += [
+                add_grid_info(res),
+                add_wind_rotation_info(res),
+            ]
+        elif gsrm_name == "scream":
+            transforms += [
+                add_grid_info(res),
+            ]
+        else:
+            raise ValueError(f"gsrm_name {gsrm_name} not supported")
```
oliverwm1, line 213:
```suggestion
        transforms.append(add_grid_info(res))
        if vcm.model_name_from_resolution_string(res) == "fv3":
            transforms.append(add_wind_rotation_info(res)])
```

oliverwm1: Could do something like this (and define appropriate function in `vcm`). It seems like we want to do this `res` -> model name transformation in various places, so we should probably make a single function to do it, and then use that function elsewhere.

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
```diff
@@ -22,6 +22,11 @@ def _stack_sample_data(ds: xr.Dataset) -> xr.Dataset:
         stack_dims = ["tile", "x", "y", DATASET_DIM_NAME]
     else:
         stack_dims = ["tile", "x", "y"]
+    if "ncol" in ds.dims:
```
oliverwm1, line 25: Is there a way to pass the horizontal dims to this function? So could just modify it to be `stack_dims = horizontal_dims + [DATASET_DIM_NAME]` etc. instead of needing a new if statement.

workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
```diff
@@ -28,3 +26,5 @@ class DiagArg:
     verification: xr.Dataset
     grid: xr.Dataset
     delp: Optional[xr.DataArray] = None
+    horizontal_dims: List[str] = field(default_factory=lambda: ["x", "y", "tile"])
+    vertical_dim: str = "z"
```
oliverwm1, line 30: Is the plan to use `z` or `lev` for the vertical dimension? There are still some places where `z` is hard-coded. Probably better to leave it hard-coded everywhere (fine with me) or make sure it is actually configurable.

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -143,11 +150,32 @@ def insert_rmse(ds: xr.Dataset):
 
 
 def load_grid_info(res: str = "c48"):
+    if res.startswith("ne"):
+        return load_grid_info_scream(res)
+    elif res.startswith("c"):
```
oliverwm1, line 155: Should consolidate this logic to one place (e.g. a `vcm` function).

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -308,9 +310,14 @@ def main(args):
     # add Q2 and total water path for PW-Q2 scatterplots and net precip domain averages
     if any(["Q2" in v for v in model.output_variables]):
         model = fv3fit.DerivedModel(model, derived_output_variables=["Q2"])
-
+    if "ncol" in evaluation_grid.dims:
+        grid_name = "ncol"
+    else:
+        grid_name = "x"
```
oliverwm1, line 316:
```suggestion
```

workflows/diagnostics/fv3net/diagnostics/offline/compute.py
```diff
@@ -308,9 +310,14 @@ def main(args):
...
     ds_predicted = get_prediction(
-        config=config, model=model, evaluation_resolution=evaluation_grid.sizes["x"]
+        config=config,
+        model=model,
+        evaluation_resolution=evaluation_grid.sizes[grid_name],
```
oliverwm1, line 320:
```suggestion
        evaluation_resolution=evaluation_grid.sizes[horizontal_dims[0]],
```
And infer `horizontal_dims` somewhere earlier in code. And then can also pass `horizontal_dims` to `_compute_diagnostics` instead of inferring it within that function.

workflows/diagnostics/fv3net/diagnostics/offline/compute_diagnostics.py
```diff
@@ -63,8 +62,12 @@ def compute_diagnostics(
     delp: xr.DataArray,
     n_jobs: int = -1,
 ):
+    if "ncol" in grid.area.dims:
+        horizontal_dims = ["ncol"]
+    else:
+        horizontal_dims = ["x", "y", "tile"]
```
oliverwm1, line 68: I suggest inferring the `horizontal_dims` once in the main function of the compute code, and then passing it down to this function. Will be less liable to bugs and easier to make changes if the dim name convention changes for SCREAM.

oliverwm1 (commented):
Thanks for the changes @elynnwu! We're getting closer. Happy to discuss on call if easier.

My main suggestions are:
1) write a new mapper function for opening SCREAM data for ML (sorry, should have realized this approach was best on the last review)
2) consolidate logic for determining if a "res" string corresponds to FV3 or SCREAM to one place
3) determine horizontal dimensions once, and then pass around to the functions that need them, instead of repeating code to infer the dimension names.

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -126,6 +126,7 @@ def open_nudge_to_fine(
         "state_after_timestep.zarr",
     ),
     cache_size_mb: Optional[float] = None,
+    gsrm_name: str = "fv3",
```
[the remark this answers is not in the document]
elynnwu: This makes sense, and it would alleviate a lot of the gsrm name issue.

workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
elynnwu: The plan is to use `z`, retained `VERTICAL_DIM="z"` and reused it here.

workflows/diagnostics/fv3net/diagnostics/offline/_input_sensitivity.py
elynnwu: Done. Needs to be sorted because `horizontal_dims` is `["x","y","tile"]` but here it needs to be tile first.

oliverwm1: Probably would be fine to just define `horizontal_dims` with the tile-first order.

external/loaders/loaders/mappers/_nudged/_nudged.py
```diff
@@ -230,3 +227,70 @@ def _get_datasets(
...
+        join="inner",
+    )
+
+    differenced_state: Dataset = {}
+    for nudging_variable in nudging_variables:
+        nudging_tendency = ds[f"{nudging_variable}_tendency_due_to_nudging"]
+        differenced_state[nudging_variable] = (
+            ds[nudging_variable] - nudging_tendency * physics_timestep_seconds
+        )
+    ds = ds.assign(differenced_state)
```
oliverwm1, line 283: Do you know where in the time loop nudging happens for SCREAM? The purpose of this code is to make sure that the ML inputs are the atmospheric state right before nudging. It's possible this code will need to change if the nudging is implemented differently than what we have in FV3GFS.

external/vcm/vcm/catalog.yaml
```diff
@@ -998,3 +998,27 @@ sources:
...
+    args:
+      urlpath: "gs://vcm-scream/latLonArea/ne30/land_sea_mask.zarr/"
```
oliverwm1, line 1023: this zarr currently has `lndgrid` as the dimension name. Should it be `ncol`?

workflows/diagnostics/fv3net/diagnostics/prognostic_run/metrics.py
```diff
@@ -150,8 +149,12 @@ def time_and_global_mean_value(diags, mask_type=mask_type):
         if len(time_mean_value) == 0:
             return xr.Dataset()
         masked_area = _mask_array(mask_type, diags["area"], diags["land_sea_mask"])
+        if "ncol" in diags["area"].dims:
+            horizontal_dims = ["ncol"]
+        else:
+            horizontal_dims = ["x", "y", "tile"]
         time_and_global_mean_value = weighted_mean(
-            time_mean_value, masked_area, HORIZONTAL_DIMS
+            time_mean_value, masked_area, horizontal_dims
```
oliverwm1, line 157: I think we should really try to avoid these deep `if` statements. Since this PR is just to make the offline diagnostics work for SCREAM, I suggest just leaving this hard-coded to `HORIZONTAL_DIMS_FV3`. Same for other cases in this module.

external/vcm/vcm/convenience.py
```diff
@@ -146,3 +146,27 @@ def shift_timestamp(time: str, seconds: Union[int, float]) -> str:
...
+    elif res.startswith("c"):
+        gsrm = "fv3"
+    else:
+        raise ValueError(
+            f"This resolution {res} can not be mapped to either scream or fv3."
+        )
+    return gsrm
+
+
+def horizontal_dims_from_resolution_string(res: str):
```
oliverwm1, line 164: I would delete this function, and instead just have code in `fv3net/diagnostics` to return the appropriate dim names based on the output from `gsrm_name_from_resolution_string`. Right now, we've defined the FV3 horizontal dims in two places (here, and as `HORIZONTAL_DIMS_FV3` in `fv3net/diagnostics/_shared/constants.py`, which is not ideal.

workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
```diff
@@ -1,9 +1,9 @@
 import xarray as xr
-from typing import Optional
-from dataclasses import dataclass
+from typing import Optional, List
+from dataclasses import dataclass, field
 import numpy as np
 
-HORIZONTAL_DIMS = ["x", "y", "tile"]
+HORIZONTAL_DIMS_FV3 = ["x", "y", "tile"]
```
oliverwm1, line 6: additionally define the SHIELD horizontal dims here.

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
```diff
@@ -143,11 +150,32 @@ def insert_rmse(ds: xr.Dataset):
...
+    )
+
+
+def load_grid_info_scream(res):
+    grid = catalog[f"grid/{res}"].read()
+    land_sea_mask = catalog[f"landseamask/{res}"].read()
+    grid_info = xr.merge([grid, land_sea_mask])
+    # TODO: currently uses the land model's grid,
+    # but should use scream's when we have it
+    grid_info = grid_info.rename({"lndgrid": "ncol"})
```
oliverwm1, line 177: should we just rename in the saved zarr?

oliverwm1 (commented):
Looking good, just a few minor comments.

external/loaders/loaders/mappers/_nudged/_nudged.py
elynnwu: Right, it is actually flexible in scream where nudging happens (the order of any process is specified in the namelist). I'm keeping it like this because we do not actually have a nudged run yet, but I added a todo note to update this once we have it solidified.

external/vcm/vcm/catalog.yaml
elynnwu: It should but I do not have the atmosphere's land mask and this is to help me remember that I need to update this once I get that file. TODO note in `load_grid_info_scream`

workflows/diagnostics/fv3net/diagnostics/offline/_helpers.py
elynnwu: Yes, I don't want to mix it here because the file needs to be updated to the atmosphere's mask instead of using land model's. Will open a separate PR as soon as I get that file.

oliverwm1 (approved):
LGTM! Make sure to update the PR description so it is accurate with current status of PR, and use that as the commit message. There is also a typo in the PR title (compatiable -> compatible).

If you'd like (optional!) you can enable the integration test before merging. It's possible this will catch something new.

elynnwu:
Integration test worked on the 3rd attempt w/o OOM error, timed out due to the 35 min limit.

PR #2189, opened 2023-03-17
Make offline create report compatible with scream

elynnwu:
In this PR, we add the capability of plotting scream data. 

Added public API:
- Added `GridMetadataFV3` and `GridMetadataScream`: for now they look rather similar just w/o `latb` and `lonb`, but in the future, we may consider retaining the native scream output dimension

Significant internal changes:
- `plot_cube`: uses grid metadata to determine whether to plot fv3 or scream data

- [x] Tests added

### DIFF
--- a/external/fv3viz/examples/plot_cube_fortran_diagnostic.py
+++ b/external/fv3viz/examples/plot_cube_fortran_diagnostic.py
@@ -11,7 +11,7 @@ import io
 import warnings
 import xarray as xr
 from fv3viz import plot_cube
-from vcm.cubedsphere import GridMetadata
+from vcm.cubedsphere import GridMetadataFV3
 
 warnings.filterwarnings(
     "ignore",
@@ -38,7 +38,7 @@ VAR = "LHTFLsfc"
 
 fortran_diagnostic_ds = get_web_dataset(DATA_URL)
 
-gfdl_grid_metadata = GridMetadata("grid_xt", "grid_yt", "grid_x", "grid_y")
+gfdl_grid_metadata = GridMetadataFV3("grid_xt", "grid_yt", "grid_x", "grid_y")
 
 # grid variables are already present in this Fortran diagnostic file
 _ = plot_cube(
--- a/external/fv3viz/examples/plot_cube_scream_diagnostics.py
+++ b/external/fv3viz/examples/plot_cube_scream_diagnostics.py
@@ -1,0 +1,52 @@
+"""
+plot_cube with DOE's SCREAM diagnostics
+===========================================
+
+Example of :py:func:`plot_cube` using SCREAM diagnostic data, with faceting
+over timesteps
+"""
+
+import requests
+import io
+import warnings
+import xarray as xr
+from fv3viz import plot_cube
+from vcm.cubedsphere import GridMetadataScream
+
+warnings.filterwarnings(
+    "ignore",
+    category=UserWarning,
+    message=(
+        "Tight layout not applied. The left and right margins cannot be made "
+        "large enough to accommodate all axes decorations."
+    ),
+)
+
+
+def get_web_dataset(url):
+    r = requests.get(url)
+    ds = xr.open_dataset(io.BytesIO(r.content))
+    return ds
+
+
+DATA_URL = (
+    "https://raw.githubusercontent.com/ai2cm/vcm-ml-example-data/"
+    "main/fv3net/fv3viz/plot_cube_scream_sample.nc"
+)
+VAR = "T_2m"
+
+
+diagnostic_ds = get_web_dataset(DATA_URL)
+
+grid_metadata = GridMetadataScream("ncol", "lon", "lat")
+
+_ = plot_cube(
+    diagnostic_ds,
+    VAR,
+    grid_metadata=grid_metadata,
+    cmap="viridis_r",
+    col="time",
+    col_wrap=2,
+    plotting_function="tripcolor",
+    gsrm_name="scream",
+)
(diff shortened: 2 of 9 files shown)


external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -543,3 +554,49 @@ def _plot_cube_axes(
...
+    array: np.ndarray,
+    lat: np.ndarray,
+    lon: np.ndarray,
+    plotting_function: str,
+    ax: plt.axes = None,
+    **kwargs,
+):
+    if ax is None:
+        ax = plt.gca()
+    if plotting_function in ["tripcolor", "tricontour", "tricontourf"]:
```
brianhenn, line 569: It'd say it's up to you if you want to maintain the different flavors of "tri" plotting; I'd say in practice we only ended up using one (pcolor).

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -159,16 +160,26 @@ def plot_cube(
...
-    _plot_func_short = partial(
-        _plot_cube_axes,
-        lat=mappable_ds.lat.values,
-        lon=mappable_ds.lon.values,
-        latb=mappable_ds.latb.values,
-        lonb=mappable_ds.lonb.values,
-        plotting_function=plotting_function,
-        **kwargs,
-    )
+    if gsrm_name == "fv3":
```
brianhenn, line 163: Glad this approach of separating the axes' plotting function from rest of the data handing seems to have worked out OK

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
```diff
@@ -396,7 +403,7 @@ def create_report(args):
         }
         wandb.init(config=wandb_config)
         wandb.log(metrics)
-
+    gsrm = loaders.BatchesLoader.from_dict(metadata["training_config"]).gsrm_name
```
brianhenn, line 406: I suspect we may want to specify the gsrm name somewhere else, since we'd run into this issue again in the prognostic run diags where we don't have access to the batches object, but I'm guessing @AnnaKwa has a better idea of where exactly.

external/fv3viz/examples/plot_cube_fortran_diagnostic.py
```diff
@@ -38,7 +38,7 @@ def get_web_dataset(url):
 
 fortran_diagnostic_ds = get_web_dataset(DATA_URL)
 
-gfdl_grid_metadata = GridMetadata("grid_xt", "grid_yt", "grid_x", "grid_y")
+gfdl_grid_metadata = GridMetadataFV3("grid_xt", "grid_yt", "grid_x", "grid_y")
```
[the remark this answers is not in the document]
elynnwu: Great idea, I should definitely add one for scream. Per discussion in #2188, I'm going to try and see if it would be possible to keep native scream output format. Will circle back to this once we confirmed which format to use.

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
elynnwu: Changed to using `ds_diags` dimension names to infer gsrm name.

external/fv3viz/examples/plot_cube_fortran_diagnostic.py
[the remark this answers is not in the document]
elynnwu: Example added under `external/fv3viz/examples/plot_cube_scream_diagnostics.py`

external/fv3viz/fv3viz/_plot_cube.py
```diff
@@ -159,16 +160,26 @@ def plot_cube(
...
+            lonb=mappable_ds.lonb.values,
+            plotting_function=plotting_function,
+            **kwargs,
+        )
+    elif gsrm_name == "scream":
+        _plot_func_short = partial(
+            _plot_scream_axes,
+            lat=mappable_ds.lat.values,
+            lon=mappable_ds.lon.values,
+            plotting_function=plotting_function,
```
brianhenn, line 178: It seems like there is a 1:1 mapping between ["pcolormesh", "contour", and "contourf"] and ["tripcolor", "tricontour", "tricontourf"]. In other words, each of these three plotting functions is the same, just with a cartesian vs unstructured grid. Rather than adding the three new "tri" functions to the acceptable list of `plotting_function` values for the top-level `plot_cube` arg, could you instead for the scream case determine which "tri" function to supply to `_plot_scream_axes` based on whether `plotting_function` is "pcolormesh", "contour", or "contourf"? That way there'd be no need to change how `plot_cube` is called with this argument.

brianhenn (commented):
Thanks @elynnwu , overall looks good. I think there is some opportunity to make fewer API changes to `plot_cube`. Maybe we can chat through my comments briefly when you're ready to work on this PR again.

external/fv3viz/examples/plot_cube_scream_diagnostics.py
```diff
@@ -0,0 +1,52 @@
+"""
```
brianhenn, line 1: Thanks for adding this!

workflows/diagnostics/fv3net/diagnostics/offline/views/create_report.py
```diff
@@ -396,7 +402,10 @@ def create_report(args):
         }
         wandb.init(config=wandb_config)
         wandb.log(metrics)
-
+    if "ncol" in ds_diags.dims:
```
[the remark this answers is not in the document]
elynnwu: Yes, actually we've added a function in vcm to determine the name based on resolution string. I'll wait for the other PR to merge first then update here.

external/fv3viz/fv3viz/_plot_cube.py
elynnwu: Good idea, done!

brianhenn (approved):
Great, thanks for the changes.

PR #2192, opened 2023-03-21
Autoencoder training function

AnnaKwa:
Adds a training function for a dense autoencoder. This currently lives in the `fv3fit.reservoir` subdirectory but does reuse a lot of the `fv3fit.keras` code, so I am open to suggestions to move it there.

Normalization and denormalization are done via layers in the encoder and decoder (similar to the fv3fit dense models). This ensures that the same scaling is used when the autoencoder is later used in reservoir model training.
 
Added public API:
- `dense-autoencoder` model type for training

  

- [x] Tests added


external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -26,20 +41,195 @@ def encode(self, x: Union[np.ndarray, tf.Tensor]) -> np.ndarray:
     def decode(self, latent_x: Union[np.ndarray, tf.Tensor]) -> np.ndarray:
         return self.decoder.predict(latent_x)
 
+
+@io.register("dense-autoencoder")
+class AutoencoderPredictor(Reloadable):
```
mcgibbon, line 46: Suggestion: I suspect you can get the same functionality without adding this new class, or at least not as hefty a version of this class.

Instead of dumping and loading the encoder and decoder, can you just dump and load the `Autoencoder`, which is itself a `tf.keras.Model`? You could also remove the `input_variables` and `output_variables` attributes, which aren't used, though I'm curious how you'll make sure the stacking/packing/unpacking behavior is consistent between this and any later reservoir code which uses it.

If you keep input and output variables, this is just a `PureKerasModel`, where you can use `predictor.model.predict()` if you want to operate on ndarrays directly and you can ignore the `predictor.predict` method. If you are removing the input and output variables attributes, that's enough of a change that it's probably worth defining a `KerasModelOnly(Reloadable)` as a more general object, which simply dumps and loads a wrapped `self.model: tf.keras.Model` and has a `predict` as here.

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -26,20 +41,195 @@ def encode(self, x: Union[np.ndarray, tf.Tensor]) -> np.ndarray:
...
-            self.encoder.save(encoder_filename)
-            decoder_filename = os.path.join(path, self._DECODER_NAME)
-            self.decoder.save(decoder_filename)
+
+@dataclasses.dataclass
+class DenseAutoencoderHyperparameters(Hyperparameters):
+    """
+    input_variables: variables to encode/decode
+    output_variables: variables to encode/decode, must match
+        input_variables
```
mcgibbon, line 104: There's no requirement that a `Hyperparameters` class has `input_variables` and `output_variables` attributes, if the inputs and outputs must be the same please use a single attribute like `state_variables`.

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -26,20 +41,195 @@ def encode(self, x: Union[np.ndarray, tf.Tensor]) -> np.ndarray:
...
+    latent_dim_size: size of latent space that encoder transforms to
+    units: number of units in each dense layer
+    n_dense_layers: number of dense layers in *each* of encoder/decoder
```
mcgibbon, line 107: Suggestion (optional): Is it intentional that the hidden dimension size does not necessarily match the output latent dimension size? And that you don't need a way to e.g. ramp up or down the number of features (e.g. 50 inputs to 40 to 30 to 20 hidden features)? If you do need this extra configuration, you could use a `layer_units` attribute of type `List[str]` which gives the number of encoder units in order and number of decoder units in reverse order (and implicitly defines the number of layers).

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -203,4 +208,32 @@ def concat_variables_along_feature_dim(
...
+            reshaped_tensors[var] = var_data
+        elif len(var_data.shape) == max_dims - 1:
+            orig_shape = var_data.shape
+            reshaped_tensors[var] = var_data.reshape(-1, *orig_shape[1:], 1)
+        else:
+            raise ValueError(
+                f"Tensor data has {len(var_data.shape)} dims, must either "
+                f"have either {max_dims} or {max_dims-1}."
+            )
+    return reshaped_tensors
```
mcgibbon, line 239: Nit: Do these changes belong in this PR? If not, can you remove them and add them in the PR where these changes are used (or in another PR that explains these changes if they're intended)?

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -26,20 +41,195 @@ def encode(self, x: Union[np.ndarray, tf.Tensor]) -> np.ndarray:
...
+    encoder, decoder = _build_autoencoder(
+        sample_data=norm_batch_concat,
+        input_size=n_features,
+        latent_dim_size=hyperparameters.latent_dim_size,
+        units=hyperparameters.units,
+        n_dense_layers=hyperparameters.n_dense_layers,
+    )
+
+    train_model = Autoencoder(encoder=encoder, decoder=decoder)
+    predict_model = Autoencoder(encoder=encoder, decoder=decoder)
```
mcgibbon, line 214: Suggestion (not required, but would improve maintainability): If you combined these into a single ModelBuilder (a callable function that takes sample data to be used for normalization/sizes as input and returns a train and predict model as output), you could use the `fv3fit.keras._models.dense.train_column_model` to avoid re-writing data processing and configuration management logic. You can see the `train_dense_model` function for an example of this.

mcgibbon:
Tests appear to be failing.

external/fv3fit/tests/training/test_train_dense_autoencoder.py
```diff
@@ -0,0 +1,67 @@
...
+    DenseAutoencoderHyperparameters,
+)
+from tests.training.test_train import (
+    get_dataset_default,
+    get_uniform_sample_func,
+)
+import fv3fit
+
+
+def test_train_autoencoder():
```
mcgibbon, line 16: You probably want to mark both of these tests with `@pytest.mark.slow`, as I expect they take more than 1s each.

external/fv3fit/fv3fit/reservoir/autoencoder.py
AnnaKwa: Yes, for the former I figured the capacitiy of the encoding/decoding hidden layers could be changed via either units or number of layer. Since our existing dense networks also are fixed to have the same units per layer I figured that was good enough for this use case as well.

external/fv3fit/fv3fit/reservoir/domain.py
AnnaKwa: Removed since this code is no longer needed after changes

external/fv3fit/fv3fit/reservoir/autoencoder.py
AnnaKwa: Updated to use train_column_model and a subclass of PureKerasModel (needed a different dump/load to avoid weird errors when calling .predict on loaded Autoencoder)

external/fv3fit/tests/training/test_train_dense_autoencoder.py
AnnaKwa: Removed this test module, with the new changes the regular tests will run for this training function.

external/fv3fit/tests/training/test_train.py
```diff
@@ -382,16 +389,18 @@ def test_dump_and_load_default_maintains_prediction(model_type):
     n_sample, n_tile, nx, ny, n_feature = 1, 6, 12, 12, 2
     sample_func = get_uniform_sample_func(size=(n_sample, n_tile, nx, ny, n_feature))
     result = train_identity_model(model_type, sample_func=sample_func)
-
     original_result = result.model.predict(result.test_dataset)
     with tempfile.TemporaryDirectory() as tmpdir:
         fv3fit.dump(result.model, tmpdir)
         loaded_model = fv3fit.load(tmpdir)
+    loaded_result = loaded_model.predict(result.test_dataset)
```
AnnaKwa, line 396: For some reason, the loaded custom keras models can't call their `.summary()` until you use them to predict something, so I moved this line up to before the assertions.

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -35,11 +63,206 @@ def load(cls, path: str) -> "Autoencoder":
             decoder = tf.keras.models.load_model(
                 os.path.join(model_path, cls._DECODER_NAME)
             )
+
         return cls(encoder=encoder, decoder=decoder)
 
+
+@io.register("keras-autoencoder")
+class PureKerasAutoencoderModel(PureKerasModel):
```
mcgibbon, line 71: There isn't much in the way of functionality that we're adopting from the base class here to justify the cognitive cost of having inheritance. Since we weren't able to just use PureKerasModel, could you remove PureKerasModel as a parent class and copy any relevant code into this class instead?

external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -35,11 +63,206 @@ def load(cls, path: str) -> "Autoencoder":
...
+    """ Mostly identical to the PureKerasModel but takes an
```
mcgibbon, line 72:
```suggestion
    """Mostly identical to the PureKerasModel but takes an
```

external/fv3fit/tests/training/test_train.py
```diff
@@ -380,16 +387,18 @@ def test_dump_and_load_default_maintains_prediction(model_type):
...
```
mcgibbon, line 394:
```suggestion
    # .predict() needs to be called before .summary() for some reason
    loaded_result = loaded_model.predict(result.test_dataset)
```

mcgibbon (approved):
One request to remove inheritance, otherwise LGTM!

external/fv3fit/fv3fit/reservoir/autoencoder.py
AnnaKwa: As discussed in slack, noted in issue https://github.com/ai2cm/fv3net/issues/2198

PR #2194, opened 2023-03-22
Ensure normalization unchanged when CycleGAN training reloaded

mcgibbon:
Currently when a CycleGAN run reloads an existing model, that reloading does not at all affect the normalization of input data. Input data is still normalized based on a random sample.

Though that random sample and as a result the normalization constants should be identical if the training data configuration is identical due to identical random seeding, this is not guaranteed if data or configuration changes. For this reason, this PR updates the CycleGAN training to reload the normalization state from the saved CycleGAN model. This means the model must be reloaded at a higher level of the training.

Refactored public API:
- Reload_path moved from CycleGANNetworkConfig to CycleGANHyperparameters.
- Fixed an issue where the normalization used when loading a reloaded CycleGAN model may not match the normalization with which that model was originally trained.

- [ ] Tests added


frodre (approved):
LGTM

PR #2199, opened 2023-03-29
Add build_model function for minimal autoencoder

AnnaKwa:
The reservoir training now relies on the autoencoder to do the feature concatentation and normalization/denormalization, which makes the autoencoder a nonoptional component in the reservoir training. This PR adds a function to the autoencoder module to build a minimal autoencoder that has no trainable parameters and only handles the necessary concatenation and normalization. It can be used in the case of no autoencoder path being specified, ex. in tests.
  

- [x] Tests added
 

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/autoencoder.py
+++ b/external/fv3fit/fv3fit/reservoir/autoencoder.py
@@ -181,6 +181,44 @@ class DenseAutoencoderHyperparameters(Hyperparameters):
         return cls(state_variables=input_variables)
 
 
+def build_concat_and_scale_only_autoencoder(
+    variables: Sequence[str], X: Tuple[tf.Tensor, ...]
+) -> tf.keras.Model:
+    """ Performs input concatenation and norm/denormalization,
+    but does not train actual encoder or decoder layers. Useful for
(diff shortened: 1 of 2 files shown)


external/fv3fit/fv3fit/reservoir/autoencoder.py
```diff
@@ -181,6 +181,44 @@ def init_testing(
...
+        )
+        start_ind += output.shape[-1]
+    denorm_output_layers = standard_denormalize(
+        names=variables, layers=decoder_outputs, arrays=X,
+    )
+    decoder = tf.keras.Model(inputs=decoder_input, outputs=denorm_output_layers)
+
+    model = Autoencoder(encoder=encoder, decoder=decoder)
+    # need to call model once so it can save without compiling
+    model([np.ones((3, arr.shape[-1])) for arr in X])
```
spencerkclark, line 218: As discussed on Slack, I like your idea of using `model.compile()` here a little better.  Could consider writing a test for savability if this is an important property of what this function returns as well.

spencerkclark (approved):
Looks good to me.  Hopefully this makes writing tests simpler going forward!

PR #2201, opened 2023-04-07
Bump fortran for prescribed sea-ice fix and new diagnostics

spencerkclark:
This PR bumps the fortran model to include an update that fixes an issue when prescribing sea-ice from a file directly in the fortran model, as well as some new physics diagnostics:

- https://github.com/ai2cm/fv3gfs-fortran/pull/362
- https://github.com/ai2cm/fv3gfs-fortran/pull/363


brianhenn (approved):
Thanks @spencerkclark

PR #2207, opened 2023-04-19
Parallelize rc training generation

AnnaKwa:
The slowest steps in training are the generation of the reservoir state sequence and the regressor matrix update. The former is already parallelized over subdomains, while the latter currently uses a for loop over subdomains. This PR parallelizes the matrix update for each subdomain's regressor to cut down time one of the slowest parts of training on each batch. Using n_jobs=4 reduces the time by roughly half. 


frodre (approved):
LGTM, nice speedup

Issue #2210, opened 2023-04-26
`fv3fit_torch` image build fails on trying to install `dgl-cu113`


spencerkclark:
The `fv3fit_torch` image build is currently failing in CI (e.g. [here](https://app.circleci.com/pipelines/github/ai2cm/fv3net/14348/workflows/1294fcf3-5372-49dc-ba12-93c82ad94c3e/jobs/135099)).  It seems to have occurred in both https://github.com/ai2cm/fv3net/pull/2208 and https://github.com/ai2cm/fv3net/pull/2209, so it may not be a transient issue.  The error has to do with installing a package:

```
#18 [14/14] RUN pip3 install dgl-cu113 dglgo -f https://data.dgl.ai/wheels/repo.html
#18 sha256:e6625a465442fa9ddc44ede99f8e292d1c06e3f08c86e4bf94fb78e728ecdb5c
#18 1.225 Looking in links: https://data.dgl.ai/wheels/repo.html
#18 1.379 ERROR: Could not find a version that satisfies the requirement dgl-cu113 (from versions: none)
#18 1.380 ERROR: No matching distribution found for dgl-cu113
#18 ERROR: executor failed running [/bin/sh -c pip3 install dgl-cu113 dglgo -f https://data.dgl.ai/wheels/repo.html]: exit code: 1
```

cc: @mcgibbon

elynnwu:
Yes, I saw this yesterday, and it went away after a while on its own.

spencerkclark:
Oh good -- thanks @elynnwu -- hopefully nothing to worry about then.

spencerkclark:
Looks like this was fixed upstream https://github.com/dmlc/dgl/issues/5625; our images appear to be building smoothly now too.

PR #2211, opened 2023-04-28
Fix array reshaping bug in prediction

AnnaKwa:
The reservoir models were working as expected when trained on a single subdomain, but predictions were way off for the same subdomain when trained on a larger domain (split into parallel trained subdomains). Models should give the same result for the same subdomain in these two cases.

I traced the source of the error to the final readout prediction step, where `np.reshape` was used improperly. For array reshaping here and in future inference code I think it would be safest to write testable functions so that a short example of the desired reshaping can unit tested. In this short PR I just added a reshaping function that calls a one-liner of np.reshape for a 2d -> 1d case, but in future PRs the reshaping is much more complex.
 
- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/_reshaping.py
+++ b/external/fv3fit/fv3fit/reservoir/_reshaping.py
@@ -1,0 +1,6 @@
+import numpy as np
+
+
+def flatten_2d_keeping_columns_contiguous(arr: np.ndarray):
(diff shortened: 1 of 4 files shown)


spencerkclark:
Sorry @frodre, I was referring to https://github.com/ai2cm/fv3gfs-fortran/pull/364, so I'm afraid I have not explored anything related to addressing #2210.  @elynnwu indicated that this may have been a transient issue, but it seems to be occurring fairly consistently now.  We may need to invest some cycles into it.

PR #2213, opened 2023-04-30
Locating vertical coordinate file (`fv_core.res.nc`) on HPC cluster

### LINKED ISSUE #2197: Prognostic run configuration hard codes cloud-based 79-level `fv_core.res.nc` asset
By way of [`fv3kube.c48_initial_conditions_overlay`](https://github.com/ai2cm/fv3net/blob/766778a1534d864e542a73c74575fb749abeb933/workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py#L88-L91), the prognostic run currently hard codes the vertical coordinate file asset to: https://github.com/ai2cm/fv3net/blob/766778a1534d864e542a73c74575fb749abeb933/external/fv3kube/fv3kube/config.py#L22-L26

This presents two possible issues (issue two being the primary current motivation for making this post):
1. If we choose to restart from files with a different number of vertical levels, this hard code will prevent it.
2. Even when restarting from files with 79 vertical levels, this file asset is hard coded to be cloud-based, which is not convenient when running on HPC.  

These can be worked around by patching the config file produced by the `prepare-config` tool with the appropriate path, but this is not a very clean solution.  Would we be open to adding another optional configuration attribute to the [`InitialCondition` dataclass](https://github.com/ai2cm/fv3net/blob/766778a1534d864e542a73c74575fb749abeb933/workflows/prognostic_c48_run/runtime/segmented_run/prepare_config.py#L69-L91) that allowed the user to specify the path to this file?

cc: @mrudko

mrudko:
The code changes in this PR aim to facilitate setting up ML workflow on HPC cluster. Preparing the full configuration file results in retaining the path to the vertical coordinate file (`fv_core.res.nc`) as specified on the Google Cloud platform (e.g. `gs://some/path`). To perform numerical simulations, the user has to modify the path manually.

Added public API:
- `c48_initial_condition_overlay` function accepts `vertical_coordinate_file` variable (`config.py`); the same variable (attribute) was added to the `InitialCondition` class (`prepare_config.py`)

Refactored public API:
- `FV_CORE_ASSET` dictionary was moved into the`c48_initial_condition_overlay` function.

- [x] Tests added
The test was added to test_config.py script. The test validates the path to vertical coordinate file (`fv_core.rs.nc`).
The regression test was modified in the `test_config.py` script.
Two tests were added to `test_prepare_config.py`.  One test validates the default value of `vertical_coordinate_file` variable, while the other one ensures the safe transfer of the same variable from minimal to full configuration files.

Resolves #2197


external/fv3kube/fv3kube/config.py
```diff
@@ -157,13 +157,24 @@ def get_full_config(
 
 
 def c48_initial_conditions_overlay(
-    url: str, timestep: str, restart_categories: RestartCategoriesConfig = None
+    url: str, timestep: str, vertical_coordinate_file: str, restart_categories: RestartCategoriesConfig = None
```
spencerkclark, line 160: Similar to what is done for `restart_categories`, I might make `vertical_coordinate_file` a keyword argument.  In this case I would use a default value of `DEFAULT_VERTICAL_COORDINATE_FILE`.  This way the [existing regression test](https://github.com/ai2cm/fv3net/blob/0deee7a220a4de6fb2d9fe807c77266def6ff956/external/fv3kube/tests/test_config.py#L35-L41), which does not pass `vertical_coordinate_file` or `restart_categories` arguments, would still pass.

external/fv3kube/tests/test_config.py
```diff
@@ -31,11 +31,21 @@ def test_merge_fv3config_overlays(mappings, expected):
...
+def test_invalid_vertical_coorinate_file():
+
+    base_url = "/some/path"
+    timestep = "20160805.000000"
+    vertical_coordinate_file = "/some/path/"
+
+    with pytest.raises(AssertionError, match='Provided vertical coordinate file is a directory'):
+        fv3kube.c48_initial_conditions_overlay(base_url, timestep, vertical_coordinate_file)
 
 def test_c48_initial_conditions_overlay(regtest):
```
spencerkclark, line 43: Indeed it may make sense to add a regression test for this new functionality, though I think preserving the old regression test, demonstrating that the same call still produces the same result, would also be useful.  To do this I would use the [ability of pytest to parametrize tests](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#parametrize).
```python
@pytest.mark.parametrize("vertical_coordinate_file", [None, "/some/path"])
def test_c48_initial_conditions_overlay(regtest, vertical_coordinate_file):
    url = "some/url"
    timestep = "20160801.000000"

    if vertical_coordinate_file is None:
        ans = fv3kube.c48_initial_conditions_overlay(url, timestep)
    else:
        ans = fv3kube.c48_initial_conditions_overlay(url, timestep, vertical_coordinate_file=vertical_coordinate_file)

    print(ans, file=regtest)
```
This preserves the regression test which calls `fv3kube.c48_initial_conditions_overlay` without a `vertical_coordinate_file` argument, which should still produce the same result as it did before.

When adding this regression test you will need to make a record of the expected output using:
```
$ pytest --regtest-reset test_config.py
```
and the add the resulting new file, which should show up in `fv3net/external/fv3kube/tests/_regtest_outputs`, to your PR.  

I think the act of parametrizing this test will also change the name of the record file for the existing regression test, so you will need to account for that as well.  The contents of that file should be identical to before, however.  To demonstrate that the contents are identical, I think you can update your PR in the following way:
```
$ pytest --regtest-reset test_config.py
$ rm _regtest_outputs/test_c48_initial_conditions_overlay[None].out
$ git mv _regtest_outputs/test_c48_initial_conditions_overlay.out _regtest_outputs/test_c48_initial_conditions_overlay[None].out
$ git add _regtest_outputs/*.out
```
By using `git mv` you are signaling to `git` that you are merely renaming a file instead of adding a completely new file and removing an old one, and so it will know in the diff to only show that the filename changed.  When updating regression tests we always want to be careful to verify that answers change only where we expect them to.  If this is done correctly you should see the regression tests passing both locally and [in our CI](https://app.circleci.com/pipelines/github/ai2cm/fv3net/14359/workflows/b1cb5125-a6b3-4ac5-8f5d-2e0b262e3b77/jobs/135241).

See [here](https://gitlab.com/uweschmitt/pytest-regtest/test) for more information on `pytest-regtest`.

spencerkclark (commented):
Thanks @mrudko -- I made some initial comments.  Note that you have a few test failures indicated by our CI.  I made some suggestions for how to address those.

Please also make sure to run the `make lint` rule and address the linting errors reported in [here](https://app.circleci.com/pipelines/github/ai2cm/fv3net/14359/workflows/b1cb5125-a6b3-4ac5-8f5d-2e0b262e3b77/jobs/135243).

workflows/prognostic_c48_run/tests/test_prepare_config.py
```diff
@@ -81,7 +84,30 @@ def test_config_high_level_duration(duration, expected):
     config = HighLevelConfig(duration=duration)
     out = config.to_fv3config()
     assert out["namelist"]["coupler_nml"]["seconds"] == expected
+def test_initial_condition_default_attribute():
```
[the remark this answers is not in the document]
mrudko: The name of the function is too long. Flake8 throws an error

flake8...................................................................Failed
- hook id: flake8
- exit code: 1

workflows/prognostic_c48_run/tests/test_prepare_config.py:99:89: E501 line too long (93 > 88 characters)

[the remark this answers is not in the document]
spencerkclark: This line is only 62 characters, so it is not the problem.  This error is referring to this section of `test_prepare_config.py` after `black` is applied (specifically line 99):
```python
    assert (
        out
        == "gs://vcm-fv3config/data/initial_conditions/fv_core_79_levels/v1.0/fv_core.res.nc"
    )
```

external/fv3kube/tests/test_config.test_c48_initial_conditions_overlay[None].out
Comment (spencerkclark) [inline external/fv3kube/tests/test_config.test_c48_initial_conditions_overlay[None].out:1]: Note this needs to be in the `fv3net/external/fv3kube/tests/_regtest_outputs` directory (currently it is in `/tests`).  In light of my comment regarding the prepare config regression tests, this file may also need to be reset to have a `source_location` without a trailing `"/"` for the vertical coordinate file.

spencerkclark (commented):
Thanks @mrudko -- I made a comment regarding the `fv3kube` regression test failure.  

Note the regression tests in the prognostic run build are also showing some new failures.  This is because [previously the vertical coordinate file asset was hardcoded to have a trailing `"/"` in for the `source_location`](https://github.com/ai2cm/fv3net/blob/ec31c34e59f21f2f03f86eff2b5b055603a6c965/external/fv3kube/fv3kube/config.py#L23).  I think it's probably safe to just reset these regression tests to the new results (without the trailing slash).  E.g. this is one of the failures:
```
____________________ test_prepare_ml_config_regression[ml] _____________________

regression test output differences for tests/test_prepare_config.py::test_prepare_ml_config_regression[ml]:

>   --- current
>   +++ tobe
>   @@ -130,7 +130,7 @@
>      target_location: INPUT
>      target_name: fv_srf_wnd.res.tile6.nc
>    - copy_method: copy
>   -  source_location: gs://vcm-fv3config/data/initial_conditions/fv_core_79_levels/v1.0
>   +  source_location: gs://vcm-fv3config/data/initial_conditions/fv_core_79_levels/v1.0/
>      source_name: fv_core.res.nc
>      target_location: INPUT
>      target_name: fv_core.res.nc
```

workflows/prognostic_c48_run/tests/test_prepare_config.py
```diff
@@ -1,16 +1,19 @@
...
+    InitialCondition,
     HighLevelConfig,
     UserConfig,
     instantiate_dataclass_from,
 )
 from runtime.diagnostics.fortran import FortranFileConfig
 from runtime.segmented_run import prepare_config
 
-TEST_DATA_DIR = "tests/prepare_config_test_data"
+TEST_DATA_DIR = "prepare_config_test_data"
```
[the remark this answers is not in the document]
mrudko: If you run the tests within the folder "tests", then it complains that it cannot find the path to the .yml files. Please see the error message below

_________________________________________________ test_prepare_ml_config_regression[ml] _________________________________________________

regtest = <pytest_regtest.RegTestFixture object at 0x1503ec02f070>, argv = ['tests/prepare_config_test_data/prognostic_config.yml']

    @pytest.mark.parametrize(
        "argv",
        [
            pytest.param([f"{TEST_DATA_DIR}/prognostic_config.yml"], id="ml",),
            pytest.param([f"{TEST_DATA_DIR}/nudge_to_fine_config.yml"], id="n2f"),
            pytest.param([f"{TEST_DATA_DIR}/nudge_to_obs_config.yml"], id="n2o"),
            pytest.param([f"{TEST_DATA_DIR}/emulator.yml"], id="emulator"),
            pytest.param([f"{TEST_DATA_DIR}/fine_res_ml.yml"], id="fine-res-ml"),
        ],
    )
    def test_prepare_ml_config_regression(regtest, argv):
        parser = prepare_config._create_arg_parser()
        args = parser.parse_args(argv)
        with regtest:
>           prepare_config.prepare_config(args)

test_prepare_config.py:33: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = Namespace(user_config='tests/prepare_config_test_data/prognostic_config.yml')

    def prepare_config(args):
        # Get model config with prognostic run updates
>       with open(args.user_config, "r") as f:
E       FileNotFoundError: [Errno 2] No such file or directory: 'tests/prepare_config_test_data/prognostic_config.yml'

../runtime/segmented_run/prepare_config.py:259: FileNotFoundError

external/fv3kube/tests/test_config.test_c48_initial_conditions_overlay[None].out
Comment (mrudko) [reply external/fv3kube/tests/test_config.test_c48_initial_conditions_overlay[None].out:1]: I removed the trailing "/" from cloud-based string in FV_CORE_ASSET dictionary. The regression tests in the prognostic run build are being passed now. Yet, the question remains if the hardcoded FV_CORE_ASSET is still necessary?

Comment (spencerkclark) [reply external/fv3kube/tests/test_config.test_c48_initial_conditions_overlay[None].out:1]: Sorry this is not what I meant.  I meant that the `test_config.test_c48_initial_conditions_overlay.out` file, which we are renaming to `test_config.test_c48_initial_conditions_overlay[None].out`, would need to be updated to not include the trailing `"/"`.  

The hard-coded `FV_CORE_ASSET` at the top of `config.py` is indeed no longer necessary, because with the changes in your PR, it is no longer used anywhere.  Please go ahead and remove it.

external/fv3kube/fv3kube/config.py
```diff
@@ -157,13 +156,25 @@ def get_full_config(
...
     """
     TIME_FMT = "%Y%m%d.%H%M%S"
     time = datetime.datetime.strptime(timestep, TIME_FMT)
     time_list = [time.year, time.month, time.day, time.hour, time.minute, time.second]
+
+    source_location, source_name = os.path.split(vertical_coordinate_file)
+
+    assert source_name != "", "Provided vertical coordinate file is a directory"
+
+    FV_CORE_ASSET = fv3config.get_asset_dict(
```
spencerkclark, line 174: nit: [based on PEP8 style](https://peps.python.org/pep-0008/#function-and-variable-names) we typically name variables (as opposed to constants) with all lowercase letters (be sure to update the name further down as well):
```suggestion
    fv_core_asset = fv3config.get_asset_dict(
```

workflows/prognostic_c48_run/tests/test_prepare_config.py
```diff
@@ -83,6 +86,40 @@ def test_config_high_level_duration(duration, expected):
...
+    base_url = "/some/path"
+    timestep = "20160805.000000"
+    test_initial_condition = InitialCondition(base_url, timestep)
+
+    out = test_initial_condition.vertical_coordinate_file
+
+    assert (
+        out == "gs://vcm-fv3config/data/initial_conditions/"
+        "fv_core_79_levels/v1.0/fv_core.res.nc"
+    )
```
spencerkclark, line 100: nit: this might be a little more concise, and make use of the constant defined in `fv3kube`:
```suggestion
    result = InitialCondition(base_url, timestep)
    assert result.vertical_coordinate_file == DEFAULT_VERTICAL_COORDINATE_FILE
```

spencerkclark (commented):
Thanks @mrudko for the discussion offline to help sort out https://github.com/ai2cm/fv3net/pull/2213#discussion_r1182507219.  I just had a couple more minor comments.  Go ahead and address those and I think this should be ready to go.  I'll give it another once over tomorrow.

spencerkclark (commented):
@mrudko note that you will need to make some additional changes so that the tests pass.  My code suggestions were illustrative, but not complete.

oliverwm1 (approved):
LGTM, pending the integration test passing. I did notice one typo in a test name. Thanks @mrudko!

external/fv3kube/tests/test_config.py
```diff
@@ -32,10 +32,31 @@ def test_merge_fv3config_overlays(mappings, expected):
     assert output == expected
 
 
-def test_c48_initial_conditions_overlay(regtest):
+def test_invalid_vertical_coorinate_file():
```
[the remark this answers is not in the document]
spencerkclark: Thanks -- good catch!

PR #2216, opened 2023-05-10
Bump version of xpartition to 0.2.1

spencerkclark:
This PR bumps the version of xpartition to 0.2.1, which was just released, and includes a safety-oriented bug fix.  This fix should eliminate the risk of corrupted coordinate data, which we only observed once on a recently created dataset, but is important to address.

xref: https://github.com/spencerkclark/xpartition/pull/17


oliverwm1 (approved):
LGTM

PR #2220, opened 2023-05-13
fv3fit: keras dense model without column data

brianhenn:
Allows the fv3fit `dense` model to work with fully stacked (no unstacked dimension) data, i.e., predicting scalars from scalars. 

Added public API:
- Added a `predict_columns` attribute to `DenseHyperparameters`, which defaults to true and specifies that dense models are saved with `z` as an unstacked dim (the prior behavior); if false models are saved with no unstacked dims.

Refactored public API:
- Renamed the package-public function `train_column_model` to `train_pure_keras_model` and added an `unstacked_dims` argument, given generalization of this function's capabilities.

- [X] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/keras/_models/dense.py
+++ b/external/fv3fit/fv3fit/keras/_models/dense.py
@@ -103,7 +103,7 @@ def train_dense_model(
         training_loop=hyperparameters.training_loop,
         build_samples=hyperparameters.normalization_fit_samples,
         callbacks=hyperparameters.callbacks,
-        predict_columns=hyperparameters.predict_columns,
+        unstacked_dims=("z",) if hyperparameters.predict_columns else (),
     )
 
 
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -48,8 +48,6 @@ class DenseHyperparameters(Hyperparameters):
             variable names and values are either a scalar referring to the
             total weight of the variable. Default is a total weight of 1
             for each variable.
-        normalize_loss: if True (default), normalize outputs by their standard
```
brianhenn, line 51: This was unused, i.e. it was not possible to compute loss on un-normalized outputs.

external/fv3fit/fv3fit/keras/_models/dense.py
```diff
@@ -93,7 +93,7 @@ def train_dense_model(
     train_batches: tf.data.Dataset,
     validation_batches: Optional[tf.data.Dataset],
 ) -> Predictor:
-    return train_column_model(
```
brianhenn, line 96: If this rename is disruptive we can leave it as before, but this is a more accurate name now and there was actually very little about this function that was column-specific before.

mcgibbon (approved):
LGTM

PR #2221, opened 2023-05-15
Add hybrid reservoir model

AnnaKwa:
Adds a `HybridReservoirComputingModel` class. It contains a base `ReservoirComputingModel` attribute which handles the methods shared between both hybrid and reservoir-only models.  The differences in the new class are 
- an additional `hybrid_variables` argument at initialization
- `HybridReservoirComputingModel.predict(hybrid_input)` requires hybrid model input for each prediction

I also added a `synchronize` method to both reservoir models. Previously the reservoir-only model called that method on its `reservoir` attribute. This change makes synchronizing the hybrid model cleaner.

Added public API:
- HybridReservoirComputingModel

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/__init__.py
+++ b/external/fv3fit/fv3fit/reservoir/__init__.py
@@ -5,5 +5,5 @@ from .config import (
 )
 from .reservoir import Reservoir
 from .readout import ReservoirComputingReadout
-from .model import ReservoirComputingModel
+from .model import ReservoirComputingModel, HybridReservoirComputingModel
 from .domain import RankDivider, stack_time_series_samples
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -47,18 +47,8 @@ class HybridReservoirComputingModel(Predictor):
         self.autoencoder = autoencoder
 
     def predict(self, hybrid_input):
-        # Returns raw readout prediction of latent state.
-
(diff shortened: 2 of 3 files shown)


frodre (commented):
Just one comment on the new hybrid reservoir model.  

I know there has been some discussion around subclassing/inheritance so I'm not sure where it stands on `fv3fit` models,  but in this case, I feel like the extension to the hybrid case would be a valid usage.   It would remove the need to re-implement some of the methods. Also, I don't think would add too much overhead to figuring out what is going on since it's in a single file and there probably won't be too many more extensions (I think).

Totally okay if there are good reasons not to do this.  Just wanted to have the conversation.

AnnaKwa:
> I know there has been some discussion around subclassing/inheritance so I'm not sure where it stands on fv3fit models, but in this case, I feel like the extension to the hybrid case would be a valid usage. It would remove the need to re-implement some of the methods. Also, I don't think would add too much overhead to figuring out what is going on since it's in a single file and there probably won't be too many more extensions (I think).

Yeah, I also was initially unsure of what the best way to create the new hybrid class would be. It shares a lot of functionality with the pure reservoir model, but has the a different init signature. I chatted with Jeremy about this and the simplest solution seemed to be this- avoiding inheritance and just making the pure reservoir model an attribute since it contains all the functionality that is shared (tracking the state and updating).

frodre (approved):
Cool. Thanks for the context, that makes sense.

PR #2224, opened 2023-05-18
Argo workflow for running scream

elynnwu:
Add argo template for running scream in the cloud, configured to run in either `highcompute-sim-pool` or `ultra-sim-pool`.

### DIFF
--- a/workflows/argo/scream-prognostic-run.yaml
+++ b/workflows/argo/scream-prognostic-run.yaml
@@ -1,0 +1,104 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: scream-prognostic-run
+spec:
+  entrypoint: driver
+  volumes:
+  - name: gcp-credentials-user-gcp-sa
+    secret:
+      secretName: gcp-key
+  - name: data-volume
+    persistentVolumeClaim:
+      claimName: scream-storage-claim
+  - name: dshm
+    emptyDir:
+      medium: Memory
+  templates:
+  - name: driver
+    inputs:
+      parameters:
+      - {name: config, value: "/tmp/180.yaml"}
+      - {name: precompiled_case, value: "true"}
+      - {name: memory, value: "700Gi"}
+      - {name: cpu, value: "180"}
+    steps:
+      - - name: choose-node-pool
+          template: choose-node-pool
+          arguments:
+            parameters:
+            - {name: cpu-request, value: "{{inputs.parameters.cpu}}"}
+            - {name: cpu-cutoff, value: "16"}
+      - - name: run-scream
+          template: run-scream
+          arguments:
+            parameters:
+            - {name: config, value: "{{inputs.parameters.config}}"}
+            - {name: precompiled_case, value: "{{inputs.parameters.precompiled_case}}"}
+            - {name: node-pool, value: "{{steps.choose-node-pool.outputs.result}}"}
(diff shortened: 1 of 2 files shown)


workflows/argo/scream-prognostic-run.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - {name: precompiled_case, value: "true"}
+      - {name: memory, value: "700Gi"}
+      - {name: cpu, value: "180"}
+    steps:
+      - - name: choose-node-pool
+          template: choose-node-pool
+          arguments:
+            parameters:
+            - {name: cpu-request, value: "{{inputs.parameters.cpu}}"}
+            - {name: cpu-cutoff, value: "16"}
```
AnnaKwa, line 31: The docs say for the c2-standard-30 machine type, vCPU=30, can the cutoff be set higher or are there other considerations that make the default choice 16?

workflows/argo/scream-prognostic-run.yaml
```diff
@@ -0,0 +1,104 @@
...
+  - name: dshm
+    emptyDir:
+      medium: Memory
+  templates:
+  - name: driver
+    inputs:
+      parameters:
+      - {name: config, value: "/tmp/180.yaml"}
+      - {name: precompiled_case, value: "true"}
+      - {name: memory, value: "700Gi"}
```
AnnaKwa, line 23: Will most SCREAM runs require the ultra sim pool? Just wondering since if not maybe use default memory and CPU for the highcompute pool.

workflows/argo/scream-prognostic-run.yaml
```diff
@@ -0,0 +1,104 @@
...
+    persistentVolumeClaim:
+      claimName: scream-storage-claim
+  - name: dshm
+    emptyDir:
+      medium: Memory
+  templates:
+  - name: driver
+    inputs:
+      parameters:
+      - {name: config, value: "/tmp/180.yaml"}
```
AnnaKwa, line 21: Does this default config file already exist on a volume that can be read when this job runs? 

Not sure if your usage is different and you don't have this issue, but one annoying difficulty for other argo templates was that it is not straightforward to pass a config file to the cluster for use in a job.
In other templates, to get around this we pass the entirety of the config's contents as the parameter (ex. [here](https://github.com/ai2cm/vcm-workflow-control/blob/1a27d5966ede468955fdc6e387eacdcadcd129dd/examples/train-evaluate-prognostic-run/run.sh#L19)).

AnnaKwa (commented):
My main question is about the config file parameter- how does the file gets onto the cluster to be used during a job?

AnnaKwa (commented):
My main question is about the config file parameter- how does the file gets onto the cluster to be used during a job?

workflows/argo/scream-prognostic-run.yaml
elynnwu: No real reason, I picked 16 because that was the number of cores I could test on my vm. Changing to 30 to match the number of cpu c2-standard has.

workflows/argo/scream-prognostic-run.yaml
elynnwu: Yes, this is a file created inside docker as part of the pre-built executables. 

And yes, passing the entirety of the config's contents as the parameter is allowed. `scream_run prepare-config` takes care of that.

AnnaKwa (approved):
Thanks for clarifying, LGTM!

PR #2225, opened 2023-05-19
Hybrid reservoir training

AnnaKwa:
This PR adds the ability to train hybrid reservoir models. Instead of adding a separate training function, I instead implemented the hybrid training as an option that is enabled if the training configuration includes a `hybrid_variables` field. If this list is provided, those variables are used as additional inputs to the readout layer training (concatenated after the reservoir states).

Added public API:
- `ReservoirTrainingConfig.hybrid_variables` config field is used to enable hybrid training
 
Significant internal changes:
- some refactors of the helper functions called in the reservoir training function to more smoothly enable hybrid training

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/config.py
+++ b/external/fv3fit/fv3fit/reservoir/config.py
@@ -103,18 +103,15 @@ class ReservoirTrainingConfig(Hyperparameters):
                 f"Output variables {self.output_variables} must be a subset of "
                 f"input variables {self.input_variables}."
             )
-
-        hybrid_and_input_vars_intersection = set(self.hybrid_variables).intersection(
-            self.input_variables
-        )
-        if (
-            self.hybrid_variables is not None
-            and len(hybrid_and_input_vars_intersection) > 0
-        ):
-            raise ValueError(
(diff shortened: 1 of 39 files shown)


external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -102,10 +103,19 @@ def __post_init__(self):
                 f"Output variables {self.output_variables} must be a subset of "
                 f"input variables {self.input_variables}."
             )
+        if (
+            self.hybrid_variables is not None
+            and len(set(self.hybrid_variables).intersection(self.input_variables)) > 0
```
frodre, line 108: nit: I tend to take these long sequences and use a well-named intermediate variables for readability

frodre (approved):
Just a general thought after looking over this.  I think the re-use of `if hybrid_variables` conditionals in the training script to slightly adjust the behavior is a bit of a flag that there is some room for building functions to construct models/data and remove the overhead of the details.

However, I think these are the only two model types we'll be using for the project so there's not the risk of this growing into a much larger issue. Just something to consider if there's need for a refactor down the line.

PR #2226, opened 2023-05-19
radiation: fix broken year warnings

brianhenn:
The radiation port, if run for certain years, had warning statements that inadvertently crashed it. This PR replaces them with print statements (the way the rest of the port outputs to console at runtime; should probably be replaced by logging throughout but that's a separate issue). 

### DIFF
--- a/external/radiation/radiation/radiation_astronomy.py
+++ b/external/radiation/radiation/radiation_astronomy.py
@@ -1,4 +1,3 @@
-import warnings
 import numpy as np
 from .phys_const import con_pi, con_solr, con_solr_old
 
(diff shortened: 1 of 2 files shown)


elynnwu (approved):
Is the intent to just allow using the closet cycle year? or should their be an updated table at some point?

brianhenn:
Fair question. Yes I think the intent is to get the year with same position in the solar cycle as the desired year, but we could also ask GFDL for an updated solar constant data file.

brianhenn:
Cool thanks @spencerkclark . Will be easy to update the GCS file that the rad port pulls its solar data from (which is just a reformatted copy of the original forcing files).

PR #2228, opened 2023-05-24
vcm: pressure and relative humidity in derived mapping

brianhenn:
Adds layer midpoint pressure and relative humidity to the `vcm.DerivedMapping`, making these variables available as ML model inputs.


AnnaKwa (approved):
LGTM!

PR #2229, opened 2023-05-24
Refactor purekeras predict

AnnaKwa:
This PR refactors `PureKerasModel.predict` out into a separate function in `fv3fit_shared.xr_prediction` so that the functionality of predicting on and returning a dataset can be shared across more model classes without inheriting from `PureKerasModel`. 

There are no code changes aside from moving the functions.

 
Added public API:
-  `fv3fit_shared.xr_prediction.predict_on_dataset`


external/fv3fit/tests/keras/test_pure_keras.py
```diff
@@ -31,38 +27,3 @@ def test_PureKerasDictPredictor_predict():
...
-
-
-@pytest.mark.parametrize(
-    ["unstacked_dims"],
-    [
-        pytest.param((), id="no_unstacked_dims"),
-        pytest.param(("z",), id="column_unstacked_dim"),
-    ],
-)
-def test_PureKerasModel_predict(unstacked_dims):
```
mcgibbon, line 43: Can we keep this test? It seems like it would still be functionality we want tested. Even if in your current code this only calls the helper function, someone else might change that in a later PR.

external/fv3fit/fv3fit/_shared/xr_prediction.py
```diff
@@ -0,0 +1,72 @@
...
+
+from fv3fit._shared.halos import append_halos, append_halos_using_mpi
+from fv3fit._shared import (
+    match_prediction_to_input_coords,
+    SAMPLE_DIM_NAME,
+    stack,
+)
+
+
+class Model(Protocol):
```
mcgibbon, line 13: Can you rename this to Predictor, which we used to avoid overloading the "model" name (even if I broke that rule when naming this directory)?

AnnaKwa: I renamed it to`ArrayPredictor` just in case someone looking at the type hint got confused with the actual `Predictor` class, which takes in and outputs datasets.

mcgibbon (approved):
LGTM, thanks for following up on this!

PR #2232, opened 2023-05-25
`DatasetPredictor`

AnnaKwa:
This adds the `DatasetPredictor` class, which uses the dataset prediction functions that were refactored out of `PureKerasModel` in a previous PR. This makes it easier to save other non-keras models as objects that still have the convenient feature of being able to take inputs/provide outputs  within prognostic runs.

The main difference between this and the `PureKerasModel` is that the dump/load is on a general base model, which should have predict, dump, and load methods, and must be in the `fv3fit.io` registry.

Added public API:
- `DatasetPredictor` 

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/xr_prediction.py
+++ b/external/fv3fit/fv3fit/_shared/xr_prediction.py
@@ -109,7 +109,7 @@ class DatasetPredictor(Predictor):
         input_variables: Iterable[Hashable],
         output_variables: Iterable[Hashable],
         model: ArrayPredictor,
-        unstacked_dims: Sequence[str],
+        unstacked_dims: Sequence[str] = ("z",),
         n_halo: int = 0,
     ):
         """Initialize the predictor
--- a/external/fv3fit/fv3fit/_shared/xr_prediction.py
+++ b/external/fv3fit/fv3fit/_shared/xr_prediction.py
@@ -105,7 +105,7 @@ class DatasetPredictor(Predictor):
         Args:
             input_variables: names of input variables
             output_variables: names of output variables
-            model: keras model to wrap
+            model: base model to wrap. Must be in fv3fit.io registry
             unstacked_dims: non-sample dimensions of model output
             n_halo: number of halo points required in input data
         """
--- a/external/fv3fit/fv3fit/_shared/xr_prediction.py
+++ b/external/fv3fit/fv3fit/_shared/xr_prediction.py
@@ -79,6 +79,7 @@ def predict_on_dataset(
     return match_prediction_to_input_coords(X, return_ds)
 
 
+@io.register("dataset-predictor")
 class DatasetPredictor(Predictor):
     """
     Base model is an ArrayPredictor which takes in a sequence of
--- a/external/fv3fit/fv3fit/_shared/xr_prediction.py
+++ b/external/fv3fit/fv3fit/_shared/xr_prediction.py
@@ -1,16 +1,23 @@
 import numpy as np
-from typing import Sequence, Iterable, Hashable, Protocol
+import os
+from typing import Sequence, Iterable, Hashable
 import xarray as xr
+import yaml
 
 from fv3fit._shared.halos import append_halos, append_halos_using_mpi
 from fv3fit._shared import (
     match_prediction_to_input_coords,
     SAMPLE_DIM_NAME,
     stack,
+    io,
+    Predictor,
+    get_dir,
+    put_dir,
 )
+from .predictor import Reloadable
 
 
-class ArrayPredictor(Protocol):
+class ArrayPredictor(Reloadable):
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/_shared/xr_prediction.py
```diff
@@ -70,3 +77,86 @@ def predict_on_dataset(
...
+            n_halo=self._n_halo,
+            unstacked_dims=self._unstacked_dims,
+        )
+
+    @classmethod
+    def load(cls, path: str) -> "DatasetPredictor":
+        """Load a serialized model from a directory."""
+        with get_dir(path) as path:
+            base_model_filename = os.path.join(path, cls._BASE_MODEL_NAME)
+            base_model: ArrayPredictor = io.load(base_model_filename)  # type: ignore
```
AnnaKwa, line 135: `ArrayPredictor` is a `Reloadable`, I'm not sure why mypy still complains here.

mcgibbon: `io.load` returns a `Reloadable`, which isn't guarunteed to be an `ArrayPredictor` from types alone. Only certain string values will give an `ArrayPredictor`. For type safety you should directly call `ArrayPredictor.load` instead of `io.load` if this truly is supposed to be an ArrayPredictor always, or type this as `Reloadable` if not.

I'm a little confused because I don't see where you defined `dump` and `load` on `ArrayPredictor`, but if you don't define them it should cause an error when initialized. Did you add a test that uses this class?

brianhenn: Or yeah, make `ArrayPredictor` another subclass of `Reloadable` that is like `Predictor` but does the type of predictions you want here.

mcgibbon: Predictor is a dataset predictor, it's predict api takes in an xr.Dataset. If you like you could rename Predictor to `DatasetPredictor` to make it clear how it relates to `ArrayPredictor`. Given 

Reloadable is just a protocol saying that the thing you return from the train function must be dumpable and loadable, it doesn't implement those methods.

I would strongly encourage implementing as little functionality as you can in an abstract base class, so that the ABC acts only as an interface/defines what should be implemented, and the code implementation itself doesn't use inheritance/isn't split between two classes.

It sounds like what you want to make is an analog to PureKerasModel, but one that wraps an ArrayPredictor to make a Predictor. You probably want to make a concrete implementation of Predictor for this, instead of defining another ABC/interface. If you do that, its functionality/code can be tested the same way we test PureKerasModel.

AnnaKwa: After discussion with Jeremy and Brian, I changed the `ArrayPredictor` to be an abstract base class and will keep the current usage of `io.load/dump` in the `DatasetPredictor`'s dump/load methods (since this class will be used for multiple types of ArrayPredictors). 

I did look at unifying these changes with `PureKerasModel` but realized that class has its own additional methods like `input_sensitivty, get_dict_compatible_model` that are very specific to having a keras model attribute, so I am leaving this new `DatasetPredictor` as a separate class. Both `PureKerasModel` and `DatasetPredictor` do at least make use of the same `predict_on_dataset` function that was refactored out in a previous PR.

external/fv3fit/tests/test_xr_prediction.py
```diff
@@ -39,3 +42,81 @@ def _get_dummy_model(input_shape):
     out_ = tf.keras.layers.Lambda(lambda x: x)(in_)
     model = tf.keras.Model(inputs=in_, outputs=out_)
     return model
+
+
+@io.register("constant-array")
```
brianhenn, line 47: Nit: We've [sometimes](https://github.com/ai2cm/fv3net/blob/eb8fc6edcf26ed4a09d74ff6edd961191813060f/external/fv3fit/tests/test_register_model.py#L17) used context managers to avoid adding test cases to registries, i.e., add it only for the time the test is run.

external/fv3fit/fv3fit/_shared/xr_prediction.py
```diff
@@ -70,3 +89,89 @@ def predict_on_dataset(
...
+            model: base model to wrap. Must be in fv3fit.io registry
+            unstacked_dims: non-sample dimensions of model output
+            n_halo: number of halo points required in input data
+        """
+        super().__init__(input_variables, output_variables)
+        self.input_variables = input_variables
+        self.output_variables = output_variables
+        self.model = model
+        self._n_halo = n_halo
+        self._unstacked_dims = unstacked_dims
```
brianhenn, line 129: Given that our use case is typically that unstacked_dims = ["z"], I'm lean towards making that the default value here.  Similarly with n_halo = 0.

brianhenn (approved):
Thanks, approving with a couple of optional suggestions below.

PR #2233, opened 2023-05-25
Sktransform

AnnaKwa:
This adds a `SkTransformer` class which applies scaling and transform / inverse transforms. Its `.predict` method is a round trip of encode/decode. In practice, this class will be used as an encoder/decoder in reservoir models.

It includes an option for limiting all outputs to be positive. This is a pretty blunt tool for preventing specific humidities from being decoded as negative values, which I observed as a common failure mode which crashed simulations where encode/decode round trips were applied to FV3GFS states every few hours.

I also moved a couple sklearn related test modules into a `tests_sklearn` dir to try organize the fv3fit tests.

 Added public API:
- `SkTransformer`

- [x] Tests added


external/fv3fit/fv3fit/sklearn/transformer.py
```diff
@@ -0,0 +1,95 @@
...
+        self,
+        transformer: TransformerMixin,
+        scaler: StandardScaler,
+        enforce_positive_outputs: bool = False,
+    ):
+        self.transformer = transformer
+        self.scaler = scaler
+        self.enforce_positive_outputs = enforce_positive_outputs
+
+    def predict(self, x: Sequence[np.ndarray]) -> Sequence[np.ndarray]:
```
frodre, line 46: I think this could use some documentation.  I don't offhand know why we might want a rountrip method named "predict".  I assume it's just to adhere to the structure of an interface?

external/fv3fit/fv3fit/sklearn/transformer.py
```diff
@@ -0,0 +1,95 @@
...
+        encoded = self.encode(x_concat)
+        decoded = self.decode(encoded)
+        decoded_split_features = np.split(
+            decoded, np.cumsum(original_feature_sizes[:-1]), axis=-1
+        )
+
+        return decoded_split_features
+
+    def encode(self, x):
+        x = _ensure_sample_dim(x)
```
frodre, line 58: Is this the right level to enforce the need for a sample dimension?  If, so I'd add some tests related to the `SkTransformer` and documentation that it expects a certain shape of data.

external/fv3fit/tests/sklearn/test_transformer.py
```diff
@@ -0,0 +1,96 @@
...
+
+class DummyTransformer:
+    def __init__(self, latent_dim, output_dim, encoded_constant, decoded_constant):
+        self.latent_dim = latent_dim
+        self.output_dim = output_dim
+        self.encoded_constant = encoded_constant
+        self.decoded_constant = decoded_constant
+
+    def transform(self, X):
+        return np.ones(self.latent_dim) + self.encoded_constant
```
frodre, line 15: To try and test a true roundtrip, you could slot the input data into a larger array and then select it out on the inverse.  That would allow for an easier to understand test below.

external/fv3fit/tests/sklearn/test_transformer.py
```diff
@@ -0,0 +1,96 @@
...
+    scaler = StandardScaler()
+    scaler.fit(np.random.rand(5, n_vars * output_dim))
+    transformer = transformer = DummyTransformer(
+        latent_dim, n_vars * output_dim, encoded_constant, decoded_constant
+    )
+    sktransformer = SkTransformer(transformer, scaler, enforce_positive_outputs=False)
+
+    x = [np.random.rand(output_dim) for i in range(n_vars)]
+    target = np.ones(n_vars * output_dim) + decoded_constant
+    outputs = sktransformer.predict(x)
```
frodre, line 37: As it currently stands, there's no true test of the outputs of the `encode` method used in `predict` roundtrip since the `DummyTransformer` doesn't use the input `X`.

external/fv3fit/tests/sklearn/test_transformer.py
```diff
@@ -0,0 +1,96 @@
...
+    transformer = transformer = DummyTransformer(
+        latent_dim, n_vars * output_dim, encoded_constant, decoded_constant
+    )
+    sktransformer = SkTransformer(transformer, scaler, enforce_positive_outputs=False)
+    sktransformer_enforce_positive = SkTransformer(
+        transformer, scaler, enforce_positive_outputs=True
+    )
+
+    x = [np.arange(output_dim) * 10.0, np.arange(output_dim) * -10.0]
+    outputs = np.concatenate(sktransformer.predict(x), axis=-1).reshape(-1)
```
frodre, line 61: Instead of all the shape manipulation, can we just check that the split outputs are equal for for the first set of features and `0.0` for the second?

frodre (changes requested):
Thanks, @AnnaKwa.  Only a few minor changes, I think there should be some documentation explaining `predict` and then I suggested some simplifications that I think would make the tests easier to understand.

external/fv3fit/fv3fit/sklearn/transformer.py
AnnaKwa: Right, since this is going to be used in the same way as the dense autoencoder, I made its `predict` method behave similarly. This method is unlikely to actually get called in the actual reservoir experiments, it's main usefulness is when testing whether the autoencoder/transformer's error will negatively affect the prognostic run (e.g. the experiments I did last week). I'll add a comment explaining this.

external/fv3fit/fv3fit/sklearn/transformer.py
AnnaKwa: Yes, the sklearn models seem to require this while IIRC keras models are smarter about figuring it out if you only give a single sample. I'll make this clearer and add a test.

external/fv3fit/tests/sklearn/test_transformer.py
AnnaKwa: Simplified the test so one variable is all negative and other is all positive

external/fv3fit/fv3fit/sklearn/transformer.py
AnnaKwa: To clarify, the behavior here is that it can either take inputs with or without a sample dim so that the behavior is similar to the base keras models (LMK if I am wrong about that).

frodre: I don't remember off the top of my head, but thanks for adding some documentation on tests around the expectation!

frodre (approved):
Thanks for the updates!  LGTM

PR #2236, opened 2023-05-31
fv3fit: squashed output model

brianhenn:
Allows outputs of an `fv3fit.Predictor` model to be squashed (set to 0 or another specified value) for all outputs of that sample, if the predicted value is below a given threshold for a specified output. For example, allows a sample in which the predicted cloud amount field is less than a threshold value to have cloud amount and other predictors (mixing ratios) set to zero. [Here's](https://github.com/ai2cm/explore/blob/master/brianh/2023-05-03-cloud-ml-NN-debugging/2023-05-24-dense-model-output-squashing.ipynb) a notebook showing these models behave as expected.

Added public API:
- `SquashedOutputModel` added to `fv3fit.io` registry

- [X] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/models.py
+++ b/external/fv3fit/fv3fit/_shared/models.py
@@ -437,3 +437,64 @@ class OutOfSampleModel(Predictor):
 
         model = cls(base_model, novelty_detector, cutoff=cutoff, taper=taper)
         return model
+
+
+@io.register("squashed_output_model")
+class SquashedOutputModel(Predictor):
+
+    _CONFIG_FILENAME = "squashed_output_model.yaml"
+
+    def __init__(
+        self,
+        base_model: Predictor,
+        squash_by_name: Optional[Hashable],
+        squash_threshold: Optional[float],
+        squash_to: Optional[float],
+    ):
+        if squash_by_name is not None:
+            if squash_threshold is None or squash_to is None:
+                raise ValueError(
+                    "If squash_by_name is specified, squash_threshold and squash_to "
+                    "must be also."
+                )
+            if squash_by_name not in base_model.output_variables:
+                raise ValueError(
+                    f"The squash by variable ({squash_by_name}) must among the "
+                    "set of model output variable names."
+                )
+
+        self._base_model = base_model
+        self._squash_by_name = squash_by_name
+        self._squash_threshold = squash_threshold
+        self._squash_to = squash_to
+
+        super().__init__(
+            input_variables=base_model.input_variables,
+            output_variables=base_model.output_variables,
+        )
+
+    def predict(self, X: xr.Dataset) -> xr.Dataset:
+        raw_predictions = self._base_model.predict(X)
+        if self._squash_by_name is not None:
+            squashed_predictions = raw_predictions.where(
+                raw_predictions[self._squash_by_name] > self._squash_threshold,
+                self._squash_to,
+            )
+        else:
+            squashed_predictions = raw_predictions
+        return squashed_predictions
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -437,3 +437,64 @@ def load(cls, path: str) -> "OutOfSampleModel":
...
+
+@io.register("squashed_output_model")
+class SquashedOutputModel(Predictor):
+
+    _CONFIG_FILENAME = "squashed_output_model.yaml"
+
+    def __init__(
+        self,
+        base_model: Predictor,
+        squash_by_name: Optional[Hashable],
```
AnnaKwa, line 450: Since this model is manually constructed, it seems like there would not be a scenario where this is `None` and the prediction is just the base model prediction. Unless there's a case where you could imagine this, I'd suggest just making the type non optional and removing the if statement.

external/fv3fit/fv3fit/_shared/models.py
```diff
@@ -437,3 +437,64 @@ def load(cls, path: str) -> "OutOfSampleModel":
...
+
+        super().__init__(
+            input_variables=base_model.input_variables,
+            output_variables=base_model.output_variables,
+        )
+
+    def predict(self, X: xr.Dataset) -> xr.Dataset:
+        raw_predictions = self._base_model.predict(X)
+        if self._squash_by_name is not None:
+            squashed_predictions = raw_predictions.where(
```
AnnaKwa, line 479: If there could be use cases where you don't want to squash everything according to the single `squash_by_name` output, I'd suggest using a pattern similar to what we've done in other models where they're initialized by a sequence of the desired transform / taper / etc configs (https://github.com/ai2cm/fv3net/blob/2a99e9db45966694db9d412dcb0887040fea0709/external/fv3fit/fv3fit/_shared/models.py#L285)

external/fv3fit/tests/test_squashed_output_model.py
```diff
@@ -0,0 +1,62 @@
...
+        pytest.param(
+            0.0,
+            1.5,
+            {"a": [[0.0, 0.0, 0.0, 0.0, 2.0]], "b": [[0.0, 0.0, 0.0, 0.0, 0.2]]},
+            id="1.5_to_zero",
+        ),
+        pytest.param(
+            0.0,
+            -1.5,
+            {"a": [[0.0, -1.0, 0.0, 1.0, 2.0]], "b": [[0.0, -0.1, 0.0, 0.1, 0.2]]},
```
AnnaKwa, line 25: The original `b` array has -0.2 as its first element, shouldn't that remain as is and not get squashed to 0.0? I could also be just missing something.

external/fv3fit/fv3fit/_shared/models.py
AnnaKwa: If not, then could you make it clear in the docs that *everything* will be squashed according to the values of `squash_by_name`? Before getting to this part of the code, I thought that only that single field would be squashed.

AnnaKwa (approved):
Thanks for making the change to the wrapped model! I had a few more comments/suggestions, you can decide whether to implement based on how you see the use case for this wrapper in the future.

external/fv3fit/tests/test_squashed_output_model.py
brianhenn: The idea is that the squashing for all variables is determined by the values in the `squash_by_name` field (here "a", which is less than the threshold, so "b" gets squashed). Since this seems not very clear, I think it's a good idea to move to configuring both the `squash_by_name` and `squash_target_names` explicitly.

PR #2237, opened 2023-05-31
Add option to coarsen restart files on pressure with limited extrapolation

spencerkclark:
This PR adds the ability to coarsen restart files on pressure levels with extrapolation via a newly added `extrapolate` keyword argument to `coarsen_restarts_on_pressure`.   The extrapolation is limited in the sense that we have only shifted the threshold for masking to the coarse grid's vertical layer midpoint pressures instead of the coarse grid's vertical layer bottom interface pressures.  Using the latter results in zero extrapolation; using the former results in allowing values filled with nearest neighbor extrapolation to be included in averages if extrapolation was not required too far.  This is mostly relevant over ocean where our default no-extrapolation approach requires masking roughly 50% of the grid cells in the lowest model level; this limited extrapolation approach enables retaining 100% of the grid cells over most ocean regions while maintaining the mask over land regions with variable topography.

2023-06-01-C3072-to-C48-unmasked-area-comparison

The above figure shows the fraction of the area of the C3072 grid cells left unmasked in the lowest model level when coarsening to C48 resolution for three different approaches.  The left and middle panels represent our default and this new limited extrapolation approach.  Unlimited extrapolation, the right panel, obviously results in no masking ever occurring.  This plot was produced in [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-01-coarsening-with-mask.ipynb) (though note that it uses a modified version of the code in this PR to enable experimenting with the different options and easy access to the `masked_area_weights` used in the process).

An accompanying PR adding this option in the fortran can be found here: ai2cm/fv3gfs-fortran#366.   [This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-01-validate-pressure-level-extrapolate-coarsening/2023-06-01-validate-pressure-level-extrapolate-coarsening.ipynb) checks that the offline approach implemented here approximately produces the same results as the online approach implemented in ai2cm/fv3gfs-fortran#366.

- [x] Tests added


external/vcm/vcm/cubedsphere/regridz.py
```diff
@@ -140,25 +164,41 @@ def _regrid_given_delp(
...
     )
 
     return ds_regrid, masked_weights
 
 
 def _mask_weights(
     weights,
+    pfull_coarse_on_fine,
     phalf_coarse_on_fine,
     phalf_fine,
```
elynnwu, line 186: [nit] type hint for these variables

external/vcm/vcm/cubedsphere/regridz.py
```diff
@@ -140,25 +164,41 @@ def _regrid_given_delp(
...
+    extrapolate=False,
 ):
-    return weights.where(
-        phalf_coarse_on_fine.isel({dim_outer: slice(1, None)}).variable
-        < phalf_fine.isel({dim_outer: -1}).variable,
-        other=0.0,
-    ).rename({dim_outer: dim_center})
+    if extrapolate:
+        return weights.where(
+            pfull_coarse_on_fine.variable < phalf_fine.isel({dim_outer: -1}).variable,
```
elynnwu, line 193: I think this is probably "safe" because -1 is always the surface in fv3. But it may be better to make a surface level variable so that it's more clear?
```
surface_level = -1
pfull_coarse_on_fine.variable < phalf_fine.isel({dim_outer: surface_level}).variable
```

external/vcm/vcm/cubedsphere/regridz.py
```diff
@@ -140,25 +164,41 @@ def _regrid_given_delp(
...
+            other=0.0,
+        )
+    else:
+        return weights.where(
+            phalf_coarse_on_fine.isel({dim_outer: slice(1, None)}).variable
+            < phalf_fine.isel({dim_outer: -1}).variable,
```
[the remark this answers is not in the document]
elynnwu: Thanks for the clarification, this makes a lot more sense now.

elynnwu (approved):
lgtm!

PR #2238, opened 2023-06-08
Allow use of multiple steppers in postphysics

AnnaKwa:
These changes allow for multiple steppers (e.g. `PureMLStepper`, `Prescriber`, etc)  to be used in the postphysics step. Currently, the `prepare_config` step does not allow user configs to contain more than one of [`bias_correction`, `nudging`, `scikit_learn`] entries. This behavior was blocking the training data generation for hybrid reservoir models, since both the interval prescriber and ML corrections need to be called in the postphysics.

Instead of disallowing any combination of postphysics steppers, I changed the prepare_config check to complain if nudging was included with another stepper.

I refactors the code in `_get_postphysics_stepper` to use the same method `_get_stepper` as `_get_prephysics_stepper` and to allow for a `CombinedStepper` to be returned.

I also added an `offset_seconds` option to the `IntervalConfig` that can be taken into account when the  `IntervalStepper` is called. The wrapper time updates in the middle of the loop (before postphysics) but the set of saved times in diagnostic files is calculated starting on the initial time. i.e. if no interval offset is specified, the reference fields are prescribed every HH:15 but the diagnostics will be saved on HH:00, so the timepoints that are needed for training are not saved. Setting the offset to -timestep in the interval allows the prescribed outputs to be applied and saved on the desired output time interval.

- [x] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/runtime/config.py
+++ b/workflows/prognostic_c48_run/runtime/config.py
@@ -62,12 +62,13 @@ class UserConfig:
     prephysics: Optional[
         List[Union[PrescriberConfig, MachineLearningConfig, IntervalConfig]]
     ] = None
+
     scikit_learn: Optional[MachineLearningConfig] = None
     nudging: Optional[NudgingConfig] = None
     tendency_prescriber: Optional[TendencyPrescriberConfig] = None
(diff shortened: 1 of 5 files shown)


brianhenn (approved):
LGTM @AnnaKwa . I have one optional suggestion.

PR #2239, opened 2023-06-08
Coarsening: add adiabatic adjustment of extrapolated temperatures

spencerkclark:
This PR is a followup to #2237.  It implements an approach to treat temperature specially during extrapolation.  Namely we start by using nearest neighbor extrapolation and propagating the lowest-level fine-grid temperature down to the needed pressure levels.  We then convert those extrapolated temperatures to potential temperatures of parcels moved adiabatically from the fine-grid lowest-level midpoint pressures to the coarse-level midpoint pressures we are extrapolating to. 

In general the effect of this is small even in the lowest model level ([global mean difference of 0.01 K](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-06-coarsening-with-adiabatic-extrapolation.ipynb)), but this type of adjustment is physically motivated and is simple enough to implement so we go forward with it here.

Added public API:
- added `temperature` argument to `regrid_to_area_weighted_pressure` for use with temperature variables.
- added `z_dim_outer` arguments to `regrid_to_area_weighted_pressure` and `regrid_to_edge_weighted_pressure` so that we can pass them down to downstream functions if needed.
- added `reference_pressure` and `poisson_constant` arguments to `vcm.calc.thermo.local.potential_temperature`.  These default to the previous hard-coded values in the function.

A notebook showing that the implementation in this PR produces matching results with the fortran can be found [here](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-08-validate-pressure-level-extrapolate-coarsening-with-adjustment/2023-06-08-validate-pressure-level-extrapolate-coarsening.ipynb).

xref: ai2cm/fv3gfs-fortran#366

- [x] Tests added


external/vcm/vcm/cubedsphere/regridz.py
```diff
@@ -196,11 +225,53 @@ def _mask_weights(
...
+    z_dim_outer: str = RESTART_Z_OUTER,
+) -> xr.DataArray:
+    strictly_interpolated = _compute_strictly_interpolated(
+        phalf_fine, phalf_coarse_on_fine, dim_center=z_dim_center, dim_outer=z_dim_outer
+    )
+
+    # Compute the potential temperature at all grid cells using the coarse model
+    # level midpoint pressures as a target and the fine lowest level midpoint
+    # pressure as a reference.
+    cp_air = 1004.6
```
spencerkclark, line 262: I would like to use: https://github.com/ai2cm/fv3net/blob/7874333ba36f5589cc43da66800c0a4f6d560727/external/vcm/vcm/calc/thermo/constants.py#L9
but it's oddly inconsistent with [what is used in FV3GFS](https://github.com/ai2cm/fv3gfs-fortran/blob/cb106f2eb806e8c635d28d8b76ee8e80a0e20bc3/FMS/constants/constants.F90#L55).  I may float the idea of changing it so that it is consistent with FV3GFS.

external/vcm/tests/test_regridz.py
```diff
@@ -125,7 +127,9 @@ def test__mask_weights(extrapolate):
     phalf_fine[1, :, 2] = 3.5
 
     # Use a simpler way to compute pfull for the sake of testing.
-    pfull_coarse_on_fine = phalf_coarse_on_fine.diff("z")
+    pfull_coarse_on_fine = _interface_to_midpoint(
+        phalf_coarse_on_fine, dim_center="z", dim_outer="z"
+    )
```
spencerkclark, line 132: This is what I intended to do in #2237...oddly enough the test still passed with `diff("z")` even though that's not really an approach for computing the pressure at level midpoints.

external/vcm/vcm/cubedsphere/regridz.py
elynnwu: Right, fv3gfs also uses a different set of constants depending on the compiler flag `GFS_PHYS`. Probably good to consolidate

external/vcm/vcm/cubedsphere/regridz.py
```diff
@@ -155,25 +171,38 @@ def _regrid_given_delp(
...
-        delp_fine, dim_center=z_dim, dim_outer=RESTART_Z_OUTER
+        delp_fine, dim_center=z_dim, dim_outer=z_dim_outer
     )
+    pfull_fine = pressure_at_midpoint_log(delp_fine, dim=z_dim)
+    pfull_coarse_on_fine = pressure_at_midpoint_log(delp_coarse_on_fine, dim=z_dim)
 
     ds_regrid = xr.zeros_like(ds)
     for var in ds:
-        ds_regrid[var] = regrid_vertical(
+        remapped = regrid_vertical(
```
elynnwu, line 184: It looks like `regrid_vertical` already has an input `iv` which supposedly indicates what kind of variables are being regridded. Though here `iv` doesn't appear to differ for different variables. I was going to suggest wrapping the boolean `temperature` in `regrid_vertical` so you don't have to have another if statement below. But then it needs the extra arguments, which is not ideal. So it's probably fine as it is, though I'm curious if `iv` should've been different?

spencerkclark: That's right -- we've always used `iv = 1` in our coarsening work.  It's been a long time since we thought about this, and I think @oliverwm1 might have been the one to initially select that. It's possible we decided to just use the generic remapping option (`iv = 1`) to keep things simple and predictable.

lharris4: The `iv = 1` in conjunction with `kord = 1`, the original monotone PPM constraint, should be fine for your purposes; this will not create new extrema that need an additional limiter to enforce positivity, and doesn't apply any special boundary conditions at the top and the bottom of the source domain.

spencerkclark:
Thanks for weighing in @lharris4 -- I'm using dry potential temperature, so a dry adiabatic lapse rate (9.8 K/km).  I suppose I could switch to using moist potential temperature, which would imply a lapse rate in the 6.5 K/km range.

lharris4:
Hi, Spencer. Thanks. This has the advantage of simplicity (theta-e is much
harder to compute) and if it works I would recommend going for it. I would
be careful to make sure that the coarsened temperatures are not too
warm/cold in areas of very steep terrain, although this could just average
out.

Lucas

spencerkclark:
Thanks @lharris4.  With this "limited" extrapolation approach, we are only extrapolating at most about half a model level, so generally we will not be extrapolating very much even in mountainous terrain.  This strategy is mainly targeted at removing biases induced by masking over ocean.  According to Chris, the lapse rate in the boundary layer of most ocean regions is approximately dry adiabatic, so he recommends sticking with this simple potential temperature approach.  Admittedly it will not make a big difference either way.

lharris4:
Hi, Spencer. Thanks for your explanation. Simpler is better in many if not
most cases and if it is just 0.5dz then it shouldn't make much of a
difference.

Lucas

elynnwu (approved):

spencerkclark:
After further discussion we have determined that since `mappm` is a finite volume vertical remapping algorithm, the potential temperature adjustment approach implemented here is not applicable.  Since `mappm` will only need to use an extrapolated temperature profile for at most half of a coarse grid vertical level, and can use an interpolation profile for at least the other half, any adjustment would have to be carefully implemented, and would likely be smaller than that implemented here.  Therefore I am closing this PR.

PR #2241, opened 2023-06-16
Use generic transform to reduce dims in reservoir

AnnaKwa:
Several changes to hybrid reservoir training with dimensionality reduction transforms: 

 - updates the reservoir training function to work with any `Transformer` class to reduce dimensionality in the inputs (can be either a dense autoencoder or sklearn transform). Adds abstract base class `Transformer` which has `encode, decode` methods and `n_latent_dims` attribute which `Autoencoder` and `SkTransformer` subclass from.
 
- Small change to `SkTransformer` so that concatenating inputs given as [variable, time, feature] -> [time, variable*feature] occurs in `encode` rather than `predict`. This is because the training, which provides data in this format, calls `encode` rather than `predict`.

- changes the hybrid inputs to also use the same dimensionality reduction transform as the reservoir inputs. Previously, the hybrid 'imperfect state' data was left as is with full vertical dimension, but this uses a lot of memory and constrains the size of reservoir that can be used. The linear regression readout training involves saving a 
a matrix X_T.X which has (reservoir_state_size + (hybrid_vars*subdomain_size)) **2 elements, which is updated for each batch of training data seen. Initializating this matrix takes ~500 Mb  per subdomain for a reservoir of state size 2000 and two full columns of hybrid input (158 vertical levels per grid point) in a 2x2 subdomain with overlap of 2. This OOM'ed on a vm with 27 Gb of memory available at the start of  training.


external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -194,7 +207,9 @@ def load(cls, path: str) -> "ReservoirComputingModel":
             rank_divider = None
 
         if fs.exists(os.path.join(path, cls._AUTOENCODER_SUBDIR)):
-            autoencoder = Autoencoder.load(os.path.join(path, cls._AUTOENCODER_SUBDIR))
+            autoencoder: ReloadableTransfomer = _load_transformer(
```
AnnaKwa, line 210: Mypy was complaining because `fv3fit.load` can return any `Reloadable`, but the autoencoder arg should only be a sklearn transform or dense autoencoder. I figured this was a sign I should have some kind of enforcement on what could get loaded as the autoencoder arg.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -1,19 +1,32 @@
...
 from .reservoir import Reservoir
 from .domain import RankDivider
 from fv3fit._shared import io
 from .utils import square_even_terms
-from .autoencoder import Autoencoder
+from .transformers import Autoencoder, SkTransformer, ReloadableTransfomer
 from ._reshaping import flatten_2d_keeping_columns_contiguous
 
 
+def _load_transformer(path) -> Union[Autoencoder, SkTransformer]:
```
frodre, line 18: return type of `ReloadableTransformer`?

external/fv3fit/fv3fit/reservoir/model.py
frodre: No need for the type hint if you specify the `_load_transformer` type, right?

external/fv3fit/fv3fit/reservoir/transformers/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .transformer import Transformer
```
frodre, line 1: I would consider removing `Transformer` at the top level if `ReloadableTransformer` is the public facing type used elsewhere.  I do understand that this is the more general type, but it might be confusing why `ReloadableTransformer` is when they all appear to be `Transformer` types.

external/fv3fit/tests/reservoir/test_sk_transformer.py
```diff
@@ -99,6 +99,34 @@ def test_sktransformer_no_sample_dim():
...
 
 
+def test_sktransformer_encode_need_to_concat_features_and_add_sample_dim():
+    output_dim = 10
+    n_vars = 3
+
+    scaler = StandardScaler()
+    scaler.fit(np.random.rand(5, n_vars * output_dim))
+    pad_size = 5
+    transformer = transformer = DummyTransformer(extra_embedded_dims=pad_size)
```
frodre, line 109: transfor
```suggestion
    transformer = DummyTransformer(extra_embedded_dims=pad_size)
```

external/fv3fit/tests/reservoir/test_sk_transformer.py
```diff
@@ -99,6 +99,34 @@ def test_sktransformer_no_sample_dim():
...
+
+
+def test_sktransformer_encode_need_to_concat_features():
+    output_dim = 10
+    n_vars = 3
+
+    scaler = StandardScaler()
+    scaler.fit(np.random.rand(5, n_vars * output_dim))
+    pad_size = 5
+    transformer = transformer = DummyTransformer(extra_embedded_dims=pad_size)
```
frodre, line 123:
```suggestion
    transformer = DummyTransformer(extra_embedded_dims=pad_size)
```

frodre (approved):
Cool, LGTM.  Thanks for bringing together a simple interface for these components.  Just a few minor comments here and there.

external/fv3fit/fv3fit/reservoir/model.py
AnnaKwa: I did try that first, but mypy was unhappy.

frodre: The altar of mypy is absolute 😂

PR #2248, opened 2023-06-21
Train and offline compute work with latest scream output

elynnwu:
Update variable names to match latest scream output. Post-process script updated also to reflect the latest changes in scream.

### DIFF
--- a/projects/scream/scream_post/format_output.py
+++ b/projects/scream/scream_post/format_output.py
@@ -206,7 +206,6 @@ if __name__ == "__main__":
     if args.split_horiz_winds:
         ds = split_horiz_winds(ds)
     if args.calc_physics_tend:
-        print(nudging_vars)
         ds = compute_tendencies_due_to_scream_physics(ds, nudging_vars)
     if args.rename_nudging_tend:
         ds = rename_nudging_tendencies(ds, nudging_vars)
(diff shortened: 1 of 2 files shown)


projects/scream/scream_post/format_output.py
```diff
@@ -72,94 +73,163 @@ def make_new_date(dates, year_offset=2015):
...
-        "y_wind_tendency_due_to_nudging",
+    return parser
+
+
+if __name__ == "__main__":
+    parser = _get_parser()
+    args = parser.parse_args()
+    ds = xr.open_mfdataset(args.input_data)
+    nudging_vars = [str(item) for item in args.nudging_variables.split(",")]
+    if args.split_horiz_winds:
```
frodre, line 206: The chain of operations seems like good opportunity for some kind of transform pipeline with a configuration spec.  I think we've generally found that using configurations makes reproducibility artifacts much easier than having runtime flags.  No need for changes though since this works too.

projects/scream/scream_post/format_output.py
```diff
@@ -72,94 +73,163 @@ def make_new_date(dates, year_offset=2015):
...
+        ds = split_horiz_winds(ds)
+    if args.calc_physics_tend:
+        print(nudging_vars)
```
frodre, line 209: Using logging anywhere else?

frodre (approved):
LGTM.

projects/scream/scream_post/format_output.py
elynnwu: Oops, that was a leftover print from debugging. Deleted the print line

PR #2249, opened 2023-06-23
Add reservoir data preprocessing opts

AnnaKwa:
This adds the following CLI options to the preprocessing script that saves reservoir training data by rank:

- `--variables, --ranks`: saves only the specified variables and ranks out of the global data
- ` --time-sample-interval`: save timesteps on this inteval. ex. if original data is every 1.5h, setting this to 2 will save rank data every 3h.
- `--additional-paths`: if variables are contained across multiple datasets, put additional dataset paths here to load.

Moves the `cubedsphere` module containing `CubedsphereDivider` to `fv3fit.reservoir`. There were import issues with running the `save_ranks` script that broke the original tests if fixed, and I didn't think it was worth turning `projects/reservoir/fv3` into a package just to address that. I deleted the cli test since those were added as part of a misunderstanding with the original PR comments.


frodre (approved):
LGTM

PR #2251, opened 2023-06-26
Add initial super-resolution example script

mcgibbon:
This PR adds an example script evaluating a nearest-neighbor super-resolution scheme with plots through fv3viz. It can serve as an entrypoint/introduction for @cynicalsavant 's work, showing some of the idiosyncracies of data loading and plotting with our tools, and demonstrating some of the ways an atmospheric scientist might want to see these models evaluated.


projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+def repeat_arr(arr, upscale_factor):
+    # This repeats values along the last dimension
+    return np.repeat(arr, upscale_factor, axis=-1)
+
+
+def upscale_nearest(ds, upscale_factor, dim):
+    ds_upscaled = xr.apply_ufunc(
+        repeat_arr,
+        ds,
+        input_core_dims=[[dim]],  # apply function along this dimension
```
gideonite, line 23: nit: I think that these comments are unnecessary. They are repeating the definition of core_dim.

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+
+DPI = 800
+
+
+def repeat_arr(arr, upscale_factor):
+    # This repeats values along the last dimension
+    return np.repeat(arr, upscale_factor, axis=-1)
+
+
+def upscale_nearest(ds, upscale_factor, dim):
```
gideonite, line 19: nit: Not sure if we are actually going to check the types of these arguments, but it is helpful to the reader to know, for example that upscale_factor is an int.

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def plot_pattern_bias(output: xr.Dataset, target: xr.Dataset, varname: str):
+    """Plot the pattern bias between two datasets
+
+    Args:
+        output (xr.Dataset): output dataset
+        target (xr.Dataset): target dataset
+    """
+    # compute the pattern bias
```
gideonite, line 44: nit: Might as well make this a separate function?

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+    out_mean: xr.Dataset = output.mean("time")
+    target_mean: xr.Dataset = target.mean("time")
+    vmin = min(
```
gideonite, line 47: If you make computing the pattern bias a separate function, I think that would helpful clarify this a lot. e.g. it looks like vmin and vmax are only used for plotting.

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+    ax[1].set_title("Time-mean Target")
+    fv3viz.plot_cube(
+        # must merge grid data into the dataset
+        # after performing bias calculation on dataset
+        bias.merge(GRID, compat="override"),
+        var_name=varname,
+        cmap="RdBu_r",
+        ax=ax[2],
+    )
+    # TODO: add area-weighting to bias calculation
```
gideonite, line 85: Interesting, I guess this isn't terribly important at the moment?

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+import numpy as np
+import fv3viz
+from vcm.catalog import catalog
+import matplotlib.pyplot as plt
+import os
+import cartopy.crs as ccrs
+
+GRID = catalog["grid/c384"].read()
+
+DPI = 800
```
gideonite, line 11: nit: I would prefer having this as an argument to the function, perhaps with this as a default value.

projects/super_res/super_res.py
```diff
@@ -0,0 +1,203 @@
...
+    plt.tight_layout()
+    fig.savefig(f"plots/{varname}-pattern_bias.png", dpi=DPI)
+
+
+def plot_first_snapshot(output: xr.Dataset, target: xr.Dataset, varname: str):
+    # must specify projection for subplots if calling plot_cube with an ax argument
+    fig, ax = plt.subplots(
+        3, 1, figsize=(6, 9), subplot_kw={"projection": ccrs.Robinson()}
+    )
+    output_snapshot = output.isel(time=0, drop=True)
```
gideonite, line 100: Maybe I'm missing something but is there something special about the first snapshot, e.g. `time=0`? Maybe it would make sense to make this an argument to this function?

projects/super_res/super_res.py
gideonite: nit: I would find it helpful to include "neighbor" in this function name.

projects/super_res/plots/.gitignore
```diff
@@ -0,0 +1 @@
+*.png
```
gideonite, line 1: nit: Not sure if this is important but there's that newline thing.

gideonite (commented):
Some comments for discussion.

I'm also a little confused about how this code actually works. The upscaling is done by repeating the data along a new axis. Is this right? How is the data reshaped to remove this new axis?

projects/super_res/super_res.py
mcgibbon: I don't expect Prakhar to know the definition of a `core_dim` as a new xarray user, so I added helpful comments like this. `apply_ufunc` in particular is pretty idiosyncratic. He can delete comments like this as desired, if he wants me to do that in this PR I can.

projects/super_res/super_res.py
mcgibbon: Maybe, but I think this is too much on the elegance side of elegant vs explicit for a single-user plotting script. When I make plotting routines I find it helpful to be able to change how and what is plotted when editing the same function, and I generally keep both those pieces of information in my working context/memory.

projects/super_res/super_res.py
mcgibbon: I moved the `bias` calculation up to make this more explicit. I generally prefer having long plotting functions, but I'm going to leave it up to @cynicalsavant how he wants to build/modify this script and the style he wants to set for it.

projects/super_res/super_res.py
mcgibbon: Not really, I left this out as it's a feature @cynicalsavant can add. It doesn't really take away from being able to use the plots to demonstrate the science.

projects/super_res/super_res.py
mcgibbon: It can. I just included this as a quick example to look at what the model does in one timestep.

mcgibbon:
> I'm also a little confused about how this code actually works. The upscaling is done by repeating the data along a new axis. Is this right? How is the data reshaped to remove this new axis?

`apply_ufunc` is a bit confusing, the code repeats the data along the existing `core_dim` axis when called. Not a new axis.

projects/super_res/super_res.py
cynicalsavant: I'd liek to leave the comments for now, they were super useful.

projects/super_res/super_res.py
cynicalsavant: Sounds good, although a separate function might make it more elegant. In any case, I'll adapt it to make it more suitable to my needs in the coming future. Nevertheless, it's super helpful as a starting point.

gideonite (approved):

projects/super_res/super_res.py
mcgibbon: Yeah, whatever you find easiest! Glad to help :)

PR #2252, opened 2023-06-27
Add mapping adapter for HybridReservoir

frodre:
The current RC model uses subdomain decomposition, feature reduction, and array reshaping to perform a single prediction.  We need a simple interface to use in the stepper prototype  that takes in a dataset-like mapping and transforms that to the expected input data.  

This PR creates an adapter to perform all of the necessary steps to take in a dataset and adds a regression test.  This is a minimal PR that provides a working adapter, but I expect much of the underlying process to be refactored away.

- [x] Tests added


frodre:
resolves #2255

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -98,6 +99,105 @@ def load(cls, path: str) -> "HybridReservoirComputingModel":
...
+class HybridDatasetAdapter:
+    def __init__(self, model: HybridReservoirComputingModel) -> None:
+        self.model = model
+        self._input_feature_sizes: Optional[Sequence] = None
+
+    def predict(self, inputs: xr.Dataset) -> xr.Dataset:
+        # TODO: centralize stacking logic for encoding decoding
+        # TODO: potentially use in train.py instead of special functions there
+        processed_inputs = self._input_data_to_array(inputs)  # x, y, feature dims
+        if self.model.rank_divider is None:
```
AnnaKwa, line 111: This check can be removed now that the rank_divider is a required arg for the models.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -98,6 +99,105 @@ def load(cls, path: str) -> "HybridReservoirComputingModel":
...
+
+    def _separate_output_variables(self, outputs: np.ndarray):
+        if self.model.autoencoder is not None:
+            ds = self._decode_ouput_variables(outputs, self.model.autoencoder)
+        else:
+            ds = self._separate_output_from_stacked_array(outputs)
+
+        return ds
+
+    def _decode_ouput_variables(self, encoded_output: np.ndarray, autoencoder):
```
AnnaKwa, line 179:
```suggestion
    def _decode_output_variables(self, encoded_output: np.ndarray, autoencoder):
```

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -98,6 +99,105 @@ def load(cls, path: str) -> "HybridReservoirComputingModel":
...
+        if encoded_output.ndim > 3:
+            raise ValueError("Unexpected dimension size in decoding operation.")
+
+        feature_size = encoded_output.shape[-1]
+        encoded_output = encoded_output.reshape(-1, feature_size)
+        decoded_arrs = autoencoder.decode(encoded_output)
+
+        return self._separate_output_from_stacked_array(decoded_arrs)
+
+    def _separate_output_from_stacked_array(self, outputs: np.ndarray):
```
AnnaKwa, line 189: I think currently the dense autoencoder transformers behave as desired (return separate arrays for each output) whereas the sklearn transformer does not. Since we decided to fix this in the near future to have all `Transformer` objects have the former behavior, I'm ok with omitting this functionality from this PR (i.e. assume that the output is already in separate arrays, and it's ok if the sklearn transform does not work with this adapter in the meantime) so that it doesn't have to be altered again in the near future when the Transformers are all made consistent.

external/fv3fit/tests/reservoir/test_model_adapter.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @property
+    def n_latent_dims(self):
+        return self._latent_dim_len
+
+    def encode(self, x):
+        return np.concatenate(x, -1)
+
+    def decode(self, latent_x):
+        return latent_x
```
AnnaKwa, line 27: Since the `Transformer`s will be all consistent in returning separate arrays for each output, could just make this test class also do that so that no changes are required after that fix.

AnnaKwa (approved):
Thanks for making the model output more usable! My suggestion to avoid having to change the output separation later on is just to assume the decoded output is in the form of the `DenseAutoencoder` output (separate arrays for each variable), since that's what we're changing to in the next round of PRs.

PR #2257, opened 2023-06-28
Fix hybrid model input concatenation

AnnaKwa:
This PR fixes a bug in the hybrid reservoir model predict method.
Previously,  the subdomain inputs the hidden states were flattened before being concatenated with the flattened hybrid inputs. This meant that the hidden state and hybrid state for each subdomain were not contiguous in the inputs being provided to the readout. These changes fix the bug by concatenating the hidden states and hybrid state along the subdomain axis before flattening.

Other updates:
- made the `rank_divider` a required arg for initializing reservoir models. This simplifies some of the logic and there's no real use case where we don't include this in the reservoir model.
- changes the hybrid inputs to exclude the overlapping cells surrounding each subdomain- this helps decrease memory usage
- Fixes a bug with the hybrid model's `increment_state` method where it was missing the input arg.

 
 

- [x] Tests added

Resolves   https://github.com/ai2cm/fv3net/issues/2259  [JIRA-TAG]

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -70,8 +70,8 @@ class HybridReservoirComputingModel(Predictor):
     def reset_state(self):
         self.reservoir_model.reset_state()
 
-    def increment_state(self):
-        self.reservoir_model.increment_state()
+    def increment_state(self, prediction_with_overlap):
+        self.reservoir_model.increment_state(prediction_with_overlap)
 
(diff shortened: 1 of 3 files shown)


frodre (approved):
LGTM.  A good start on removing things needed for the adapter!

It might have implications for parallelism if we ever need to farm out subdomains vs. using the big readout because each reservoir will need the full rank input.  But I think we could provide a method that generates a model for a specific subdomain based on this, like we discussed yesterday.

PR #2258, opened 2023-06-28
fv3net environment on apple chips

brianhenn:

(Delete unused sections)
Added public API:

Refactored public API:

Significant internal changes:

Requirement changes:

- [ ] Tests added


environment.yml
```diff
@@ -1,9 +1,6 @@
 channels:
   - conda-forge
   - defaults
-platforms:
```
brianhenn, line 4: The version of conda-lock we use was not actually reading this

Makefile
```diff
@@ -238,10 +238,9 @@ update_submodules:
 ############################################################
 
 lock_deps: lock_pip
-	conda-lock -f environment.yml
-	conda-lock render
```
brianhenn, line 242: The render argument and the output it produced (the multi-platform conda-lock.yml file) was not actually used in making our environments.

pip-requirements.txt
```diff
@@ -46,7 +46,7 @@ xpartition>=0.2.1
 xarray >= 0.19.0
 
 # developer tools
-conda-lock
+conda-lock<1.0.0
```
brianhenn, line 49: Requires that we use a version of conda-lock that uses the -p flags to make platform-specific lock files. 

Alternatively, we could switch to using a versions >=1.0, in which we'd need to use a different conda-lock command (render) to implement the approach we've used. See [docs](https://github.com/conda/conda-lock#pre-10-compatible-usage-explicit-per-platform-locks).

PR #2261, opened 2023-06-28
Refactor sklearn transformer to return sequence on decode

frodre:
The `SkTransformer` class took a sequence of arrays into the `encode` method but a returned stacked feature dimension array for the `decode` method-- requiring some extra work to complete the round trip.

This PR fixes that inconsistency.

Refactored public API:
- Added type hints to the `encode` and `decode`
- `decode` returns a sequence of ndarrays

- [x] Tests adjusted


AnnaKwa (approved):
Thanks, LGTM!

Issue #2253, opened 2023-06-27
Remove vertical dim info from RankDivider


AnnaKwa:

AnnaKwa:
Test the increase in training time if the slicing of the time series is done in a for loop by a spatial-dim-only RankDivider, as opposed to the current slicing on array with time dim.

PR #2262, opened 2023-06-28
Use xy dims only in RankDivider

AnnaKwa:
Since the `RankDivider` is only concerned with dividing/reconstituting along horizontal x and y dims, it shouldn't require information about the z dimension size. This PR updates the `RankDivider` so that it only accepts rank extent information along x and y dimensions. In instances where the object is used to reshape arrays into (x, y, z) dims, the z dimension size can be inferred from the input array. An error will be raised upon initialization if z information is provided.

   
Resolves https://github.com/ai2cm/fv3net/issues/2253

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/domain.py
+++ b/external/fv3fit/fv3fit/reservoir/domain.py
@@ -47,10 +47,10 @@ class RankDivider:
             )
         """
         self.subdomain_layout = subdomain_layout
-        if "time" in rank_dims:
+        if "time" in rank_dims or "z" in rank_dims:
             raise ValueError(
-                "Do not include time in dimension information. "
-                "Only spatial dimensions (x, y, z) should be provided."
+                "Do not include time or z in dimension information. "
+                "Only spatial dimensions (x, y, ) should be provided."
             )
         if not {"x", "y"}.issubset(rank_dims):
             raise ValueError(
@@ -79,8 +79,8 @@ class RankDivider:
         )
 
     @property
-    def n_subdomain_features(self):
-        # number of total features (nx * ny * nz) in one subdomain
+    def subdomain_size_with_overlap(self):
+        # number of total features (nx * ny) in one subdomain
         return int(np.prod(self.get_subdomain_extent(with_overlap=True)))
 
     def get_subdomain_extent(self, with_overlap: bool):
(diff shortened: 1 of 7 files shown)


external/fv3fit/tests/reservoir/test_model_adapter.py
```diff
@@ -34,10 +34,15 @@ def decode(self, latent_x):
...
 
 def get_initialized_hybrid_model():
+    # expects rank size (including halos) in latent space
+    divider = RankDivider((2, 2), ["x", "y"], [8, 8], 2)
+    autoencoder = DoNothingAutoencoder(6)
+    input_size = 6 * 6 * autoencoder.n_latent_dims
+    hybrid_input_size_per_subdomain = (
+        divider.subdomain_xy_size_without_overlap ** 2 * autoencoder.n_latent_dims
+    )
+    output_size = hybrid_input_size_per_subdomain
```
AnnaKwa: These are the same values as before, just some edits from while I was trying to debug what broke in the adapter.

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -202,7 +209,7 @@ def load(cls, path):
             metadata = yaml.safe_load(f)
         return cls(**metadata)
 
-    def merge_subdomains(self, flat_prediction: np.ndarray):
+    def merge_subdomains(self, flat_prediction: np.ndarray) -> np.ndarray:
```
frodre, line 212: The initial comment in this method kind of alludes to this, but I would add a small docstring description that this rebuilds flattened readout individual subdomains back into the original x, y, z layout.  `merge_subdomains` makes sense in context of comments and the plural use of `subdomains` but it can be a bit misleading at the higher level since `unstack_subdomain` also exists and the difference is subtle.

frodre: It also seems like `unstack_subdomain` method could be used somewhere here with the similarity in calculating the vertical dimension size, but there is the slight difference of adding the subdomain size and the overlap always being `False`.  Nothing required for this comment.  Just making a public note of it.

frodre (approved):
LGTM, a few minor suggestions

PR #2264, opened 2023-06-29
Fix use of ReloadableTransformer in reservior model module

frodre:
The definition of `ReloadableTransformer` was duplicated in `reservoir.model` to conform with some mypy type constraints.  This PR uses `cast` on the loaded `fv3fit.load` call because the only saved model for the `autoencoder` argument must be of that type by construction.  This removed the need for that small duplication.

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -1,7 +1,7 @@
 import fsspec
 import numpy as np
 import os
-from typing import Optional, Iterable, Hashable, Union, Sequence
+from typing import Optional, Iterable, Hashable, Sequence, cast
 import xarray as xr
 import yaml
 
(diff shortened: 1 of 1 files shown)


external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -308,19 +296,18 @@ def load(cls, path: str) -> "ReservoirComputingModel":
         with fsspec.open(os.path.join(path, cls._METADATA_NAME), "r") as f:
             metadata = yaml.safe_load(f)
 
-        fs: fsspec.AbstractFileSystem = fsspec.get_fs_token_paths(path)[0]
-
-        if fs.exists(os.path.join(path, cls._RANK_DIVIDER_NAME)):
```
frodre, line 313: `RankDivider` is now a required component so removed the conditional here too.

AnnaKwa (approved):
LGTM!

PR #2267, opened 2023-06-30
Create `xtorch_harmonics` package

spencerkclark:
This PR adds a package containing an xarray wrapper of a spherical harmonic transform roundtrip that can be completed with NVIDIA's `torch_harmonics` package.  The roundtrip transform code itself is derived from [this notebook](https://github.com/ai2cm/explore/blob/master/oliwm/2023-04-16-fme-analysis/2023-06-16-fregrid-to-gaussian-example.ipynb) of @oliverwm1.  The xarray wrapper is based on that developed in [this notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-06-22-wrap-spherical-harmonic-transform-roundtrip.ipynb) of mine.  

Some updates from that development notebook include:
- Code to ensure that DataArray and Dataset level attributes are preserved in round trips.
- A split of the `grid` parameter into `forward_grid` and `inverse_grid` components to allow for more flexibility.
- Optional grid validation code to ensure that the coordinates of the grid match the expectation of the `forward_grid` specified.  This can be disabled by setting `unsafe=True`.
- Updating of the latitude coordinate in the case that the `forward_grid` does not equal the `inverse_grid`.

Added public API:
- `xtorch_harmonics.roundtrip`: a function to apply a roundtrip spherical harmonic transform to all horizontally varying variables in a Dataset or to a single DataArray.

Requirement changes:
- `torch-harmonics@git+https://github.com/NVIDIA/torch-harmonics.git@8826246cacf6c37b600cdd63fde210815ba238fd`
- Install `git` and `git-lfs` in the fv3net Docker image to enable installation of `torch_harmonics` from GitHub.
- Pin `pip` to version 20.2.4 ([consistent with what we do in `environment.yml` in fv3net](https://github.com/ai2cm/fv3net/blob/4b0e9f00c9a061f650e84b08d5b63bb7739baeb3/environment.yml#L19)) in the `lint` build in CircleCI to avoid errors like [this](https://app.circleci.com/pipelines/github/ai2cm/fv3net/14735/workflows/142bbcce-efe1-46df-a8cd-c6757a8beb03/jobs/139771).

- [x] Tests added


constraints.txt
```diff
@@ -288,6 +288,7 @@ tf-estimator-nightly==2.8.0.dev2021122109
 threadpoolctl==3.1.0
 toml==0.10.1
 toolz==0.10.0
+git+https://github.com/NVIDIA/torch-harmonics.git@8826246cacf6c37b600cdd63fde210815ba238fd
```
spencerkclark, line 291: `torch_harmonics` uses a BSD-3-Clause license (see [here](https://github.com/NVIDIA/torch-harmonics/blob/main/LICENSE)).

docker/fv3net/Dockerfile
```diff
@@ -22,6 +22,10 @@ RUN chmod +x /usr/local/bin/argo
 # install gcloud sdk
 RUN gcloud config set project vcm-ml
 
+# install git and git-lfs (needed to install torch_harmonics from GitHub)
```
oliverwm1, line 25: do we actually need git-lfs? I know you can clone the repo even without git-lfs installed (the "lfs" parts just don't get pulled, I think) although maybe there is some issue with pip installing from github?

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
```diff
@@ -0,0 +1,208 @@
...
+        raise ValueError(
+            f"Latitude coordinate {obj[lat_dim]} does not match the expected "
+            f"quadrature points for the provided forward grid {forward_grid!r}."
+        )
+
+    try:
+        if lon_dim in obj.coords:
+            # TODO: I suppose for this check we really only need to ensure that
+            # points are equally spaced and global.  Whether they exactly align
+            # with a regular grid from 0 to 360 degrees E is not necessarily
```
oliverwm1, line 70: I agree. E.g. having a longitude grid that goes from -180 to 180 is pretty common, and I would expect to still be able to a forward transform in that case so I'm not sure this check makes sense.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
```diff
@@ -0,0 +1,208 @@
...
+
+def _roundtrip_numpy(
+    array: np.array, forward_grid: T_Grid, inverse_grid: T_Grid
+) -> np.ndarray:
+    *_, n_lat, n_lon = array.shape
+    tensor = torch.tensor(array).type(torch.double)
+    forward_transform = torch_harmonics.RealSHT(n_lat, n_lon, grid=forward_grid)
+    inverse_transform = torch_harmonics.InverseRealSHT(n_lat, n_lon, grid=inverse_grid)
+    roundtripped = inverse_transform(forward_transform(tensor))
+    return np.array(roundtripped.type(torch.float))
```
oliverwm1, line 89: Although torch harmonics requires input to be double precision, not sure we always want to convert back to float. Can you do something like `roundtripped.astype(array.type)` where array is the input numpy array?

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
```diff
@@ -0,0 +1,208 @@
...
+            results.append(result)
+    return xr.merge(results).assign_attrs(ds.attrs)
+
+
+def roundtrip(
+    obj: T_XarrayObject,
+    forward_grid: T_Grid = LEGENDRE_GAUSS_GRID,
+    inverse_grid: T_Grid = LEGENDRE_GAUSS_GRID,
+    lat_dim: Hashable = LAT_DIM,
+    lon_dim: Hashable = LON_DIM,
```
oliverwm1, line 144: nit: minor preference to not have defaults for these args, since this package is very general purpose.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
```diff
@@ -0,0 +1,208 @@
...
+    n_lat = obj.sizes[lat_dim]
+    n_lon = obj.sizes[lon_dim]
+
+    expected_latitudes = compute_quadrature_latitudes(n_lat, forward_grid)
+    expected_longitudes = compute_quadrature_longitudes(n_lon)
+
+    try:
+        if lat_dim in obj.coords:
+            np.testing.assert_allclose(obj[lat_dim].data, expected_latitudes)
+            # TODO: could warn or raise if lat_dim not in obj.coords.
```
oliverwm1, line 59: I think this is fine, but maybe print warning that no validation was done since there is no coordinate to validate?

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
```diff
@@ -0,0 +1,208 @@
...
+    return (bounds[:-1] + bounds[1:]) / 2
+
+
+def _validate_quadrature_points(
+    obj: T_XarrayObject, forward_grid: T_Grid, lat_dim: Hashable, lon_dim: Hashable
+):
+    n_lat = obj.sizes[lat_dim]
+    n_lon = obj.sizes[lon_dim]
+
+    expected_latitudes = compute_quadrature_latitudes(n_lat, forward_grid)
```
oliverwm1, line 53: What if a dataset has latitudes in opposite order? I think it's still well posed to compute SHT, although for sure the sign of (some) coefficients will change.

oliverwm1 (commented):
In general looks great! The coordinate validation is tricky... In general I think it's fine to leave in what you have but it is a bit unclear to me whether a shifted/flipped coordinate should be valid.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
spencerkclark: Exactly, yeah; I modified the check to be more general.  Let me know what you think.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
spencerkclark: Good catch!  Yes, that is important.  I modified things accordingly.  I think it's easier to handle once it is back as a NumPy array.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
spencerkclark: Sounds good.  I raise warnings now when validation cannot be performed.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
spencerkclark: Yeah, that's probably better.  The coarsening functions can be awkward to deal with sometimes because of this.

external/xtorch_harmonics/xtorch_harmonics/xtorch_harmonics.py
spencerkclark: Good point, yes, this is another example where we might want more flexibility in validation.  I updated the check to allow this case too (in general this flexibility at least seems fair in the `roundtrip` case).

spencerkclark (commented):
Thanks for the initial review @oliverwm1.  I tried to address all of your comments; see my responses below.

docker/fv3net/Dockerfile
oliverwm1: I see, thanks for checking.

oliverwm1 (approved):
LGTM, thanks for the changes.

PR #2270, opened 2023-07-06
Add reservoir stepper to fv3gfs runtime

frodre:
To integrate a reservoir models into fv3gfs this PR introduces a new steppers that functions across different `TimeLoop` checkpoints (i.e., the beginning and end of a timestep) with the use of a shared model object and a state machine for consistency.

Added public API:
* `get_reservoir_steppers`: return a tuple of steppers, one for the increment of the underlying reservoir (`ReservoirIncrementOnlyStepper`) and one for prediction (`ReservoirPredictStepper`)
* `ReservoirConfig`: Simple configuration to load a reservoir model with a specified number of steps to synchronize the reservoir and the reservoir timestep duration.

Significant internal changes:
* `loop.py` now includes the reservoir steppers in the loop.  The increment takes place before any state updates such that it can grab the input state to increment the reservoir to t+1.  The prediction step takes place after all other updates such that we see the t+1 of the final model outputs (including any ML corrections) for the readout layer.  The reservoir increment will take place on the subsequent timestep.

- [x] Tests added


workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -0,0 +1,233 @@
...
+    def to_predicted(self):
+        # predict only allowed after increment has been called
+        if self._last_called != self.INCREMENT:
+            raise ValueError("Must call increment before next prediction")
+        self._last_called = self.PREDICT
+
+    def set_init_time(self, time: cftime.DatetimeJulian):
+        self._init_time = time
+
+    def __call__(self, state: str):
```
AnnaKwa, line 64: Just curious- why this pattern (calling with hard coded options) rather than directly calling `to_incremented()` etc.?

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @property
+    def init_time(self):
+        return self._state_machine.init_time
+
+    @property
+    def completed_sync_steps(self):
+        return self._state_machine.completed_increments
+
+    def increment_reservoir(self, time, state):
```
AnnaKwa, line 105: Is there a reason why this method and `predict` are on the base class rather than their respective child classes?

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -0,0 +1,233 @@
...
+
+        updated_state = {}
+        diags = {}
+
+        # won't evaluate to true until we've reached the step before the next increment
+        # e.g., if fv3 has k timesteps between rc timestep, on t + k - 1, the timestep
+        # at the end will have ticked over to t + k, we'll maybe use the integrated
+        # hybrid quantites from t -> t + k, make the rc prediction for t + k, and then
+        # increment during the next time loop based on those outputs.
+        if self._is_rc_update_step(time):
```
AnnaKwa, line 144: Is it possible to move this check up a level to the increment/predict stepper classes' `__call__`? That way, `increment_state` and `predict` wouldn't need the time arg and their signatures would be more similar to the lower level methods of the same name.

AnnaKwa (changes requested):
Thanks Andre! I had a few minor requests but the structure is really clear. The separation of the increment and predict steppers is nice.

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -0,0 +1,233 @@
...
+        self._last_called = self.INCREMENT
+        self._num_increments_completed += 1
+
+    def to_predicted(self):
+        # predict only allowed after increment has been called
+        if self._last_called != self.INCREMENT:
+            raise ValueError("Must call increment before next prediction")
+        self._last_called = self.PREDICT
+
+    def set_init_time(self, time: cftime.DatetimeJulian):
```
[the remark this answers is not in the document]
frodre: Good call.  Done.

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
frodre: This is to align with the defined `Stepper` protocol  defined in `runtime.steppers.stepper`

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
frodre: Mainly for ease of testing, but I will co-locate the code with the point of use!

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -0,0 +1,233 @@
...
+import dataclasses
+from datetime import timedelta
+import pandas as pd
+from typing import Optional, MutableMapping, Hashable, Sequence, Union, cast
+import xarray as xr
+
+import fv3fit
+from fv3fit._shared import put_dir
+from fv3fit._shared.halos import append_halos_using_mpi
+from fv3fit.reservoir.model import HybridReservoirComputingModel
```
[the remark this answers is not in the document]
frodre: I originally left both in as a union, since either works, but I'll move it to just a single expectation for clarity with the TimeLoop usage.

AnnaKwa (approved):
LGTM!

PR #2274, opened 2023-07-11
Move encode/decode into base models

AnnaKwa:
Refactored public API:
-  This PR moves array reshaping and encoding/decoding of inputs and outputs into the base reservoir models. Previously these were done by the hybrid adapter. Having the base models handle these operations makes offline predictions easier and simplifies the public API for the base models. Input into the hybrid model `predict` method is assumed to be in the native (x,y,z) format instead of a flattened array. Now, `predict` returns np arrays for each variable in the native (x, y, z) coords.
- Made the `autoencoder` a required arg instead of optional to the reservoir models. All models trained using the training code have an autoencoder by default.
  
Significant internal changes:
- Refactored above functionality out of the hybrid adapter, now it handles converting xr datasets to arrays but otherwise relies on its base model to handle reshaping and encoding/decoding.
- Moved the `DoNothingAutoencoder` to `reservoir.transformers.transformer` since it is a generally useful class for tests and even training (if you do not want to standardize inputs). Added dump/load methods to it.

Updated tests to expect the new 
- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/_reshaping.py
+++ b/external/fv3fit/fv3fit/reservoir/_reshaping.py
@@ -1,4 +1,5 @@
 import numpy as np
+import tensorflow as tf
 
 
 def flatten_2d_keeping_columns_contiguous(arr: np.ndarray):
@@ -34,3 +35,9 @@ def split_1d_samples_into_2d_rows(
         return np.reshape(arr, (time_dim_size, n_rows, -1), order="C")
     else:
         return np.reshape(arr, (n_rows, -1), order="C")
+
+
+def stack_array_preserving_last_dim(data):
+    original_z_dim = data.shape[-1]
+    reshaped = tf.reshape(data, shape=(-1, original_z_dim))
+    return reshaped
--- a/external/fv3fit/fv3fit/reservoir/domain.py
+++ b/external/fv3fit/fv3fit/reservoir/domain.py
@@ -2,7 +2,6 @@ import fsspec
 import numpy as np
 import tensorflow as tf
 from typing import Sequence, Iterable
-import xarray as xr
 import yaml
 from ._reshaping import stack_data, split_1d_samples_into_2d_rows
 import pace.util
@@ -15,22 +14,6 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
     sl = [slice(None)] * arr.ndim
     sl[axis] = inds
     return arr[tuple(sl)]
-
-
-def _transpose_xy_dims(ds: xr.Dataset, rank_dims: Sequence[str]):
-    # Useful for transposing the x, y dims in a dataset to match those in
-    # RankDivider.rank_dims, and leaves other dims in the same order
-    # relative to x,y. Dims after the first occurence of one of the rank_dims
-    # are assumed to be feature dims.
-    # e.g. (time, y, x, z) -> (time, x, y, z) for rank_dims=(x, y)
-    leading_non_xy_dims = []
-    for dim in ds.dims:
-        if dim not in rank_dims:
-            leading_non_xy_dims.append(dim)
-        if dim in rank_dims:
-            break
-    ordered_dims = (*leading_non_xy_dims, *rank_dims)
-    return ds.transpose(*ordered_dims, ...)
 
 
 class RankDivider:
--- a/external/fv3fit/fv3fit/reservoir/domain.py
+++ b/external/fv3fit/fv3fit/reservoir/domain.py
@@ -2,6 +2,7 @@ import fsspec
 import numpy as np
 import tensorflow as tf
 from typing import Sequence, Iterable
+import xarray as xr
 import yaml
 from ._reshaping import stack_data, split_1d_samples_into_2d_rows
 import pace.util
@@ -14,6 +15,22 @@ def slice_along_axis(arr: np.ndarray, inds: slice, axis: int = 0):
     sl = [slice(None)] * arr.ndim
     sl[axis] = inds
     return arr[tuple(sl)]
+
+
+def _transpose_xy_dims(ds: xr.Dataset, rank_dims: Sequence[str]):
+    # Useful for transposing the x, y dims in a dataset to match those in
+    # RankDivider.rank_dims, and leaves other dims in the same order
+    # relative to x,y. Dims after the first occurence of one of the rank_dims
+    # are assumed to be feature dims.
+    # e.g. (time, y, x, z) -> (time, x, y, z) for rank_dims=(x, y)
+    leading_non_xy_dims = []
+    for dim in ds.dims:
+        if dim not in rank_dims:
+            leading_non_xy_dims.append(dim)
+        if dim in rank_dims:
+            break
+    ordered_dims = (*leading_non_xy_dims, *rank_dims)
+    return ds.transpose(*ordered_dims, ...)
 
 
 class RankDivider:
--- a/external/fv3fit/fv3fit/reservoir/domain.py
+++ b/external/fv3fit/fv3fit/reservoir/domain.py
@@ -65,16 +65,14 @@ class RankDivider:
         self._partitioner = pace.util.TilePartitioner(subdomain_layout)
 
         # dimensions of rank data without the halo points. Useful for slice calculation.
-        self._rank_extent_without_overlap = self._get_rank_extent_without_overlap(
+        self.rank_extent_without_overlap = self._get_rank_extent_without_overlap(
             rank_extent, overlap
         )
 
     @property
     def subdomain_xy_size_without_overlap(self):
         # length of one side of subdomain along x/y axes
-        return (
-            self._rank_extent_without_overlap[self._x_ind] // self.subdomain_layout[0]
-        )
+        return self.rank_extent_without_overlap[self._x_ind] // self.subdomain_layout[0]
 
     @property
     def subdomain_size_with_overlap(self):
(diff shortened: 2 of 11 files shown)


external/fv3fit/fv3fit/reservoir/_reshaping.py
```diff
@@ -34,3 +35,9 @@ def split_1d_samples_into_2d_rows(
         return np.reshape(arr, (time_dim_size, n_rows, -1), order="C")
     else:
         return np.reshape(arr, (n_rows, -1), order="C")
+
+
+def stack_array_preserving_last_dim(data):
```
AnnaKwa, line 40: Relocated from `train.py`

external/fv3fit/fv3fit/reservoir/domain.py
```diff
@@ -65,16 +65,14 @@ def __init__(
         self._partitioner = pace.util.TilePartitioner(subdomain_layout)
 
         # dimensions of rank data without the halo points. Useful for slice calculation.
-        self._rank_extent_without_overlap = self._get_rank_extent_without_overlap(
+        self.rank_extent_without_overlap = self._get_rank_extent_without_overlap(
```
AnnaKwa, line 68: Made this a public attribute since it is useful info for other functions to grab

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -28,8 +44,8 @@ def __init__(
         reservoir: Reservoir,
         readout: ReservoirComputingReadout,
         rank_divider: RankDivider,
+        autoencoder: ReloadableTransfomer,
```
AnnaKwa, line 47: All trained models have an autoencoder saved with them, and the `predict` method now assumes that there is an autoencoder, so I made this a required arg.

external/fv3fit/fv3fit/reservoir/train.py
```diff
@@ -156,14 +137,22 @@ def train_reservoir_model(
     readout = combine_readouts(subdomain_readouts)
 
     model: Union[ReservoirComputingModel, HybridReservoirComputingModel]
+
+    # After training, the data used in inference does not have a time dimension
+    rank_divider_spatial_only = RankDivider(
```
AnnaKwa, line 142: The `TimeSeriesRankDivider` is only useful for manipulating training data with a time dim, for prediction using a saved model we only work with spatial dimensions.

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
```diff
@@ -20,3 +27,85 @@ def encode(self, x: Sequence[ArrayLike]) -> np.ndarray:
...
+        return cls(
+            latent_dim_len=config["latent_dim_len"],
+            original_feature_sizes=config["original_feature_sizes"],
+        )
+
+
+def decode_columns(
+    encoded_output: np.ndarray, transformer: Transformer, xy_shape: Sequence[int]
+) -> Sequence[np.ndarray]:
+    # Differs from encode_columns as the decoder expects a single input array
```
frodre, line 84: Please use docstrings for encode/decode instead of a comment, so this description will show up in an IDE.

external/fv3fit/tests/reservoir/test_transformer.py
```diff
@@ -0,0 +1,60 @@
...
+        (*expected_shapes[0][:-1], nz * nvars) if nz is not None else expected_shapes[0]
+    )
+    transformer = DoNothingAutoencoder(nz * nvars)
+
+    # need to call encode before decode
+    data_arrs = [np.random.rand(*shape) for shape in expected_shapes]
+    transformer.encode(data_arrs)
+
+    encoded_input = np.random.rand(*encoded_input_shape)
+    # pdb.set_trace()
```
frodre, line 55: Remove pdb comment

external/fv3fit/tests/reservoir/test_transformer.py
```diff
@@ -0,0 +1,60 @@
...
+    transformer = DoNothingAutoencoder(nz)
+    data_arrs = [np.random.rand(*shape) for var in range(nvars)]
+
+    encoded = encode_columns(data_arrs, transformer=transformer)
+
+    assert encoded.shape == expected_shape
+
+
+@pytest.mark.parametrize(
+    "nx, ny, nz, nvars", [(4, 4, 3, 2), (2, 2, 1, 1), (2, 2, 1, 2)]
```
frodre, line 38: What's the purpose of the last test vs. the first test parameter set?

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
```diff
@@ -20,3 +27,85 @@ def encode(self, x: Sequence[ArrayLike]) -> np.ndarray:
...
+    def load(cls, path: str) -> "DoNothingAutoencoder":
+        with fsspec.open(os.path.join(path, cls._CONFIG_NAME), "r") as f:
+            config = yaml.safe_load(f)
+        return cls(
+            latent_dim_len=config["latent_dim_len"],
+            original_feature_sizes=config["original_feature_sizes"],
+        )
+
+
+def decode_columns(
```
frodre, line 81: I think this works since the encode/decode of the two transformers share the same call signature, but what was the intent for providing a wrapper function vs. letting the transformers perform the ops to meet the expectations?

frodre (changes requested):
Thanks, @AnnaKwa 

Just one requested change to move the increment encode/decode requirement out of adapter into the models as well.   Everything else was pretty minor!

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -106,94 +131,57 @@ def __init__(self, model: HybridReservoirComputingModel) -> None:
...
-        return self._output_array_to_ds(unstacked_arr, dims=list(inputs.dims))
+        xy_input_arrs = self._input_dataset_to_arrays(inputs)  # x, y, feature dims
+
+        prediction_arr = self.model.predict(xy_input_arrs)
+        return self._output_array_to_ds(prediction_arr, dims=list(inputs.dims))
 
     def increment_state(self, inputs: xr.Dataset):
-        processed_inputs = self._input_data_to_array(inputs)
+        xy_input_arrs = self._input_dataset_to_arrays(inputs)  # x, y, feature dims
+        encoded_xy_input_arrs = encode_columns(xy_input_arrs, self.model.autoencoder)
```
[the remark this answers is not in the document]
AnnaKwa: No, I tunnel visioned on the `predict` method but it makes sense to do it for the state increment as well.

external/fv3fit/fv3fit/reservoir/train.py
```diff
@@ -193,7 +182,7 @@ def _process_batch_Xy_data(
 
     # Concatenate features, normalize and optionally convert data
     # to latent representation
-    batch_data_encoded = _encode_columns(batch_X, autoencoder)
+    batch_data_encoded = encode_columns(batch_X, autoencoder)
```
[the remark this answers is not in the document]
AnnaKwa: Correct

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
```diff
@@ -20,3 +27,85 @@ def encode(self, x: Sequence[ArrayLike]) -> np.ndarray:
...
+@io.register("do-nothing-transformer")
+class DoNothingAutoencoder(Transformer, Reloadable):
+    _CONFIG_NAME = "mock_transformer.yaml"
+
+    """Useful class for tests. Encode just concatenates input
+    variables. Decode separates them back into individual arrays.
+    """
+
+    def __init__(
+        self, latent_dim_len, original_feature_sizes: Optional[Sequence[int]] = None
```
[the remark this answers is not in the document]
AnnaKwa: I think it's still needed for `Transformer` objects since the training code uses the `n_latent_dims` property to determine the size of the reservoir matrices.

external/fv3fit/tests/reservoir/test_transformer.py
AnnaKwa: I wanted to test the case of nz=1 with both nvars=1 and >1.

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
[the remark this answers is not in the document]
frodre: Sorry, I was thinking that instead of a separate `latent_dim_len` attribute, it would be switched to the sum of feature sizes.  You're right about the need for the property to adhere to the base class.

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
AnnaKwa: Since `RankDivider.merge_subdomains` already exists and operates on the raw fully flattened prediction to reshape into (x, y, latent_feature) dims, I found it easiest to have a function with a similar input data format as `encode_columns` that stacks this output with using the latent_feature dim as its unstacked dim. The alternative of reshaping the raw flat prediction array to [sample, latent_feature] dims and then just calling `Transformer.decode` on that would also work, but would require reflattening the decoded data before passing to `merge_subdomains` to get the final (x, y, z_feature) dims.

frodre (approved):
Thanks for the changes.  Just one small change request to remove the extra attribute for the `DoNothingAutoencoder` since the latent dim property can be calculated from the feature sizes.  Otherwise, LGTM!

Issue #2275, opened 2023-07-11
Simplify RankDivider usage by only dealing with spatial dims


AnnaKwa:
The existence of both `TimeSeriesRankDivider` and `RankDivider` can be confusing and already led to some additional debugging during refactors. 

Test whether it's a significant slowdown in training to just use the regular `RankDivider` iterating over the time dimension, rather than the 'slice all dims at once' approach that the `TimeSeriesRankDivider` uses. If it's not too bad, it would streamline the code to just have a single `RankDivider` class that only slices on spatial dims.

Update- I tested this and it's actually slightly faster to iterate over timesteps with just the spatial slicing `RankDivider` as opposed to slicing time and xy dims with the `TimeSeriesRankDivider`:
```
TimeSeriesRankDivider processing time: 0.11544919013977051 s
RankDivider processing time: 0.10854530334472656 s
2023-07-11 23:11:05 [INFO] train.py::L131 : Synchronizing on batch 2
TimeSeriesRankDivider processing time: 0.11331391334533691 s
RankDivider processing time: 0.10267090797424316 s
2023-07-11 23:11:06 [INFO] train.py::L157 : Fitting on batch 3
TimeSeriesRankDivider processing time: 0.18235254287719727 s
RankDivider processing time: 0.12042832374572754 s
2023-07-11 23:11:06 [INFO] train.py::L157 : Fitting on batch 4
TimeSeriesRankDivider processing time: 0.1820662021636963 s
RankDivider processing time: 0.10405874252319336 s
2023-07-11 23:11:07 [INFO] train.py::L157 : Fitting on batch 5
TimeSeriesRankDivider processing time: 0.1683943271636963 s
RankDivider processing time: 0.11531591415405273 s
```

PR #2276, opened 2023-07-12
Simplify rankdividers

AnnaKwa:
In investigating [this issue](https://github.com/orgs/ai2cm/projects/4/views/1?pane=issue&itemId=31907382) I found that using the spatial-only `RankDivider` class to slice data into subdomains within a loop over timesteps was actually faster than using the `TimeSeriesRankDivider` class to do the slicing directly on the full array with a time dim (not sure why- maybe the slicing implementation is weird). See issue for timing details.

Therefore I decided to just remove the `TimeSeriesRankDivider` in order to simplify the codebase. This also allows for some helper functions to be deleted or simplifed since we don't need to provide the option of operating on an array with a time dimension- it is now assumed that the input arrays will have dims (x, y, z_feature/encoded_feature).

I updated the function `process_batch_Xy_data` to handle the data processing inside a loop over timesteps, and moved this function into `utils.py` so the tests were more organized.

There's a decent number of files with changes but most are just related to the deleted or simplified helpers, plus moving `process_batch_Xy_data`  to the utils module. The main change to review is within `process_batch_Xy_data`  where the loop over timesteps was added. This is hard to see in the file diff, since that function was moved to another module, so here is a diff of what changed in that function: https://www.diffchecker.com/FKtr1Z8o/

- [x] Tests added

Resolves https://github.com/ai2cm/fv3net/issues/2275

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/_reshaping.py
+++ b/external/fv3fit/fv3fit/reservoir/_reshaping.py
@@ -14,27 +14,11 @@ def concat_inputs_along_subdomain_features(a, b):
     return np.concatenate([a, b], axis=1)
 
 
-def stack_data(tensor, keep_first_dim: bool):
-    # Used to reshape a subdomains into a flat columns.
-    # Option to keep first dim, used in the case where time is the first dimension
-    if keep_first_dim is True:
-        n_samples = tensor.shape[0]
-        return np.reshape(tensor, (n_samples, -1))
-    else:
-        return np.reshape(tensor, -1)
-
-
-def split_1d_samples_into_2d_rows(
-    arr: np.ndarray, n_rows: int, keep_first_dim_shape: bool
-) -> np.ndarray:
+def split_1d_samples_into_2d_rows(arr: np.ndarray, n_rows: int) -> np.ndarray:
     # Consecutive chunks of 1d array form rows of 2d array
     # ex. 1d to 2d reshaping (8,) -> (2,4) for n_rows=2
     # [1,2,3,4,5,6,7,8] -> [[1,2,3,4], [5,6,7,8]]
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/reservoir/_reshaping.py
```diff
@@ -14,27 +14,11 @@ def concat_inputs_along_subdomain_features(a, b):
...
-        n_samples = tensor.shape[0]
-        return np.reshape(tensor, (n_samples, -1))
-    else:
-        return np.reshape(tensor, -1)
-
-
-def split_1d_samples_into_2d_rows(
-    arr: np.ndarray, n_rows: int, keep_first_dim_shape: bool
-) -> np.ndarray:
+def split_1d_samples_into_2d_rows(arr: np.ndarray, n_rows: int) -> np.ndarray:
```
AnnaKwa, line 17: Since there's no more use case for these helper functions that involves a time dim, they can be simplified or deleted.

external/fv3fit/fv3fit/reservoir/utils.py
```diff
@@ -12,3 +16,54 @@ def _square_evens(v: np.ndarray) -> np.ndarray:
 
 def square_even_terms(v: np.ndarray, axis=1) -> np.ndarray:
     return np.apply_along_axis(func1d=_square_evens, axis=axis, arr=v)
+
+
+def get_ordered_X(X_mapping, variables):
```
frodre, line 21: If this is more than a helper function, I would use variable name `X' with type hints here instead of `X_mapping`.

frodre (approved):
Nice work checking out the efficiency of this! Leads to a lot of nice simplification. LGTM

PR #2277, opened 2023-07-12
Add ERA5 dataflow regridding workflow

frodre:
This PR adds a sample workflow to resample and regrid hourly ERA5 data hosted on GCS
using `xarray-beam` and Google's Dataflow. This is an alternative to downloading data from the Copernicus web API, which takes about 20+ hrs per variable.  This does not perform the final step of regular lat/lon grid -> cubed sphere grid.

- [x] Tests added


projects/reservoir/era5/daily_avg.py
```diff
@@ -0,0 +1,115 @@
...
+        data.time.isel(time=-1).values,
+        np.timedelta64(1, "D"),
+    )
+    num_times = len(time_template)
+    template = xr.concat([single_day] * num_times, "time").assign_coords(
+        {"time": time_template}
+    )
+    template = preserve_attrs(template, for_pred)
+    template = xbeam.make_template(template)
+    print(template)
```
AnnaKwa, line 91: Leftover print statement

projects/reservoir/era5/daily_avg.py
```diff
@@ -0,0 +1,115 @@
...
+
+    beam_options = PipelineOptions(beam_args)
+    with beam.Pipeline(options=beam_options) as p:
+        (
+            p
+            | xbeam.DatasetToChunks(for_pred, chunks={"time": 24})
+            | beam.MapTuple(
+                lambda k, v: (
+                    k.with_offsets(time=k.offsets["time"] // 24),
+                    v.resample(time="1D").mean().drop(drop_vars).compute(),
```
AnnaKwa, line 101: Can `drop_vars` be dropped earlier on before the beginning of the pipeline (when `for_pred` is created)?

projects/reservoir/era5/daily_avg.py
```diff
@@ -0,0 +1,115 @@
...
+from apache_beam.options.pipeline_options import PipelineOptions
+import numpy as np
+import xarray as xr
+import xarray_beam as xbeam
+import argparse
+
+"""
+This script is tailored to provide daily averages from ERA5 data for a
+specific subset of variables used for creating an ocean RC model. The
+chunking and machines used are tailored to the specific use case so adjustments
```
AnnaKwa, line 11: I think it's fine to hard code stuff in this script since our use case is very specific, but could you add a few more specifics about the hard coded assumptions here at the top? e.g. regrids to 1 deg x 1 deg, variables saved, etc.

AnnaKwa (approved):
Been a while since I've dealt with dataflow- thanks for adding this to speed up the data processing! The xarray-beam package is neat, it would've been nice to have back in the early days of our dataflow pipelines. Just a few minor comments.

PR #2278, opened 2023-07-12
create conda env file

gideonite:
This is the first step towards making this code reproducible for the future. The environment file contains a lot of cruft, e.g. higher order deps that are probably brittle.


projects/super_res/environment.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - kornia
+  - accelerate
+  - xarray
+  - seaborn
+  - zarr
+  - google-cloud-sdk
+  - fsspec
+  - gcsfs
+  - joblib
+  - tensorboard
```
cynicalsavant, line 28: I think we can remove tensorboard since I'm using wandb

projects/super_res/access_gcp.py
cynicalsavant, line 1: What seems to be the purpose of this file?

projects/super_res/environment.yaml
cynicalsavant: Also, could you run python trainer.py without any issues?

projects/super_res/access_gcp.py
gideonite: My fault, this was part of my experiment which is no longer needed. I just removed it. Apologies for the confusion.

projects/super_res/environment.yaml
gideonite: Right, good catch. Meanwhile, `wandb` isn't there.

Issue #2273, opened 2023-07-11
Standardize array inputs to always have a z dimension, even if it is size 1


AnnaKwa:
Various places in the reservoir code assume the last dim of input arrays is the feature dimension, which is not the case when the input is a 2D field. The training code calls the `assure_same_dims` function which adds a z dim of size 1 if other 3D fields are present, but this doesn't handle the edge case where only 2D inputs are present (this would probably be the format of data for SST models). At inference, the adapter assumes the last dim of each variable is the feature dim. For both of these cases we should add some code early on in the array creation to reshape to add a feature dim of size 1 for 2D fields.

PR #2279, opened 2023-07-12
Ensure all input arrays to reservoir have a feature dimension

AnnaKwa:
The changes in this PR ensure that the arrays in the reservoir training and reservoir model prediction always have a feature dimension, even if the variables are 2D (feature dim is size 1 in that case).

See [this issue](https://github.com/ai2cm/fv3net/issues/2273) for more explanation of why these changes are needed.

Resolves https://github.com/ai2cm/fv3net/issues/2273
  


frodre (approved):
LGTM!

PR #2280, opened 2023-07-13
fv3net: shorten e2e test by using saved nudged run data

brianhenn:
The e2e test fails frequently by exceeding its 35 minute limit; this PR speeds up the version of it run on master by ~30% by removing the nudge-to-fine prognostic run and instead using saved GCS nudged run data. The full version will now run weekly on master automatically with a longer time limit, and can also still be run by invoking it manually on a branch.

Also fixes a bug where we were not using a current `fv3fit` image in the e2e test.


tests/end_to_end_integration/run_test.sh
```diff
@@ -2,14 +2,14 @@
...
-if [[ $# != 2 ]]
-then
-    echo "usage: tests/end_to_end_integration_argo/run_test.sh <registry> <version>"
-    exit 1
-fi
-
 SLEEP_TIME=15
 
+function printUsage {
+    echo "usage: tests/end_to_end_integration_argo/run_test.sh [-r] <registry> <version>"
```
elynnwu, line 8: [nit] I think `-r` is fine, but it would be easier to follow if you use `--runNudgeToFine=True`, you can have it default to either True or False. Then you don't have to manually change `runNudgeToFine` later in this script on line 100.

elynnwu: I think it makes circleci config a bit easier to read too:
```
    parameters:
      runNudgeToFine:
        default: false
        type: boolean
```

elynnwu (approved):

PR #2282, opened 2023-07-13
Minor cleanup to diff code

gideonite:
- rename initial script to something more meaningful
- add doc string to script file
- wrap main method in a function and use the __main__ call.


cynicalsavant (approved):
Looks good!

PR #2283, opened 2023-07-14
make results dir if it doesn't exist

gideonite:
- Make results dir if it doesn't exist.
- Unrelated: update requirements file with the latest dep.


cynicalsavant (approved):
looks good

PR #2285, opened 2023-07-17
log scaling and logging

cynicalsavant:
just added some log scaling adn additional logging before running the new experiment

### DIFF
--- a/projects/super_res/config.py
+++ b/projects/super_res/config.py
@@ -21,7 +21,7 @@ results_folder = "./results"
 tensorboard_dir = "./tensorboard"
 milestone = 1
 
-batch_size = 2
+batch_size = 4
 data_config = {
     "dataset_name": "c384",
     "length": 7,
@@ -29,7 +29,8 @@ data_config = {
     "channels": ["PRATEsfc_coarse"],
     #"img_channel": 2,
     "img_channel": 1,
-    "img_size": 384
+    "img_size": 384,
+    "logscale": True
 }
 
 data_name = f"{data_config['dataset_name']}-{data_config['channels']}-{objective}-{loss}-d{dim}-t{diffusion_steps}{additional_note}"
--- a/projects/super_res/data/vsrdata.py
+++ b/projects/super_res/data/vsrdata.py
@@ -4,7 +4,7 @@ from torch.utils.data import Dataset
 
 class VSRDataset(Dataset):
     
-    def __init__(self, channels, mode, length):
+    def __init__(self, channels, mode, length, logscale = False):
         '''
         Args:
             channels (list): list of channels to use
@@ -21,6 +21,10 @@ class VSRDataset(Dataset):
         # shape : (tile, time, channel, y, x)
         c384_np = np.stack([c384[channel].values for channel in channels], axis = 2)
         c48_np = np.stack([c48[channel].values for channel in channels], axis = 2)
+
+        if logscale:
+            c384_np = np.log(c384_np - c384_np.min() + 1e-14)
+            c48_np = np.log(c48_np - c48_np.min() + 1e-14)
 
         # calculate split (80/20)
         split = int(c384_np.shape[1] * 0.8)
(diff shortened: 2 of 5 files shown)


projects/super_res/config.py
```diff
@@ -21,15 +21,16 @@
 tensorboard_dir = "./tensorboard"
 milestone = 1
 
-batch_size = 2
+batch_size = 4
```
gideonite, line 24: nit: Please revert this change. I don't think this needs to be committed.

projects/super_res/data/vsrdata.py
```diff
@@ -22,6 +22,10 @@ def __init__(self, channels, mode, length):
         c384_np = np.stack([c384[channel].values for channel in channels], axis = 2)
         c48_np = np.stack([c48[channel].values for channel in channels], axis = 2)
 
+        if logscale:
+            c384_np = np.log(c384_np - c384_np.min() + 1e-14)
```
gideonite, line 26: Do you know what floating point precision we are using? It's probably better to make this more robust via finfo, e.g. if we are using `torch.float32`

```
 torch.finfo(torch.float32).resolution
```
This returns 1e-06.

projects/super_res/model/autoreg_diffusion.py
```diff
@@ -1312,6 +1311,7 @@ def sample(self):
         self.ema.ema_model.eval()
 
         cmap = mpl.colormaps['viridis']
+        sm = smap(None, cmap)
```
gideonite, line 1314: I'm a little confused here. What if `config.data_config.logscale = False`? Does this mean that we don't want to plot in logscale anymore?

projects/super_res/model/autoreg_diffusion.py
```diff
@@ -1329,9 +1329,6 @@ def sample(self):
                     #videos, base, nsteps, flows = self.ema.ema_model.sample(lres, hres, True)
                     videos, base, nsteps, flows = self.ema.ema_model.sample(lres, hres)
 
-                    vmax = torch.max(torch.max(videos.max(), hres.max()), lres.max()).cpu().numpy()
```
gideonite, line 1332: nit: I know you are removing this code, but I think you can use python's native `max` function here, e.g. 
```
max([videos.max(), hres.max(), lres.max()])
```
which I think would help clarify this line.

gideonite (changes requested):
This is a great and simple idea which seems to have made a big difference. 

A few "nits" which are just small comments that I'm making which I think would help improve this PR, but which you can ignore if you choose. 

I am little confused about the logging and setting the jiggle parameter (`eps`) to match the floating point precision that we are using in the code. Once these two things are addressed, I think we'll be good to go.

projects/super_res/model/autoreg_diffusion.py
cynicalsavant: that's just the default value in the function. it takes the value from config file where it's set to be true

projects/super_res/model/autoreg_diffusion.py
cynicalsavant: aah okay

projects/super_res/data/vsrdata.py
gideonite: Ok got it. Then you can do `np.finfo(arr.dtype).resolution`. What do you think?

projects/super_res/model/autoreg_diffusion.py
gideonite: Sorry, I still don't get it. What happens if the `config.logscale` is False? I think the data will not be plotted in logscale but we always want to plot logscale, right?

cynicalsavant: not just for plotting but as an input to the model itself

gideonite:
We are going to forgo having plots with backward compatibility, assuming that moving forward we are going to log scaling the relevant features. 

I'm going to approve now just to unblock you, but would appreciate the change regarding the jiggle parameter `eps` to the log transform of the data.

gideonite:
There were some conflicts with my PR for changing the config. In sympathy, I resolved them since I know that this can be super frustrating. Please double check that I did a proper job. @cynicalsavant

gideonite (approved):

PR #2286, opened 2023-07-17
Diff code/wandb improvements

gideonite:
Small improvements to wandb setup.
- define wandb run name and description
- set the step in the validation videos so that they line up properly

Misc:
- fix requirements.txt for missing deps (discovered at validation time)

### DIFF
--- a/projects/super_res/model/autoreg_diffusion.py
+++ b/projects/super_res/model/autoreg_diffusion.py
@@ -1296,13 +1296,13 @@ class Trainer(object):
                                 videos, base, res, flows = self.ema.ema_model.sample(lres, hres)
                                 psnr_index = piq.psnr(hres[:,2:,0:1,:,:], videos.clamp(0.0, 1.0)[:,:,0:1,:,:], data_range=1., reduction='none')
                                 
-                                accelerator.log({"true_high": wandb.Video((hres[:,2:,0:1,:,:].repeat(1,1,3,1,1).cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"true_low": wandb.Video((lres[:,2:,0:1,:,:].repeat(1,1,3,1,1).cpu().numpy()*255).astype(np.uint8))})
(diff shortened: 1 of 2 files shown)


gideonite:
You can see an example of how this works [here](https://wandb.ai/ai2cm/vsr-orig-autoreg-hres/runs/yns8u0ea?workspace=user-gideond). The results are garbage but you should see that the video indexes line up and there's a name for this run which is not automatically generated.

projects/super_res/model/autoreg_diffusion.py
```diff
@@ -1288,13 +1296,13 @@ def train(self):
...
-                                accelerator.log({"true_low": wandb.Video((lres[:,2:,0:1,:,:].repeat(1,1,3,1,1).cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"pred": wandb.Video((base.clamp(0.0, 1.0)[:,:,0:1,:,:].repeat(1,1,3,1,1).detach().cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"samples": wandb.Video((videos.clamp(0.0, 1.0)[:,:,0:1,:,:].repeat(1,1,3,1,1).detach().cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"res": wandb.Video((res.clamp(0.0, 1.0)[:,:,0:1,:,:].repeat(1,1,3,1,1).detach().cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"flows": wandb.Video((flows.clamp(0.0, 1.0).detach().cpu().numpy()*255).astype(np.uint8))})
-                                accelerator.log({"psnr": psnr_index.mean()})
+                                accelerator.log({"true_high": wandb.Video((hres[:,2:,0:1,:,:].repeat(1,1,3,1,1).cpu().numpy()*255).astype(np.uint8))}, step=self.step)
```
cynicalsavant, line 1299: Interesting that setting explicit step worked because it was failing for me. I notice however that thee's a log at each training step. That's because you changed the config file right? to save and sample after every step?

gideonite: >That's because you changed the config file right? to save and sample after every step?

Right, I did this just to accelerate the feedback for development. I didn't want to wait many steps to see the outputs. But that change should not be in this PR.

PR #2288, opened 2023-07-18
Add offline reservoir prediction script

AnnaKwa:
This adds a short workflow to get reservoir models' offline (single step) predictions on a validation or test set. This PR just adds the code to produce the prediction, a following PR will add the metrics saving.

 
Added public API:
- fv3net.diagnostics.reservoir.compute


workflows/diagnostics/fv3net/diagnostics/reservoir/compute.py
```diff
@@ -0,0 +1,184 @@
...
+    n_timesteps = states_with_overlap_time_series[0].shape[0]
+    for t in range(n_timesteps):
+        state = [
+            variable_time_series[t]
+            for variable_time_series in states_with_overlap_time_series
+        ]
+        model.increment_state(state)
+        predict_kwargs = {}
+        if hybrid_inputs_time_series is not None:
+            predict_kwargs["hybrid_input"] = [
```
frodre, line 84: Should this be t+1 for the hybrid variables?

workflows/diagnostics/fv3net/diagnostics/reservoir/compute.py
```diff
@@ -0,0 +1,184 @@
...
+    one_step_prediction_time_series = []
+    target_time_series = []
+    for batch_data in val_batches:
+        states_with_overlap_time_series = get_ordered_X(
+            batch_data, model.input_variables
+        )
+
+        if isinstance(model, HybridReservoirComputingModel):
+            hybrid_inputs_time_series = get_ordered_X(
+                batch_data, model.hybrid_variables
```
frodre, line 134: _states_without_overlap() needed here?

frodre (approved):
Looks like a good start to me with your selection of basic metrics.  I might add one more, the time-mean squared error, since that one can't be calculated offhand with what's saved so far.

Other than that, just check that the hybrid time and extent are correct for the metric calculation.  I was basing this off of my understanding of how the rc works, so let me know if I need to update my mental model.

workflows/diagnostics/fv3net/diagnostics/reservoir/compute.py
frodre: Or perhaps the hybrid variables saved are the post t+1 data already? (For a 15 minute fv3gfs step)

AnnaKwa: You're right- that's a subtle but important point for the imperfect 'prediction' to also be for the same t+1 timestep as the reservoir prediction. It's a bug in the training too. I also fixed that line in the training in the review response.

AnnaKwa: Just refreshed myself on the [training data generation](https://github.com/ai2cm/vcm-workflow-control/blob/2023-06-23-reservoir-workflow-example/examples/generate_hybrid_training/prescribe_reference_3h.yaml) to double check your second point- both the prescribed state (target) and the hybrid input (ML prediction) will both be for the same end of timestep as long as the state is saved to state_after_timestep.zarr and the prescribing is done via the bias_correction config field (naming is a relic from when the feature was introduced, should probably update that)

AnnaKwa: On second thought- 
You are correct that it should be the ML model prediction at time t+1 used as input, but in correcting this I'm thinking that it might be best to take care of this shift in the data preprocessing. e.g. a timestep in the data might have timestamp 0100, with data air_temperature (at 0100) and air_temperature_predicted_by_corrective_ML_subsequent_timestep (prediction for 0400).

The advantage of this would be that 
i) all the data needed for a prediction would be contained in a single batch, no need to grab the first ML-prediction from the next batch when at the end of a batch
ii) allows for hybrid prediction cases like Paula's where the additional inputs aren't necessarily an imperfect prediction for the predicted timestep, but are just other fields at the current input state time

PR #2290, opened 2023-07-20
`HybridReservoirComputingModel` takes rank data without halo points as input

AnnaKwa:
The implementation of `HybridReservoirComputingModel.predict` didn't match the intended usage in the docstring. This method should take in hybrid rank data *without* the halo points, since in online usage that would require an additional halo update to retrieve those points. I added a check that the input data is the expected shape, without halo points. 

This was enabled by a lower down bug in `RankDivider.get_subdomain_tensor_slice`. The slicing methods in that class are intended only to be used on the full input (i.e. rank plus halo points) but it was possible to call that method on an array with smaller x/y dim size. I added a check to this method that raises a ValueError if the input array not the expected shape. 

Since `RankDivider.get_subdomain_tensor_slice` is still the simplest method to use in `predict` to divide the xy array into subdomains, I added a step in `HybridReservoirComputingModel.predict` which pads the input hybrid data (which has no halo points) with a halo of zeros so that it is compatible with `RankDivider.get_subdomain_tensor_slice`. Using the arg `overlap=False` when dividing into subdomains omits the zero padded points from the subdomain slices.
 
A few tests were amended to reflect these changes in requirements.

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/domain.py
+++ b/external/fv3fit/fv3fit/reservoir/domain.py
@@ -25,6 +25,9 @@ class RankDivider:
         overlap: int,
     ):
         """ Divides a rank of data into subdomains for use in training.
+        When dividing a tensor into subdomains, it is assumed that the input rank
+        data always includes <overlap> number of halo points.
+
         Args:
             subdomain_layout: layout describing subdomain grid within the rank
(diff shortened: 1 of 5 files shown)


frodre (approved):
LGTM, thanks for the quick fix

PR #2292, opened 2023-07-20
Fix prognostic scream run unit test

elynnwu:
Unit tests broke due to missing `py` package and outdated `pytest`. `py` was already installed but for some reason python did not recognize it, uninstall and install fixed the issue. 


mcgibbon (approved):
LGTM.

PR #2293, opened 2023-07-21
Count reservoir synchronization by timestep instead of batch during training

AnnaKwa:
This PR updates how the synchronization is tracked in training. Previously the amount of data at the start of the training set to use for synchronization was specified as a number of batches. This was inflexible since the synchronization had to be a multiple of the data batch size, and in some cases the entire training dataset is a single large batch. Now, the synchronization can be set as a number of time steps. To enable this, a `SynchronziationTracker` class is added which counts the number of times the reservoir has been incremented. If the number of increments is less than the synchronization length, the data is cut out of the training set.

 
Added public API:
- replaces the reservoir model config parameter `n_batches_burn` with `n_timesteps_synchronize`, which allows for more control over how many timesteps are used for reservoir  synchronization.
  

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/utils.py
+++ b/external/fv3fit/fv3fit/reservoir/utils.py
@@ -6,6 +6,11 @@ from fv3fit.reservoir.domain import RankDivider, assure_txyz_dims
 
 
 class SynchronziationTracker:
+    """Counts the number of times a reservoir has been incremented,
+    and excludes time series data from training set if the number of
+    incrments is less than the specified synchronization length.
+    """
+
     def __init__(self, n_synchronize: int):
         self.n_synchronize = n_synchronize
         self.n_steps_synchronized = 0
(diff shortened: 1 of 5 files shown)


external/fv3fit/fv3fit/reservoir/utils.py
```diff
@@ -5,6 +5,35 @@
 from fv3fit.reservoir.domain import RankDivider, assure_txyz_dims
 
 
+class SynchronziationTracker:
+    def __init__(self, n_synchronize: int):
```
frodre, line 9: Add a docstring description

external/fv3fit/fv3fit/reservoir/utils.py
```diff
@@ -5,6 +5,35 @@
...
+        self.n_steps_synchronized = 0
+
+    @property
+    def completed_synchronization(self):
+        if self.n_steps_synchronized > self.n_synchronize:
+            return True
+        else:
+            return False
+
+    def count(self, n_samples: int):
```
frodre, line 20: Perhaps a clearer name would be helpful.  "count_sync_steps"?

frodre (approved):
LGTM!

PR #2295, opened 2023-07-21
off the shelf model first draft

cynicalsavant:
A few things are bundled with this PR:
- SwinIR model for the cold start problem
- CRPS metrics
- pattern bias and related histogram-based metrics


projects/super_res/model/autoreg_diffusion_mod.py
```diff
@@ -0,0 +1,1442 @@
...
+def normalize_to_neg_one_to_one(img):
+    return img * 2 - 1
+
+def unnormalize_to_zero_to_one(t):
+    return (t + 1) * 0.5
+
+# ssf modules
+
+def gaussian_pyramids(input, base_sigma = 1, m = 5):
+
```
gideonite, line 106: nit: I think it would make sense to remove this spacing. Maybe it's copied from another codebase where this is standard? Still, it doesn't seem to be standard in many codebases and IMHO makes it hard to read.

projects/super_res/model/autoreg_diffusion_mod.py
```diff
@@ -0,0 +1,1442 @@
...
+from matplotlib.cm import ScalarMappable as smap
+
+from tqdm.auto import tqdm
+from ema_pytorch import EMA
+
+import flow_vis
+
+from accelerate import Accelerator
+
+from .network_swinir import SwinIR as net
```
gideonite, line 36: Given that we have a number of neural networks, I think it would make sense to give this a more specific name. I would be fine with "swinir_net" or "context_net" but am open to whatever you think makes the most sense. Still, I think that "net" is too ambiguous. Would you agree?

projects/super_res/model/autoreg_diffusion_mod.py
```diff
@@ -0,0 +1,1442 @@
...
+                                pred = np.zeros((num_samples, 1, num_videos_per_batch, num_frames, img_channels, img_size, img_size), dtype = 'uint8')
+                                truth[0,:,:,:,:,:] = (hres[:,2:,:,:,:].repeat(1,1,1,1,1).cpu().numpy()*255).astype(np.uint8)
+                                for k in range(num_samples):
+                                    videos, base, res, flows = self.ema.ema_model.sample(lres)
+                                    pred[k,0,:,:,:,:] = (videos.clamp(0.0, 1.0)[:,:,0:1,:,:].repeat(1,1,1,1,1).detach().cpu().numpy()*255).astype(np.uint8)
+                                
+                                crps_index = calculate_crps(truth, pred, num_samples, num_videos_per_batch, num_frames, img_channels, img_size)
+                                psnr_index = piq.psnr(hres[:,2:,0:1,:,:], videos.clamp(0.0, 1.0)[:,:,0:1,:,:], data_range=1., reduction='none')
+                                
+                                accelerator.log({"true_high": wandb.Video((hres[:,2:,0:1,:,:].repeat(1,1,3,1,1).cpu().numpy()*255).astype(np.uint8))}, step=self.step)
```
gideonite, line 1263: Let's take this opportunity to change these names into the interpretable ones that you are using your wandb [workspace](https://wandb.ai/ai2cm/vsr-orig-autoreg-hres/runs/hpc6lshz?workspace=user-prakhars). Does this sound reasonable?

projects/super_res/fv3net.yml
```diff
@@ -0,0 +1,297 @@
+name: fv3net
```
gideonite, line 1: I see that this has more specified deps, which makes it more brittle for others but makes your workflow a bit faster. I hope that you won't need to do this step very often anyway but in the meantime -- could you help me understand why this is called fv3net? Is just the deps for your code?

projects/super_res/model/autoreg_diffusion_mod.py
```diff
@@ -0,0 +1,1442 @@
...
+
+# constants
+
+ModelPrediction =  namedtuple('ModelPrediction', ['pred_noise', 'pred_x_start'])
+
+# helpers functions
+
+def calculate_crps(truth, pred, num_samples, num_videos_per_batch, num_frames, img_channels, img_size):
+    truth_cdf = np.zeros((256, 1, num_videos_per_batch, num_frames, img_channels, img_size, img_size), dtype = 'uint8')
+    for i in range(256):
```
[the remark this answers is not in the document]
cynicalsavant: So all the images are stored with 8 bit precision, hence the 256. Leads to natural binning of the pixel space so calculating the cdf becomes very trivial in native numpy, otherwise there's an unnecessary layer of complexity

projects/super_res/model/autoreg_diffusion_mod.py
cynicalsavant: Aah of course

projects/super_res/model/autoreg_diffusion_mod.py
cynicalsavant: sure, i'll change it

projects/super_res/model/autoreg_diffusion_mod.py
cynicalsavant: We can easily edit them in WandB, the only reason I didn't change it here was because the code becomes unnecessarily wordy

projects/super_res/fv3net.yml
cynicalsavant: yeah yeah I just named my env that way, no specific reason

gideonite: nit: I think it's a bit confusing. Best change it to something like `vsr`.

gideonite (approved):
There are a lot of changes bundled into this PR. In the interest of not blocking progress, I am just going to approve this. There are some exciting new results and figures that need to get merged in.

PR #2297, opened 2023-07-26
preprocessing of era5 sst data format changed for fv3net reservoir computing

paulaharder:
I changed era_process.py to
- have weekly mean quantities instead of daily
- take start and end dates as arguments to create a dataset for a specific time period
- not have a time dim issue that requires decode_times=False when loading with xarray
- have separate files for different tiles
- enable separate files for train/val/test
- include all given variables (sst,t2m,u10,v10,sst_tend) in one file
- shift t2m,u10,v10
- have a new field "sst_tendency", the sst tendency

### DIFF
--- a/projects/reservoir/fv3/era_process.py
+++ b/projects/reservoir/fv3/era_process.py
@@ -14,8 +14,23 @@ FREGRID_EXAMPLE_SOURCE_DATA = (
     "gs://vcm-ml-raw/2020-11-12-gridspec-orography-and-mosaic-data/C48/*.nc"
 )
 
-var_id_mapping = {"sst": "sst", "temp2m": "t2m", "u_wind": "u10", "v_wind": "v10"}
+var_id_mapping = {
+    "sst": "sst",
+    "sst_tendency": "sst_tendency",
+    "temp2m": "t2m",
+    "u_wind": "u10",
+    "v_wind": "v10",
+}
 
+shift_rename_mapping = {
+    "sst": "sst",
+    "sst_tendency": "sst_tendency",
+    "t2m": "t2m_at_next_timestep",
+    "u10": "u10_at_next_time_step",
+    "v10": "v10_at_next_time_step",
+}
+
+logging.basicConfig()
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
@@ -26,7 +41,17 @@ def add_arguments():
         "--in_path", help="path to directory /var/downloaded which contains raw data"
     )
     parser.add_argument("--out_path", help="path to save regridded data")
+    parser.add_argument(
+        "--train_val_test", help="whether you are creating train, val or test data"
+    )
+    parser.add_argument(
+        "--start_date", help="start date as a string in the format of YYYY-MM-DD"
+    )
+    parser.add_argument(
+        "--end_date", help="end date as a string in the format of YYYY-MM-DD"
+    )
     parser.add_argument("--variables", nargs="+", help="list of variables to process")
+    # dates to split val, test
     return parser.parse_args()
 
 
@@ -37,18 +62,34 @@ def main(args):
     setup_fregrid()
     for var in variables:
         check_time_steps_complete(path, var)
-        weekdays = ["mon", "tue", "wed", "thu", "fri", "sat", "sun"]
-        for i, day in enumerate(weekdays):
-            merge_into_file(path, var, i, day)
-            logger.info(f"{day} is merged")
-            regrid_to_cubed_sphere(path, var, day)
-            logger.info(f"{day} is regridded")
-            interpolated = interpolate_nans(var, day)
-            logger.info(f"{day} is interpolated")
-            masked = mask(interpolated)
-            logger.info(f"{day} is masked")
-            masked.to_netcdf(args.out_path + var + "_" + day + ".nc")
-            logger.info(f"{day} is done")
+        merge_into_weekly_file(path, var, args)
+        logger.info(f"{var} is merged into weekly file")
+        regrid_to_cubed_sphere(path, var)
+        logger.info(f"{var} is regridded")
+        interpolated = interpolate_nans(var)
+        interpolated.to_netcdf("interpolated.nc")
+        logger.info(f"{var} is interpolated")
+        masked = mask(interpolated)
+        masked.to_netcdf("temp.nc")
+        logger.info(f"{var} is masked")
+        if var in ["temp2m", "u_wind", "v_wind"]:
+            shifted = shift_variable(masked)
+        else:
+            shifted = masked
+        shifted.to_netcdf(
+            args.out_path + var + "_" + args.train_val_test + "_regridded.nc"
+        )
+        logger.info(f"{var} is done")
+    if "sst" in variables:
+        calculate_sst_tendency(args)
+        variables.append("sst_tendency")
+    merged = merge_variables(variables, args)
+    for var in variables:
+        var_id = var_id_mapping[var]
+        merged = merged.rename_vars({var_id: shift_rename_mapping[var_id]})
+    del merged.time.attrs["units"]
+    del merged.time.attrs["calendar"]
+    save_data_as_tiles(merged, args)
 
 
 def download_source_data():
@@ -97,7 +138,7 @@ def check_time_steps_complete(path, var):
     for f in glob.glob(path + var + "/downloaded/*.nc"):
         dataset = xr.open_dataset(f)
         time_step_list.append(dataset["time"].data)
-    dataset.close()
+        dataset.close()
     arrs = list(np.concatenate(time_step_list))
     arrs.sort()
     for i, date in enumerate(arrs[1:]):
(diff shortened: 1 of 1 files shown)


projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
+        logger.info(f"{var} is merged into weekly file")
+        regrid_to_cubed_sphere(path, var)
+        logger.info(f"{var} is regridded")
+        interpolated = interpolate_nans(var)
+        interpolated.to_netcdf("interpolated.nc")
+        logger.info(f"{var} is interpolated")
+        masked = mask(interpolated)
+        masked.to_netcdf("temp.nc")
+        logger.info(f"{var} is masked")
+        if var in ["temp2m", "u_wind", "v_wind"]:
```
AnnaKwa, line 75: For visibility, hard code the list of variables that get shifted as `TIME_SHIFTED_VARIABLES` at the top of the module and refer to that variable here.

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
+            shifted = shift_variable(masked)
```
AnnaKwa, line 76: Can just call `xr.DataArray.shift` directly here and remove the `shift_variable` function.

projects/reservoir/fv3/era_process.py
```diff
@@ -108,26 +149,21 @@ def check_time_steps_complete(path, var):
...
+        [xr.open_dataset(f) for f in glob.glob(path + var + "/downloaded/*.nc")],
+        dim="time",
+    )
+    merged_data = merged_data.sortby("time")
+    # get time chunk according to inout args todo
+    sorted_data = merged_data.sel(time=slice(args.start_date, args.end_date))
+    # merge into weekly mean
+    weekly_merged = calculate_weekly_mean(sorted_data)
     current_working_directory = os.getcwd()
+    weekly_merged.to_netcdf(path + var + "/merged/" + var + ".nc")
```
AnnaKwa, line 163: Can define this netcdf name as a variable since it's used again later in the subsequent function call. For extra clarity I'd add a `weekly_mean` descriptor to the filenames.

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
+        interpolated.to_netcdf("interpolated.nc")
+        logger.info(f"{var} is interpolated")
+        masked = mask(interpolated)
+        masked.to_netcdf("temp.nc")
+        logger.info(f"{var} is masked")
+        if var in ["temp2m", "u_wind", "v_wind"]:
+            shifted = shift_variable(masked)
+        else:
+            shifted = masked
+        shifted.to_netcdf(
```
AnnaKwa, line 79: There's a number of temporary netcdfs that are written in this script before the final merged output is saved. For the intermediate netcdfs that are not needed, I would create them in a[ temporary directory](https://stackoverflow.com/questions/3223604/how-do-i-create-a-temporary-directory-in-python) so they are cleaned up at the end.

projects/reservoir/fv3/era_process.py
```diff
@@ -26,7 +41,17 @@ def add_arguments():
         "--in_path", help="path to directory /var/downloaded which contains raw data"
     )
     parser.add_argument("--out_path", help="path to save regridded data")
+    parser.add_argument(
```
AnnaKwa, line 44: I'd just omit this arg and have this information in the `out_path` arg instead.

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
     setup_fregrid()
```
AnnaKwa, line 62: I missed this in the previous PR, but the download locations for the data saved in this function are also hard coded. I'd change this to save them in a temporary directory (see other comment about this).

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
+        )
+        logger.info(f"{var} is done")
+    if "sst" in variables:
+        calculate_sst_tendency(args)
+        variables.append("sst_tendency")
+    merged = merge_variables(variables, args)
+    for var in variables:
+        var_id = var_id_mapping[var]
+        merged = merged.rename_vars({var_id: shift_rename_mapping[var_id]})
+    del merged.time.attrs["units"]
```
AnnaKwa, line 90: Can move these lines modifying attrs into the `merge_variables` function.

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
-            logger.info(f"{day} is merged")
-            regrid_to_cubed_sphere(path, var, day)
-            logger.info(f"{day} is regridded")
-            interpolated = interpolate_nans(var, day)
-            logger.info(f"{day} is interpolated")
-            masked = mask(interpolated)
-            logger.info(f"{day} is masked")
-            masked.to_netcdf(args.out_path + var + "_" + day + ".nc")
-            logger.info(f"{day} is done")
+        merge_into_weekly_file(path, var, args)
```
AnnaKwa, line 65: Instead of calling the sequence of regridding and masking for each variable, can they all just be merged at the start? e.g. something like
```
for var in variables:
    check_time_steps_complete(...)
full_variables = merge_variables(path, variables)
# do time shift, add tendency by doing xarray operations on the full dataset, no need to write it in between steps
full_variables["sst_tendency"] = calculate_sst_tendency(...)
for var in TIME_SHIFT_VARIABLES:
    full_variables[var] = full_variables[var].shift(time=-1)
# masking, etc.
...
```

projects/reservoir/fv3/era_process.py
```diff
@@ -37,18 +62,34 @@ def main(args):
...
+        logger.info(f"{var} is merged into weekly file")
+        regrid_to_cubed_sphere(path, var)
+        logger.info(f"{var} is regridded")
+        interpolated = interpolate_nans(var)
+        interpolated.to_netcdf("interpolated.nc")
```
AnnaKwa, line 70: With the exception of the netcdfs used as input to fregrid, you don't have to save netcfs in between operations. Instead, just pass the dataset to the next function and return the modified dataset.

AnnaKwa (changes requested):
Thanks for the updates! High level comments:
- There are a few intermediate steps that save netcdfs when they could skip this and just pass a dataset to the next operation. I recommend removing these. For the cases where a temporary netcdfs does need to be written, i.e. the fregrid input data, I'd use a temporary directory that gets cleaned up after the script is done.
- If you move the merging of all the variables to the beginning of the operations, after the times are trimmed to the specified range, you can do the regridding, masking, etc. once and don't have to loop over variables

projects/reservoir/fv3/era_process.py
```diff
@@ -212,6 +243,62 @@ def mask(coarsened):
...
+            decode_times=False,
+        )
+        data_list.append(data)
+    return xr.merge(data_list)
+
+
+def save_data_as_tiles(data, args):
+    # save each tile individually
+    for i in range(len(data.tile)):
+        data.isel(tile=i).to_netcdf(
```
AnnaKwa, line 297: The data loaders that are used in training take in a directory which is assumed to contain all the netcdfs for a single rank (tile). Could you change the output path to match this structure (one tile data per directory)?

projects/reservoir/fv3/era_process.py
paulaharder: Actually, I removed the attrs stuff

projects/reservoir/fv3/era_process.py
paulaharder: interpolated.nc of a leftover from debugging, I'll remove it

projects/reservoir/fv3/era_process.py
paulaharder: Maybe, but I was worried that fregrid is going to cause problems

projects/reservoir/fv3/era_process.py
paulaharder: I tried this, but it caused many issues with making directories or writing files, so I just created a /temp folder and removed it after

projects/reservoir/fv3/era_process.py
paulaharder: Okay, that#s done. It worked

projects/reservoir/fv3/era_process.py
```diff
@@ -79,15 +87,22 @@ def main(args):
...
+    interpolated["sst"] = mask(interpolated["sst"], args)
+    if "sst" in variables:
+        variables.append("sst_tendency")
+        interpolated["sst_tendency"] = calculate_sst_tendency(interpolated.sst)
+        interpolated["sst_tendency"].fillna(args.fill_value_sst_tend)
+    for var in variables:
+        var_id = var_id_mapping[var]
+        interpolated = interpolated.rename_vars({var_id: shift_rename_mapping[var_id]})
+    save_data_as_tiles(interpolated, args)
+    atexit.register(shutil.rmtree, tempdirname)
```
AnnaKwa, line 105: Move this line further up right after the temporary directory is created

projects/reservoir/fv3/era_process.py
```diff
@@ -52,16 +54,22 @@ def add_arguments():
...
         "--start_date", help="start date as a string in the format of YYYY-MM-DD"
     )
     parser.add_argument(
         "--end_date", help="end date as a string in the format of YYYY-MM-DD"
     )
     parser.add_argument("--variables", nargs="+", help="list of variables to process")
+    parser.add_argument(
+        "--fill_value_sst", default=291, type=float, help="value to fill sst nans"
+    )
+    parser.add_argument(
```
AnnaKwa, line 67: Can be removed since now the reservoir training can save a mask array to deal with unwanted inputs. However since that would require additional changes to this PR, just leave this for now and add a TODO comment above it mentioning that it can be removed in a subsequent PR.

AnnaKwa (approved):
Approving with a couple remaining minor comments

projects/reservoir/fv3/era_process.py
```diff
@@ -10,62 +10,103 @@
...
+        "--end_date", help="end date as a string in the format of YYYY-MM-DD"
+    )
     parser.add_argument("--variables", nargs="+", help="list of variables to process")
+    # dates to split val, test
     return parser.parse_args()
 
 
 # raw nc input files are expected to be in path+var+'/downloaded/
 def main(args):
+    tempdirname = "temp"
```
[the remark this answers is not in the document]
paulaharder: That didn't work, we tried it a couple weeks ago

PR #2302, opened 2023-07-28
Coarsening: expose `toa_pressure` as parameter

spencerkclark:
We are more frequently running with different vertical level configurations than the 79-level GSRM configuration of X-SHiELD.  These configurations importantly have different top of model pressures.  If this is not properly accounted for when computing pressure at the interfaces and midpoints during the coarsening process, errors are introduced.  These lower-level functions generally all have `toa_pressure` as a parameter; however this parameter is not exposed at the higher level of our restart coarsening routines.  This PR fixes that, making it possible to apply these routines reliably to sets of restart files produced with different model configurations.

[This notebook](https://github.com/ai2cm/explore/blob/master/spencerc/2023-07-28-illustrate-toa-pressure-error.ipynb) illustrates the type of errors one experiences when using the wrong TOA pressure.  `phis` after hydrostatic adjustment is the most sensitive variable.

Added public API:
- A required `toa_pressure` parameter to:
  - `vcm.calc.thermo.vertically_dependent.hydrostatic_dz`
  - `vcm.cubed_sphere.coarsen_restarts.coarsen_restarts_on_sigma`
  - `vcm.cubed_sphere.coarsen_restarts.coarsen_restarts_on_pressure`
  - `vcm.cubed_sphere.coarsen_restarts.coarsen_restarts_via_blended_method`
  - `vcm.cubed_sphere.regridz.regrid_to_area_weighted_pressure`
  - `vcm.cubed_sphere.regridz.regrid_to_edge_weighted_pressure`
- An optional `--toa_pressure` flag to the coarsen restarts dataflow pipeline (default `vcm.calc.thermo.constants.TOA_PRESSURE)`.

Significant internal changes:
- `toa_pressure` argument threaded through numerous internal coarsening functions.

- [x] Tests added

### DIFF
--- a/external/vcm/tests/test_coarsen_restarts.py
+++ b/external/vcm/tests/test_coarsen_restarts.py
@@ -7,6 +7,7 @@ import pytest
 import xarray as xr
 
 from pathlib import Path
+from vcm.calc.thermo.constants import TOA_PRESSURE
 from vcm.cubedsphere.coarsen_restarts import (
     coarsen_restarts_on_sigma,
     coarsen_restarts_on_pressure,
@@ -112,7 +113,11 @@ def test_coarsen_restarts(tag):
     restart_data, grid_data = generate_synthetic_data(restart_schemas, grid_spec_schema)
 
     func, kwargs = REGRESSION_TESTS[tag]
-    result = func(FACTOR, grid_data, restart_data, **kwargs)
+    if func == coarsen_restarts_on_sigma:
+        # coarsen_restarts_on_sigma does not require the toa_pressure as an argument
+        result = func(FACTOR, grid_data, restart_data, **kwargs)
+    else:
+        result = func(FACTOR, grid_data, TOA_PRESSURE, restart_data, **kwargs)
     result = {category: ds for category, ds in result.items()}
 
     # To reset the reference data, run this module as a script:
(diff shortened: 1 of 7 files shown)


external/vcm/tests/test_coarsen_restarts.py
```diff
@@ -112,7 +113,11 @@ def test_coarsen_restarts(tag):
     restart_data, grid_data = generate_synthetic_data(restart_schemas, grid_spec_schema)
 
     func, kwargs = REGRESSION_TESTS[tag]
-    result = func(FACTOR, grid_data, restart_data, **kwargs)
+    if func == coarsen_restarts_on_sigma:
+        # coarsen_restarts_on_sigma does not require the toa_pressure as an argument
+        result = func(FACTOR, grid_data, restart_data, **kwargs)
+    else:
+        result = func(FACTOR, grid_data, TOA_PRESSURE, restart_data, **kwargs)
```
elynnwu, line 120: [optional] the other option is to have `coarsen_restarts_on_sigma` still take in `toa_pressure` and do nothing. I find this if statement to be a bit awkward, or perhaps this test should be split into different tests.

elynnwu (approved):

oliverwm1 (approved):
LGTM

PR #2312, opened 2023-08-15
Decoder always returns list of arrays

AnnaKwa:
One-liner fix for the case where there is only one output variable. Keras models return a list of arrays for >2 outputs, but just an array for 1 output. This PR ensures a list is always returned.


frodre (approved):
LGTM.  Maybe add a small test that ensures this.  I'm not sure if this will break if we ever update TF version; I couldn't find anything in the current docs about the function.

PR #2317, opened 2023-08-23
Reservoir option to use input mask

AnnaKwa:
This adds the optional config parameter `mask_variable` to mask reservoir training inputs according to the field specified by that arg. A mask array is saved with the reservoir that zeroes out any input points where the mask is zero. The mask is applied right before the inputs are multiplied with the input reservoir matrix that transforms the physical input into the higher dim hidden state,  ensuring the masked points don't have any effect on the hidden state. This could be useful in the case of masking out land and corner points in SST training. 

Additional minor changes:
- refactored a helper function `assure_txyz_dims` to operate on single tensors, since it is useful for ensuring the mask has the same dimensions as the input.
- fixed the dacite config in the training config so it can load tuples (used in rank extent of the reservoir subdomain config)

In addition to the unit test, I trained models on mock SST data using the land mask as the input mask, then loaded them and tested the input mask was working in the trained models' predictions. I altered the values over land in the test data to be way out of sample from the original data, and verified that the models' predictions did not change when using the altered data as input. (bottom of [this notebook](https://github.com/ai2cm/explore/blob/master/annak/2023-07-17-example-reservoir-training/2023-08-15-global-rollout.ipynb))
 
- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/_shared/training_config.py
+++ b/external/fv3fit/fv3fit/_shared/training_config.py
@@ -110,7 +110,7 @@ class TrainingConfig:
         hyperparameter_class = get_hyperparameter_class(kwargs["model_type"])
         # custom enums must be specified for dacite to handle correctly
         dacite_config = dacite.Config(
-            strict=True, cast=[bool, str, int, float, StdDevMethod, MeanMethod]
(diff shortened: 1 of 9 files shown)


frodre (approved):
LGTM

PR #2318, opened 2023-08-25
Add vcm.DerivedMapping keys for lowest level T, q, u, v

AnnaKwa:
The SST hybrid reservoir model uses several lowest level atmospheric inputs, which are saved in fortran diagnostics as 2D fields but only available at runtime via the full 3D fields. This adds entries to `vcm.DerivedMapping` so that just the lowest level 2D fields can be retrieved at runtime to use as inputs.

### DIFF
--- a/external/vcm/vcm/derived_mapping.py
+++ b/external/vcm/vcm/derived_mapping.py
@@ -432,13 +432,6 @@ def lowest_level_air_temperature(self):
 
 
 @DerivedMapping.register(
-    "lowest_level_specific_humidity", required_inputs=["specific_humidity"]
-)
-def lowest_level_specific_humidity(self):
-    return self["specific_humidity"].isel(z=-1).squeeze()
-
-
-@DerivedMapping.register(
     "lowest_level_northward_wind", required_inputs=["northward_wind"]
 )
 def lowest_level_northward_wind(self):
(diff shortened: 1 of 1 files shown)


frodre (approved):
Thanks @AnnaKwa! I'm curious how much of a functional difference the lowest level fields will be compared to the diagnostic 2m T U10/V10, etc.  I'm guessing it's more of a hassle to grab those "diagnostic" variables from the wrapper scope, so we should only cross that bridge if necessary.

AnnaKwa:
That's a good point. I checked out your concerns in this [notebook](https://github.com/ai2cm/explore/blob/master/annak/2023-07-17-example-reservoir-training/2023-08-25-check-fortran-near-sfc-vs-lowest-level.ipynb). It looks like the difference in training fields vs lowest model level would be a source of systematic bias when testing out the hybrid SST reservoir online.

AnnaKwa:
I looked into Andre's comment and the difference between the T2m, 10m u/v wind fields (used in training) and the lowest model level fields is pretty significant. https://github.com/ai2cm/explore/blob/master/annak/2023-07-17-example-reservoir-training/2023-08-25-check-fortran-near-sfc-vs-lowest-level.ipynb

 I made a different PR to instead expose 10m winds (T2m is already available in the wrapper). Closing this PR.

PR #2319, opened 2023-08-28
expose u10m and v10m to derived state

AnnaKwa:
Exposes 10m u and v winds to the wrapper so they can be used at runtime for inputs to SST models.


spencerkclark (approved):
Looks good to me!  

I suggested adding word of caution to the comment for anyone adding fields to this dictionary in the future; this is powerful functionality, but it needs to be used with care (before recommending this approach I double checked that [`u10m` and `v10m` are indeed instantaneous diagnostics](https://github.com/ai2cm/fv3gfs-fortran/blob/b3793303df9800de341a3f8e8b234c6f4dc2b7a1/FV3/gfsphysics/GFS_layer/GFS_diagnostics.F90#L1661-L1685)).

PR #2320, opened 2023-08-28
Script to split reservoir model into subdomain models

frodre:
Added functions to take a reservoir model for a full tile and multiple subdomains and split each subdomain out into its own model artifact.  The fv3gfs configuration for reservoir models uses a mapping from rank integer -> reservoir model to load for that rank.  Since we defined the subdomain division using the tile partitioner it's easy to keep the ranks and subdomain models aligned with the model cpu layout.

Added public API:
* `adapters.split_multi_subdomain_model`: splits a single model object into its component subdomains as models
* `adapters.generate_subdomain_models_from_all_tiles`: take model rank to save path mapping and expand each model into subdomain models saved in an output directory
* `projects/reservoir/scripts/split-rc-model.py`: A script that splits models and uploads them to a new directory using an input yaml (e.g., `models.yaml`)

- [x] Tests added


external/fv3fit/tests/reservoir/_regtest_outputs/test_model_adapter.test_adapter_predict.out
```diff
@@ -2,5 +2,5 @@
 Dimensions:  (x: 4, y: 4, z: 3)
 Dimensions without coordinates: x, y, z
 Data variables:
-    a        (x, y, z) float64 -1.967 0.5274 -1.969 -1.259 ... 5.788 4.426 4.439
-    b        (x, y, z) float64 -0.5784 -0.6441 -0.739 ... 4.685 4.086 2.695
+    a        (x, y, z) float64 1.486 -3.29 2.696 ... -0.3318 7.511 -0.06362
```
frodre, line 5: These changed because the rng state was not pinned for generating the coefficient matrices and I switched to use a centralized helper function to grab the models during tests.

external/fv3fit/tests/reservoir/helpers.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
```
frodre, line 1: Moving the model object instance generation to a shared location for test files.

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -276,6 +276,8 @@ def predict(self, inputs, state):
         # no halo necessary for potential hybrid inputs
         # +1 to align with the necessary increment before any prediction
         if self._state_machine.completed_increments >= self.synchronize_steps + 1:
+
+            logger.info(f"Predicting rc model")
```
frodre, line 280: Move the notification within step where it actually predicts.

external/fv3fit/fv3fit/reservoir/adapters.py
```diff
@@ -198,3 +206,97 @@ def load(cls, path: str) -> "HybridReservoirDatasetAdapter":
...
+        "reservoir": model.reservoir,
+        "transformers": model.transformers,
+        "square_half_hidden_state": model.square_half_hidden_state,
+    }
+
+    if isinstance(model, HybridReservoirComputingModel):
+        new_kwargs["hybrid_variables"] = model.hybrid_variables
+
+    new_models = []
+    for i in range(rank_divider.n_subdomains):
```
AnnaKwa, line 257: This doesn't have to happen in this PR if it's a pain, but  it may be useful to be able to split into smaller ranks that still contain >1 subdomain. e.g. original model covering a full tile contains 36 subdomains of 8x8 each and is split into 4 models, each with 9 subdomains.

AnnaKwa (approved):
LGTM! I have a few suggestions but nothing major.

external/fv3fit/fv3fit/reservoir/adapters.py
```diff
@@ -198,3 +206,97 @@ def load(cls, path: str) -> "HybridReservoirDatasetAdapter":
...
         )
         return adapter
+
+
+ReservoirAdapterType = Union[ReservoirDatasetAdapter, HybridReservoirDatasetAdapter]
+ReservoirModelLike = Union[ReservoirModelType, ReservoirAdapterType]
+
+
+@typing.no_type_check
+def split_multi_subdomain_model(
```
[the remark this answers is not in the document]
frodre: Initially this was an attempt to contain the changes rather than add another method to the 4 model classes, but I agree the conditionals are cumbersome.  I made the change to move it into the pure reservoir and did my best not to repeat too much code.  I think this points to the need for some consolidation of the reservoir models and adapters, but I'll leave that for a future refactor.

external/fv3fit/fv3fit/reservoir/adapters.py
frodre: I think this is an interesting idea.  I think it can be accomplished using the rank divider on a subdomain layout indices to keep the coefficients grouped properly for the readout, but too complicated for this PR.

projects/reservoir/scripts/models.yaml
```diff
@@ -0,0 +1,6 @@
+0: gs://vcm-ml-scratch/paulah/sst-only-full-sub-24-halo-4-masked-tile-0
```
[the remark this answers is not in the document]
frodre: Yep, forgot to delete this.

PR #2321, opened 2023-08-29
Record states prior to being overwritten by interval stepper

AnnaKwa:
This PR enables saving the baseline FV3GFS coarse model state as the 'imperfect' hybrid input set. The state is prescribed to the nudged reference state every 3 hours, but we need to record the model state immediately before this correction is applied.
 
Added public API:
- optional field `IntervalConfig.record_fields_before_update` takes a list of state variables to save in diagnostics right before the postphysics stepper updates are applied. These are saved as `x_before_interval_update`. They are also saved at timesteps where the interval stepper does not overwrite the state, since the diagnostics need to be available at all timesteps.
 

frodre (approved):
LGTM

PR #2323, opened 2023-09-01
Reservoir diagnostic adjustment

frodre:
Fixes diagnostic names for inputs and hybrid inputs to match the fv3gfs standard names.  Switch output diagnostic to always be a prediction from the reservoir so we can investigate spinup during synchronization. Add offset to facilitate saving reservoir inputs since the increment happens before the timestep ticks over to the next time (diagnostics recorded at end of timestep).

Adjusted the workflow labels to adhere to the pod disruption budgets we have set on k8s.  This should prevent long-running pods from being evicted (hopefully)

Public API Changes:
* Added `offset` to `IntervalTimes` container to facilitate saving outputs at a frequency offset from the initial time

- [x] tests added


AnnaKwa (approved):
LGTM

PR #2326, opened 2023-09-06
Reservoir training on sequence of time series

AnnaKwa:
This adds the option to train a reservoir model on multiple time series. New, larger training datasets consist of an ensemble of simulations with different initial conditions in the C384 reference data. Reservoir model training requires specialized handling of multiple time series since each must have its own synchronization period at the start.

The main adjustment is that the `reservoir` training function can now take in either a `tf.Dataset` or `Sequence[tf.Dataset]`. If a sequence of datasets is provided it is assumed that each member in the sequence is a time series of training data, and synchronization will be done at the start of each time series before the sequence of hidden states used in training is generated.

In order to minimize change to the existing training code in `fv3fit.train` I just copied that script to the reservoir subdirectory and updated the parts relevant to data loading.  The adjusted training script allows a sequence of datasets to be loaded from a training config file if it contains multiple configuration entries. This avoids a lot of edits to deal with typing issues (other training functions only allow a single `tf.Dataset` for training batches).

Added public API:
- `fv3fit.reservoir.train_multiple_dataset_sequence`
- optional parameter `reservoir` in argo `training` WorkflowTemplate- if "true", will use the above training script to load sequence of datasets. Defaults to 'false', which preserves prior behavior of using `fv3fit.train`.

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/domain2.py
+++ b/external/fv3fit/fv3fit/reservoir/domain2.py
@@ -306,7 +306,7 @@ class RankXYDivider:
             dim_slices = self._add_potential_leading_dim_to_slices(
                 subdomain.shape, dim_slices
             )
-            merged[dim_slices] = subdomain
+            merged[tuple(dim_slices)] = subdomain
 
         return merged
 
--- a/external/fv3fit/fv3fit/reservoir/train.py
+++ b/external/fv3fit/fv3fit/reservoir/train.py
@@ -7,7 +7,7 @@ from fv3fit.reservoir.readout import (
 )
 import numpy as np
 import tensorflow as tf
-from typing import Optional, List, Union, cast, Mapping
+from typing import Optional, List, Union, cast, Mapping, Sequence
 from .. import Predictor
 from .utils import (
     square_even_terms,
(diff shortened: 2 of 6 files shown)


external/fv3fit/fv3fit/reservoir/train.py
```diff
@@ -138,81 +140,85 @@ def train_reservoir_model(
...
-            rank_divider=_output_rank_divider_with_overlap,
-            autoencoder=transformers.output,
-            trim_halo=True,
-        )
 
-        # reservoir increment occurs in this call, so always call this
-        # function even if X, Y are not used for readout training.
-        reservoir_state_time_series = _get_reservoir_state_time_series(
-            input_time_series, hyperparameters.input_noise, reservoir
+    for train_batches in train_batches_sequence:
```
AnnaKwa, line 144: All the code that iterates over a time series' batches within this loop is the same as before.

external/fv3fit/fv3fit/reservoir/domain2.py
```diff
@@ -306,7 +306,7 @@ def merge_all_subdomains(self, data: np.ndarray) -> np.ndarray:
             dim_slices = self._add_potential_leading_dim_to_slices(
                 subdomain.shape, dim_slices
             )
-            merged[dim_slices] = subdomain
+            merged[tuple(dim_slices)] = subdomain
```
AnnaKwa, line 309: This addresses a FutureWarning that using non-tuple indices may lead to unexpected results

frodre: Thank you!

frodre (approved):
LGTM, I agree with the separation for something separated like this.  Ends up getting messy trying to support so many specific use cases.

PR #2327, opened 2023-09-06
Reservoir warm start

frodre:
With the long-diagnostic runs we now have the ability to pre-synchronize a model.  This PR adds the ability to load models at runtime without resetting the reservoir state.  Additionally, the state is now a field dumped/loaded with a reservoir model.

Refactored public API:
- The runtime config for reservoirs now includes a `warm_start` option which will forego state resets on the first increment

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/reservoir.py
+++ b/external/fv3fit/fv3fit/reservoir/reservoir.py
@@ -35,6 +35,7 @@ class Reservoir:
     _RESERVOIR_WEIGHTS_NAME = "reservoir_W_res.npz"
     _METADATA_NAME = "metadata.bin"
     _INPUT_MASK_NAME = "input_mask.npy"
+    _STATE_NAME = "state.npy"
 
     def __init__(
         self,
@@ -43,6 +44,7 @@ class Reservoir:
         W_in: Optional[scipy.sparse.csc_matrix] = None,
         W_res: Optional[scipy.sparse.csc_matrix] = None,
         input_mask_array: Optional[np.ndarray] = None,
+        state: Optional[np.ndarray] = None,
     ):
         """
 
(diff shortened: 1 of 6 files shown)


external/fv3fit/fv3fit/reservoir/reservoir.py
```diff
@@ -140,6 +142,10 @@ def dump(self, path: str) -> None:
         with fs.open(os.path.join(path, self._RESERVOIR_WEIGHTS_NAME), "wb") as f:
             scipy.sparse.save_npz(f, self.W_res)
 
+        if self.state is not None:
```
AnnaKwa, line 145: Could you move the state dumping into a separate `dump_state` method that would be called separately from the rest of the model dump? That way, the regular training would still just save a model without a warm start.

workflows/prognostic_c48_run/runtime/steppers/reservoir.py
```diff
@@ -153,18 +155,25 @@ def __init__(
...
         self.diagnostic = diagnostic_only
         self.input_averager = input_averager
+        self.warm_start = warm_start
 
         if state_machine is None:
             state_machine = _FiniteStateMachine()
         self._state_machine = state_machine
 
+        if self.warm_start:
+            self.synchronize_steps = 0
```
AnnaKwa, line 173: Maybe also raise an error if `synchronize_steps != 0` and is going to be overwritten, just to ensure users know what is happening.

AnnaKwa (changes requested):
LGTM for the most part, though could you move the dumping of the state into its own method so that it's optional and not automatically done for all model saves?

AnnaKwa (approved):

PR #2328, opened 2023-09-08
Fix dimension ordering of 3D inputs for online reservoir models

AnnaKwa:
Reservoir models with 3D input states run into errors online when the state data array has the z as the first dimension. This PR adds z to the ordered list of variables to transpose data arrays to.


frodre (approved):
LGTM.  Nice on the separation of concerns for the dim ordering and associated tests!

PR #2330, opened 2023-09-12
Use ultra-node-pool for fv3gfs-run if memory exceeds cutoff

AnnaKwa:
This edits the `choose-node-pool` step to use the highmem-sim-pool (N2 highmem w/ 32 cores and 256 GB memory) if a larger amount of memory than available on climate-sim-pool (>30Gi) is requested.


brianhenn (approved):
Thanks @AnnaKwa. Approving with one required change and an optional one below. Probably also worth triggering the e2e test before merging just to be sure, since it can be hard to anticipate everything the workflows do.

PR #2332, opened 2023-09-15
Add wandb logging of validation skill to reservoir training

AnnaKwa:
This adds optional wandb logging of validation skill for reservoir training.

The code in `fv3fit.reservoir.validation` `_get_predictions_over_batch, validation_prediction` is mostly copy/pasted from its previous home in `workflows/diagnostics/fv3net/reservoir` so it can be used in both the training and offline diagnostics. This PR inserts some additional calculations for the baseline error that uses the 'imperfect' hybrid input if it exists in the data.

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/validation.py
+++ b/external/fv3fit/fv3fit/reservoir/validation.py
@@ -148,6 +148,8 @@ UNITS = {
 
 
 def log_rmse_z_plots(ds_val, variables):
+    # Note: spatial mean is not area-averaged. If reporting metrics in publication
+    # will need to change this.
     for var in variables:
         rmse = {}
         for comparison in ["persistence", "prediction", "imperfect_prediction"]:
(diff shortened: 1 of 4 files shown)


external/fv3fit/fv3fit/reservoir/validation.py
```diff
@@ -0,0 +1,204 @@
...
+}
+
+
+def log_rmse_z_plots(ds_val, variables):
+    for var in variables:
+        rmse = {}
+        for comparison in ["persistence", "prediction", "imperfect_prediction"]:
+            mse_key = f"time_mean_{comparison}_mse_{var}"
+            if mse_key in ds_val:
+                rmse[comparison] = np.sqrt(ds_val[mse_key].mean(["x", "y"])).values
```
frodre, line 156: Doesn't matter for comparison, but if we ever use these values for a publication we should make a note at least to switch to area-weighted spatial averages.

frodre (approved):
LGTM

PR #2333, opened 2023-09-19
Time shift hybrid data during preprocessing

AnnaKwa:
The hybrid reservoir model takes in the state at time t, updates its hidden state, and then uses its hidden state plus the 'imperfect' (i.e. coarse model FV3GFS) prediction at time **t+1** to predict the corrected state at time t+1. We [previously discussed](https://github.com/ai2cm/fv3net/pull/2288#discussion_r1270052043) how the hybrid inputs should be shifted by an index -1 to match up with the hidden states at time t. However, I did not merge the agreed upon solution (do the shifting in the preprocessing) so my latest results from hybrid models have the incorrect time index of hybrid inputs.

This PR corrects the issue by shifting hybrid inputs in the preprocessing step.
 
It also includes a fix to only reset the reservoir state at the start of each initial condition's time series. Previously it was resetting every batch. Previously trained models used a branch with this fix so they weren't affected.

### DIFF
--- a/projects/reservoir/fv3/save_ranks.py
+++ b/projects/reservoir/fv3/save_ranks.py
@@ -13,6 +13,11 @@ logging.basicConfig()
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+# Shift hybrid data by this number of timesteps.
+# -1 is common use case for saving the next timestep's ML prediction at
+# the same sample index as reservoir input.
+TIME_SHIFT_HYBRID_DATA = -1
+
 
 def _get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


projects/reservoir/fv3/save_ranks.py
```diff
@@ -13,6 +13,11 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+# Shift hybrid data by this number of timesteps.
+# -1 is common use case for saving the next timestep's ML prediction at
+# the same sample index as reservoir input.
+TIME_SHIFT_HYBRID_DATA = -1
```
frodre, line 19: Is there any benefit to adding this as a configuration option with a default as opposed to a hard-code?

frodre (approved):
LGTM.  Glad you found the bugs!

projects/reservoir/fv3/save_ranks.py
AnnaKwa: In practice we'll always have the t+1 hybrid input available at run time, so I'm not sure if we'd want to train models with different time shifts. It'd be a good reminder of what the code is doing though to have this in the args with a default of -1.

PR #2334, opened 2023-09-25
Prognostic run: check surface flux overrides are provided if flag set to true

AnnaKwa:
This adds a check in the prephysics step that the surface flux overrides are provided by the stepper if `override_surface_radiative_fluxes` is set to True in the namelist. We've accidentally run simulations with this flag on but without prescribing values, which results in zero fluxes throughout the run. 

I tested in local runs that this functions as intended.


brianhenn (approved):
Thanks!

PR #2337, opened 2023-10-03
Add sphum limiter and precipitation update to reservoir stepper

AnnaKwa:
This adds the positive specific humidity limiter and a precipitation update to the reservoir stepper. It also adds the associated diagnostics for column net heating/moistening from the reservoir step.

The existing functions for these computations already existed for the ML stepper but took in tendency inputs only. To minimize changes to the existing code I added a conversion of the reservoir's state prediction to the equivalent tendencies, which are now returned by the reservoir stepper instead of an empty dict. The tendencies are used in the reservoir step of the loop file to compute diagnostics.  

Significant internal changes:
- added `hydrostatic` and `mse_conserving_limiter` optional fields in the reservoir config. These are used to compute diagnostics in the same way as in the ML stepper, where they are also config fields.
- added `model_timestep` attribute to the reservoir stepper so it can compute the equivalent tendencies from the predicted and input state.
 
- [x ] Tests added

### DIFF
--- a/workflows/prognostic_c48_run/runtime/diagnostics/compute.py
+++ b/workflows/prognostic_c48_run/runtime/diagnostics/compute.py
@@ -280,3 +280,25 @@ def compute_baseline_diagnostics(state: State) -> Diagnostics:
             description="surface precipitation rate due to parameterized physics"
         ),
     )
+
+
+def tendencies_from_state_updates(
+    initial_state: State, updated_state: State, dt: float
+):
+    """Compute tendencies from state updates
+
+    Args:
+        initial_state: initial state
+        updated_state: updated state
+        variables: variables to compute tendencies for
+
+    Returns:
+        tendencies: tendencies computed from state updates
+    """
+    tendencies = {}
+    for variable in updated_state:
+        tendency_var = STATE_NAME_TO_TENDENCY[variable]
+        tendencies[tendency_var] = (
+            updated_state[variable] - initial_state[variable]
+        ) / dt
+    return tendencies
--- a/workflows/prognostic_c48_run/runtime/diagnostics/compute.py
+++ b/workflows/prognostic_c48_run/runtime/diagnostics/compute.py
@@ -245,12 +245,12 @@ def compute_ml_momentum_diagnostics(state: State, tendency: State) -> Diagnostic
     )
 
 
-def rename_diagnostics(diags: Diagnostics):
+def rename_diagnostics(diags: Diagnostics, label: str = "machine_learning"):
     """Postfix ML output names with _diagnostic and create zero-valued outputs in
     their stead. Function operates in place."""
     ml_tendencies = {
-        "net_moistening_due_to_machine_learning",
-        "net_heating_due_to_machine_learning",
+        f"net_moistening_due_to_{label}",
+        f"net_heating_due_to_{label}",
         "column_integrated_dQu",
         "column_integrated_dQv",
         "override_for_time_adjusted_total_sky_downward_shortwave_flux_at_surface",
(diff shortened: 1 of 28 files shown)


workflows/prognostic_c48_run/runtime/diagnostics/compute.py
```diff
@@ -24,8 +24,9 @@ def enforce_heating_and_moistening_tendency_constraints(
     timestep: float,
     hydrostatic: bool,
     mse_conserving: bool,
-    temperature_tendency_name="dQ1",
-    humidity_tendency_name="dQ2",
+    temperature_tendency_name: str = "dQ1",
+    humidity_tendency_name: str = "dQ2",
+    zero_fill_missing_tendencies: bool = False,
```
AnnaKwa, line 29: This is added as an optional arg to preserve prior behavior in the ML postphysics stepper where these diags aren't output if dQ1, dQ2 are not present. If this isn't important I can remove it and the zero-filled fields will always be in the diags.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -574,12 +578,43 @@ def _increment_reservoir(self) -> Diagnostics:
...
 
     def _apply_reservoir_update_to_state(self) -> Diagnostics:
-        # TODO: handle tendencies
+        # TODO: handle tendencies. Currently the returned tendencies
+        # are only used for diagnostics and are not used in updating state
         if self._reservoir_predict_stepper is not None:
-            [_, diags, state] = self._reservoir_predict_stepper(
-                self._state.time, self._state
+            [
+                tendencies_from_state_prediction,
```
AnnaKwa, line 585: The reservoir stepper now returns a non empty tendency dict (they are calculated using the state prediction). These are used to compute diagnostics, but are not used in the fv3 state update.

workflows/prognostic_c48_run/runtime/diagnostics/compute.py
frodre: 👍 I think it's best to preserve previous behavior when possible.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -574,12 +578,43 @@ def _increment_reservoir(self) -> Diagnostics:
...
+                diags,
+                state_updates,
+            ] = self._reservoir_predict_stepper(self._state.time, self._state)
+            (
```
frodre, line 589: The smallest of small nits,  consistent use of `[` vs `(` to group variable assignments.

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -574,12 +578,43 @@ def _increment_reservoir(self) -> Diagnostics:
...
+                stepper_diags,
+                net_moistening,
+            ) = self._reservoir_predict_stepper.get_diagnostics(
+                self._state, tendencies_from_state_prediction
+            )
+            diags.update(stepper_diags)
+            if self._reservoir_predict_stepper.diagnostic is True:  # type: ignore
```
frodre, line 596: nit: `if self._reservoir_predict_stepper.diagnostic:`

frodre: Could rename the underlying attribute to `is_diagnostic` to make it clearer that it's a bool

workflows/prognostic_c48_run/runtime/loop.py
```diff
@@ -574,12 +578,43 @@ def _increment_reservoir(self) -> Diagnostics:
...
+                rename_diagnostics(diags, label="reservoir_predictor")
+
+            state_updates[TOTAL_PRECIP] = precipitation_sum(
```
frodre, line 599: Just checking that it's okay to always provide `TOTAL_PRECIP` updates.  I'm guessing based on the signature that an update with 0-tendency net moistening returns the same precip from the state.

frodre (approved):
LGTM, just a couple very minor adjustments and one question.

workflows/prognostic_c48_run/runtime/loop.py
AnnaKwa: Correct.

PR #2340, opened 2023-10-03
Update scream workflow for latest changes

elynnwu:
Two minor updates to scream workflow:
1. wind components are now `U` and `V`
2. format scream output takes in chunk size when converting to zarr


frodre (approved):
LGTM

PR #2341, opened 2023-10-04
Fix reservoir renaming duplication

AnnaKwa:
While testing a hybrid reservoir model locally, I was running into an issue during the increment stepper's renaming of inputs where the hybrid input variables `[air_temperature_before_interval_update, ...]` and reservoir input variables `[air_temperature, ...]` loaded and then the renaming would fail because both sets would try rename to `[air_temperature_rc_in, ...]`.

The problem occurs because underlying model expects "air_temperature" for the reservoir input and "air_temperature_before_interval_update" for the hybrid input. The same field air_temperature is used for both cases but is renamed to "air_temperature_before_interval_update" before the model uses it as the hybrid input. However, the increment stepper loads both sets of input variables and does `rename_dataset_members(inputs, {k: f"{self.rename_mapping.get(k, k)}_rc_in" for k in inputs})` which results in a renaming mapping with collisions when creating some of its diagnostics e.g. ` {'specific_humidity_before_interval_update': 'specific_humidity_rc_in', 'air_temperature': 'air_temperature_rc_in', 'air_temperature_before_interval_update': 'air_temperature_rc_in', 'specific_humidity': 'specific_humidity_rc_in'}`. If we only load the nonhybrid/hybrid inputs that the respective increment/predict steppers need then we can avoid this name collision

This PR adjusts the increment stepper to only load the reservoir input variables and not the hybrid inputs. This necessitated adding an additional `nonhybrid_input_variables` attribute to the adapter.

I'm not sure why this error occurs in the current master but not the older image (4e893f5) I was using in my vcm-workflow-control experiments- maybe one of the components changed in the meantime.

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/adapters.py
+++ b/external/fv3fit/fv3fit/reservoir/adapters.py
@@ -103,6 +103,7 @@ class ReservoirDatasetAdapter(Predictor):
         self.model = model
         self.input_variables = model.input_variables
         self.output_variables = model.output_variables
+        self.nonhybrid_input_variables = model.input_variables
         self.model_adapter = DatasetAdapter(
             input_variables=self.input_variables,
             output_variables=self.output_variables,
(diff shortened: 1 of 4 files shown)


frodre (approved):
Thanks again for adding in the explanation.  Fix looks good to me.

PR #2344, opened 2023-10-09
Add input mask option for hybrid readout data

frodre:
In the reservoir models, we have options to mask the reservoir inputs using a specified mask (used for SST runs, currently) but the hybrid inputs had no such option.  This PR takes that same mask and also applies it to the readout training by multiplication with the post-encoded spatial gridpoints.  It is then saved as a field with the hybrid model which is loaded and applied to any hybrid input data for prediction.  This allows us to explore how much of the model behavior is tied to continental atmospheric information vs. the ocean areas where the model should actually experience the coupling.

- [ ] Tests added

### DIFF
--- a/external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
+++ b/external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
@@ -9,6 +9,7 @@ from fv3fit.reservoir import (
     ReservoirComputingModel,
     Reservoir,
     ReservoirHyperparameters,
+    HybridReservoirComputingModel,
 )
 
 from helpers import get_reservoir_computing_model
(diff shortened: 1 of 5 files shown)


external/fv3fit/tests/reservoir/test_reservoir_computing_model.py
```diff
@@ -9,6 +9,7 @@
     ReservoirComputingModel,
     Reservoir,
     ReservoirHyperparameters,
+    HybridReservoirComputingModel,
 )
```
AnnaKwa, line 13: Optional: you could add a test similar to the reservoir input mask [test](https://github.com/ai2cm/fv3net/blob/3556fa956f434b6ee89549354af9404472214c33/external/fv3fit/tests/reservoir/test_reservoir.py#L141) that changing the hybrid data in a masked region doesn't affect the final output.

AnnaKwa (approved):
LGTM! One test needs a small fix, one optional suggestion.

PR #2352, opened 2023-10-18
Include reservoir predictions in dQ1/2 diags

AnnaKwa:
The existing online diags only check for column heating/moistening due to the postphysics stepper. This adds the reservoir stepper label to a list of suffixes to check for when grabbing these diagnostics for the report.


frodre (approved):
LGTM.  Will PRATEsfc always be in the dataset?  Assuming it's just a standard fortran state output.

AnnaKwa:
> LGTM. Will PRATEsfc always be in the dataset? Assuming it's just a standard fortran state output.

Yeah, I guess if the fortran diagnostics are not configured to save `PRATEsfc` this breaks, but the existing version of this function already assumes that field is available so I just went with it.

PR #2355, opened 2023-10-25
Ignore nan in prognostic run diagnostic rms calculation

elynnwu:
Scream prognostic outputs often have nans to denote invalid data points (e.g., temperature at 850mb when surface pressure at that point is lower than 850mb). While fv3gfs run extrapolates those points, scream masks them. This causes rms calculation to report inf. This PR adds the capability of ignoring nan values.

Significant internal changes:
- `prognostic_run/compute.py`: `rms` now checks if nan exists in either inputs and excludes them from calculation

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
@@ -143,8 +143,20 @@ def _is_empty(data: Union[xr.Dataset, xr.DataArray]) -> bool:
         return False
 
 
+def ignore_nan_values(x, y, w):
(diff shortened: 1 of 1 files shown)


AnnaKwa (approved):

AnnaKwa:
LGTM. Does the rms function get used on non-time-averaged quantities for the SCREAM diags? I believe we only calculate RMS of already time averaged arrays for the FV3 online diagnostics. If that is also the case you could drop that line.

elynnwu:
It looks like `rms_errors` should have been time averaged based on the transform, but when I put a breakpoint inside for debugging, I'm still getting non-time-averaged fields. I'll test it again without the breakpoint.

PR #2356, opened 2023-10-30
Allow reservoir model keys to be string repr of integers

AnnaKwa:
JSON and YAML don't write integer keys. This is an issue when autogenerating configs for prognostic runs where the script fills in the rank models in the resevoir corrector field with integer keys but `json.dump` saves the keys as strings. This PR allows these keys to be string representations of integers, which are converted to integers after the config is loaded.


frodre (approved):
LGTM.  Might be worthwhile to add a simple test to `test_reservoir_stepper.py` that checks the error raising code since we're no longer in the dacite-only instantiation.

PR #2360, opened 2023-10-31
`IntervalStepper.n_calls` option 

AnnaKwa:
This adds the optional config arg n_calls to the IntervalStepper class. It can be used to prescribe the first N reservoir timesteps to a reference dataset during synchronization, after which the interval stepper will not be used to set the state.

It is up to the user to properly configure the interval stepper `n_calls` to be the same value as the synchronization steps for the reservoir.


frodre (approved):
Looks good!  Thanks for adding a test.  I'd add some description for the `n_calls` configuration at least.

PR #2361, opened 2023-11-01
Add 2d variance metrics to reservoir training

AnnaKwa:
This adds scalar metrics for the mean vertically-summed grid-scale variance of outputs in the x, y plane. Since the prognostic run reservoir predictions have issues with too much grid-scale noise in column-integrated quantities, I would like to see how the hyperparameters affect this in offline evaluation. I don't have area or pressure thicknesses saved in the data, so this is a very rough way of estimated the variance in column-integrated quantities.

During synchronization of the reservoir the `_rc_out` precipitable water field has a higher variance than the `_hyb_in` field, which suggests that this should be visible in offline evaluation.

image


frodre (approved):
LGTM

PR #2362, opened 2023-11-03
Prognostic run: refactor `main.py` to import wrapper depending on the config

spencerkclark:
This PR is an initial piece broken out of #2350.  It adds a `wrapper` attribute to the `UserConfig` class that is used to determine which Python wrapper to import within the `main.py` run file.  As of this PR, the only supported option is the default `"fv3gfs.wrapper"`, but in the abstract this adds a hook for us to ultimately also support `"shield.wrapper"`.

Added public API:
- A `wrapper` can now be specified within a prognostic run `UserConfig`.  The only currently supported option is the default of `"fv3gfs.wrapper"`.  Because of this default, this PR is fully backwards compatible (no need to change already-defined configs).

Significant internal changes:
- The Python wrapper is now imported within the `main` function defined in `main.py` within the prognostic run instead of the top of the module.
- The wrapper is initialized within the `main` function, and thus after all top-level imports.  Initial tests (including with a 2x2 layout in the cloud) suggest the odd issue that motivated the previous approach may be resolved.  Perhaps this is related to the fact that we no longer use OpenMPI in the prognostic run image (we use MPICH).  This logic had originally been added in https://github.com/ai2cm/fv3net/pull/706.
- The garbage collection step added in https://github.com/ai2cm/fv3net/pull/1440 has also been removed, since it appears it is no longer needed.

- [x] Tests added


workflows/prognostic_c48_run/runtime/config.py
```diff
@@ -123,3 +127,15 @@ def get_model_urls(config_dict: dict) -> List[str]:
         for entry in prephysics_config:
             urls += entry.get("model", [])
     return urls
+
+
+def get_wrapper(config: UserConfig):
+    """Import the appropriate wrapper based on the config"""
+    module = f"{config.fortran_model}.wrapper"
```
oliverwm1, line 134: Are we guaranteed these modules will always have the format `{fortran_model}.wrapper`? I guess realistically we have control here and it's not that likely we'll have more than two.

workflows/prognostic_c48_run/runtime/main.py
```diff
@@ -72,11 +70,12 @@ def main():
         diag_file.flush()
 
     loop.log_global_timings()
+    return wrapper
 
 
 if __name__ == "__main__":
 
-    main()
+    wrapper = main()
```
oliverwm1, line 78: Feels a bit odd to return a module in this way, but not necessarily an issue. Do you know if the cleanup is still necessary?

oliverwm1 (commented):
Once `fortran_model='shield'` is available as an option, do you envision always maintaining both options?

Also, does a user have to ensure that the `fortran_model` that they select is consistent with (for example) the docker image they are using? Would we have a docker image with both Fortran models installed so that they could actually switch between them?

spencerkclark (commented):
Thanks for taking a look @oliverwm1 and for these high-level design questions.

> Once `fortran_model='shield'` is available as an option, do you envision always maintaining both options?

"Always" is a strong word, but in the near future, yes, I think we will need to maintain both options to support reservoir computing work.

> Also, does a user have to ensure that the `fortran_model` that they select is consistent with (for example) the docker image they are using? Would we have a docker image with both Fortran models installed so that they could actually switch between them?

My approach so far has been to build a separate docker image for each fortran model and have logic in the prognostic run argo workflow decide which image to use based on the config (i.e. the user would not need to specify anything more than they do now; the argo workflow can figure out the image to use).  I have something that works in #2350, though I want to iterate on it a little more.

I have been hesitant to try building an image with both fortran models installed, since there are overlapping dependencies (multiple versions of NCEPlibs and FMS), and [we do things like add FMS to the `LD_LIBRARY_PATH` in the FV3GFS prognostic run image](https://github.com/ai2cm/fv3net/blob/0de31080f1680d1bd26ac2310d98448c74c6c026/docker/prognostic_run/base.Dockerfile#L61).  Maybe it's possible to manage these dependencies without conflict, but it feels a little tricky.  If there is a strong argument for this versus building two images, however, I'm happy to think about it more (the main disadvantages of two images I can think of are larger total space required in the container registry, and more complexity in argo).

workflows/prognostic_c48_run/runtime/main.py
spencerkclark: Locally it seems like the garbage collecting step is no longer required for the regression tests.  I pushed an update with these changes and triggered the integration tests to see if they cause any issues there.

workflows/prognostic_c48_run/runtime/config.py
spencerkclark: Switched to this approach in d510c3b.

workflows/prognostic_c48_run/runtime/main.py
oliverwm1: Okay sounds good, thanks.

oliverwm1:
> "Always" is a strong word, but in the near future, yes, I think we will need to maintain both options to support reservoir computing work.

Okay sounds good.

> My approach so far has been to build a separate docker image for each fortran model and have logic in the prognostic run argo workflow decide which image to use based on the config (i.e. the user would not need to specify anything more than they do now; the argo workflow can figure out the image to use). I have something that works in https://github.com/ai2cm/fv3net/pull/2350, though I want to iterate on it a little more.

I see. That seems like a good approach. Definitely would be helpful to not require the user to have to specify the fortran model in two places.

oliverwm1 (approved):
LGTM

workflows/prognostic_c48_run/runtime/main.py
oliverwm1: We should keep an eye on this change, maybe issues will crop up for longer runs.

spencerkclark: Sounds good.  I just submitted a 10-day nudged run off the latest commit in this branch, which is maybe a little bit more of a stress test.  Will merge if that completes successfully.

spencerkclark: The 10-day nudged run seems to have gone off without a hitch: 

```
$ podlogs nudged-fv3gfs-6603-2081873082
```

I am going to go ahead and merge.

spencerkclark:
Thanks @oliverwm1!

PR #2365, opened 2023-11-06
Add Dockerfile for `prognostic_run_shield` image

spencerkclark:
This PR adds a Dockerfile for building a `prognostic_run_shield` image, analogous to the `prognostic_run` image, but with SHiELD-wrapper instead of fv3gfs-wrapper installed.  There are a few differences from the `prognostic_run` image:

- GNU 10 compilers are used instead of GNU 9, since latest FMS / SHiELD require GNU 10+ compilers to build.
- SHiELD-wrapper (and its dependencies) are installed instead of fv3gfs-wrapper.  In terms of compiled dependencies, this happens under the hood within the `COMPILE` script in the SHiELD_build repo.
- call_py_fort is omitted, since we have not made an effort to port the microphysics-emulation-specific call_py_fort hooks into SHiELD, which are somewhat orthogonal to the Python wrapper.

To make things easier to review, for now I have just skipped any tests that require running prognostic simulations.  A preview of the changes required for doing that with SHiELD-wrapper can be found in #2350 (nothing major within the time loop; most changes are needed to parametrize / configure tests).  In a followup PR I will add the changes needed to actually test `shield.wrapper` simulations in the same way we test `fv3gfs.wrapper`, which will depend on this PR as well as the changes in #2362.

Added public API:
- Added a `prognostic_run_shield` image, which will eventually be used for prognostic runs using SHiELD-wrapper.

Significant internal changes:
- Added logic for skipping specific regression tests if `fv3gfs.wrapper` is not installed in the image.
- Added an `authenticate` call to the `test_read_last_segment_gcs` test such that it can be run without relying on previous regression tests being run before it; thanks @frodre for help debugging this.

Requirement changes:
- Added a SHiELD-wrapper submodule, pointing to latest main.

### DIFF
--- a/workflows/prognostic_c48_run/tests/segmented_run/test_append.py
+++ b/workflows/prognostic_c48_run/tests/segmented_run/test_append.py
@@ -1,6 +1,7 @@
 import sys
 from pathlib import Path
 import subprocess
+from fv3post.gsutil import authenticate
 from runtime.segmented_run.append import read_last_segment
 from runtime.segmented_run.run import compose_simulation_command, runfile
 from vcm.cloud import get_fs
@@ -26,6 +27,12 @@ def test_read_last_segment_gcs(tmp_path: Path):
     # (https://cloud.google.com/storage/docs/consistency)
     # and also any integrations with fsspec which has several layers of internal
     # caching which can break this.
+
+    # Note that we must ensure that we are authenticated
+    # with gcloud, which can be accomplished by calling the authenticate
+    # function from fv3post.gsutil.
+    authenticate()
+
     id_ = uuid.uuid4().hex
     url = "gs://vcm-ml-scratch/test_data" + "/" + id_
     last_segment = None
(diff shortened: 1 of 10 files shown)


workflows/prognostic_c48_run/tests/segmented_run/test_append.py
```diff
@@ -21,11 +22,14 @@ def test_read_last_segment(tmpdir):
...
-    # and also any integrations with fsspec which has several layers of internal
-    # caching which can break this.
+    # This test requires network access, but so do some others This tests the
+    # consistency of GCS in the sense of "reading our own writes"
+    # (https://cloud.google.com/storage/docs/consistency) and also any
+    # integrations with fsspec which has several layers of internal caching
+    # which can break this.  Note that we must ensure that we are authenticated
+    # with gcloud, which can be accomplished by calling the authenticate
+    # function from fv3post.gsutil.
+    authenticate()
```
spencerkclark, line 32: Previously we were relying on `authenticate` being called within the post-processing step of regression tests, which are skipped for the time being in this PR.  If `authenticate` is not called then this test fails, so we ensure this never happens by including an `authenticate` call within the test itself.

AnnaKwa (approved):
LGTM

PR #2366, opened 2023-11-08
Use encoder/decoder call directly

AnnaKwa:
Making a separate PR using @frodre 's commit that greatly speeds up training.
Directly calling the encoder vs `.predict` reduces batch fitting time by >2-10x.


frodre (approved):
Awesome! Glad it sped up your training too!

PR #2369, opened 2023-11-13
Add `--catalog_path` argument to `prognostic_run_diags shell`

### LINKED ISSUE #2368: `prognostic_run_diags` tries to use network to load grid data
By way of the following globally-scoped code within the definition of the `shell` subcommand, calling any subcommand under `prognostic_run_diags` causes network / authentication access to try and load grid data from our default `vcm` data catalog (even if we are not trying to use `shell`):

https://github.com/ai2cm/fv3net/blob/cf4706703531eaa2f5b6a222547609820d6b9f7d/workflows/diagnostics/fv3net/diagnostics/prognostic_run/shell.py#L111-L113

This is problematic on machines that do not have network access, e.g. nodes on HPC systems.  As we've done in this context before, it is best to enable providing a path to a local custom catalog, which has entries that point to local rather than remote data.

Here the natural solution appears to be to add an optional `catalog_path` argument to the `shell` subcommand and `State` class constructor.  When initializing the `State` we could then load the grid (once we know the catalog), rather than load the grid when initializing the module (making an assumption about the catalog).

spencerkclark:
This PR removes automatic network / authentication access when using the `prognostic_run_diags` command line tool.

Added public API:
- Added an optional `--catalog_path` argument to the `prognostic_run_diags shell` interface to enable providing an alternative to the default `vcm` cloud-based data catalog for grid and land mask data.

Significant internal changes:
- The catalog and grid are no longer loaded globally within the `shell.py` module within the prognostic run diagnostics codebase.  Instead they are loaded within the `State` constructor, which is after we know which catalog to use.

- [x] Tests added

Resolves #2368

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/shell.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/shell.py
@@ -65,16 +65,18 @@ class ItermTape:
 
 
 class State:
-    def __init__(self):
+    def __init__(self, catalog_path=vcm.catalog.catalog_path):
         self.data_3d = None
         self.data_2d = None
         self.tape = OneFileTape()
         self.state = {}
+        self.catalog = intake.open_catalog(catalog_path)
+        self.grid = load_run_data.load_grid(self.catalog)
 
     def load(self, url):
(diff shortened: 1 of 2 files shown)


elynnwu (approved):
lgtm

PR #2371, opened 2023-11-13
Make `prognostic_run_diags` compatible with scream dataset

elynnwu:
For convenience, we infer whether scream data is used based on whether `ncol` is in the dataset. 

(Delete unused sections)
Added public API:
- `vcm.check_if_scream_dataset`: infer if scream data is used
- Added `vcm.scream.standardize_scream_diagnostics`

Refactored public API:
- `prognostic_run_diags`: takes in `gsrm`, defaults to `fv3gfs`
- `load_run_data.py`: added custom code for loading scream data 

- [ ] Tests added

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
+++ b/workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
@@ -1,6 +1,6 @@
 import xarray as xr
-from typing import Optional, List
-from dataclasses import dataclass, field
+from typing import Optional
+from dataclasses import dataclass
 import numpy as np
 
 HORIZONTAL_DIMS_FV3 = ["x", "y", "tile"]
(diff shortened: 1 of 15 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -101,7 +102,7 @@ def _merge_diag_computes(
             )
             continue
 
-        diag_arg = DiagArg(prog, verif, grid)
+        diag_arg = DiagArg(prog, verif, grid, gsrm=gsrm)
```
AnnaKwa, line 105: Is it possible for the `DiagArg` class to keep its old signature (without gsrm arg) and just figure out what its horizontal_dims attribute should be based off the grid or prediction datasets provided? I see there's already a `check_if_scream_dataset` function added for convenience.

AnnaKwa (approved):
Overall LGTM! Minor comments
1) `model`  would be technically more correct term to distinguish the scream/fv3gfs options, though that word is kind of overloaded
2) Are there edge cases where `DiagArg` cannot figure out the horizontal dims from the datasets provided as args? If not, I'd just use those to figure out the dims instead of the additional str arg.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
elynnwu: Good point, we can actually keep the old signature and not have `gsrm`or `model` floating around at all.

workflows/diagnostics/fv3net/diagnostics/_shared/constants.py
```diff
@@ -29,5 +29,15 @@ class DiagArg:
     verification: xr.Dataset
     grid: xr.Dataset
     delp: Optional[xr.DataArray] = None
-    horizontal_dims: List[str] = field(default_factory=lambda: HORIZONTAL_DIMS_FV3)
+    gsrm: str = "fv3gfs"
```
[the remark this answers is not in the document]
elynnwu: Keeping the original signature, no gsrm or model.

PR #2372, opened 2023-11-13
Make report plotting compatible with scream data

elynnwu:
Prognostic run reporting can now plot scream format data. Occasionally the long name and units are missing, we revert back to the variable name in that case.

Refactored public API:
- `plot_cubed_sphere_map`: takes in gsrm type and plot scream formatted data

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
@@ -123,7 +123,6 @@ def plot_cubed_sphere_map(
     run_metrics: RunMetrics,
     varfilter: str,
     metrics_for_title: Mapping[str, str] = None,
-    gsrm: str = "fv3gfs",
 ) -> str:
     """Plot horizontal maps of cubed-sphere data for diagnostics which match varfilter.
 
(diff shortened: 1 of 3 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/static_report.py
```diff
@@ -725,16 +761,16 @@ def register_parser(subparsers):
 
 def main(args):
     computed_diagnostics = ComputedDiagnosticsList.from_directory(args.input)
-    make_report(computed_diagnostics, args.output)
+    make_report(computed_diagnostics, args.output, args.gsrm)
```
brianhenn, line 764: Overall this looks good, but I am wondering if it would be preferable to avoid adding an API-level argument (`args.gsrm`), in favor of trying to infer the needed grid variables from the data at the lower level where it's actually used, for example in `plot_cubed_sphere_maps` by wrapping the `ds = run_diags.get_variables(run, COORD_VARS + [varname])`  with a try-except or something.

elynnwu: Good idea, I think we can actually just check if `ncol` is in the dataset to switch to scream specific plotting. Will make the change.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
```diff
@@ -148,14 +148,30 @@ def plot_cubed_sphere_map(
...
+                COORD_VARS = ["lon", "lat"]
             ds = run_diags.get_variables(run, COORD_VARS + [varname])
             plot_title = _render_map_title(
                 run_metrics, shortname, run, metrics_for_title
             )
             fig, ax = plt.subplots(
                 figsize=(6, 3), subplot_kw={"projection": ccrs.Robinson()}
             )
-            fv3viz.plot_cube(ds, varname, ax=ax, vmin=vmin, vmax=vmax, cmap=cmap)
+            if "ncol" in ds.sizes:
```
brianhenn, line 162: :+1:

brianhenn (approved):
LGTM

PR #2374, opened 2023-11-17
Add `google-cloud-sdk` to bare-metal prognostic run environments

spencerkclark:
It occurred to me that not everyone will already have `google-cloud-sdk` (and namely `gsutil`) installed in their search `PATH` on Stellar or Gaea, so this adds `google-cloud-sdk` to the conda environment created in the `install_base_software.sh` script.  `gsutil` is used within the post-processing step of the prognostic run, and without it one will encounter errors.  No authentication is needed if just moving files around locally (as is the case when running on HPC).  

Note we similarly install `google-cloud-sdk` in the prognostic run docker image, so this is not anything particularly new:

https://github.com/ai2cm/fv3net/blob/61f3a55ddd707d7093149dea7b28c266ed774b76/.environment-scripts/gnu_docker/install_base_software.sh#L50


elynnwu (approved):
lgtm

PR #2376, opened 2023-11-30
Add SHiELD-wrapper regression tests

spencerkclark:
This PR builds on #2365 and makes the changes necessary to add `predictor` and `nudging` regression tests for SHiELD-wrapper-based prognostic runs.  In so doing it refactors `test_regression.py` to parametrize over the wrapper type, and move the base fortran configs into YAML files in their own subdirectory.

Note that unlike in #2350 no changes are required to the time loop of the prognostic run, thanks to ai2cm/SHiELD-wrapper#18, which has now been merged upstream.  I updated the SHiELD-wrapper submodule accordingly in this PR.

Significant internal changes:
- Added a `wrapper` parameter to `file_configs_to_namelist_settings`, since one of the namelist parameters used to control the output frequency of physics diagnostics in SHiELD is different than that in FV3GFS.
- Moved the base fortran config for the FV3GFS prognostic run regression tests into a YAML file in a subdirectory alongside the base fortran config for the SHiELD prognostic run regression tests.

- [x] Tests added


AnnaKwa (approved):
LGTM

PR #2378, opened 2023-12-04
Add capability to specify date range in `prognostic_run_diags save`

elynnwu:
Sometimes prognostic run is incomplete and we will still like to look at error statistics for a specific time frame. This PR adds the capability of specifying the start and end date when saving prognostic run diagnostics.

Added public API:
- `prognostic_run_diags save`: additional flags `--start-date` and `--end-date` 

Significant internal changes:
- When start/end date is specified, we slice the loaded data.

- [x] Tests added

### DIFF
--- a/workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
+++ b/workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
@@ -615,6 +615,15 @@ def add_catalog_and_verification_arguments(parser: ArgumentParser):
     )
 
 
+def add_time_range_arguments(parser: ArgumentParser):
+    parser.add_argument(
+        "--start-date", type=str, help="Start date for diagnostics", default=""
+    )
+    parser.add_argument(
+        "--end-date", type=str, help="End date for diagnostics", default=""
+    )
+
+
 def register_parser(subparsers):
     parser: ArgumentParser = subparsers.add_parser(
         "save", help="Compute the prognostic run diags."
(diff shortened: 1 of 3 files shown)


workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -615,13 +615,23 @@ def add_catalog_and_verification_arguments(parser: ArgumentParser):
     )
 
 
+def add_time_range_arguments(parser: ArgumentParser):
+    parser.add_argument(
+        "--start-date", type=str, help="Start date for diagnostics", default=""
+    )
+    parser.add_argument(
+        "--end-date", type=str, help="End date for diagnostics", default=""
```
spencerkclark, line 623: suggestion: `None` happens to be a valid input as a `start` and/or `end` to `slice`, which means no bound (e.g. `slice(None, None)` returns the full data).  Using it as the default value for `--start-date` and `--end-date` would remove the need for the conditional logic in `evaluation_pair_to_input_data`, and would allow for the case of setting just one of the dates (e.g. maybe one only needs to set an `--end-date` bound).  

```suggestion
        "--start-date", type=str, help="Start date for diagnostics", default=None
    )
    parser.add_argument(
        "--end-date", type=str, help="End date for diagnostics", default=None
```

spencerkclark (commented):
Nice!  One minor suggestion, but otherwise this looks good to me.

spencerkclark (approved):

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
elynnwu: Great suggestion, thank you!

PR #2382, opened 2023-12-09
Add transformer for normalizing individual spatial grid points

frodre:
Add a new encoder `_ScaleSpatialConcatZTransformer` built using `build_scale_spatial_concat_z_transformer` to perform a per grid-point normalization.  This matches the normalization used by Acomano et al. in their coupled atmosphere--ocean hybrid reservoir modeling.

Also adds a workaround for the batch processing in the training to account for the fact that an encoder may expect data with or without a halo.  Previously we assumed that when a halo was provided in the data to process and the output should be trimmed, the encoder would accept halo data.  We should probably standardize this.

Added public API:
- `build_scale_spatial_concat_z_transformer` - a function to create a transformer object that normalizes per grid point and combines along the z features.  (Currently requires all z features to be equivalent across included variables)

Refactored public API:
- `encode_columns` and `decode_columns` removed in favor of adding `encode/decode_txyz` methods to the base `Transformer` class.

Significant internal changes:
- Added a workaround for transformer use in processing tfdataset batch data that may or may not expect data with a halo

- [x] Tests added

### DIFF
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -94,7 +94,9 @@ class HybridReservoirComputingModel(Predictor):
     def predict(self, hybrid_input: Sequence[np.ndarray]):
         # hybrid input is assumed to be in original spatial xy dims
         # (x, y, feature) and does not include overlaps.
-        encoded_hybrid_input = self.transformers.hybrid.encode_txyz(hybrid_input)
+        encoded_hybrid_input = self.transformers.hybrid.encode_unstacked_xyz(
+            hybrid_input
+        )
 
         flat_hybrid_in = self._hybrid_rank_divider.get_all_subdomains_with_flat_feature(
             encoded_hybrid_input
@@ -111,7 +113,7 @@ class HybridReservoirComputingModel(Predictor):
         prediction = self._output_rank_divider.merge_all_flat_feature_subdomains(
             flat_prediction
         )
-        decoded_prediction = self.transformers.output.decode_txyz(prediction)
+        decoded_prediction = self.transformers.output.decode_unstacked_xyz(prediction)
         return decoded_prediction
 
     def _concatenate_readout_inputs(self, hidden_state_input, flat_hybrid_input):
@@ -231,7 +233,7 @@ class ReservoirComputingModel(Predictor):
         prediction = self._output_rank_divider.merge_all_flat_feature_subdomains(
             flat_prediction
         )
-        decoded_prediction = self.transformers.output.decode_txyz(prediction)
+        decoded_prediction = self.transformers.output.decode_unstacked_xyz(prediction)
         return decoded_prediction
 
     def reset_state(self):
@@ -243,7 +245,7 @@ class ReservoirComputingModel(Predictor):
 
     def increment_state(self, prediction_with_overlap: Sequence[np.ndarray]) -> None:
         # input array is in native x, y, z_feature coordinates
-        encoded_xy_input_arrs = self.transformers.input.encode_txyz(
+        encoded_xy_input_arrs = self.transformers.input.encode_unstacked_xyz(
             prediction_with_overlap
         )
         encoded_flat_sub = self.rank_divider.get_all_subdomains_with_flat_feature(
@@ -253,7 +255,7 @@ class ReservoirComputingModel(Predictor):
 
     def synchronize(self, synchronization_time_series):
         # input arrays in native x, y, z_feature coordinates
-        encoded_timeseries = self.transformers.input.encode_txyz(
+        encoded_timeseries = self.transformers.input.encode_unstacked_xyz(
             synchronization_time_series
         )
         encoded_flat = self.rank_divider.get_all_subdomains_with_flat_feature(
--- a/external/fv3fit/fv3fit/reservoir/model.py
+++ b/external/fv3fit/fv3fit/reservoir/model.py
@@ -12,7 +12,7 @@ from .reservoir import Reservoir
 from .domain2 import RankXYDivider
 from fv3fit._shared import io
 from .utils import square_even_terms
-from .transformers import encode_columns, decode_columns, TransformerGroup
+from .transformers import TransformerGroup
 
 DIMENSION_ORDER = ("x", "y")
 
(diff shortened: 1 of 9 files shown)


external/fv3fit/fv3fit/reservoir/utils.py
```diff
@@ -107,18 +105,41 @@ def process_batch_data(
...
+        encoder_inputs = pre_trimmed_data
+
     if autoencoder is not None:
-        data_encoded = encode_columns(data, autoencoder)
+        try:
+            data_encoded = autoencoder.encode_txyz(encoder_inputs)
+        except ValueError as e:
+            # TODO: there is a chicken/egg problem here in that no
+            # specification of transforms creates an autoencoder that
+            # expects halo during training, while pre-trained encoder
```
AnnaKwa, line 123: Oh, that's annoying... have you been loading any existing transforms in your training? Since I switched to just using a normalization without vertical dimensionality reduction I haven't used that feature. If you haven't been loading transforms either the follow-on PR would be pretty low priority.

external/fv3fit/tests/reservoir/test_transformer.py
```diff
@@ -23,37 +23,134 @@ def test_DoNothingAutoencoder(nz, nvars):
...
+def _get_sample_xyz_data(nx, ny, nz):
+    xyz_base = np.ones((1, nx, ny, nz))
+    a = np.concatenate([-10 * xyz_base, 10 * xyz_base], axis=0)  # mean = 0, std = 10
+    b = np.concatenate([xyz_base, 2 * xyz_base], axis=0)  # mean = 1, std = 1
+    return [a, b]
+
+
+# test encode w/ specified arrays
+def test_scale_per_feature_concat_z_transform():
+    nx, ny, nz = 3, 4, 1
```
AnnaKwa, line 66: Could you add a test case with z>1?

external/fv3fit/tests/reservoir/test_transformer.py
```diff
@@ -23,37 +23,134 @@ def test_DoNothingAutoencoder(nz, nvars):
...
+
+    # test that it normalizes
+    nsamples, nvars = 2, 2
+    assert encoded.shape == (nsamples, 3, 4, 1 * nvars)
+    np.testing.assert_allclose(encoded, normalized_and_stacked, rtol=1e-6)
+
+    # test round trip
+    decoded = transformer.decode_txyz(encoded)
+    np.testing.assert_allclose(decoded, [a, b], rtol=1e-6)
+
```
AnnaKwa, line 84: Just to cover bases, could you add a test for the new transformer decoding an encoded input that has no time dim? I think this is how it is used in the prognostic run.

AnnaKwa (commented):
Looks mostly good to me! Just a few minor request for additional tests. The workaround for whether the encoder in training expects trimmed/untrimmed data looks fine to me.

external/fv3fit/fv3fit/reservoir/utils.py
frodre: Yes, I've been using the trained transformer in this PR.  I trained the hybrid SST encoder without any halo, and ran into this issue.  But I agree the follow-on is low priority given that this works

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
```diff
@@ -32,7 +38,250 @@ def decode(self, x: ArrayLike) -> Sequence[ArrayLike]:
...
+
+    def encode_txyz(self, input_arrs: Sequence[ndarray]) -> ndarray:
+        # Overloads the original transformer encode_txyz since we want to
+        # handle the xyz dimensions differently.
+
+        self._check_consistent_xyz(input_arrs)
+
+        leading_dims = input_arrs[0].shape[:-3]
+        # stack xyz
+        spatial_last_dim = [tf.reshape(arr, (*leading_dims, -1)) for arr in input_arrs]
```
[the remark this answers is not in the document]
frodre: Since this is already a fairly singularly focused class, I'm hesitant to write out too many subunits for small tasks that wont be generalized.  I agree about the non-uniform arrays not being informative of correctness.  Instead, to verify that everything is encoded/decoded properly, I adjusted the sample arrays to use `np.arange`.  Each xyz element will be unique in this case, so we can't pass tests without correct reshaping.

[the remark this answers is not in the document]
frodre: See d2d64696f

external/fv3fit/tests/reservoir/test_transformer.py
frodre: Added

external/fv3fit/tests/reservoir/test_transformer.py
frodre: Added

external/fv3fit/fv3fit/reservoir/transformers/transformer.py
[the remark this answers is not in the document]
frodre: See [d2d6469](https://github.com/ai2cm/fv3net/pull/2382/commits/d2d64696f520d2d95dd34f78483735c2bb60a354)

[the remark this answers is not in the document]
frodre: Thanks for this suggestion.  I adjusted all the shared reshaping into helper methods (some static) and adjusted the test to use non-uniform data.  It now tests the encoding shape, that each spatial point normalizes to +/- 1 for the forward encoding, the round trip, and that we recover the correct shape and data for a non-uniform decode input.

external/fv3fit/fv3fit/reservoir/model.py
```diff
@@ -113,9 +111,7 @@ def predict(self, hybrid_input: Sequence[np.ndarray]):
         prediction = self._output_rank_divider.merge_all_flat_feature_subdomains(
             flat_prediction
         )
-        decoded_prediction = decode_columns(
-            encoded_output=prediction, transformer=self.transformers.output,
-        )
+        decoded_prediction = self.transformers.output.decode_txyz(prediction)
```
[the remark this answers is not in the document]
frodre: I adjusted the names to `_unstacked_xyz`

AnnaKwa (approved):

PR #2383, opened 2023-12-09
Reservoir SST local training, sweep, and wandb validation

frodre:
Add validation specific to the SST reservoir training along with wanb logging and the capability to update configuration via wandb (for sweeps).  Also add sweep initialization, job submission, and configuration scripts.

Example [wandb sweep](https://wandb.ai/ai2cm/sst-reservoir-tuning/sweeps/wqjswb1a)
Example [wandb training validation](https://wandb.ai/ai2cm/sst-reservoir-training/groups/20231206-v1/workspace?workspace=user-aperkins)

Refactored public API:
- `ReservoirTrainingConfig` now takes two more flags: 
  - `mask_readout`: option to mask readout alongside reservoir input if `mask_field` is present in the data
  - `validate_sst_only`: enables the SST-specific wandb validation instead of the validation used for the atmospheric reservoir
- `Transformer` paths in the config can now be remote paths
- `fv3fit.dump` can now dump to remote paths even when subcomponents are not remote compatible

Significant internal changes:
- Adds a `set state` method to the reservoir for usage in testing/notebooks

- [ ] Tests added


external/fv3fit/fv3fit/reservoir/config.py
```diff
@@ -91,6 +91,8 @@ class ReservoirTrainingConfig(Hyperparameters):
     mask_variable: if specified, save mask array that is multiplied to the input array
         before multiplication with W_in. This applies a mask using the
         mask_variable field.
+    mask_readout: if mask_variable is specified, apply that mask to the hybrid inputs
```
frodre, line 94: Add `validate_sst_only`

external/fv3fit/fv3fit/reservoir/domain2.py
```diff
@@ -54,10 +54,14 @@ def __init__(
         if overlap < 0:
             raise ValueError("Overlap must be non-negative")
 
+        self._rank_dims = ["x", "y"]
```
frodre, line 57: These were hardcoded since everything is set up to transpose to this form.  Needed since the partitioner expects `[y, x]` shape.  Maybe in the long-run it would be easier to switch to `[y, x]` since that's the default in the wrapper runtime.

external/fv3fit/fv3fit/reservoir/train.py
```diff
@@ -83,10 +89,23 @@ def _get_transformers(
     return TransformerGroup(**transformers)
 
 
+def _expand_mask_zdim(mask: tf.Tensor, z_dim_len: int) -> tf.Tensor:
```
frodre, line 92: Check if still used.  Remove if not.

external/fv3fit/fv3fit/reservoir/utils.py
```diff
@@ -121,6 +126,30 @@ def process_batch_data(
         return rank_divider.get_all_subdomains_with_flat_feature(data_trimmed)
 
 
+def process_validation_batch_data_to_dataset(
```
frodre, line 129: Reason for a new tfdataset batch processing function: validation using an adapter model needs a xarray dataset as input and there is no need to convert to latent space.

external/fv3fit/fv3fit/reservoir/validation.py
```diff
@@ -251,3 +389,138 @@ def log_rmse_scalar_metrics(ds_val, variables):
...
+    model: ReservoirAdapterType,
+    reservoir_inputs: xr.Dataset,
+    hybrid_inputs: Optional[xr.Dataset],
+    n_sync_steps: int,
+    targets: xr.Dataset,
+    mask=None,
+    area=None,
+):
+
+    # want to do the index handling in this function
```
frodre, line 404: Adjust comment

projects/reservoir/train/train_local.sh
```diff
@@ -0,0 +1,42 @@
...
+# Loop through each tile and submit the specified number of jobs
+for tile in {0..0}; do
+  # Create a temporary directory for the updated configuration files
+  export TILE=$tile
+
+  tmpdir=$(mktemp -d)
+  envsubst < $training_data > $tmpdir/$training_data
+  envsubst < $validation_data > $tmpdir/$validation_data
+  envsubst < $train_config > $tmpdir/$train_config
+  # cp $train_config $tmpdir/$train_config
```
frodre, line 27: Remove

external/fv3fit/fv3fit/reservoir/domain2.py
AnnaKwa: I ran into this issue with the wrapper dim order being different, I think currently this is handled by  transposing to the wrapper's order  in the adapter.

external/fv3fit/fv3fit/reservoir/validation.py
```diff
@@ -1,9 +1,147 @@
...
+    else:
+        raise ValueError("src_len must be <= dst_len")
+
+    return sl
+
+
+def _insert_tile_to_overlap(current: xr.DataArray, previous: xr.DataArray):
+    # we can't grab the halos for offline rollouts because there is no prediction
+    # for other tiles.  Instead just grab the original overlap, which is *very*
+    # optimistic for forecasting...
```
AnnaKwa, line 55: Is this a large enough concern that we should just return constant values for rollout error if overlap>0?

AnnaKwa (commented):
LGTM, just some minor comments. It might make future code tidying easier in the future if the SST-specific validation code lives in its own module for now, since that will probably be the default use going forward.

external/fv3fit/fv3fit/reservoir/validation.py
frodre: I don't think it's a super large concern.  I usually am not comparing inter-experiment offline skill to a fine degree since the spread on similar hyperparameters is pretty wide.  I just check that the rollouts aren't blowing up or in the 75th percentile of RMSE.

external/fv3fit/fv3fit/reservoir/validation.py
```diff
@@ -1,9 +1,147 @@
...
+from fv3fit.reservoir.adapters import ReservoirAdapterType
+
+
+UNITS = {
+    "sst": "K",
+}
+
+logger = logging.getLogger(__name__)
+
+
```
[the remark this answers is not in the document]
frodre: Moved to a new file.

external/fv3fit/fv3fit/reservoir/train.py
```diff
@@ -275,25 +293,65 @@ def train_reservoir_model(
...
+            if adapter_model.is_hybrid:
+                adapter_model = cast(HybridReservoirDatasetAdapter, adapter_model)
+                hybrid_data = process_validation_batch_data_to_dataset(
+                    data, adapter_model.hybrid_variables, trim_divider=rank_divider
+                )
+            else:
+                hybrid_data = None
+
+            output_vars = list(adapter_model.output_variables)
+            if "mask_field" in data:
```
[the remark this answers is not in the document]
frodre: Good catch.  I was using this because setting the hyperparameters configuration forced masking of reservoir inputs.  I switched to applying the mask in the transforms so I could be more particular about which fields are masked, even if they are all included in the same encoding.  I still wanted the validation output masked and this was a mechanism for that.  

I switched the implementation up so that we have to specify boolean flags to mask either/both of the reservoir and hybrid inputs.  Setting the `mask_variable` will now apply to the validation when specified.

AnnaKwa (approved):
Thanks for the updates, LGTM.

PR #2377, opened 2023-12-02
Enable SHiELD runs in argo workflows

spencerkclark:
This PR builds on #2376 and splits out from #2350 what is necessary to run SHiELD-wrapper-based prognostic simulations through our standard prognostic run argo workflow.  No changes to the frontend API are needed; the prognostic run workflow is modified to infer which template (`run-fv3gfs` or `run-shield`) to run based on the input config.

For convenience this also adds a starter base config for SHiELD, which is based on the configuration used in the PIRE simulations (but for simplicity with the mixed layer ocean turned off).  I have tested the `prognostic-run` and `restart-prognostic-run` workflows using a SHiELD-based config offline.  I'm not sure if we want to add an integration test yet or not.

Significant internal changes:
- Refactored the `prognostic-run` workflow to infer whether to use FV3GFS or SHiELD based on the config.
- Refactored the `restart-prognostic-run` workflow to infer whether to use FV3GFS or SHiELD based on the config at the provided URL.
- Refactored the directory structure of the base config YAMLs in `fv3kube` to better accommodate SHiELD configs.  No user-facing changes to the FV3GFS configs are made.

Note this PR makes use of YAML anchors and aliases to reduce the amount of duplicate configuration code.  Some illustration of how these work can be found [here](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/).  Use of this concept was already introduced in https://github.com/ai2cm/fv3net/pull/2103 within the `training.yaml` template, though this is the first time using it in the prognostic run.

-----

To illustrate the updated workflows I have included some example step outputs from `argo get` below (we ran the `prognostic-run` workflow for two segments and then ran one more segment via the `restart-prognostic-run` workflow).

### prognostic-run

```
STEP                                     TEMPLATE                                PODNAME                                        DURATION  MESSAGE
 ✔ 2023-12-21-baseline-shield-example    prognostic-run
 ├───✔ resolve-output-url                resolve-output-url/resolve-output-url   2023-12-21-baseline-shield-example-681218977   3s
 ├───✔ convert-input-config-to-artifact  convert-input-config-to-artifact        2023-12-21-baseline-shield-example-117187868   3s
 ├───✔ infer-wrapper                     infer-wrapper                           2023-12-21-baseline-shield-example-3326365873  3s
 ├───○ prepare-config-fv3gfs             prepare-config-fv3gfs                                                                            when ''shield.wrapper' == 'fv3gfs.wrapper'' evaluated false
[...]

PR #2386, opened 2024-01-05
fv3kube: update `package_data` in `setup.py`

spencerkclark:
#2377 reorganized the `base_yamls` directory in fv3kube to make room for SHiELD reference configurations, but neglected to update the `package_data` parameter in `setup.py` accordingly.  Without this change, installing `fv3kube` via something like:

```
$ pip install git+https://github.com/ai2cm/fv3net.git@b8e2f83b5206539724a4d096d0433ceeb3bc805a#egg=fv3kube&subdirectory=external/fv3kube
```

does not include the `base_yamls` files, which are an important component of the library.  I've tested this locally and it fixes the issue, e.g. use:

```
$ pip install git+https://github.com/ai2cm/fv3net.git@8fe01cd6c49ea635a1b07afd4ee4615db7555ce6#egg=fv3kube&subdirectory=external/fv3kube
```

and check for the existence of the `base_yamls` directory (note the different SHA from the original example).

### DIFF
--- a/external/fv3kube/setup.py
+++ b/external/fv3kube/setup.py
@@ -9,6 +9,6 @@ setup(
     author_email="oliverwm@allenai.org",
     packages=find_packages(),
     package_dir={"": "."},
-    package_data={"fv3kube": ["base_yamls/*/fv3config.yml"]},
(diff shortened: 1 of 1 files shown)


oliverwm1 (approved):
LGTM

PR #2392, opened 2024-06-11
fv3fit prediction on GPU

elynnwu:
This PR extends `xr_prediction` to work for GPU through the use of `cupy`. 

Significant internal changes:
- `DatasetPredictor` now calls either `_cpu_predict` or `_gpu_predict`

Requirement changes:
- Optional dependency `cupy`

- [ ] Tests added

GPU unit test currently works on perlmutter:

image


external/fv3fit/fv3fit/testing.py
```diff
@@ -9,6 +9,11 @@
 from fv3fit._shared.novelty_detector import NoveltyDetector
 from ._shared import Predictor, io, SAMPLE_DIM_NAME
 import numpy as np
+
+try:
+    import cupy as cp
```
frodre, line 14: Might also need `cupy_xarray` import here.

external/fv3fit/tests/test_xr_prediction.py
```diff
@@ -131,3 +141,25 @@ def test_DatasetPredictor_dump_load(tmpdir):
         np.testing.assert_allclose(
             loaded_predictor.predict(in_xr)["out0"], in_xr["in0"]
         )
+
+
+@unittest.skipIf(not cupy_available, "cupy not available")
```
frodre, line 146: Since we don't have CI w/ GPUs set up, could you paste a local test passing of this on Perlmutter?  

We could try setting up a github action that connects to our cluster as in the `full-model` repo and add a mark for `gpu-only` to run GPU-specific tests.  I think that would involve some additional dependency installation pathways for automated env setup for with GPU-specific modules (cupy and cupy_xarray).  I'm fine doing this in a separate PR though.

frodre (commented):
Looks good.  Just wondering whether the test passes without `cupy_xarray` on a GPU machine.

elynnwu:
The test passed without importing cupy_xarray, but from reading the doc, importing cupy_xarray "registers the `DataSet.cupy` and `DataArray.cupy` namespaces but is not used directly". So adding it back in as suggested.

external/fv3fit/tests/test_xr_prediction.py
elynnwu: Good idea. Opening a separate issue for enabling GPU-specific test.

frodre (approved):
Thanks for pulling the GPU prediction into fv3net! LGTM

PR #2394, opened 2024-06-13
Remove microphysics integration test

elynnwu:
Remove microphysics integration test since we no longer do development on it.


oliverwm1 (approved):
LGTM.

PR #2396, opened 2024-06-21
Optimize diagnostics for larger HPC nodes

frodre:
This PR optimizes the online diagnostic calculation and report generation for use on HPC where there are typically many more cores.  To achieve this I standardized some of the common aggregation functions for taking means and switched to use of dask distributed, which has more fine-grained control over workers and memory usage.  Additionally, the diagnostic functions are batched to occur concurrently using `joblib` but the individual dask tasks are elevated to the top-level of the scheduler to properly track individual task operations and memory usage.

Refactored public API:
- new flags for controlling the compute parallelism to the online prognostic run `compute.py`
  - `--scheduler-file` provide a dask scheduler file for an external dask distributed cluster when doing multi-node processing
  - `--num-concurrent-2d-calcs` number of concurrent 2D diagnostics to execute (i.e., diagnostic functions), defaults to 8
  - `--num-concurrent-3d-calcs` number of concurrent 3D diagnostics to execute (i.e., diagnostic functions), defaults to 2

Significant internal changes:
- Online diagnostics are now calculated using a dask distributed cluster with progress visible via `localhost:8787` and parallelized using batches of function calls with `joblib`
- Generated diagnostic plots has been parallelized using `joblib`
- Diurnal cycle now uses `flox` to speed up the dask groupby operation
- `load_scream_data` now separates to physics-only tendencies for `due_to_scream_physics` variables, whereas previously, it included both the nudging/ML update
  - `load_scream_data` now removes the ML precipitation from `PRATEsfc` since the diagnostics expects this to be a physics only quantity

Requirement changes:

- [ ] Tests added


workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -146,34 +162,61 @@ def _is_empty(data: Union[xr.Dataset, xr.DataArray]) -> bool:
...
+    w = w.where(mask, drop=True)
     return x, y, w
 
 
 def rms(x, y, w, dims):
     with xr.set_options(keep_attrs=True):
-        if np.isnan(x).any() or np.isnan(y).any():
-            x, y, w = ignore_nan_values(x, y, w)
-        return np.sqrt(((x - y) ** 2 * w).sum(dims) / w.sum(dims))
+        return xr.ufuncs.sqrt(((x - y) ** 2 * w).sum(dims) / w.sum(dims))
```
elynnwu, line 174: It looks like xarray has dropped all ufuncs in all favor of numpy’s ufuncs after 2022.06. So I think numpy sqrt would be the same already. Does this make much difference here performance wise?

workflows/diagnostics/fv3net/diagnostics/prognostic_run/diurnal_cycle.py
```diff
@@ -32,18 +33,28 @@ def _calc_ds_diurnal_cycle(ds):
     """
     Calculates the diurnal cycle for all variables.  Expects
     time dimension and longitude variable "lon".
+
+    Uses flox.xarray.xarray_reduce to speed up the diurnal cycle calc
+    See https://xarray.dev/blog/flox
```
elynnwu, line 38: Nice find!

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
```diff
@@ -143,39 +195,38 @@ def plot_cubed_sphere_map(
...
-            else:
-                fv3viz.plot_cube(ds, varname, ax=ax, vmin=vmin, vmax=vmax, cmap=cmap)
-            ax.set_title(plot_title)
-            plt.subplots_adjust(left=0.01, right=0.75, bottom=0.02)
-            data[varname][run] = MatplotlibFigure(fig, width="500px")
-            plt.close(fig)
+            )
+
+    jobs = [delayed(_plot_cubed_sphere_panel)(*args) for args in job_args]
+    with Parallel(n_jobs=8, verbose=10) as parallel:
```
elynnwu, line 223: Do we need to pass n_jobs? or are we okay hard coding it to 8?

projects/scream/scream_post/translate_prognostic_output.py
```diff
@@ -288,20 +305,61 @@ def rename_ML_correction_to_machine_learning(ds: xr.Dataset):
...
+            variable_names.append(match.group(1))
+
+    return variable_names
+
+
+def determine_tendency_variables(ds: xr.Dataset, run_label: str):
+    # from original scream output files
+    tendency_variables = _get_variable_names(ds, rf"^{run_label}_(.*)_tend$")
+    # from zarr files produced by format_output.py
+    tendency_variables += _get_variable_names(ds, rf"^(.*?)_tendency_due_to_nudging")
```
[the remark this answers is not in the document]
frodre: No, this is a specific add to include the nudging tendencies output by `format_output.py`

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
```diff
@@ -146,34 +162,61 @@ def _is_empty(data: Union[xr.Dataset, xr.DataArray]) -> bool:
 def ignore_nan_values(x, y, w):
     if "time" not in w.dims and "time" in x.dims:
         w = w.expand_dims(dim={"time": x.time})
-    mask = np.isnan(x) | np.isnan(y)
-    x = x.where(~mask, drop=True)
-    y = y.where(~mask, drop=True)
-    w = w.where(~mask, drop=True)
+    mask = (x.notnull() | y.notnull()).compute()
```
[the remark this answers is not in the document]
frodre: I wanted the mask to be computed up front so it doesn't get triggered by each subsequent `where` call.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
frodre: I believe I actually got some kind of error here or was debugging when I made the switch.  Or it could've been something related to me trying to remove explicit numpy references.  I'm fine switching back to see if it works.

workflows/diagnostics/fv3net/diagnostics/prognostic_run/views/matplotlib.py
frodre: I think since it's just figure generation and unlikely to be resource hungry, I'm fine with hardcoding.  it was implicitly hardcoded to n_jobs=1 🐢

workflows/diagnostics/fv3net/diagnostics/prognostic_run/compute.py
elynnwu: I'll try a run with np.sqrt and see if it makes a big difference. I would advocate for sticking with numpy in prep for the day when we have to upgrade xarray and everything breaks...

elynnwu: New run with np.sqrt did not show obvious slowdown. I would opt for keeping the original form. Also, we should keep the ignore nan check because it would nan out variables like T850 where there're actual nans.

frodre: I switched it back to including the NaN masking (without drop since xarray sum will ignore NaN) and using `np.sqrt`

elynnwu (approved):

PR #2399, opened 2024-07-03
Novelty detection compatibility with cupy GPU arrays

frodre:
This PR adds some basic infrastructure for handling Cupy arrays when using novelty detection. Unfortunately due to some of the underlying components, the novelty detection model cannot easily be moved to GPU-only calculations.  So for now, we are transferring to CPU where necessary to run the calculation.

### DIFF
--- a/external/fv3fit/fv3fit/_shared/packer.py
+++ b/external/fv3fit/fv3fit/_shared/packer.py
@@ -18,6 +18,13 @@ import xarray as xr
 import pandas as pd
 import tensorflow as tf
 
+try:
+    import cupy_xarray  # noqa
+
+    CUPY_AVAILABLE = True
+except ImportError:
+    CUPY_AVAILABLE = False
+
 
 def _feature_dims(data: xr.Dataset, sample_dims: Sequence[str]) -> Sequence[str]:
     return [str(dim) for dim in data.dims.keys() if dim not in sample_dims]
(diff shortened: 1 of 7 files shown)


external/fv3fit/fv3fit/_shared/packer.py
```diff
@@ -120,9 +127,16 @@ def pack(
 
     data_clipped = xr.Dataset(clip(data, config.clip))
     stacked = data_clipped.to_stacked_array(feature_dim_name, sample_dims=sample_dims)
-    stacked = stacked.dropna(feature_dim_name)
+    if CUPY_AVAILABLE and stacked.cupy.is_cupy:
+        # dropna features a numpy function which breaks on device cupy expectations
+        stacked = stacked.cupy.as_numpy()
+        stacked = stacked.dropna(feature_dim_name)
+        stacked = stacked.cupy.as_cupy()
```
elynnwu, line 134: It looks like we can use [`apply_ufunc`](https://cupy-xarray.readthedocs.io/source/apply-ufunc.html) to make `dropna` to work with cupy arrays. I couldn't get it to work right away but putting a note here in case we want to revisit later.

external/fv3fit/fv3fit/sklearn/_ocsvm_novelty_detector.py
```diff
@@ -128,7 +138,11 @@ def predict(self, data: xr.Dataset) -> xr.Dataset:
         X, _ = pack(
             stacked_data[self.input_variables], [SAMPLE_DIM_NAME], self.packer_config
         )
+
+        if CUPY_LOADED:
```
elynnwu, line 142: I think you also need `isinstance(X, cp.ndarray)` here

external/loaders/loaders/batches/save.py
```diff
@@ -39,31 +40,50 @@ def get_parser() -> argparse.ArgumentParser:
...
+    )
     return parser
 
 
-def main(data_config: str, output_path: str, variable_names: Sequence[str]):
+def main(
+    data_config: str,
+    output_path: str,
+    variable_names: Sequence[str],
+    num_workers: int = 1,
```
elynnwu, line 56:
```suggestion
    num_workers: int = 8,
```

to be consistent with the default num-workers

frodre: I left it as `num_workers=1` since the argument parser defines the default for external calls to this, and I didn't want the tests attached to `main` to use multiprocessing

elynnwu (approved):
lgtm

PR #2401, opened 2024-07-08
Add SCREAM prescribed radiation to online diagnostic variables

frodre:
Adds the prescribed/predicted radiation variables net sfc SW and net LW down to the online diagnostics to compare to the physics rather than adding more confusion to the renaming matrix.


elynnwu (approved):
lgtm

PR #2403, opened 2024-07-09
Tapering compatibility with cupy GPU arrays

elynnwu:
This PR adds compatibility for vertical tapering with cupy GPU arrays.

### DIFF
--- a/external/vcm/vcm/calc/calc.py
+++ b/external/vcm/vcm/calc/calc.py
@@ -1,13 +1,16 @@
+import numpy as np
+
 try:
     import cupy as cp
     import cupy_xarray  # noqa
 
-    CUPY_AVAILABLE = True
-
+    if cp.cuda.is_available():
+        xp = cp
+    else:  # fallback to numpy if cupy is installed but no GPU is available
+        xp = np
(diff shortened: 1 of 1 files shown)


external/vcm/vcm/calc/calc.py
```diff
@@ -43,7 +52,11 @@ def weighted_average(
 
 
 def vertical_tapering_scale_factors(n_levels: int, cutoff: int, rate: float):
-    z_arr = np.arange(n_levels)
-    scaled = np.exp((z_arr[slice(None, cutoff)] - cutoff) / rate)
-    unscaled = np.ones(n_levels - cutoff)
-    return np.hstack([scaled, unscaled])
+    if CUPY_AVAILABLE and cp.cuda.is_available():
```
frodre, line 55: Since we don't need to check an input array, I would remove the `CUPY_AVAILABLE` and just do the check of `cuda.is_available()` on import and set `xp` to `np` or `cp` depending on that.

frodre (approved):
Looks good. I made a suggestion for the import check and setting of `xp`

PR #2404, opened 2024-10-28
`fv3kube` test environment fix

brianhenn:
Restores passing unit and end-to-end integration tests:

- Fixes a `tox` environment problem that cropped up in the `fv3kube` tests due to unpinned test packages and pinned repo code. 
- Fixes a CircleCI environment problem installing dependencies necessary to launch the integration test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,9 +158,12 @@ jobs:
           name: Install dependencies
           command: |
             apt-get install -y wget jq make gettext gzip
-            python3 -m venv .venv
-            source .venv/bin/activate
+            apt install python3.11-venv
+            VENV_DIR="$(pwd)/.venv"
+            python3 -m venv $VENV_DIR
+            source "${VENV_DIR}/bin/activate"
             pip3 install yq==2.11.0
+            echo "source ${VENV_DIR}/bin/activate" >> $BASH_ENV
 
             # install kustomize
             wget "https://raw.githubusercontent.com/\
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -158,7 +158,12 @@ jobs:
           name: Install dependencies
           command: |
             apt-get install -y wget jq make gettext gzip
+            apt install -y python3.11-venv
+            VENV_DIR="$(pwd)/.venv"
+            python3 -m venv $VENV_DIR
+            source "${VENV_DIR}/bin/activate"
             pip3 install yq==2.11.0
```
brianhenn, line 165: The installation of `yq` had failed because CircleCI or the image was no longer allowing system-level installs of python packages this way.

external/fv3kube/tox.ini
```diff
@@ -4,6 +4,8 @@ skipsdist = true
 
 [testenv]
 deps =
+    pytest==4.6.11
```
brianhenn, line 7: These are now pinned to versions that are known to work in our other external packages that use tox. If left unpinned, tox now installs versions that conflict with python 3.8 that's in `fv3net`.

.circleci/config.yml
```diff
@@ -158,7 +158,12 @@ jobs:
           name: Install dependencies
           command: |
             apt-get install -y wget jq make gettext gzip
+            apt install -y python3.11-venv
```
elynnwu, line 161: Does it have to be 3.11 because the earlier version is not available?

elynnwu (approved):

.circleci/config.yml
brianhenn: It's the version of python pre-installed in the image we're using. It might be possible to make the installation less specific (`install python3-venv` instead), will investigate.

PR #2405, opened 2024-10-29
Remove scream weekly test

elynnwu:
We no longer support docker on the scream repository since it is not used. Removing the weekly build test here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -322,7 +322,6 @@ workflows:
   weekly-workflow:
     when: << pipeline.parameters.run-weekly-workflow >>
     jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -322,7 +322,6 @@ workflows:
   weekly-workflow:
     when: << pipeline.parameters.run-weekly-workflow >>
     jobs:
-      - build_and_test_scream
```
brianhenn, line 325: My preference would be for deleting the job definition for `build_and_test_scream` as well.

brianhenn (approved):

